PR #33, opened 2021-06-09
Feature/generate deb packages

lwronski:
Add support for building deb, dmg, pkg packages using Scala-cli.


README.md
```diff
@@ -40,6 +40,19 @@ $ ./mill integration.jvm.test 'scala.cli.integration.RunTests.Multiple scripts'
...
+
+Build native packagers:
+* `debian` for linux
+* `dmg` and `pkg` for centOS
+
+(generating native packager for specified format)
+```text
+$ ./mill -i 'cli.run' package ..arguments... --debian --output-package-path 'path.deb'
+$ ./mill -i 'cli.run' package ..arguments... --dmg --output-package-path 'path.dmg'
+$ ./mill -i 'cli.run' package ..arguments... --pkg --output-package-path 'path.pkg'
```
alexarchambault: The single quotes around `cli.run` are not necessary, I removed them from the README elsewhere.
```suggestion
$ ./mill -i cli.run package ..arguments... --debian --output-package-path 'path.deb'
$ ./mill -i cli.run package ..arguments... --dmg --output-package-path 'path.dmg'
$ ./mill -i cli.run package ..arguments... --pkg --output-package-path 'path.pkg'
```

build.sc
```diff
@@ -83,7 +83,8 @@ class Build(val crossScalaVersion: String) extends CrossSbtModule with ScalaCliP
     Deps.scalaJsTestAdapter,
     Deps.scalametaTrees,
     Deps.scalaparse,
-    Deps.swoval
+    Deps.swoval,
+    Deps.scalaPackager
```
alexarchambault: Maybe insert this right above `scalaparse`, to keep dependencies sorted.

modules/cli-core/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -97,10 +100,28 @@ object Package extends ScalaCommand[PackageOptions] {
...
   }
 
+  private def buildNativePackage(options: PackageOptions, sourceAppPath: os.Path) = {
+
+    def resolveNativePackagePath(nativePackagerType: NativePackagerType): os.Path =
+      os.Path(options.outputPackagePath.getOrElse(nativePackagerType.defaultNativePackageName), Os.pwd)
+
+    val force = options.force
+
+    import NativePackagerType._
```
alexarchambault: Couldn't `Debian` and all just extend `PackageType`, instead of introducing a new `NativePackagerType`? This way, they could be handled a few lines above, when we match `options.packageType`.

modules/integration-core/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -0,0 +1,74 @@
...
+          cwd = root,
+          stdin = os.Inherit,
+          stdout = os.Inherit,
+          stderr = os.Inherit
+        )
+
+        val pkgAppPath = root / pkgAppFile
+        expect(os.isFile(pkgAppPath))
+
+        if( TestUtil.isCI) {
```
alexarchambault: NIT
```suggestion
        if (TestUtil.isCI) {
```

modules/integration-core/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -0,0 +1,74 @@
...
+          os.proc("installer", "-pkg", pkgAppFile, "-target", "CurrentUserHomeDirectory").call(
+            cwd = root,
+            stdin = os.Inherit,
+            stdout = os.Inherit,
+            stderr = os.Inherit
+          )
+
+          val output = os.proc(s"/Users/runner/Applications/$appName.app/Contents/MacOS/$appName").call(cwd = os.root).out.text.trim
```
alexarchambault: Just in case, we could get the home directory, rather than hard-coding it:
```suggestion
          val home = sys.props("user.home")
          val output = os.proc(s"$home/Applications/$appName.app/Contents/MacOS/$appName").call(cwd = os.root).out.text.trim
```

.github/scripts/generate-packages-linux.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+# Build scala-cli native image
+./mill -i show cli-core.nativeImage
+# Build deb package for scala-cli
+cs launch org.virtuslab::scala-packager-cli:0.1.3 -- --source-app-path ./out/cli-core/nativeImage/dest/scala --output scala-cli.deb --debian
```
alexarchambault: I think you can get the launcher path directly from mill, rather than hard-coding it. Something like this ought to work:
```suggestion
LAUNCHER="$(./mill -i show cli-core.nativeImage | jq -r)"
# Build deb package for scala-cli
cs launch org.virtuslab::scala-packager-cli:0.1.3 -- --source-app-path "$LAUNCHER" --output scala-cli.deb --debian
```

.github/scripts/generate-packages-linux.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
```
alexarchambault: [It'd be safer](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/) to add at least `set -e`, so that the script will exit if any command fails. With the suggestion below, `set -o pipefail` is necessary too. And we can add `set -u` just in case.
```suggestion
#!/usr/bin/env bash
set -euo pipefail
```

.github/scripts/generate-packages-linux.sh
```diff
@@ -0,0 +1,7 @@
...
+# Build scala-cli native image
+./mill -i show cli-core.nativeImage
+# Build deb package for scala-cli
+cs launch org.virtuslab::scala-packager-cli:0.1.3 -- --source-app-path ./out/cli-core/nativeImage/dest/scala --output scala-cli.deb --debian
+# Copy deb package to artifacts
+mkdir artifacts && cp ./scala-cli.deb artifacts/scala-cli.deb
```
alexarchambault: Thanks to `set -e`, we don't need the `&&` anymore. And passing `-p` to `mkdir` is safer, in case the directory already exists (if we move this step around for example).
```suggestion
mkdir -p artifacts
cp ./scala-cli.deb artifacts/scala-cli.deb
```

.github/scripts/generate-packages-linux.sh
lwronski: This didn't work because .mill returns the path in this format `"ref:a423c713:/Users/lwronski/projects/scalacli/out/cli-core/nativeImage/dest/scala"`

I added `grep -oE "[^:]+$")` to extracting cli-core path.

modules/cli-core/src/main/scala/scala/cli/commands/Package.scala
lwronski: Now `NativePackagerType` extends `PackageType`. I need to create a bootstrap file for each nativePackager type, so it can be handled for this case `nativePackagerType: PackageOptions.NativePackagerType` and then the appropriate builder can be called from scala-packager:  `nativePackagerType match {
          case PackageOptions.PackageType.Debian => ... DebianPackage(...).build()
           case PackageOptions.PackageType.Dmg => ... DmgPackage(...).build()`

README.md
```diff
@@ -40,6 +40,19 @@ $ ./mill integration.jvm.test 'scala.cli.integration.RunTests.Multiple scripts'
 $ ./mill integration.native.test
 ```
 
+#### Generate native packages
+
+Build native packagers:
+* `debian` for linux
+* `dmg` and `pkg` for centOS
```
alexarchambault: macOS?

modules/cli-core/src/main/scala/scala/cli/commands/PackageOptions.scala
```diff
@@ -25,27 +25,45 @@ final case class PackageOptions(
...
+  @Group("Package")
+  @HelpMessage("Build dmg package, available only on centOS")
+    dmg: Boolean = false,
+  @Group("Package")
+  @HelpMessage("Build pkg package, available only on centOS")
+    pkg: Boolean = false,
 ) {
   import PackageOptions.PackageType
-  def packageType: PackageType =
+  def packageType: PackageType = {
```
alexarchambault: No need to add braces here I think.

project/deps.sc
```diff
@@ -49,6 +49,7 @@ object Deps {
   def svmSubs = ivy"org.scalameta::svm-subs:20.2.0"
   def swoval = ivy"com.swoval:file-tree-views:2.1.5"
   def testInterface = ivy"org.scala-sbt:test-interface:1.0"
+  def scalaPackager = ivy"org.virtuslab::scala-packager:0.1.3"
```
alexarchambault: Let's add this one jsut before `scalaparse` above, to preserve the ordering.

modules/cli-core/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -2,19 +2,21 @@ package scala.cli.commands
 
 import caseapp._
 import coursier.launcher.{AssemblyGenerator, BootstrapGenerator, ClassPathEntry, Parameters, Preamble}
-import scala.build.{Build, Inputs, Os}
-import scala.scalanative.{build => sn}
-import scala.scalanative.util.Scope
+import packager.BuildSettings
+import packager.macOS.dmg.DmgPackage
```
alexarchambault: One last thing: it's not common to use upper case letters in package names. `packager.macos` (or `packager.mac`) would feel more standard.

PR #55, opened 2021-07-19
First implementation of landing page

romanowski:


alexarchambault (commented):
I'm fine with the overall points.

But are we sure we want a markdown page as a landing page? Wouldn't the layout used by [scalameta](https://scalameta.org) ([source](https://github.com/scalameta/scalameta/blob/a62b0f31f2fc6cf3faa81a5caaac48fa6206c9de/website/pages/en/index.js)) or [Metals](https://scalameta.org/metals) ([source](https://github.com/scalameta/metals/blob/13e669b5419020945a217cf8bfb9545557c9c1dd/website/src/pages/index.js)) be better?

romanowski:
I've started with `.md` to focus more on the content than layout but I will rewrite it into Metals-like layout.

PR #63, opened 2021-07-23
Static binaries

alexarchambault:


alexarchambault:
Going to merge, but I might restrict those new jobs to master, if ever they're a hindrance in the future…

PR #71, opened 2021-07-28
Enable specify lanucher app name and create msi installer with logo

### LINKED ISSUE #49: Packaging should require configuration
Currently `scala-cli package --deb` create a valid deb package that adds `app` command to PATH.

We should have it configurable and also have interactive way to configure it

lwronski:
- Enable to configure launcher app name which will be added to PATH
- Enable to create msg installer with logo


.github/workflows/ci.yml
```diff
@@ -31,7 +31,6 @@ jobs:
     - name: Build OS packages
       run: .github/scripts/generate-os-packages.sh
       shell: bash
-      if: runner.os == 'Windows'
```
alexarchambault: Remove this line?

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -27,6 +27,9 @@ final case class BuildOptions(
                    mainClass: Option[String]              = None,
                  testOptions: TestOptions                 = TestOptions(),
               packageOptions: PackageOptions              = PackageOptions(),
+               debianOptions: DebianOptions               = DebianOptions(),
+               redHatOptions: RedHatOptions               = RedHatOptions(),
+              windowsOptions: WindowsOptions              = WindowsOptions(),
```
alexarchambault: Maybe these should be fields of `PackageOptions`, rather than being here.

website/docs/packages/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
+    "label": "Native Packages",
+    "position": 18
+  }
```
alexarchambault: Couldn't we add the content below as sections in the `package.md` page? The extra sidebar section feels odd to me.

.github/workflows/ci.yml
lwronski: In last week I want to generate os packages for linux and test them, but I was blocked, because the stage `CI / native-tests (ubuntu-latest) ` was failing.  So I removed this line to run os package generation on `jvm-tests (ubuntu-latest)`.

And then I forgot to restore this line, I will do it in next commit.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
alexarchambault: I mean, you can keep the 3 classes, `DebianOptions`, etc., but move the 3 fields, `debianOptions`, etc., to `PackageOptions`.

alexarchambault (commented):
Looks good!

But can you squash the commits you added since the previous review with the relevant earlier commits, before merging?

PR #78, opened 2021-08-04
Feature/build docker images

lwronski:
Building docker images on CI and used them for testing.


.github/scripts/docker/DockerFile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/java-debian10
```
alexarchambault: I think `gcr.io/distroless/base-debian10` should work as a base, doesn't it?

.github/workflows/ci.yml
```diff
@@ -10,6 +10,7 @@ on:
 jobs:
   jvm-tests:
     runs-on: ${{ matrix.OS }}
+    if: 1 == 2
```
alexarchambault: Remove this?

.github/workflows/ci.yml
```diff
@@ -45,6 +46,7 @@ jobs:
 
   native-tests:
     runs-on: ${{ matrix.OS }}
+    if: 2 == 3
```
alexarchambault: Remove this too?

.github/workflows/ci.yml
```diff
@@ -141,6 +148,7 @@ jobs:
 
   reference-doc:
     runs-on: ubuntu-latest
+    if: 2 == 1
```
alexarchambault: Remove this?

.github/workflows/ci.yml
```diff
@@ -101,6 +107,7 @@ jobs:
 
   native-static-tests:
     runs-on: ubuntu-latest
+    if: 2 == 1
```
alexarchambault: Remove this?

build.sc
```diff
@@ -357,6 +367,11 @@ trait JvmIntegrationCore extends CliIntegrationCore {
   def cliKind = "jvm"
 }
 
+trait DockerIntegrationCore extends CliIntegrationCore {
+  def cliKind = "docker"
+  def testLauncher = `cli-core`.launcher()
```
alexarchambault: Do we need to pass a test launcher to the docker tests?

modules/integration-core/src/test/scala/scala/cli/integration/RunDockerTests.scala
```diff
@@ -0,0 +1,45 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+import java.io.{ByteArrayOutputStream, File}
+import java.nio.charset.Charset
+
+import scala.util.Properties
+
+class RunDockerTests extends munit.FunSuite {
```
alexarchambault: This file has 4 space indent, right? We use 2 space indent most of the time.

modules/integration-core/src/test/scala/scala/cli/integration/RunDockerTests.scala
```diff
@@ -0,0 +1,45 @@
...
+                System.err.write(b, 0, len)
+                },
+                mergeErrIntoOut = true
+            )
+            val output = new String(rawOutput.toByteArray, Charset.defaultCharset())
+            expect(output.linesIterator.toVector.last == message)
+        }
+    }
+
+    if (Properties.isLinux && TestUtil.cliKind == "docker")
```
alexarchambault: We shouldn't need this check… We'll only run these tests on systems that support it. (Incidentally, running them on macOS or Windows might work, as the mostly static binary is built from docker too.)

.github/scripts/docker/DockerFile
lwronski: If I use `gcr.io/distroless/base-debian10`, the test is fail -
` /usr/local/bin/scala: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory`

build.sc
lwronski: Probably not, so how should I implement method testLauncher in DockerIntegrationCore?

.github/scripts/docker/DockerFile
alexarchambault: The mostly-static binary is precisely made not to depend on `libstdc++`. It works fine [here](https://github.com/VirtuslabRnD/scala-cli/blob/c01a212af47e61a2c530f7bf0b4186c4a58bf991/modules/integration-core/src/test/scala/scala/cli/integration/RunTestDefinitions.scala#L809-L811).

.github/workflows/ci.yml
alexarchambault: This still appear in the first commit. Same for the other `if`s.

build.sc
alexarchambault: Just don't use `CliIntegrationBase` in `DockerIntegrationCore`.

modules/integration-core/src/test/scala/scala/cli/integration/RunDockerTests.scala
alexarchambault: I still see 4 space indents in the first commit.

modules/integration-core/src/test/scala/scala/cli/integration/RunDockerTests.scala
```diff
@@ -0,0 +1,45 @@
...
+import com.eed3si9n.expecty.Expecty.expect
+
+import java.io.{ByteArrayOutputStream, File}
+import java.nio.charset.Charset
+
+import scala.util.Properties
+
+class RunDockerTests extends munit.FunSuite {
+
+    def simplRunAppDocker(): Unit = {
```
alexarchambault: `s/simplRunAppDocker/simpleRunAppDocker/g`?

.github/workflows/ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
     - name: Generate docker image
       run: .github/scripts/generate-docker-images-ubuntu.sh
     - name: Docker integration tests
-      run: ./mill integration.docker.test 'scala.cli.integration.RunDockerTests.*'
+      run:  ./mill integration.docker.test
```
alexarchambault: Unnecessary space before `./mill`

.github/workflows/ci.yml
lwronski: I have just squashed all commits and remove this change from history

.github/scripts/docker/DockerFile
lwronski: It's works. Locally, I was using different binary with scala-cli so it throws an error.

build.sc
```diff
@@ -53,7 +53,22 @@ object `integration-core` extends Module {
   }
 }
 
+object `integration-docker` extends CliIntegrationDockerCore {
```
alexarchambault: Do we really need that one? I think by renaming `modules/integration-docker` to `modules/integration/docker`, `integration.docker` below should automatically pick the right sources.

lwronski: Yes, we don't need this definition of module. I moved `modules/integration-docker` to `modules/integration/docker` and it correct pick the right sources.

alexarchambault (approved):

PR #102, opened 2021-08-21
Automatically install Visual C++ redist from MSI

alexarchambault:


alexarchambault:
Tested locally, and it worked fine 🎉

PR #104, opened 2021-08-21
fmt everything

alexarchambault:


alexarchambault:
Unlike what it may seems, many changes here are _not_ the result of just running `scalafmt` at the root of the repo. `scalafmt` tends to break long lines at odd places, so many of its changes were manually tweaked (while ensuring scalafmt is still fine with the resulting code). Which might make rebasing other outgoing PRs a problem.

PR #105, opened 2021-08-26
Fix issues in framework detection logic

Baccata:
Hey folks. 

Let me start by saying that I'm utterly excited about this project. It had been a while since I had been hyped about anything, and I'm really really happy that this is a thing. I've not communicated its existence to other people, as I've been informed that you may want to keep it low profile for the time being. But it looks awesome 👍 

Anyway, I've played with it a little bit and found that the `test` command wasn't working for a test-framework [I maintain](https://github.com/disneystreaming/weaver-test). As far I can tell, there were two reasons for it : 

* for some reason, a `module-info.class` appears to be on the classpath when attempting to run a test with weaver. I have no idea where it comes from, but java8 isn't happy with trying to load it. 
* the test framework in weaver has several constructors. 

These changes seem to do the trick when running the cli from sources. I have however ran into the following message when running scala-cli from the nightly build, after switching my JRE to 9+ to avoid the `UnsupportedClassVersionError` :

```scala 
java.lang.NoClassDefFoundError: module-info is not a class because access_flag ACC_MODULE is set
```

Since I'm not managing to reproduce when running from sources, and that `NoClassDefFound` is typically a fatal error, I'm not sure I should try to catch it, or whether I should apply a more precise filter somehow. Maybe just filtering out `module-info` might do the trick ? WDYT ? 

#### Side question regarding running scala-cli from sources : 

Running `mill -i scala ...` doesn't work for me : the `publishLocalNoFluff` seem to publish the artifacts in a cache local to the project, but the `cli.run` task attempts to pull them from my `$HOME/.ivy2/local`. I worked-around it by running `mill __.publishLocal` manually before running `mill -i scala ...`, but I was wondering if there's something I've missed that would avoid the workaround. 

#### PS 

I've updated the [vscode-scaladex-search]( https://github.com/Baccata/vscode-scaladex-search) extension to make it scala-cli friendly : using it when a .conf file is open gives the correct formatting, and using it when a `scala` file is open but no sbt file is present in the workspace formats the deps using the ammonite syntax. 

Kind regards


alexarchambault (commented):
Thanks @Baccata! It'd be nice to add an integration test for weaver-test, in [`TestTestDefinitions.scala`](https://github.com/VirtuslabRnD/scala-cli/blob/ab84f810c753447940cf3ebdf8d793ebf83bee92/modules/integration/src/test/scala/scala/cli/integration/TestTestDefinitions.scala), to prevent regressions. With just a small `.scala` file with some weaver-test tests that always pass, and running it for both JVM and Scala.JS (see how the other tests in `TestTestDefinitions.scala` do - [this section](https://github.com/VirtuslabRnD/scala-cli#run-integration-tests-with-the-jvm-launcher) of the README shows how to run specific integration tests.)

Baccata:
Added an integration test in 0fb7df7

Baccata:
Ironically it looks like this integration test has uncovered a bug in weaver (on windows, specifically), where the path relativisation (which happens when capturing source locations) seems to blow on windows. I've created an [issue](https://github.com/disneystreaming/weaver-test/issues/364) to investigate. 

In the meantime, should I add a line to ignore the test on windows for the time being, or would you rather I fixed weaver first and came back to this PR ?

alexarchambault:
About https://github.com/VirtuslabRnD/scala-cli/pull/105/commits/25971dc2cd819b82a078cd4eaa4f6eb6a534ce31, it should be necessary when we tag releases, as sonatype snapshots shouldn't be automatically added for releases.

alexarchambault (approved):
Merging, thanks again!

PR #111, opened 2021-09-02
Update brew formula and deb/yum repository in react of releasing scal…

lwronski:


project/publish.sc
```diff
@@ -76,6 +76,34 @@ trait ScalaCliPublishModule extends PublishModule with PublishLocalNoFluff {
...
+      os.proc("git", "status").call(cwd = repoDir).out.text().trim.contains("nothing to commit")
+    ) {
+      println("Nothing Changes")
+    }
+    else {
+      os.proc("git", "add", "-A").call(cwd = repoDir)
+      os.proc("git", "commit", "-am", name).call(cwd = repoDir)
+      os.proc("git", "push", "origin", branch).call(cwd = repoDir)
+    }
+  }
```
alexarchambault: Is there a reason to put those in `PublishModule`? I'm not sure it's the right place for those (just leaving them in `build.sc`, or putting them at the root of `settings.sc`, would have worked better IMO).

lwronski: There is no reason to put this logic in `Publish Module`. So I'm going to move this logic to `build.sc` except method `getToken()` which I will move to `settings.sc`.

.github/workflows/ci.yml
```diff
@@ -226,8 +226,47 @@ jobs:
...
+          fetch-depth: 0
+          submodules: true
+      - uses: actions/download-artifact@v2
+        with:
+          name: launchers
+          path: artifacts/
+      - name: Display structure of downloaded files
+        run: ls -R
+        working-directory: artifacts/
+      - uses: webfactory/ssh-agent@v0.5.3
```
alexarchambault: Just in case, let's use full commit numbers rather than tags, like
```suggestion
      - uses: webfactory/ssh-agent@5f066a372ec13036ab7cb9a8adf18c936f8d2043
```

It's safer, securitywise (a tag can be erased, while a commit number can't really).

.github/workflows/ci.yml
```diff
@@ -226,8 +226,47 @@ jobs:
...
+        with:
+          ssh-private-key: |
+            ${{ secrets.SCALA_CLI_PACKAGES_KEY }}
+            ${{ secrets.HOMEBREW_SCALA_CLI }}
+      - uses: webfactory/ssh-agent@v0.5.3
```
alexarchambault: Same as above.

project/publish.sc
alexarchambault: It seems these are still in `ScalaCliPublishModule`, right?

lwronski: Right, I copy the logic to build.sc from settings.sc, but I forgot to remove it from settings.sc

.github/workflows/ci.yml
```diff
@@ -226,8 +226,47 @@ jobs:
...
+          submodules: true
+      - uses: actions/download-artifact@v2
+        with:
+          name: launchers
+          path: artifacts/
+      - name: Display structure of downloaded files
+        run: ls -R
+        working-directory: artifacts/
+      - uses: webfactory/ssh-agent@5f066a372ec13036ab7cb9a8adf18c936f8d2043
+      - uses: webfactory/ssh-agent@5f066a372ec13036ab7cb9a8adf18c936f8d2043
```
alexarchambault: Duplicated line?
```suggestion
      - uses: webfactory/ssh-agent@5f066a372ec13036ab7cb9a8adf18c936f8d2043
```

alexarchambault (approved):

PR #113, opened 2021-09-03
Create docker images via package sub-command

lwronski:
It will be added new command to generate docker images from the package sub-command:
`scala-cli package helloWorld.scala --docker --image-repository hello-scala`

It works also for ScalaJs application:
`scala-cli package helloWorld.scala --js --docker --image-repository hello-scala`


modules/cli/src/main/scala/scala/cli/commands/PackagerOptions.scala
```diff
@@ -60,7 +60,31 @@ final case class PackagerOptions(
...
+    suppressValidation: Option[Boolean] = None,
+  @Group("Windows")
+  @HelpMessage("Path to extra WIX config content")
+  @ValueDescription("path")
+    extraConfig: List[String] = Nil,
+  @Group("Docker")
+  @HelpMessage(
+    "Building the container from base image"
+  )
+  from: Option[String] = None,
```
alexarchambault: Maybe prefix the docker options with `docker`, like `dockerFrom`, `dockerImageRegistry`, etc.

alexarchambault (commented):
LGTM overall, but the CI error seem legit.

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -202,4 +202,80 @@ class NativePackagerTests extends munit.FunSuite {
...
+        )
+
+      val expectedImage =
+        s"$imageRepository:$imageTag"
+
+      val output = os.proc("docker", "run", expectedImage).call(cwd = os.root).out.text.trim
+      expect(output == message)
+
+      // clear
+      os.proc("docker", "rmi", "-f", expectedImage).call(cwd = os.root)
```
alexarchambault: Same as above: this could be done from a `finally` clause.

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -202,4 +202,80 @@ class NativePackagerTests extends munit.FunSuite {
...
```
alexarchambault: Maybe it'd be safer to clean-up the image from a `finally` clause:
```suggestion
      try {
        val output = os.proc("docker", "run", expectedImage).call(cwd = os.root).out.text.trim
        expect(output == message)
      }
      finally {
        // clear
        os.proc("docker", "rmi", "-f", expectedImage).call(cwd = os.root, check = false)
      }
```

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
lwronski: Thanks for the review. Right, it will be a better solution.

modules/cli/src/main/resources/META-INF/native-image/org.virtuslab/scala-cli-core/reflect-config.json
```diff
@@ -793,5 +793,153 @@
...
+  {
+    "name":"com.google.api.client.util.GenericData",
+    "allDeclaredFields":true
+  },
+  {
+    "name":"java.lang.String"
+  },
+  {
+    "name":"java.util.Date"
+  },
```
alexarchambault: These are usually not necessary, unless gson uses those at some point…

lwronski: If I remove `"name":"java.lang.String"`, then it throws following error:
```
Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.NoClassDefFoundError: java.lang.String
        at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:566)
```

alexarchambault (approved):

PR #116, opened 2021-09-05
Fix reference to macOS

kubukoz:


alexarchambault:
@kubukoz Thanks! The messages you edited in your first commit actually originate from the sources, and `./mill generate-reference-doc.run` updates the `.md` files. I opened https://github.com/VirtuslabRnD/scala-cli/pull/118 to update the messages in the sources too.

kubukoz:
doh, I now see this diff is +0 -0, guess that makes sense as the second commit reverted my only change 😂

Of course it's from the sources... I'll do it right later today.

alexarchambault:
> Of course it's from the sources... I'll do it right later today.

https://github.com/VirtuslabRnD/scala-cli/pull/118 does it, I just merged it.

kubukoz:
Thanks :)

PR #121, opened 2021-09-08
Ignore shebang at the top of source files

lwronski:


alexarchambault (approved):
Let's merge, we can always think of more complex cases later (if ever there are some).

PR #123, opened 2021-09-09
Update Scala to 3.0.2

alexarchambault:


alexarchambault:
The Ammonite version here isn't published for all Scala versions we support, so CI is going to be red. https://github.com/com-lihaoyi/Ammonite/pull/1215 should fix that.

PR #130, opened 2021-09-13
Add export command

alexarchambault:


alexarchambault:
This is still a bit of a work in progress, but basic export to sbt works. Export to Mill is missing, and more options could be handled (see the `TODO`s in `Sbt.scala`).

PR #132, opened 2021-09-14
Accept github gists

lwronski:


modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -217,41 +222,50 @@ object Inputs {
     cwd: os.Path,
     directories: Directories,
     baseProjectName: String,
-    download: String => Either[String, Array[Byte]],
+    download: String => Either[String, List[UrlPath]],
```
alexarchambault: I think we shouldn't need to change that, and add `UrlPath`, here. The `Array[Byte]` could be changed to `os.Path` if needed, but that would only be a nice-to-have (see my comment below).

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -217,41 +222,50 @@ object Inputs {
...
+          Right(Seq(VirtualScript(stdinOpt0.get, "<stdin>", os.sub / "stdin.sc")))
         else if (arg.contains("://"))
-          download(arg).map { content =>
-            val wrapperPath = {
-              val u = new URI(arg) // FIXME Ignore parsing errors?
-              val it = Option(u.getScheme).iterator ++
-                Option(u.getAuthority).iterator ++
-                Option(u.getPath).iterator.flatMap(_.split('/').iterator)
-              os.sub / it.filter(_.nonEmpty).toVector
+          download(arg).map { files =>
```
alexarchambault: The Gist detection could be done around here. If a Gist URL is detected, it can be changed to the Gist zip URL (see the "Download ZIP" button on every Gist), and the zip file be downloaded here.

With the zip file, we can have a look at its content without unzipping it on disk, using the `java.util.zip.*` APIs (`ZipInputStream`, that can be passed a `ByteArrayInputStream` built from the `Array[Byte]`, or `ZipFile`, that can be passed a `File` if we change `download` to give us an `os.Path`)

This way, we can peek files from the Gist zip, and create `Virtual*` instances out of it, with a sensible `source` field (for `source`, we can try to build URLs like [this one](https://gist.github.com/alexarchambault/7b4ec20c4033690dd750ffd601e540ec#file-main-sc), pointing at the Gist individual files).

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -185,14 +186,36 @@ final case class SharedOptions(
...
+
+      val artifact = Artifact(parsedUrl).withChanging(true)
       val res = coursierCache.logger.use {
         coursierCache.file(artifact).run.unsafeRun()(coursierCache.ec)
       }
       res
         .left.map(_.describe)
-        .map(f => os.read.bytes(os.Path(f, Os.pwd)))
+        .map { f =>
+          // trying to unpack archive, support for downloading tar.gz files
```
alexarchambault: We shouldn't need to unpack things on disk, see my comments above.

alexarchambault (approved):

PR #133, opened 2021-09-15
Explicitly pass errors around

alexarchambault:
This makes some internal methods (mostly in the `build` module) return errors explicitly via `Either[BuildException, ?]`, rather than throwing an exception. These errors go up to the "leaf" modules (cli, tests), which can then report errors properly (by printing a message to stderr, rather than throwing, which prints a stack trace).

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -29,13 +29,14 @@ import _root_.scala.util.Properties
 implicit def millModuleBasePath: define.BasePath =
   define.BasePath(super.millModuleBasePath.value / "modules")
 
-object cli           extends Cli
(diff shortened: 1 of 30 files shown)


alexarchambault:
Only a handful of errors are reported via this channel for now, but with this PR, everything's ready for more errors to be reported this way.

alexarchambault:
This relies on the `-Xasync` option, available since `2.12.12` in scala `2.12`, to easily deal with eithers.

It should work as is when we'll port things to `2.13` (we're waiting on some Scala Native-related stuff for that). For Scala 3, we'll likely need CPS stuff.

PR #134, opened 2021-09-16
Add  command to generate scala project template

lwronski:
`scala-cli new` generate template scala-cli project from this repo https://github.com/VirtuslabRnD/scala-cli.g8


alexarchambault:
We should have merged this IMO. And hide the `new` command by default.

PR #135, opened 2021-09-17
Add docker package logging and generate bootstrap file in temp directory for docker

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -246,29 +246,45 @@ object Package extends ScalaCommand[PackageOptions] {
               imageResizerOpt = imageResizerOpt
             ).build()
         }
-    }
+      case PackageType.Docker =>
+        val exec = if (build.options.scalaJsOptions.enable) "node" else "sh"
+        val from = build.options.packageOptions.dockerOptions.from match {
+          case Some(baseImage) => baseImage
+          case None => if (build.options.scalaJsOptions.enable) "node" else "openjdk:8-jre-slim"
```
alexarchambault, line 253: What about the Scala Native case? It would make sense on Linux at least. Maybe [`frolvlad/alpine-glibc`](https://hub.docker.com/r/frolvlad/alpine-glibc) could work as a base image?

But maybe that can be done in another PR…

PR #137, opened 2021-09-17
Add install home command

lwronski:


modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,39 @@
...
+import caseapp._
+import coursier.env.{EnvironmentUpdate, ProfileUpdater}
+
+import java.io.File
+import scala.util.Properties
+
+object InstallHome extends ScalaCommand[InstallHomeOptions] {
+  override def hidden = true
+  override def names = List(
+    List("install", "home")
```
alexarchambault: Let's not add this name here. I believe the default would give `scala-cli install-home` without the override (while it's `scala-cli install home` with the override here, so 3 args instead of 2).

That should allow not to conflict with an `install` command, if we were to add one later on.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,39 @@
...
+  )
+  def run(options: InstallHomeOptions, args: RemainingArgs): Unit = {
+
+    if (Properties.isWin)
+      System.err.println(s"Install home is not supported yet on windows")
```
alexarchambault: NIT
```suggestion
      System.err.println("Install home is not supported yet on Windows")
```

alexarchambault: But we could make that work on Windows I think, just like in [`AddPath`](https://github.com/VirtuslabRnD/scala-cli/blob/d324ee94766bbbf585cab6304e4044dd9d117cd5/modules/cli/src/main/scala/scala/cli/commands/AddPath.scala#L21-L28).

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    // copy scala-cli to binary directory
+    val binDirPath      = scala.build.Directories.default().binRepoDir
+    val scalaCliBinPath = binDirPath / "scala-cli"
+
+    if (os.exists(scalaCliBinPath)) os.remove(scalaCliBinPath)
+    os.makeDir.all(binDirPath)
+    os.copy(os.Path(options.scalaCliBinaryPath, os.pwd), scalaCliBinPath)
```
alexarchambault: Should we really wipe all of `scalaCliBinPath` here? It might be a problem if we decide to put some more binaries there later on.

Maybe we should ask the user if (s)he wants to proceed if a launcher already exists in the bin dir.

alexarchambault: I think the filename needs to be specified, like:
```suggestion
    os.copy(os.Path(options.scalaCliBinaryPath, os.pwd), scalaCliBinPath / "scala-cli")
```

`os.copy` also accepts a `createFolders` argument, so that the `os.makeDir.all` shouldn't be necessary.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    val update    = EnvironmentUpdate(Nil, Seq("PATH" -> scalaCliBinPath.toString()))
+    val updater   = ProfileUpdater()
+    val didUpdate = updater.applyUpdate(update)
+
+    if (didUpdate) {
+      println("Successfully installed scala-cli")
+    }
+    else {
+      System.err.println(s"ScalaCli is already installed")
```
alexarchambault: NIT

```suggestion
      System.err.println("scala-cli is already installed")
```

modules/cli/src/main/scala/scala/cli/commands/InstallHomeOptions.scala
```diff
@@ -0,0 +1,19 @@
+package scala.cli.commands
+
+import caseapp._
+
+// format: off
+@HelpMessage(
+  """Linking scala-cli binary to home
```
alexarchambault: ?
```suggestion
  """Install scala-cli in a sub-directory of the home directory
```

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
lwronski: Good point, we should also support windows

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,55 @@
...
+object InstallHome extends ScalaCommand[InstallHomeOptions] {
+  override def hidden: Boolean = true
+  def run(options: InstallHomeOptions, args: RemainingArgs): Unit = {
+
+    val binDirPath      = scala.build.Directories.default().binRepoDir
+    val scalaCliBinPath = binDirPath / "scala-cli"
+
+    if (os.exists(scalaCliBinPath)) {
+      println("scala-cli already exists. Do you want to override it in local repo directory [Y/n]")
+      val replace = readLine()
```
alexarchambault: We don't want to call `readLine` in non-interactive environments (like CIs). This can be detected in most cases with `System.console() == null`, but it's not enough sometimes IIRC. So `coursier.paths.Util.useAnsiOutput()` could be used (it's misnamed, it's not really about ansi stuff…).

You may also want to add a `-f` / `--force` option, to force erasing things no matter what.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,55 @@
...
+        sys.exit(1)
+      }
+    }
+
+    os.copy(
+      from = os.Path(options.scalaCliBinaryPath, os.pwd),
+      to = scalaCliBinPath / "scala-cli",
+      createFolders = true
+    )
+    os.perms.set(scalaCliBinPath / "scala-cli", os.PermSet.fromString("rwxrwxr-x"))
```
alexarchambault: This shouldn't be called on Windows.

alexarchambault (approved):

kubukoz:
Random thought now that I stumbled into this, is there going to be an `install` command to `package` an app and also put it into a directory on the `$PATH`? (I think `stack` and possibly other tools do this too)

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -0,0 +1,67 @@
...
+object InstallHome extends ScalaCommand[InstallHomeOptions] {
+  override def hidden: Boolean = true
+  def run(options: InstallHomeOptions, args: RemainingArgs): Unit = {
+
+    val binDirPath      = scala.build.Directories.default().binRepoDir
+    val scalaCliBinPath = binDirPath / "scala-cli"
+
+    if (os.exists(scalaCliBinPath)) {
+      if (options.force) os.remove.all(scalaCliBinPath)
+      else if (coursier.paths.Util.useAnsiOutput()) {
```
tpasternak, line 18: EDIT: What if the session is not interactive? I guess Util.useAnsiOutput() covers it, but I'm not sure

alexarchambault: It should. It doesn't have the right name, it's not really about ANSI output. Its value is computed [here](https://github.com/coursier/coursier/blob/f9246587f62020afa152f67dfd14927d57eae02d/modules/paths/src/main/java/coursier/paths/Util.java#L121-L146). It returns false if `System.console() == null` in particular, or on CIs.

PR #140, opened 2021-09-20
Add support for compiler plugins

lwronski:


modules/build/src/main/scala/scala/build/options/ClassPathOptions.scala
```diff
@@ -8,7 +8,8 @@ final case class ClassPathOptions(
   extraCompileOnlyJars: Seq[os.Path] = Nil,
   extraSourceJars: Seq[os.Path] = Nil,
   fetchSources: Option[Boolean] = None,
-  extraDependencies: Seq[AnyDependency] = Nil
+  extraDependencies: Seq[AnyDependency] = Nil,
+  compilerPlugins: Seq[AnyDependency] = Nil
```
alexarchambault: Let's rather put `compilerPlugins` in `ScalaOptions` (along with `scalaVersion`, `scalacOptions`, …).

modules/cli/src/main/scala/scala/cli/commands/SharedDependencyOptions.scala
```diff
@@ -15,7 +15,10 @@ final case class SharedDependencyOptions(
   @HelpMessage("")
   @Name("repo")
   @Name("r")
-    repository: List[String] = Nil
+    repository: List[String] = Nil,
+  @Group("Dependency")
```
alexarchambault: Group `Scala` would be a better fit I believe.

modules/integration/src/test/scala/scala/cli/integration/CompileTests212.scala
```diff
@@ -3,5 +3,59 @@ package scala.cli.integration
...
+        * The above copyright notice and this permission notice shall be included in all copies or
+        * substantial portions of the Software.
+        *
+        * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+        * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+        * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+        * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+        * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+        * DEALINGS IN THE SOFTWARE.
+        *
```
alexarchambault: I'm not sure we need to add this… The few lines below are not a "substantial portions of the Software".

modules/integration/src/test/scala/scala/cli/integration/CompileTests212.scala
```diff
@@ -3,5 +3,59 @@ package scala.cli.integration
...
+        * Originally copied from
+        * (https://github.com/typelevel/kind-projector/blob/main/src/test/scala/polylambda.scala)
```
alexarchambault: Let's put `https://github.com/typelevel/kind-projector/blob/00bf25cef1b7d01d61a3555cccb6cf38fe30e117/src/test/scala/polylambda.scala` instead (got it by pressing Y on the link above, it's less likely to break in the future)

modules/integration/src/test/scala/scala/cli/integration/CompileTests212.scala
lwronski: It's only one fragment from test, it is potential use case of kind-projector library and not "substantial portions of the Software". 

So I deleted the entry of license and added link to the test in kind-project repo.

website/docs/compile.md
```diff
@@ -70,6 +70,13 @@ scala-cli compile Hello.scala -O -deprecation -O -Xlint:infer-any
 
 `-O` accepts both options with the prefixes above and those without such a prefix.
 
+## Scala compiler plugins
+To add compiler plugins dependencis you'd like to use --compiler-plugin:
```
alexarchambault:
```suggestion
Use `--compiler-plugin` to add compiler plugin dependencies:
```

modules/cli/src/main/scala/scala/cli/commands/SharedDependencyOptions.scala
```diff
@@ -15,7 +15,10 @@ final case class SharedDependencyOptions(
   @HelpMessage("")
   @Name("repo")
   @Name("r")
-    repository: List[String] = Nil
+    repository: List[String] = Nil,
+  @Group("Scala")
+  @HelpMessage("Add compiler plugins dependencies")
```
alexarchambault, line 22: Maybe we could add `"plugin"` and `"P"` aliases (via `@Name` annotations), which are more convenient to type than `--compiler-plugin`.

website/docs/compile.md
```diff
@@ -70,6 +70,13 @@ scala-cli compile Hello.scala -O -deprecation -O -Xlint:infer-any
...
+
+```bash
+scala-cli compile Plugin.scala --compiler-plugin org.typelevel:::kind-projector:0.13.2 --scala 2.12.14
```
alexarchambault: NIT (using `Plugin.scala` is confusing IMO, as it's not a plugin itself)
```suggestion
scala-cli compile Main.scala --compiler-plugin org.typelevel:::kind-projector:0.13.2 --scala 2.12.14
```

alexarchambault (commented):
(Only added minor comments.)

website/docs/compile.md
```diff
@@ -70,6 +70,13 @@ scala-cli compile Hello.scala -O -deprecation -O -Xlint:infer-any
 
 `-O` accepts both options with the prefixes above and those without such a prefix.
 
+## Scala compiler plugins
+Use `--compiler-plugin` to add compiler plugin dependencies:
```
alexarchambault, line 74: Seems `--compiler-plugin` was removed. (We should keep it IMO, along with `-P` and `--plugin`.)

alexarchambault (approved):

PR #141, opened 2021-09-20
Resolve dependencies in setup-ide step

tpasternak:
Fail only when dependencies passed via CLI cannot be resolved

Fail when someone passes '--dependency <invalid_dependency>
Do not fail when there is an invalid '$ivy.<dep>' in code

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -148,6 +148,40 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  val importPprintOnlyProject = TestInputs(
+    Seq(
+      os.rel / "simple.sc" -> s"import $$ivy.`com.lihaoyi::pprint:${Constants.pprintVersion}`"
+    )
+  )
+
+  test("setup-ide should succeed for valid dependencies") {
+    importPprintOnlyProject.fromRoot { root =>
+      val p = os.proc(
+        TestUtil.cli,
+        "setup-ide",
+        ".",
+        extraOptions,
+        "--dependency",
+        s"org.scalameta::munit:${Constants.munitVersion}"
+      ).call(cwd = root, check = false)
+      expect(p.exitCode == 0)
+    }
+  }
+
+  test("setup-ide should fail for invalid dependencies") {
+    importPprintOnlyProject.fromRoot { root =>
+      val p = os.proc(
+        TestUtil.cli,
+        "setup-ide",
+        ".",
+        extraOptions,
+        "--dependency",
+        "org.scalameta::munit:0.7.119999"
+      ).call(cwd = root, check = false)
+      expect(p.exitCode == 1)
+    }
+  }
+
   test("simple") {
     val inputs = TestInputs(
       Seq(
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -163,8 +163,7 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
         extraOptions,
         "--dependency",
         s"org.scalameta::munit:${Constants.munitVersion}"
-      ).call(cwd = root, check = false)
-      expect(p.exitCode == 0)
+      ).call(cwd = root)
     }
   }
 
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -156,7 +156,7 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
   test("setup-ide should succeed for valid dependencies") {
     importPprintOnlyProject.fromRoot { root =>
-      val p = os.proc(
+      os.proc(
         TestUtil.cli,
         "setup-ide",
         ".",
@@ -169,6 +169,7 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
   test("setup-ide should fail for invalid dependencies") {
     importPprintOnlyProject.fromRoot { root =>
+      val invalidMunitVersion = "0.7.119999"
       val p = os.proc(
         TestUtil.cli,
         "setup-ide",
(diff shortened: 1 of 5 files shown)


alexarchambault (commented):
Just pushed a commit with minor adjustments (using `options.buildOptions.classPathOptions.extraDependencies` rather than `options.shared.dependencies.dependency` - the latter might have empty strings, which are ignored, while the former doesn't).

Maybe some tests could be added. There's already a `setup-ide` test in [`BspTestDefinitions.scala`](https://github.com/VirtuslabRnD/scala-cli/blob/23571f78d9cbf81316ba57f8c3299204e3b8791e/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala). We could ensure that a faulty dependency on the command-line makes `setup-ide` fails, while the same as an `import $ivy` doesn't.

alexarchambault:
(We could switch to Java 11 if Java 8 is a problem… We already use a Java 11 GraalVM.)

alexarchambault (approved):

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,6 +125,38 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+    )
+    inputs.fromRoot { root =>
+      val p = os.proc(
+        TestUtil.cli,
+        "setup-ide",
+        extraOptions,
+        "--dependency",
+        "com.lihaoyi::cask:0.7.11"
+      ).call(cwd = root, check = false)
+      expect(p.exitCode == 1)
```
alexarchambault: Isn't it supposed to succeed here?
```suggestion
      os.proc(
        TestUtil.cli,
        "setup-ide",
        extraOptions,
        "--dependency",
        "com.lihaoyi::cask:0.7.11"
      ).call(cwd = root)
```

alexarchambault (changes requested):

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
tpasternak: Thanks for finding this, the tests were failing, because in both cases the setup-cli command was failing earlier ("No inputs found"). Now I think it's correct. I also had to adapt to `CrossSources` returning `Either` after rebase.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -149,6 +149,40 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  val importPprintOnlyProject = TestInputs(
+    Seq(
+      os.rel / "simple.sc" -> s"import $$ivy.`com.lihaoyi::pprint:${Constants.pprintVersion}`"
```
alexarchambault, line 153: No need to get this one via constants IMO… It's fine to hardcode dependency versions in the tests.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -149,6 +149,40 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+
+  test("setup-ide should succeed for valid dependencies") {
+    importPprintOnlyProject.fromRoot { root =>
+      val p = os.proc(
+        TestUtil.cli,
+        "setup-ide",
+        ".",
+        extraOptions,
+        "--dependency",
+        s"org.scalameta::munit:${Constants.munitVersion}"
```
alexarchambault, line 165: Same as above, no need to get this version via the build.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -149,6 +149,40 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+      ).call(cwd = root, check = false)
+      expect(p.exitCode == 0)
```
alexarchambault: Just removing `check = false` ought to be enough, no need to check for a `0` exit code.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
tpasternak: I did it to make it easy to track all the dependencies that are used during the build/testing. If we will use a tool to scan vulnerable dependencies in future, it won't find the hardcoded ones

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
tpasternak: Done

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
tpasternak: (but ofc I can revert it if you have a strong opinion)

alexarchambault: There's already many hardcoded dependencies in the tests… 

Let's either revert, or put test dependencies aside from the main ones, in the build.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -149,6 +149,39 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
 
+  val importPprintOnlyProject = TestInputs(
+    Seq(
+      os.rel / "simple.sc" -> s"import $$ivy.`com.lihaoyi::pprint:${Constants.pprintVersion}`"
+    )
+  )
+
+  test("setup-ide should succeed for valid dependencies") {
+    importPprintOnlyProject.fromRoot { root =>
+      val p = os.proc(
```
alexarchambault: NIT
```suggestion
      os.proc(
```

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -149,6 +149,39 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+        TestUtil.cli,
+        "setup-ide",
+        ".",
+        extraOptions,
+        "--dependency",
+        "org.scalameta::munit:0.7.119999"
+      ).call(cwd = root, check = false)
+      expect(p.exitCode == 1)
```
alexarchambault, line 182: By passing `stderr = os.Pipe` to `call`, we can get the errors printed by scala-cli, and ensure that the error originate from a not found dependency and not something else.

alexarchambault (approved):
Let's merge if that's fine for you, @tpasternak.

PR #143, opened 2021-09-21
Remove conf file

alexarchambault:


alexarchambault:
The docs need to be updated too…

PR #144, opened 2021-09-22
Fix - piping input corresponds to virtual scala scripts

lwronski:


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -611,10 +611,21 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
-  if (!Properties.isWin)
+  if (!Properties.isWin) {
     test("piping") {
       piping()
     }
+    test("Piping input should corresponds to scripts") {
```
alexarchambault: NIT
```suggestion
    test("Scripts accepted as piped input") {
```

PR #147, opened 2021-09-23
Fix - replace organization name to 'virtuslab'

lwronski:


alexarchambault (approved):
So no more ghcr.io for docker images?

lwronski:
> So no more ghcr.io for docker images?

Yes, I think we should publish docker image only to [docker hub](https://hub.docker.com/u/virtuslab)

alexarchambault:
Is it good to merge, @lwronski?

PR #148, opened 2021-09-24
Add scalafix, remove unused imports

lwronski:
The final solution should be to run `./mill all __. fix`.  There is a problem that several modules are compiling to scala 3 and in this version there is no flag `-Ywarn-unused `yet. 

If we run `./mill all __. fix`,it tries to run scalafix on all modules in the defined scala version, but there are modules that compile to every scala version, so it throws an error that `-Ywarn-unused` flag not exists in scala 3. 

I haven't found a solution that can easily exclude scalafix from running for modules compiling to scala 3, therefore I created a scalafix () method in which I explicitly run fix() on each module.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,13 @@ jobs:
       run: ./mill -i __.checkNativeImageConfFormat
     - name: Check Ammonite availability
       run: ./mill -i 'dummy.amm[_].resolvedRunIvyDeps.'
+    - name: Scalafix check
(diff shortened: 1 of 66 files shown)


alexarchambault:
I opened https://github.com/VirtusLab/scala-cli/pull/158 with just https://github.com/VirtusLab/scala-cli/pull/148/commits/8cc457c0f688fe7ebf3153c14eeeae0432a285c0. Once it's merged, we can just squash-and-merge the PR here.

PR #150, opened 2021-09-24
Finish export feature

alexarchambault:


modules/integration/src/main/scala/scala/cli/integration/TestInputs.scala
```diff
@@ -39,17 +39,17 @@ object TestInputs {
...
-        setDaemon(true)
-        override def run(): Unit =
-          try os.remove.all(d)
-          catch {
-            case NonFatal(e) =>
-              System.err.println(s"Could not remove $d, ignoring it.")
-          }
-      }
-    )
+    // Runtime.getRuntime.addShutdownHook(
```
romanowski: How we deal with left over tmp dirs now?

alexarchambault: My mistake, I only meant to comment that out locally.

alexarchambault:
(Merging to unblock other PRs…)

PR #152, opened 2021-09-27
Add first cookbook

romanowski:


romanowski:
I wonder what do you think about such form of cookbook? I am adding an automatic checking scripts at this moment.

website/docs/cookbooks/scala-versions.md
```diff
@@ -0,0 +1,71 @@
+# Picking Scala version with scala-cli
+
+Scala cli by default runs latest stable Scala version.
+
+Here is an unversal piece of code that detect Scala version in runtime
+
+```scala
+object ScalaVersion extends App {
```
lwronski, line 8: I run locally this example, it throws me the following error:
```
[error] ./ScalaVersion.scala:7:38: Not found: versions
[error]                                      ^^^^^^^^^
Error compiling project (Scala 3.0.2, JVM)
Compilation failed
```

lwronski (commented):
I think, it's very valuable to create short cookbook presenting the most useful features and describing them.

website/docs/cookbooks/scala-versions.md
romanowski: Yes, that is why I added automated validation as a followup :)

docs_checker/check.scala
```diff
@@ -0,0 +1,122 @@
...
+        else 
+            val toCheck =  Files.list(path).iterator.asScala.filterNot(_.getFileName.toString.startsWith("."))
+            toCheck.toList.flatMap(checkPath)
+    catch
+        case e: Throwable =>
+            e.printStackTrace()
+            Seq(TestCase(path, Some(e)))
+
+def checkFile(file: Path) = 
+    val content = Files.lines(file).iterator.asScala.toList
```
alexarchambault, line 77: NIT Specify a charset just in case? (`StandardCharsets.UTF_8`)

.github/scripts/check_docs.sh
```diff
@@ -0,0 +1,11 @@
...
+
+dest=$(pwd)/.scala/bin
+
+./mill -i copyTo cli.launcher $dest/scala-cli 
+
+export PATH=$dest:$PATH
+echo Adding $dest to classpath
+ls $dest
+
+./mill -i scala docs_checker/check.scala -- docs/cookbooks
```
alexarchambault, line 11: We could make this whole file a Mill command, with something along those lines in `build.sc`:
```scala
def checkDocs(args: String*) = T.command {
  val launcher = cli.launcher().path
  os.proc(launcher, "docs_checker/check.scala", "--", if (args.isEmpty) Seq("docs/cookbooks") else args)
    .call(stdout = os.Inherit)
}
```

Run with:
```text
$ ./mill -i checkDocs
```

Watch mode should work for the Scala CLI launcher itself, when we edit its sources:
```text
$ ./mill -i -w checkDocs
```

With some clever use of Mill `T.sources`, we could even get watch mode to work for the cookbook themselves. Maybe with something like
```scala
def cookbookSources = T.sources {
  Seq(PathRef(os.pwd / "docs" / "cookbooks"))
}
def cookbookScalaSources = T.sources {
  Seq(PathRef(os.pwd / "docs_checker" / "check.scala"))
}
def checkDocs() = T.command {
  val launcher = cli.launcher().path
  val sources = cookbookSources().map(_.path)
  val scalaSources = cookbookScalaSources().map(_.path)
  os.proc(launcher, scalaSources, "--", sources)
    .call(stdout = os.Inherit)
}
```

alexarchambault (commented):
I added a few comments. But they can be addressed or discussed in another PR, if we'd like to get that in sooner than later.

PR #164, opened 2021-09-29
Add support for 'in' clause in require directives

alexarchambault:


alexarchambault:
This is missing glob support (`*` to match file / dire names), but passing a directory to `in` already makes it apply to all files in it and its sub-directories.

PR #165, opened 2021-09-29
Turn cookbook recepies into examples

romanowski:


docs_checker/check.scala
```diff
@@ -58,68 +60,105 @@ def parse(content: Seq[String], currentCommands: Seq[Commands], context: Context
...
+                lines.exists(_.contains(pattern)),
+                msg(s"Pattern: $pattern does not exisits in  any line in:\n$lastOutput")
+              )
+            }
+    }
+  finally if dest.isEmpty then ops.rm(ammPath(out))
+
+  // remove empty space at begining of all files
+  if dest.nonEmpty then
+    val header = s"File was generated from based on ${file}, do not edit manually!"
```
alexarchambault: NIT
```suggestion
    val header = s"This file was generated from $file, do not edit manually!"
```

docs_checker/check.scala
```diff
@@ -1,5 +1,6 @@
 // using scala 3.0.2
 // using "org.scalameta::munit:0.7.29"
+// using com.lihaoyi:ammonite-ops_2.13:2.4.0
```
alexarchambault: Wouldn't [os-lib](https://github.com/com-lihaoyi/os-lib) (`com.lihaoyi::os-lib:0.7.5`) be enough? Ammonite-ops is more or less deprecated IIRC.

romanowski: I just pick a random lib, I will change that

docs_checker/check.scala
```diff
@@ -58,68 +59,118 @@ def parse(content: Seq[String], currentCommands: Seq[Commands], context: Context
...
         Files.list(path).iterator.asScala.filterNot(_.getFileName.toString.startsWith("."))
-      toCheck.toList.flatMap(checkPath)
+      toCheck.toList.flatMap(checkPath(dest))
   catch
     case e: Throwable =>
       e.printStackTrace()
       Seq(TestCase(path, Some(e)))
 
-def checkFile(file: Path) =
+def ammPath(p: Path) = os.Path(p.toAbsolutePath)
```
lwronski: I think, this method can be removed if you migrate from `Java - Paths.*` to `os.lib`.

docs_checker/check.scala
```diff
@@ -58,68 +59,118 @@ def parse(content: Seq[String], currentCommands: Seq[Commands], context: Context
...
+
+val fakeLineMarker = "//fakeline"
+
+def checkFile(file: Path, dest: Option[Path]) =
   val content  = Files.lines(file).iterator.asScala.toList
```
lwronski:
```suggestion
  val content  = os.read.lines(file)
```
but, you have to replace `Path` to `os.Path`

PR #171, opened 2021-09-30
Enable some scalafmt rewrites

alexarchambault:
This enables
- `RedundantBraces`
- `RedundantParens`
- `SortModifiers`

~I'm not too happy with how `RedundantBraces` handles `try`, `finally`, and `yield` blocks, so I'm not sure we want to merge this right now (maybe we can enable only the two other rules first).~ (a few `// format: off` or manual reflowing of the code just do)

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -14,6 +14,16 @@ newlines.alwaysBeforeElseAfterCurlyIf = true
 
 runner.dialect = scala213
 
+rewrite.rules = [
+  RedundantBraces
+]
+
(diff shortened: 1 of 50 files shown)


alexarchambault:
Not sure `DisableSyntax.noIsInstanceOf = true` work fine. There's still one [here](https://github.com/VirtusLab/scala-cli/blob/dce5a890028446218cd69d4f61e0922eaf3fef04/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/internal/Operations.scala#L237) for example.

PR #178, opened 2021-10-01
[wip] Cli args in scripts

tpasternak:


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -211,7 +230,9 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
       )
     )
     inputs.fromRoot { root =>
-      val output = os.proc(TestUtil.cli, extraOptions, "dir", "--main-class", "print").call(cwd =
+      // TODO: It's probably impossible to handle `--main-class print`, because we put main method in the other
```
tpasternak: @alexarchambault @lwronski I couldn't find any way to fix it. This means the scripts users will have to pass `--main-class <somethnig>_sc`, or we will have to add `_sc` suffix silently in options engine

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -203,6 +203,16 @@ object Build {
   def classesDir(root: os.Path, projectName: String): os.Path =
     root / ".scala" / projectName / "classes"
 
+  def scalaNativeSupported(options: BuildOptions, inputs: Inputs): Boolean = {
+    options.scalaOptions.scalaVersion match {
+      case None                            => false
+      case Some(v) if v.startsWith("3")    => false
+      case Some(v) if v.startsWith("2.13") => true
+      case Some(v) if v.startsWith("2.12") =>
+        !inputs.sourceFiles().exists(_.isInstanceOf[Inputs.ScriptBase])
```
alexarchambault: Let's not use `isInstanceOf` (if ever the type returned by `sourceFile` changes, we'll get no compiler error here). Better use a match:
```scala
        !inputs.sourceFiles().exists {
          case _: Inputs.ScriptBase => true
          case _ => false
        }
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -127,6 +127,8 @@ object Inputs {
 
   sealed trait SingleElement extends Element
 
+  sealed trait ScriptBase
```
alexarchambault: `AnyScript` instead of `ScriptBase`? (like the existing `AnyScalaFile`)

modules/build/src/main/scala/scala/build/errors/ScalaNativeCompatibilityError.scala
```diff
@@ -0,0 +1,10 @@
+package scala.build.errors
+
+case object ScalaNativeCompatibilityError
```
alexarchambault: Exceptions shouldn't be case classes or case objects. For case objects, this gives them a wrong stack trace. For case classes, this breaks the case class semantics, see the discussion around https://github.com/scalameta/metals/pull/3145#discussion_r712800065.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -713,6 +713,32 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+  test("interconnection between scripts") {
+    val inputs = TestInputs(
+      Seq(
+        os.rel / "f.sc"     -> "def f(x: String) = println(x + x + x)",
+        os.rel / "main0.sc" -> "f.f(args(0))"
+      )
+    )
+    inputs.fromRoot { root =>
+      val p =
+        os.proc(TestUtil.cli, "main0.sc", "f.sc", "--", "20").call(cwd = root, stdout = os.Pipe)
```
alexarchambault: `stdout = os.Pipe` is the default, no need to specify it.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -713,6 +713,32 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+    }
+  }
+  test("CLI args passed to script") {
+    val inputs = TestInputs(
+      Seq(
+        os.rel / "f.sc" -> "println(args(0))"
+      )
+    )
+    inputs.fromRoot { root =>
+      val p = os.proc(TestUtil.cli, "f.sc", "--", "16").call(cwd = root, stdout = os.Pipe)
```
alexarchambault: Same as above.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -118,10 +118,29 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+        Seq(
+          os.rel / "a.sc" -> "println(1)"
+        )
+      )
+      inputs.fromRoot { root =>
+        val output = os.proc(TestUtil.cli, extraOptions, "--native", "a.sc").call(
+          cwd =
+            root,
+          check = false,
+          stderr = os.Pipe
```
alexarchambault, line 136: Same as above.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -118,10 +118,29 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+    test("Descriptive error message for unsupported native/script configurations") {
+      val inputs = TestInputs(
+        Seq(
+          os.rel / "a.sc" -> "println(1)"
+        )
+      )
+      inputs.fromRoot { root =>
+        val output = os.proc(TestUtil.cli, extraOptions, "--native", "a.sc").call(
+          cwd =
+            root,
```
alexarchambault: Why a new line?

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
alexarchambault: `--main-class print_sc` is fine for me. I wish it could have been `print.sc`, with the right back-ticks added for it to work.

alexarchambault (commented):
I didn't look closely at `MainClassRunner` yet… I'll get back to it a bit later.

modules/build/src/main/scala/scala/build/errors/ScalaNativeCompatibilityError.scala
```diff
@@ -0,0 +1,10 @@
+package scala.build.errors
+
+final class ScalaNativeCompatibilityError
+    extends BuildException(
+      """scala-cli: invalid option: '--native' for scripts is supported only for scala 2.13.*
+        |Please try one of the following combinations:
+        |  scala-cli --native -S 2.12 <...> (for *.sc files)
```
romanowski:
```suggestion
        |  scala-cli --native -S 2.13 <...> (for *.sc files)
```

romanowski (changes requested):
Please include documentation or even better a cookbook

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
tpasternak: I can do it in a separate PR. The problem is that we don't support special characters in --main-class at all. `scala.sc` class is translated to `scala$u002Esc`, but arg passed to `--main-class` remains unchanged. We would have find the code in scala compiler that is used to transform the name and apply this also here. As a quick hack we can just translate some arbitrary symbols (dot, space etc.)

modules/build/src/main/scala/scala/build/errors/ScalaNativeCompatibilityError.scala
tpasternak: For 2.12 we support only .scala files
For 2.13 both .scala and .sc

Hopefully will be fixed in scala-native 4.1

website/docs/cookbooks/scripting.md
```diff
@@ -0,0 +1,28 @@
+# Develop Scripts in Scala
```
lwronski, line 1: Maybe, we should merge scripting cookbook with this [one](https://github.com/VirtusLab/scala-cli/blob/master/website/docs/cookbooks/scala-scripts.md).

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -127,6 +127,8 @@ object Inputs {
 
   sealed trait SingleElement extends Element
 
+  sealed trait AnyScript
```
alexarchambault: NIT Make that extend `Element`, so that we know it's related to it.
```suggestion
  sealed trait AnyScript extends Element
```

website/docs/cookbooks/scripting.md
tpasternak: Can I do it in a follow-up PR?

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -316,40 +310,54 @@ object Build {
...
-    ) { bloopServer =>
-      build(
-        inputs = inputs,
-        options = options,
-        logger = logger,
-        buildClient = buildClient,
-        bloopServer = bloopServer,
-        crossBuilds = crossBuilds
+  ): Either[BuildException, Builds] =
+    if (options.platform == Platform.Native && !scalaNativeSupported(options, inputs))
```
alexarchambault: This check should be done in `buildOnce` rather than here. Some new options may come up from the sources, and can set or change the platform later.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -316,40 +310,54 @@ object Build {
   def classesDir(root: os.Path, projectName: String, scope: Scope): os.Path =
     classesRootDir(root, projectName) / scope.name
 
+  def scalaNativeSupported(options: BuildOptions, inputs: Inputs): Boolean =
+    options.scalaOptions.scalaVersion match {
```
alexarchambault: Getting the Scala version via `options.scalaParams` would give just a `String`, rather than an `Option[String]`, and should be the way to get the Scala version from `BuildOptions`. We may change the default Scala version to a `2.13` one, and `options.scalaParams` would reflect that.

`options.scalaParams` returns an either, but you can use `either` / `value` to deal with it, like other methods in this file.

romanowski (approved):

modules/build/src/main/scala/scala/build/errors/ScalaNativeCompatibilityError.scala
```diff
@@ -0,0 +1,10 @@
+package scala.build.errors
+
+final class ScalaNativeCompatibilityError
+    extends BuildException(
+      """scala-cli: invalid option: '--native' for scripts is supported only for scala 2.13.*
```
alexarchambault, line 5: **For later**: we shouldn't mention command-line options from `BuildException` message. Options are managed from the `cli` module, that the `build` module knows nothing about. And these options might not be available for all commands.

I'd like to add a way in each command to print a specific message if some particular exceptions happen, to suggest options there.

Also, the "name" of the application should be `(new scala.cli.internal.Argv0).get("scala-cli")` rather than just `scala-cli`. The former relies on more idiomatic native stuff. If the `scala-cli` file is renamed to `scala` or `scala-cli-1.0`, its value will be `scala` or `scala-cli-1.0` for example.

We should be be able to handle that if we print specific messages for some exceptions, like I mentioned just above.

modules/build/src/main/scala/scala/build/internal/CustomClassCodeWrapper.scala
```diff
@@ -1,135 +0,0 @@
-package scala.build.internal
```
alexarchambault, line 1: NIT Could this class be removed in a specific commit, doing only that?

modules/build/src/test/scala/scala/build/tests/TestUtil.scala
```diff
@@ -17,7 +17,7 @@ object TestUtil {
     def assertGeneratedEquals(expected: String*): Unit = {
       val generated0 = generated()
       assert(
-        generated0.map(_.toString).toSet == expected.toSet, {
+        expected.toSet == generated0.map(_.toString).toSet, {
```
alexarchambault: This doesn't change anything, does it?

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -107,10 +107,6 @@ final case class SharedOptions(
 
   def logger = logging.logger
 
-  private def codeWrapper: Option[CodeWrapper] =
-    if (classWrap) Some(CustomCodeClassWrapper)
```
alexarchambault, line 111: I think we should remove the `classWrap` option here, and the associated CLI option too if there's one.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -713,6 +713,32 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
       }
     }
   }
+
+  test("interconnection between scripts") {
```
alexarchambault: Why is this in `BspTestDefinitions`? `RunTestDefinitions` would be a better fit for this.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -25,7 +25,7 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
     )
     val launcherName = {
       val ext = if (Properties.isWin) ".bat" else ""
-      fileName.stripSuffix(".sc") + ext
+      fileName.stripSuffix(".sc") + "_sc" + ext
```
alexarchambault: We shouldn't add a `_sc` to the default package name. Maybe it could just be stripped in the `package` command itself?

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -85,7 +86,7 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
              |""".stripMargin
       )
     )
-    val destName = fileName.stripSuffix(".sc") + ".js"
+    val destName = fileName.stripSuffix(".sc") + "_sc.js"
```
alexarchambault: Same as above, no `_sc` in the default output file name.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -125,7 +126,7 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
     )
     val destName = {
       val ext = if (Properties.isWin) ".exe" else ""
-      fileName.stripSuffix(".sc") + ext
+      fileName.stripSuffix(".sc") + "_sc" + ext
```
alexarchambault: Same as above, no `_sc`.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -162,7 +163,7 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
              |""".stripMargin
       )
     )
-    val launcherName = fileName.stripSuffix(".sc") + ".jar"
+    val launcherName = fileName.stripSuffix(".sc") + "_sc.jar"
```
alexarchambault: Same as above, no `_sc`.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -118,10 +118,26 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
     }
+  else
+    test("Descriptive error message for unsupported native/script configurations") {
+      val inputs = TestInputs(
+        Seq(
+          os.rel / "a.sc" -> "println(1)"
+        )
+      )
+      inputs.fromRoot { root =>
+        val output = os.proc(TestUtil.cli, extraOptions, "--native", "a.sc").call(
```
alexarchambault, line 133: Maybe we should just default to `2.13` for Scala Native, and not fail that particular command…

alexarchambault:
I pushed an extra commit getting rid of the `MainClassRunner` and `_trait` stuff, that I think are unnecessary (and make scala-cli generate too many classes for single basic scripts.)

tpasternak:
Will it work for files named `main.sc`? Btw scala compiler escapes dot symbols with $002e, we would have to escape them also in main class detection code

alexarchambault:
> Will it work for files named `main.sc`?

Good point, I think adding the package name when we refer to that object from the `main` method should do.

alexarchambault:
> Btw scala compiler escapes dot symbols with $002e, we would have to escape them also in main class detection code

We can try that later, it might not be trivial overall.

alexarchambault:
> Good point, I think adding the package name when we refer to that object from the main method should do.

Quickly tried my suggestion ^, and it seems it doesn't work when we're in the root package… (but it works fine else)

modules/build/src/main/scala/scala/build/internal/CustomCodeWrapper.scala
```diff
@@ -1,23 +1,49 @@
...
+    val name      = mainClassObject(indexedWrapperName).backticked
+    val traitName = Name(s"${indexedWrapperName.raw}_trait").backticked
+    val mainObjectCode = s"""|
+                             |object $name {
+                             |  private var argsOpt0 = Option.empty[Seq[String]]
+                             |  def setArgs(args: Seq[String]): Unit = {
+                             |    argsOpt0 = Some(args)
+                             |  }
+                             |  def argsOpt: Option[Seq[String]] = argsOpt0
+                             |  def args: Seq[String] = argsOpt.getOrElse {
```
alexarchambault: Let's keep `Array[String]` instead of `Seq[String]` here. We're getting a warning about the `Array` to `Seq` conversion in Scala 3 else.

alexarchambault: About the issue if the script is called `main.sc`: IMO we should just refuse to have a script named `main.sc` at the root like this. It seems we can't refer to other objects at the root with `_root_.ObjectName` (I ran into this limitation too in the `repl` command…). We could have put the scripts in a default package, but this raises a number of other issues (if we have both `.sc` and `.scala` files at the root, they can't reference each other then…)

We should also probably add some `$` signs to the other variables and methods in this object, so that they don't clash with the script name either (like `args.sc` or `argsOpt.sc`).

tpasternak: Ok, but we already have `main.sc` cases in tests and in documentation

We load this gist in one of the tests https://gist.github.com/alexarchambault/7b4ec20c4033690dd750ffd601e540ec

alexarchambault: We should change that then (I just fixed the gist).

romanowski:
It seems that @tpasternak solurion works for now and the only downside is a bit complicated but I am fine with having that in our codebase. 

As for the @alexarchambault concerns: I think we can live with so many generated classes for now.

Let's merge it for now and we can eventually fix it later.

alexarchambault:
Why did you revert my commit? The tests only needed minor changes to pass after it.

alexarchambault:
The multiple classes isn't the only problem of the `MainClassRunner` stuff. It basically creates two instances of the user variables (one in `new MainClassRunner`, one in the object extending the trait where the user code is put), which runs the entry point script twice in particular:
```text
$ cat test-proj/foo.sc
var messages = List.empty[String]
bar.init()
println("Messages:")
for (m <- messages)
  println(m)

$ cat test-proj/bar.sc
def init(): Unit =
  foo.messages = "Hello" :: foo.messages

$ ./mill -i cli.run test-proj -M foo_sc
Messages:
Hello
Messages:
```

tpasternak:
Ok, I reverted it because it did not support main.sc, but you're right my approach was wrong, code cannot be executed twice

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -488,34 +506,36 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
              |""".stripMargin.linesIterator.toVector
       if (exceptionLines != expectedLines) {
-        pprint.log(exceptionLines)
-        pprint.log(expectedLines)
+        println(exceptionLines.mkString("\n"))
+        println(expectedLines)
       }
-      expect(exceptionLines == expectedLines)
+      expect(exceptionLines.length == expectedLines.length)
+      for { i <- Range(0, exceptionLines.length) } expect(exceptionLines(i) == expectedLines(i))
```
alexarchambault, line 538: NIT, more idiomatic way of doing this:
```suggestion
      for (i <- 0 until exceptionLines.length) expect(exceptionLines(i) == expectedLines(i))
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -553,17 +573,18 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
            |Caused by: java.lang.RuntimeException: nope
            |${tab}at scala.sys.package$$.error(package.scala:27)
            |${tab}at throws$$.something(throws.sc:3)
            |${tab}at throws$$.<clinit>(throws.sc:5)
-           |$tab... 1 more
-           |""".stripMargin.linesIterator.toVector
-      expect(exceptionLines == expectedLines)
+           |$tab... 2 more""".stripMargin.linesIterator.toVector
+      expect(exceptionLines.length == expectedLines.length)
+      for { i <- Range(0, exceptionLines.length) } expect(exceptionLines(i) == expectedLines(i))
```
alexarchambault, line 587: NIT, like above
```suggestion
      for (i <- 0 <- exceptionLines.length) expect(exceptionLines(i) == expectedLines(i))
```

alexarchambault (approved):
Looks good to me!

(I only added two NIT comments. Take them into account if you're ok with them, but it's fine merging as is, else…)

PR #182, opened 2021-10-04
Fix CI - head_ref return empty string on master

lwronski:
`github.head_ref` only returns a non-empty value for pull request events, therefore master was [corrupted](https://github.com/VirtusLab/scala-cli/actions/runs/1294281003) by an empty value for the `group` parameter in `concurrency`.

Using github.ref should fix it because it return the branch name or tag ref that triggered the workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


tpasternak (approved):
:+1:

alexarchambault (approved):

PR #185, opened 2021-10-05
Add standalone launcher for Scala CLI - MacOs/Linux

lwronski:


scala-cli.sh
```diff
@@ -0,0 +1,32 @@
...
+  mkdir -p "$(dirname "$CACHE_DEST")"
+  TMP_DEST="$CACHE_DEST.tmp-setup"
+  echo "Downloading $SCALA_CLI_URL"
+  curl -fLo "$TMP_DEST" "$SCALA_CLI_URL"
+  mv "$TMP_DEST" "$CACHE_DEST"
+  gunzip -k $CACHE_DEST
+  chmod +x $SCALA_CLI_BIN_PATH
+fi
+
+exec $SCALA_CLI_BIN_PATH "$@"
```
alexarchambault: Misses quotes around variables?

romanowski (changes requested):
We need a documentation entry, tests and later a cookbook (please create a issue for one)

scala-cli.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -eu
+
+# This script automatically download scala-cli from latest release and cache via coursier
+
+# https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c
+LATATEST_TAG=$(
```
alexarchambault: Typo?
```suggestion
LATEST_TAG=$(
```

scala-cli.sh
```diff
@@ -0,0 +1,32 @@
...
+    curl --silent "https://api.github.com/repos/VirtusLab/scala-cli/releases/latest" | grep '"tag_name":' | sed 's/.*"\(v.*\)",/\1/'
```
alexarchambault: It seems that means every run of this script will run a `curl` command upfront… I don't think we should do that. (If `cs` was installed, we could use `cs get --changing` instead of curl, and benefit from caching. But let's not assume `cs` is installed either.)

We should rather hardcode the latest Scala CLI version here, and automatically update it from the CI upon new releases.

scala-cli.sh
```diff
@@ -0,0 +1,32 @@
...
+)
+
+if [ "$(expr substr $(uname -s) 1 5 2>/dev/null)" == "Linux" ]; then
+  SCALA_CLI_URL="https://github.com/VirtusLab/scala-cli/releases/download/$LATATEST_TAG/scala-cli-x86_64-pc-linux.gz"
+  CACHE_BASE="$HOME/.cache/coursier/v1"
+else
```
alexarchambault: Let's check for `Darwin` too. If it's neither Linux nor Darwin, either exit with an error message, or default to Linux with a warning.

scala-cli.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -eu
+
+# This script automatically download scala-cli from latest release and cache via coursier
```
alexarchambault: Let's add a quick description of what this script is, ideally between the shebang and `set -eu`, like
```bash
#!/usr/bin/env bash

# This is the launcher script of Scala CLI (https://github.com/VirtusLab/scala-cli).
# This script downloads and runs the Scala CLI version set by VERSION below.
#
# Download the latest version of this script at https://github.com/VirtusLab/scala-cli/raw/master/scala-cli.sh


SCALA_CLI_VERSION="0.0.4"

set -eu

```

scala-cli.sh
```diff
@@ -0,0 +1,32 @@
...
+
+LATEST_TAG="0.0.4"
```
alexarchambault: Better use `SCALA_CLI_VERSION`, which is more explicit if users want to manually change it.

build.sc
```diff
@@ -671,6 +670,32 @@ private def commitChanges(name: String, branch: String, repoDir: os.Path): Unit
...
+
+    val branch = "master"
+    val repo   = s"https://oauth2:${ghToken()}@github.com/VirtusLab/scala-cli.git"
+
+    // Cloning
+    gitClone(repo, branch, targetDir)
+    setupGithubRepo(scalaCliDir)
+
+    val launcherScript = os.read(standAloneLauncherPath)
+    val latestTagRegex = "LATEST_TAG=\".*\"".r
```
alexarchambault: SCALA_CLI_VERSION rather than LATEST_TAG?

build.sc
```diff
@@ -671,6 +670,32 @@ private def commitChanges(name: String, branch: String, repoDir: os.Path): Unit
...
+    val updatedLauncherScript =
+      latestTagRegex.replaceFirstIn(launcherScript, s"LATEST_TAG=$version")
```
alexarchambault: Same as above.

alexarchambault (approved):

PR #186, opened 2021-10-05
Add more interactive install-home command

lwronski:
1. If scala-cli is up-to-date, we abort installation.
2. If scala-cli is out-of-date, we automatically update scala-cli 
3. If someone want to downgrade to scala-cli, we ask user to confirm this operation.
4. If someone put `--force` flag, we install scala-cli without any checking, if scala-cli is up-to-date.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
@@ -3,12 +3,7 @@ package scala.cli.commands
 import caseapp._
 
 @HelpMessage("Print details about this application")
-final case class AboutOptions(
-  @Group("About")
-  @Name("v")
-  @HelpMessage("Print only the scala-cli version")
-    version: Boolean = false
-)
+final case class AboutOptions()
 
 object AboutOptions {
   implicit lazy val parser: Parser[AboutOptions] = Parser.derive
--- a/modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
@@ -3,7 +3,12 @@ package scala.cli.commands
 import caseapp._
 
 @HelpMessage("Print details about this application")
-final case class AboutOptions()
+final case class AboutOptions(
+  @Group("About")
+  @Name("v")
+  @HelpMessage("Print only the scala-cli version")
+    version: Boolean = false
+)
 
 object AboutOptions {
   implicit lazy val parser: Parser[AboutOptions] = Parser.derive
--- a/modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
@@ -15,22 +15,21 @@ object InstallHome extends ScalaCommand[InstallHomeOptions] {
     Version(newVersion) > Version(oldVersion)
   }
 
-  private def loggingEqual(version: String) = {
+  private def logEqual(version: String) = {
     System.err.println(
       s"Scala-cli $version is already installed and up-to-date."
     )
     sys.exit(0)
   }
 
-  private def loggingUpdate(env: Boolean, newVersion: String, oldVersion: String) = {
+  private def logUpdate(env: Boolean, newVersion: String, oldVersion: String) =
     if (!env) println(
       s"""scala-cli $oldVersion is already installed and out-of-date.
          |scala-cli will be updated to version $newVersion
          |""".stripMargin
     )
-  }
 
-  private def loggingDowngrade(env: Boolean, newVersion: String, oldVersion: String) = {
+  private def logDowngrade(env: Boolean, newVersion: String, oldVersion: String) =
     if (!env && coursier.paths.Util.useAnsiOutput()) {
       println(s"scala-cli $oldVersion is already installed and up-to-date.")
       println(s"Do you want to downgrade scala-cli to version $newVersion [Y/n]")
@@ -46,7 +45,6 @@ object InstallHome extends ScalaCommand[InstallHomeOptions] {
       )
       sys.exit(1)
     }
-  }
 
   def run(options: InstallHomeOptions, args: RemainingArgs): Unit = {
 
@@ -62,13 +60,12 @@ object InstallHome extends ScalaCommand[InstallHomeOptions] {
       os.proc(binDirPath / options.binaryName, "version").call(cwd = os.pwd).out.text.trim
     }.toOption.getOrElse("0.0.0")
 
-    if (os.exists(binDirPath)) {
+    if (os.exists(binDirPath))
       if (options.force) () // skip logging
-      else if (newVersion == oldVersion) loggingEqual(newVersion)
+      else if (newVersion == oldVersion) logEqual(newVersion)
       else if (isOutOfDate(newVersion, oldVersion))
-        loggingUpdate(options.env, newVersion, oldVersion)
-      else loggingDowngrade(options.env, newVersion, oldVersion)
-    }
+        logUpdate(options.env, newVersion, oldVersion)
+      else logDowngrade(options.env, newVersion, oldVersion)
 
     os.remove.all(binDirPath)
 
@@ -80,9 +77,8 @@ object InstallHome extends ScalaCommand[InstallHomeOptions] {
     if (!Properties.isWin)
       os.perms.set(binDirPath / options.binaryName, os.PermSet.fromString("rwxr-xr-x"))
 
-    if (options.env) {
-      println(s""" export PATH="$$PATH:$binDirPath" """)
-    }
+    if (options.env)
+      println(s"""export PATH="$binDirPath:$$PATH"""")
     else {
 
       val update = EnvironmentUpdate(Nil, Seq("PATH" -> binDirPath.toString()))
--- a/modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
@@ -10,8 +10,42 @@ object InstallHome extends ScalaCommand[InstallHomeOptions] {
   override def hidden: Boolean = true
 
   private def isOutOfDate(newVersion: String, oldVersion: String): Boolean = {
-    val versionOrdering = Ordering.by { (_: String).split("""\.""").map(_.toInt).toIterable }
-    versionOrdering.gt(newVersion, oldVersion)
+    import coursier.core.Version
+
+    Version(newVersion) > Version(oldVersion)
+  }
+
+  private def loggingEqual(version: String) = {
+    System.err.println(
+      s"Scala-cli $version is already installed and up-to-date."
+    )
+    sys.exit(0)
+  }
+
+  private def loggingUpdate(env: Boolean, newVersion: String, oldVersion: String) = {
+    if (!env) println(
+      s"""scala-cli $oldVersion is already installed and out-of-date.
+         |scala-cli will be updated to version $newVersion
+         |""".stripMargin
+    )
+  }
+
+  private def loggingDowngrade(env: Boolean, newVersion: String, oldVersion: String) = {
+    if (!env && coursier.paths.Util.useAnsiOutput()) {
+      println(s"scala-cli $oldVersion is already installed and up-to-date.")
+      println(s"Do you want to downgrade scala-cli to version $newVersion [Y/n]")
+      val response = readLine()
+      if (response != "Y") {
+        System.err.println("Abort")
+        sys.exit(1)
+      }
+    }
+    else {
+      System.err.println(
+        s"Error: scala-cli is already installed $oldVersion and up-to-date. Downgrade to $newVersion pass -f or --force."
+      )
+      sys.exit(1)
+    }
   }
 
   def run(options: InstallHomeOptions, args: RemainingArgs): Unit = {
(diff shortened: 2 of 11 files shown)


modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
```diff
@@ -3,7 +3,12 @@ package scala.cli.commands
 import caseapp._
 
 @HelpMessage("Print details about this application")
-final case class AboutOptions()
+final case class AboutOptions(
+  @Group("About")
+  @Name("v")
+  @HelpMessage("Print only the scala-cli version")
+    version: Boolean = false
```
tpasternak: I'd prefer to name it `--version-only`

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,96 @@ import caseapp._
...
 
 import scala.io.StdIn.readLine
-import scala.util.Properties
+import scala.util.{Properties, Try}
 
 object InstallHome extends ScalaCommand[InstallHomeOptions] {
   override def hidden: Boolean = true
+
+  private def isOutOfDate(newVersion: String, oldVersion: String): Boolean = {
+    val versionOrdering = Ordering.by { (_: String).split("""\.""").map(_.toInt).toIterable }
```
tpasternak: We can use `coursier.core.Version` for this

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
```diff
@@ -0,0 +1,98 @@
...
+      cwd = root,
+      stdin = os.Inherit,
+      stdout = os.Inherit
+    )
+  }
+
+  private def installScalaCli(root: os.Path, binVersion: String, force: Boolean) = {
+    // format: off
+    val cmdInstallVersion = Seq[os.Shellable](
+      TestUtil.cli, "install-home", "--scala-cli-binary-path", binVersion , "--binary-name", dummyScalaCliBinName, "--bin-dir", binDirPath
```
tpasternak: Does this test modify files in user's home? I'd avoid it.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,96 @@ import caseapp._
...
-      to = scalaCliBinPath / "scala-cli",
+      from = newScalaCliBinPath,
+      to = binDirPath / options.binaryName,
       createFolders = true
     )
     if (!Properties.isWin)
-      os.perms.set(scalaCliBinPath / "scala-cli", os.PermSet.fromString("rwxrwxr-x"))
-
-    val update = EnvironmentUpdate(Nil, Seq("PATH" -> scalaCliBinPath.toString()))
+      os.perms.set(binDirPath / options.binaryName, os.PermSet.fromString("rwxrwxr-x"))
```
tpasternak: This adds permissions for group, user and other. Isn't this too much? How about just +x?

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,96 @@ import caseapp._
...
+      else {
+        if (!options.env && coursier.paths.Util.useAnsiOutput()) {
+          println(s"scala-cli $oldVersion is already installed and up-to-date.")
+          println(s"Do you want to downgrade scala-cli to version $newVersion [Y/n]")
+          val response = readLine()
+          if (response != "Y") {
+            System.err.println("Abort")
+            sys.exit(1)
+          }
+          else
```
tpasternak: we could handle incorrect response here and ask again

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,96 @@ import caseapp._
...
-      else if (coursier.paths.Util.useAnsiOutput()) {
-        println(
-          "scala-cli already exists. Do you want to override it [Y/n]"
+    // Backward compatibility - previous versions not have the `--version` parameter
+    val oldVersion: String = Try {
+      os.proc(binDirPath / options.binaryName, "about", "--version").call(cwd =
+        os.pwd).out.text.trim
+    }.toOption.getOrElse("0.0.0")
+
+    if (os.exists(binDirPath)) {
```
tpasternak: Would it be possible to flatten these nested ifs? The logic here is quite sophisticated and it might be easy to get lost in futre

modules/cli/src/main/scala/scala/cli/commands/AboutOptions.scala
alexarchambault: I kind of agree that `scala-cli about --version` looks odd…

Maybe we should just have a `version` sub-command, like `scala-cli version`.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,96 @@ import caseapp._
...
+      if (options.force) os.remove.all(binDirPath)
+      else if (newVersion == oldVersion) {
+        System.err.println(
+          s"Scala-cli $newVersion is already installed and up-to-date."
         )
-        val replace = readLine()
-        if (replace == "Y") {
-          os.remove.all(scalaCliBinPath)
+        sys.exit(1)
```
alexarchambault: Should we really error out here? `sys.exit(0)` would be fine IMO.

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
```diff
@@ -0,0 +1,98 @@
...
+           |  println(\"$firstVersion\")
+           |}""".stripMargin,
+      os.rel / dummyScalaCliSecondName ->
+        s"""
+           |object DummyScalaCli extends App {
+           |  println(\"$secondVersion\")
+           |}""".stripMargin
+    )
+  )
+  val binDirPath = os.home / ".scala" / "scala-cli"
```
alexarchambault: We shouldn't use the local `os.home` in the tests here (this is going to bork any Scala CLI installed there).

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
```diff
@@ -0,0 +1,98 @@
...
+      cwd = root,
+      stdin = os.Inherit,
+      stdout = os.Inherit
+    )
+  }
+
+  if (!Properties.isWin && TestUtil.isCI) {
+    test("updating and downgrading dummy scala-cli using install-home command") {
+
+      testInputs.fromRoot { root =>
```
alexarchambault: NIT: When enabling / disabling tests on specific OSes, other test suites put the body of the test in a method of the test class, and then call that method inside the `if`, like
```scala
def testThing(): Unit = {
  // run the test
}

if (!Properties.isWin && TestUtil.isCI)
  test("test thing") {
    testThing()
  }
```

It makes it easier to enable the test for other OSes or other configs later on, or just scrap the condition with a minimal diff.

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
lwronski: Almost, it modifies `~/.profile`. 
Right, I will change it by using `--env` flag, which enforces, that scala-cli prints only `export $PATH=$PATH/../scala-cli` to updating PATH.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
lwronski: I want to make something similar to `apt` installer. There, if you don't accept operating with passing 'Y', it will abort process and you have to run again. 

Downgrade to scala-cli should be very rarely situation and we should help user to avoid this operation. So I think, that passing any other sign then 'Y' should reject this operation.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,102 @@ import caseapp._
...
-      else {
-        val updater = ProfileUpdater()
-        updater.applyUpdate(update)
-      }
+      os.perms.set(binDirPath / options.binaryName, os.PermSet.fromString("rwxr-xr-x"))
 
-    if (didUpdate) {
-      println("Successfully installed scala-cli")
+    if (options.env) {
+      println(s""" export PATH="$$PATH:$binDirPath" """)
```
alexarchambault: Don't add spaces around the command (NIT), and maybe put the directory first here (it's meant to apply only to the current session, so it should be safe to put our directory first here)
```suggestion
      println(s"""export PATH="$binDirPath:$$PATH"""")
```

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,102 @@ import caseapp._
...
-        System.err.println(
-          s"Error: scala-cli already exists. Pass -f or --force to force erasing it."
-        )
+  private def isOutOfDate(newVersion: String, oldVersion: String): Boolean = {
+    import coursier.core.Version
+
+    Version(newVersion) > Version(oldVersion)
+  }
+
+  private def loggingEqual(version: String) = {
```
alexarchambault: NIT `log` (verb) sounds better than `logging` (name)
```suggestion
  private def logEqual(version: String) = {
```

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,102 @@ import caseapp._
...
+    System.err.println(
+      s"Scala-cli $version is already installed and up-to-date."
+    )
+    sys.exit(0)
+  }
+
+  private def loggingUpdate(env: Boolean, newVersion: String, oldVersion: String) = {
```
alexarchambault: Same as above
```suggestion
  private def logUpdate(env: Boolean, newVersion: String, oldVersion: String) = {
```

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,64 +4,102 @@ import caseapp._
...
+    if (!env) println(
+      s"""scala-cli $oldVersion is already installed and out-of-date.
+         |scala-cli will be updated to version $newVersion
+         |""".stripMargin
+    )
+  }
+
+  private def loggingDowngrade(env: Boolean, newVersion: String, oldVersion: String) = {
```
alexarchambault: Same as above
```suggestion
  private def logDowngrade(env: Boolean, newVersion: String, oldVersion: String) = {
```

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,60 +4,98 @@ import caseapp._
...
     os.copy(
-      from = os.Path(options.scalaCliBinaryPath, os.pwd),
-      to = scalaCliBinPath / "scala-cli",
+      from = newScalaCliBinPath,
+      to = binDirPath / options.binaryName,
       createFolders = true
     )
     if (!Properties.isWin)
-      os.perms.set(scalaCliBinPath / "scala-cli", os.PermSet.fromString("rwxrwxr-x"))
+      os.perms.set(binDirPath / options.binaryName, os.PermSet.fromString("rwxr-xr-x"))
```
tpasternak, line 78: This still may add `r` premissions to `group`. Would `os.PermSet.fromString("u+x")` work here?

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
```diff
@@ -4,60 +4,98 @@ import caseapp._
...
+      if (response != "Y") {
+        System.err.println("Abort")
         sys.exit(1)
       }
+    }
+    else {
+      System.err.println(
+        s"Error: scala-cli is already installed $oldVersion and up-to-date. Downgrade to $newVersion pass -f or --force."
+      )
+      sys.exit(1)
```
tpasternak, line 46: I'd prefer not to put `exit` in a function named `logDowngrade`. Minor issue, can be done in a follow-up PR

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
```diff
@@ -0,0 +1,111 @@
...
+    ) ++ (if(force) Seq[os.Shellable]("--force") else Seq.empty)
+    // format: on
+    os.proc(cmdInstallVersion).call(
+      cwd = root,
+      stdin = os.Inherit,
+      stdout = os.Inherit
+    )
+  }
+
+  def runInstallHome(): Unit = {
```
tpasternak, line 64: Very sophisticated test, took me some time that we build fake scala-cli executables that can print only version and juggle them. Maybe we could comment it (in at all)

tpasternak (commented):
I'd only change the chmod parameter to `u+x` or sth like this. The two other comments may be ignored.

modules/cli/src/main/scala/scala/cli/commands/InstallHome.scala
tpasternak: Nvm, now i see that oslib does not support `+` notation. Approved

tpasternak (approved):

modules/cli/src/main/scala/scala/cli/commands/VersionOptions.scala
```diff
@@ -0,0 +1,11 @@
+package scala.cli.commands
+
+import caseapp._
+
+@HelpMessage("Print scala-cli version")
+final case class VersionOptions()
+
+object VersionOptions {
+  implicit val parser = Parser[VersionOptions]
```
alexarchambault: NIT
Let's add explicit types to these, like done [here](https://github.com/VirtusLab/scala-cli/blob/afe3262cdf1a11de5114e1a192d0dc55de972c66/modules/cli/src/main/scala/scala/cli/commands/RunOptions.scala#L44-L45) (https://github.com/VirtusLab/scala-cli/pull/202 added explicit type annotations to all of those).

alexarchambault (approved):

PR #188, opened 2021-10-05
Setup-ide should run before compile, run and test

MarcinAman:


modules/cli/src/main/scala/scala/cli/commands/BspOptions.scala
```diff
@@ -2,31 +2,21 @@ package scala.cli.commands
...
+import java.nio.file.Path
 
 // format: off
 final case class BspOptions(
   // FIXME There might be too many options in SharedOptions for the bsp command…
   @Recurse
-    shared: SharedOptions = SharedOptions()
+    shared: SharedOptions = SharedOptions(),
+  
+  @Group("Compilation server")
```
alexarchambault: The "Compilation server" group is more for Bloop options. We don't necessarily need a group here…
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/BspOptions.scala
```diff
@@ -2,31 +2,21 @@ package scala.cli.commands
...
+  @HelpMessage("Path to a json file containing BSP configuration")
```
alexarchambault: It's not a "BSP configuration". It's more command-line options.
```suggestion
  @HelpMessage("Command-line options JSON file")
```

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -10,6 +10,7 @@ object Compile extends ScalaCommand[CompileOptions] {
   override def group                                  = "Main"
   override def sharedOptions(options: CompileOptions) = Some(options.shared)
   def run(options: CompileOptions, args: RemainingArgs): Unit = {
+    SetupIde.run(SetupIdeOptions(shared = options.shared), args, previousCommandName = Some(name))
```
alexarchambault: Calling `SetupIde.run` as is is a problem. It calls `sys.exit` sometimes, or prints messages, which we don't want here. Ideally, we'd like the setup-ide stuff to be run silently, not exit under any condition, and only print messages if verbosity is > 0.

`SetupIde.run` should be refactored, so that `SetupIde` exposes another method, doing the setup-ide stuff silently for the other commands.

modules/cli/src/main/scala/scala/cli/commands/CoursierOptions.scala
```diff
@@ -23,6 +24,7 @@ final case class CoursierOptions(
 }
 
 object CoursierOptions {
-  implicit val parser = Parser[CoursierOptions]
-  implicit val help   = Help[CoursierOptions]
+  implicit val parser                  = Parser[CoursierOptions]
+  implicit val help                    = Help[CoursierOptions]
+  implicit val rw: RW[CoursierOptions] = macroRW
```
alexarchambault: NIT: Let's call this… `jsonCodec` maybe? `rw` is too terse and doesn't really say what it is.
```suggestion
  implicit val jsonCodec: RW[CoursierOptions] = macroRW
```

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -37,6 +36,7 @@ object Run extends ScalaCommand[RunOptions] {
       )
 
     val cross = options.compileCross.cross.getOrElse(false)
+    SetupIde.run(SetupIdeOptions(shared = options.shared), args, previousCommandName = Some(name))
```
alexarchambault: Same as above (`SetupIde.run` might exit or print unwanted messages).

modules/cli/src/main/scala/scala/cli/commands/ScalaJsOptions.scala
```diff
@@ -41,6 +42,7 @@ final case class ScalaJsOptions(
 }
 
 object ScalaJsOptions {
-  implicit val parser = Parser[ScalaJsOptions]
-  implicit val help   = Help[ScalaJsOptions]
+  implicit val parser                 = Parser[ScalaJsOptions]
+  implicit val help                   = Help[ScalaJsOptions]
+  implicit val rw: RW[ScalaJsOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
```diff
@@ -53,6 +54,7 @@ final case class ScalaNativeOptions(
 }
 
 object ScalaNativeOptions {
-  implicit val parser = Parser[ScalaNativeOptions]
-  implicit val help   = Help[ScalaNativeOptions]
+  implicit val parser                     = Parser[ScalaNativeOptions]
+  implicit val help                       = Help[ScalaNativeOptions]
+  implicit val rw: RW[ScalaNativeOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/ScalacOptions.scala
```diff
@@ -53,5 +54,6 @@ object ScalacOptions {
         NilParser
     baseParser.to[ScalacOptions]
   }
-  implicit val help = Help[SharedOptions]
+  implicit val help                  = Help[ScalacOptions]
+  implicit val rw: RW[ScalacOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
alexarchambault: Maybe we should pass it our `inputs` instance too, so that it doesn't have to parse them again itself.

modules/cli/src/main/scala/scala/cli/commands/SharedCompilationServerOptions.scala
```diff
@@ -224,8 +225,9 @@ final case class SharedCompilationServerOptions(
 }
 
 object SharedCompilationServerOptions {
-  implicit val parser = Parser[SharedCompilationServerOptions]
-  implicit val help   = Help[SharedCompilationServerOptions]
+  implicit val parser                                 = Parser[SharedCompilationServerOptions]
+  implicit val help                                   = Help[SharedCompilationServerOptions]
+  implicit val rw: RW[SharedCompilationServerOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/SharedDependencyOptions.scala
```diff
@@ -25,6 +26,7 @@ final case class SharedDependencyOptions(
 // format: on
 
 object SharedDependencyOptions {
-  implicit val parser = Parser[SharedDependencyOptions]
-  implicit val help   = Help[SharedDependencyOptions]
+  implicit val parser                          = Parser[SharedDependencyOptions]
+  implicit val help                            = Help[SharedDependencyOptions]
+  implicit val rw: RW[SharedDependencyOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/SharedDirectoriesOptions.scala
```diff
@@ -22,6 +23,7 @@ final case class SharedDirectoriesOptions(
 }
 
 object SharedDirectoriesOptions {
-  implicit val parser = Parser[SharedDirectoriesOptions]
-  implicit val help   = Help[SharedDirectoriesOptions]
+  implicit val parser                           = Parser[SharedDirectoriesOptions]
+  implicit val help                             = Help[SharedDirectoriesOptions]
+  implicit val rw: RW[SharedDirectoriesOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/SharedJvmOptions.scala
```diff
@@ -44,6 +45,7 @@ final case class SharedJvmOptions(
 }
 
 object SharedJvmOptions {
-  implicit val parser = Parser[SharedJvmOptions]
-  implicit val help   = Help[SharedJvmOptions]
+  implicit val parser                   = Parser[SharedJvmOptions]
+  implicit val help                     = Help[SharedJvmOptions]
+  implicit val rw: RW[SharedJvmOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -222,8 +223,9 @@ final case class SharedOptions(
 }
 
 object SharedOptions {
-  implicit val parser = Parser[SharedOptions]
-  implicit val help   = Help[SharedOptions]
+  implicit val parser                = Parser[SharedOptions]
+  implicit val help                  = Help[SharedOptions]
+  implicit val rw: RW[SharedOptions] = macroRW
```
alexarchambault: NIT Same as above (`jsonCodec`).

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -37,51 +45,57 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
-      else withBspCommand
-    }
+    val dir = options.bspDirectory
+      .filter(_.nonEmpty)
+      .map(os.Path(_, Os.pwd))
+      .getOrElse(inputs.workspace / ".bsp")
+
+    val bspName            = options.bspName.map(_.trim).filter(_.nonEmpty).getOrElse("scala-cli")
+    val bspJsonDestination = dir / s"$bspName.json"
+    val scalaCliBspJsonDestination = inputs.workspace / ".scala" / "scala-cli-bsp.json"
```
alexarchambault: `ide-options.json`?

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -37,51 +45,57 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+      else rawArgv(0)
 
     val details = new BspConnectionDetails(
-      name,
-      argv.toList.asJava,
+      bspName,
+      List(
+        progName,
+        "bsp",
+        inputs.workspace.toNIO.toAbsolutePath.toString(),
```
alexarchambault: We should pass things from `args: RemainingArgs` rather than `inputs.workspace`. The inputs may be different (several directories, individual files, …).

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -37,51 +45,57 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+        "--json-options",
+        scalaCliBspJsonDestination.toNIO.toAbsolutePath().toString()
```
alexarchambault: No need to call `.toNIO.toAbsolutePath().toString()`, `os.Path`s are already absolute (and normalized).
```suggestion
        scalaCliBspJsonDestination.toString
```

modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -9,6 +9,7 @@ object Test extends ScalaCommand[TestOptions] {
   override def group                               = "Main"
   override def sharedOptions(options: TestOptions) = Some(options.shared)
   def run(options: TestOptions, args: RemainingArgs): Unit = {
+    SetupIde.run(SetupIdeOptions(shared = options.shared), args, previousCommandName = Some(name))
```
alexarchambault: Same as above (no `SetupIde.run` here).

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,35 +125,70 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     expect(expectedPrefixes.exists(uri.startsWith))
   }
 
-  test("setup-ide") {
```
alexarchambault, line 128: Let's keep the former test as is, just in case.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,35 +125,70 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+    // check that we can decode the connection details
+    val details = upickle.default.read(json)(detailsCodec)
+    f(details)
+  }
+
+  Seq(
+    "setup-ide",
+    "compile",
+    "run"
+  ).foreach(command => {
```
alexarchambault: ?
```suggestion
  for (command <- Seq("setup-ide", "compile", "run"))
```

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,35 +125,70 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
-  }
+  })
 
   val importPprintOnlyProject = TestInputs(
     Seq(
       os.rel / "simple.sc" -> s"import $$ivy.`com.lihaoyi::pprint:${Constants.pprintVersion}`"
     )
   )
 
+  test("setup-ide should have only absolute path if relative one was specified") {
```
alexarchambault: I'm not sure what this tests. Plus the handling of the `os.Path`s looks suspicious (all `os.Path`s are absolute and normalized).

MarcinAman: the issue that i wanted to solve is whenever i run something like `scala-cli setup-ide ../other/folder -S 3.0.1` in the directory that is not within a project then the `scala-cli.json` contains relative path that i posted. This is wrong as the paths (if we want relative one) should be resolved from the workspace not from the place where i run scala-cli. Therefore i thought that it would be safe to always include absolute paths just like we do when path to `scala-cli` is relative.

MarcinAman: To be exact this is what scala-cli outputs: 
```json
{
  "name": "sbt",
  "argv": [
    "scala-cli",
    "bsp",
    "../example/scala3",
    "-S",
    "3.0.1",
    "--bsp-name",
    "sbt"
  ],
  "version": "0.0.4-SNAPSHOT",
  "bspVersion": "2.0.0-M13",
  "languages": [
    "scala",
    "java"
  ]
}⏎       
```

When i run from: `~/Documents/scala-cli` and have an example project in `~/Documents/example/scala3`

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
```diff
@@ -1,33 +1,43 @@
...
       pprint.stderr.log(args)
 
-    val buildOptions     = options.buildOptions
-    val bloopRifleConfig = options.shared.bloopRifleConfig()
-    val logger           = options.shared.logging.logger
+    val sharedOptions: SharedOptions =
+      options.jsonOptions.map { optionsPath =>
+        val source = Source.fromFile(optionsPath.toFile())
+        try read[SharedOptions](source.mkString)
+        finally source.close()
```
alexarchambault: We can use os-lib here, like
```suggestion
        val source = os.read(os.Path(optionsPath, os.pwd))
        read[SharedOptions](source)
```

modules/cli/src/main/scala/scala/cli/commands/BspOptions.scala
```diff
@@ -2,31 +2,20 @@ package scala.cli.commands
...
 final case class BspOptions(
   // FIXME There might be too many options in SharedOptions for the bsp command…
   @Recurse
-    shared: SharedOptions = SharedOptions()
+    shared: SharedOptions = SharedOptions(),
+  
+  @HelpMessage("Command-line options JSON file")
+  @ValueDescription("path")
+  @Hidden
+  jsonOptions: Option[Path] = None
```
alexarchambault: A string is fine here (rather than NIO Path), that we give as input to `os.Path`.

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -40,63 +41,103 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
-
-    val charset = options.charset
-      .map(_.trim)
-      .filter(_.nonEmpty)
-      .map(Charset.forName)
-      .getOrElse(Charset.defaultCharset()) // Should it be UTF-8?
+      // Ensure the path to the CLI is absolute
+      val absolutePathToScalaCli: String = {
+        if (rawArgv(0).contains(File.separator))
+          os.FilePath(rawArgv(0)).resolveFrom(Os.pwd).toString
```
alexarchambault: NIT
```suggestion
          os.Path(rawArgv(0), Os.pwd).toString
```

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -40,63 +41,103 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+        else {
+          val absoluteScalaCliPath =
+            this.getClass.getProtectionDomain.getCodeSource.getLocation.toURI.getPath.trim
```
alexarchambault: I'm not sure this is going to work from the native executables… (it should throw an NPE I think).

Using just `rawArgv(0)` should be fine here ([these](https://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe/1024937#1024937) would be even better, but that can wait…)

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -40,63 +41,103 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
 
-    os.write.over(dest, json.getBytes(charset), createFolders = true)
+      val unparsedArgs =
+        if (args.unparsed.isEmpty)
+          args.unparsed
+        else "--" +: args.unparsed
+      val remainingArgs = args.remaining.map { arg =>
+        if (arg.contains(File.separator))
+          os.FilePath(arg).resolveFrom(Os.pwd).toString
+        else arg
```
alexarchambault: It seems this won't handle file arguments like in `scala-cli setup-ide Foo.scala` or `scala-cli setup-ide ..`. I think we should resolve the path in all cases, like
```suggestion
        os.Path(arg, Os.pwd).toString
```

(This won't handle URIs or `-` - for stdin - fine, but we can deal with those later, by rejecting them I guess.)

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
alexarchambault: (Weird, I thought I had answered this in an earlier review, but I can't see my answers here… Anyway, that was a good point ^^)

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,35 +125,64 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+    importPprintOnlyProject.fromRoot { root =>
+      os.proc(TestUtil.cli, "setup-ide", "./" + root.last, extraOptions).call(
```
alexarchambault: Let's not get out of `root` from the tests. Just put the sources inside a sub-directory, pass the directory name here and use `root` as cwd below.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,35 +125,64 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+        cwd = os.Path(root.toIO.getParentFile().toPath()),
+        stdout = os.Inherit
+      )
+      readBspConfig(
+        root,
+        details =>
+          expect(details.argv.find(_ == root.toIO.getAbsolutePath()).isDefined)
```
alexarchambault: ?
```suggestion
          expect(details.argv.contains(root.toString))
```

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -125,6 +126,17 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     expect(expectedPrefixes.exists(uri.startsWith))
   }
 
+  def readBspConfig(root: os.Path, f: (Details) => Unit): Unit = {
```
alexarchambault: NIT Why not just returning the `Details`? And let the tests below do whatever they want with them?

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -136,24 +148,76 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+          details => {
+            expect(details.argv.length >= 2)
+            expect(details.argv(1) == "bsp")
+            expect(details.argv(3) == "--json-options")
+            expect(Paths.get(details.argv(4)).isAbsolute)
+          }
+        )
+
+        val scalaCliBspConfigFile = root / ".scala" / "ide-options.json"
+        expect(scalaCliBspConfigFile.toIO.exists())
```
alexarchambault: NIT os-lib can do that:
```suggestion
        expect(os.exists(scalaCliBspConfigFile))
```

alexarchambault:
Just pushed a NIT commit, with mostly minor syntax changes or dummy refactorings. Otherwise, LGTM.

alexarchambault:
CI is green 🎉

PR #191, opened 2021-10-06
Better errors

alexarchambault:


modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -78,15 +78,24 @@ final class BspImpl(
       )
     }
 
-    (sources, options0, classesDir0, artifacts, project, generatedSources, buildChanged)
```
romanowski, line 81: Shoudn't we create a case class for this long tuple so parameters etc. are named?

romanowski (approved):

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
alexarchambault: We probably should at that point, yes.

PR #195, opened 2021-10-07
Docs improvements

MarcinAman:
- Fixed some typos
- Bumped docusaurus to newest version because they had issues with scala highlighting (https://github.com/facebook/docusaurus/issues/2564#issuecomment-906623985). I mean they still have but if we have java before scala in additionalLanguages then it works: `additionalLanguages: ['java', 'scala', 'bash'],`
- added some themes. They are up to discussion, let me know if you like them or not

### DIFF
--- a/website/docs/compile.md
+++ b/website/docs/compile.md
@@ -101,7 +101,7 @@ You can add dependencies on the command-line, via `--dependency`:
 scala-cli compile Hello.scala --dependency dev.zio::zio:1.0.9
 ```
 
-Note that `--dependency` is only meant as a convenience. You should favor
+Note that `--dependency` is only meant as a convenience. You should favour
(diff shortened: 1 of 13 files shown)


website/docs/compile.md
```diff
@@ -101,7 +101,7 @@ You can add dependencies on the command-line, via `--dependency`:
 scala-cli compile Hello.scala --dependency dev.zio::zio:1.0.9
 ```
 
-Note that `--dependency` is only meant as a convenience. You should favor
```
alexarchambault, line 104: This is valid US english :)

We have to decide if we'd rather default to British or US english.

alexarchambault (commented):
Apart from the US / UK english point, LGTM 👍

romanowski (approved):

PR #205, opened 2021-10-11
Publish scala-cli on snap in react of release scala-cli

lwronski:


.github/scripts/snapcraft.yaml.template
```diff
@@ -0,0 +1,24 @@
...
+description: |
+  The scala-cli CLI aims at making it easier to run, test, and package, Scala sources.
+  It can:
+    * compile Scala code
+    * run it
+    * package it as a JAR or in formats such as deb, rpm, MSI, …
+    * fire up a REPL allowing you to manually run your code
+    * compile and run tests suites
+base: core18
+grade: stable
```
tpasternak, line 13: I'd put `devel` grade here, and in future add a logic that switches between `devel/stable` (stable only after tag pushes). Is SHOULD_PUBLISH on only for the official releases? If yes, then nvm.

.github/scripts/snapcraft.yaml.template
```diff
@@ -0,0 +1,24 @@
...
+confinement: strict
```
tpasternak: Will it work without defining `interfaces`? If I understand correctly, `strict` means no access at all

.github/scripts/snapcraft.yaml.template
```diff
@@ -0,0 +1,24 @@
+name: scala-cli
+version: "@LAUNCHER_VERSION@"
+summary: ScalaCLI combines all the features you need to learn and use Scala
+description: |
+  The scala-cli CLI aims at making it easier to run, test, and package, Scala sources.
```
tpasternak: @romanowski This will be publicly visible, just fyi

.github/scripts/snapcraft.yaml.template
```diff
@@ -0,0 +1,24 @@
+name: scala-cli
+version: "@LAUNCHER_VERSION@"
+summary: ScalaCLI combines all the features you need to learn and use Scala
```
tpasternak: We use different names ("ScalaCLI", "scala-cli CLI", "scala-cli"). Probably should be unified. Feel free to ignore this comment.

.github/scripts/snapcraft.yaml.template
```diff
@@ -0,0 +1,24 @@
...
+  It can:
+    * compile Scala code
+    * run it
+    * package it as a JAR or in formats such as deb, rpm, MSI, …
+    * fire up a REPL allowing you to manually run your code
+    * compile and run tests suites
+base: core18
+grade: stable
+confinement: strict
+
```
tpasternak, line 15: Missing things like `website`, `issues`, `icon`, `license`, `architectures` etc. Can be done in another PR

.github/scripts/snapcraft.yaml.template
alexarchambault: ?
```suggestion
  Scala CLI aims at making it easier to run, test, and package, Scala sources.
```

.github/scripts/snapcraft.yaml.template
alexarchambault:
```suggestion
summary: Scala CLI combines all the features you need to learn and use Scala
```

build.sc
```diff
@@ -836,6 +836,56 @@ object ci extends Module {
...
+      "docker", "run",
+      "-v", s"$targetDir:/snapcraft",
+      "-w", "/snapcraft",
+      "--env", "SNAPCRAFT_LOGIN_FILE",
+      "snapcore/snapcraft",
+      "bash", "update-snapcraft.sh"
+    )
+    // format: on
+
+    os.proc(dockerCmd).call(cwd = targetDir)
```
tpasternak, line 887: Could we compute sha256 from the snap file and explicitly print it in logs? (Just to make it possible to prove the build comes form GitHub Actions)

build.sc
```diff
@@ -836,6 +836,56 @@ object ci extends Module {
...
+    os.makeDir.all(snapcraftDir)
+
+    val templatePath         = os.pwd / ".github" / "scripts" / "snapcraft.yaml.template"
+    val updateScriptPath     = os.pwd / ".github" / "scripts" / "update-snapcraft.sh"
+    val scalaCliLauncherPath = targetDir / "scala-cli-x86_64-pc-linux.gz"
+    val launcherBinPath      = snapcraftDir / "scala-cli"
+
+    // copy scala-cli launcher to snapcraft directory
+    os.copy(
+      os.Path("artifacts", os.pwd) / "scala-cli-x86_64-pc-linux.gz",
```
alexarchambault: NIT
```suggestion
      os.pwd / "artifacts" / "scala-cli-x86_64-pc-linux.gz",
```

build.sc
```diff
@@ -836,6 +836,56 @@ object ci extends Module {
...
+
+    os.makeDir.all(snapcraftDir)
+
+    val templatePath         = os.pwd / ".github" / "scripts" / "snapcraft.yaml.template"
+    val updateScriptPath     = os.pwd / ".github" / "scripts" / "update-snapcraft.sh"
+    val scalaCliLauncherPath = targetDir / "scala-cli-x86_64-pc-linux.gz"
+    val launcherBinPath      = snapcraftDir / "scala-cli"
+
+    // copy scala-cli launcher to snapcraft directory
+    os.copy(
```
tpasternak, line 856: Maybe I lost it, where we call build? Do we assume `build` has been called before and the publishSnapcraft is not run in a clean git tree?

.github/scripts/snapcraft.yaml.template
romanowski: not a problem :)

.github/scripts/snapcraft.yaml.template
romanowski: Agree. Generally if you name the tool in general please use `Scala CLI` and `scala-cli` if we are naming the specific command.

build.sc
lwronski: Exactly, we run `publishSnapcraft` only on CI, and the assumption is that we need to call the scala-cli launcher build before publishing steps.

build.sc
lwronski: Sure, we can. I think I can do it in a separate PR

.github/scripts/snapcraft.yaml.template
lwronski: In my opinion, for now, we should publish snapcraft package only for the official releases.

alexarchambault (approved):
Don't know for @tpasternak's points, but AFAIC, looks good to me.

tpasternak:
My only concern is classic confinement

PR #206, opened 2021-10-11
Allow shebangs and using directives in scripts

tpasternak:
Previously, shebangs were replaced whith empty strings what resulted
with whitespace lines before the `using` directive. Now, we ignore
whitespaces in the beginning and in the end while running scripts.


lwronski:
I think, this may change the content of the exception. The line number where the error occurred will be shifted by one or more.

alexarchambault:
I think https://github.com/VirtusLab/scala-cli/pull/200/commits/6fc273fa7460ac5433f46fa9b725e752b232915a should fix that issue. (I mean, it should fix the overall issue with empty lines / shebang at the beginning of files.)

It makes the current `using` directives parser just accept those empty lines.

PR #212, opened 2021-10-13
Build scala-cli with jvm 11

tpasternak:
Let's check what happens

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -33,12 +33,16 @@ implicit def millModuleBasePath: define.BasePath =
 object cli            extends Cli
 object `build-macros` extends Cross[BuildMacros](Scala.defaultInternal)
 object build          extends Cross[Build](Scala.defaultInternal)
-object stubs          extends JavaModule with ScalaCliPublishModule
 object runner         extends Cross[Runner](Scala.all: _*)
 object `test-runner`  extends Cross[TestRunner](Scala.all: _*)
 object `bloop-rifle`  extends Cross[BloopRifle](Scala.allScala2: _*)
 object `tasty-lib`    extends Cross[TastyLib](Scala.all: _*)
 
+object stubs extends JavaModule with ScalaCliPublishModule {
+  def javacOptions = T {
+    super.javacOptions() ++ Seq("-target", "8", "-source", "8")
+  }
+}
 object integration extends Module {
   object docker extends CliIntegrationDocker {
     object test extends Tests {
@@ -529,7 +533,7 @@ class TastyLib(val crossScalaVersion: String) extends CrossSbtModule with ScalaC
     super.scalacOptions() ++ {
       if (scalaVersion().startsWith("2.")) Seq("-Ywarn-unused")
       else Nil
-    } ++ Seq("-release", "8")
+    }
   )
 
 }
--- a/build.sc
+++ b/build.sc
@@ -435,7 +435,8 @@ class Runner(val crossScalaVersion: String) extends CrossSbtModule with ScalaCli
     super.scalacOptions() ++ {
       if (scalaVersion().startsWith("2.")) Seq("-Ywarn-unused")
       else Nil
-    }
+    } ++ Seq("-release", "8")
+
   }
   def mainClass = Some("scala.cli.runner.Runner")
   def ivyDeps =
@@ -474,7 +475,7 @@ class TestRunner(val crossScalaVersion: String) extends CrossSbtModule with Scal
     super.scalacOptions() ++ {
       if (scalaVersion().startsWith("2.")) Seq("-Ywarn-unused", "-deprecation")
       else Nil
-    }
+    } ++ Seq("-release", "8")
   }
   def ivyDeps = super.ivyDeps() ++ Agg(
     Deps.asm,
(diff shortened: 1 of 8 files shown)


alexarchambault (commented):
We need to make sure that the `runner`, `test-runner`, and `stub` modules are still fine with Java 8, by passing them the right javac and scalac options (see [this SO question](https://stackoverflow.com/questions/25926111/how-to-force-sbt-to-use-java-8), I think we need both its javac and scalac options). These are the modules that are sometimes added to the user class path by Scala CLI.

To be sure it works, we can add a simple test in `RunTestDefinitions` say, that passes `--jvm 8` to `scala-cli`.

MarcinAman:
Wouldn't upgrading to latest LTS (JDK 17) make sense? Afaik this is quite a problematic release as oracle decided to encapsulate some of its internals therefore some tooling may not be compatible. So if it works on JDK 17 it will probably run on older JDK-s

alexarchambault:
For now, we're using the Java 11 version of GraalVM to build the native launcher, so switching to Java 11 isn't a problem. Maybe the upcoming GraalVM version will have Java 17 versions, and allow to switch to Java 17 then.

tpasternak:
Graal 17 is out https://phoronix.com/scan.php?page=news_item&px=GraalVM-21.3-Released

build.sc
```diff
@@ -524,12 +525,13 @@ class BloopRifle(val crossScalaVersion: String) extends CrossSbtModule with Scal
...
 class TastyLib(val crossScalaVersion: String) extends CrossSbtModule with ScalaCliPublishModule
     with ScalaCliScalafixModule {
-  def scalacOptions = T {
+  def scalacOptions = T(
     super.scalacOptions() ++ {
       if (scalaVersion().startsWith("2.")) Seq("-Ywarn-unused")
       else Nil
-    }
-  }
+    } ++ Seq("-release", "8")
```
alexarchambault: I think that should have been added to the `stubs` module. This one doesn't need it.

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -214,7 +214,7 @@ class NativePackagerTests extends munit.FunSuite {
       // format: off
       val cmd = Seq[os.Shellable](
         TestUtil.cli,
-        "package", helloWorldFileName,
+        "package", "--jvm", "8", helloWorldFileName,
```
alexarchambault: Maybe we should just change the base image in [`Package.scala`](https://github.com/VirtusLab/scala-cli/blob/a95f8925f636cc8bb4efadeae9c0c7942c85465f/modules/cli/src/main/scala/scala/cli/commands/Package.scala#L355).

modules/integration/src/test/scala/scala/cli/integration/TestTestDefinitions.scala
```diff
@@ -509,5 +509,4 @@ abstract class TestTestDefinitions(val scalaVersionOpt: Option[String])
         expect(output.contains("Hello from native"))
     }
   }
-
```
alexarchambault, line 512: It'd be nice to check that tests run fine with `--jvm 8` too (this should ensure that `test-runner` is fine on Java 8).

website/docs/commands/package.md
```diff
@@ -114,7 +114,7 @@ The docker image name parameter `--docker-image-repository` is mandatory.
 The following command generates a `hello-docker` image with the `latest` tag:
 
 ```bash
-scala-cli package --docker HelloDocker.scala --docker-image-repository hello-docker
+scala-cli package --jvm 8 --docker HelloDocker.scala --docker-image-repository hello-docker
```
alexarchambault: Same as above, this shouldn't be needed with a change of base image.

build.sc
tpasternak: Done

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
tpasternak: Done, but this does not fully resolve the problem. I opened an issue #413

modules/integration/src/test/scala/scala/cli/integration/TestTestDefinitions.scala
tpasternak: Done

website/docs/cookbooks/scala-docker.md
```diff
@@ -18,7 +18,7 @@ Passing `--docker` to the `package` sub-command generates a Docker image. When c
 The following command generates a `hello-docker` image with the `latest` tag:
 
 ```bash
-scala-cli package --docker HelloDocker.scala --docker-image-repository hello-docker
+scala-cli package --jvm 8 --docker HelloDocker.scala --docker-image-repository hello-docker
```
alexarchambault: Do we really need to pass that option here? It should work fine without it (doesn't it?)

website/docs/cookbooks/scala-docker.md
```diff
@@ -51,12 +51,9 @@ Built docker image, run it with
 This command creates a Scala Native Docker image:
 
 ```bash ignore
-scala-cli package --native --docker HelloDocker.scala --docker-image-repository hello-docker
+scala-cli package --native --jvm 8 --docker HelloDocker.scala --docker-image-repository hello-docker
```
alexarchambault: Same as above.

alexarchambault (approved):

PR #214, opened 2021-10-13
Download inputs each time

tpasternak:
Not sure yet, but looks like this change causes:
1. Inputs are downloaded each time
2. If download fails, coursier gets input from cache


alexarchambault (approved):
It'd be nice to have an option like `--cache-downloads`, that would just not call `withTtl`. But I guess that can be done in a later PR.

tpasternak:
Hey, I was going to do it, but I noticed it works in that way if there's no internet connection

alexarchambault:
Merging, seems the CI error is transient (and a bit worrying), but unrelated to these changes.

PR #215, opened 2021-10-13
Add cookbook for docker

lwronski:


website/docs/cookbooks/scala-docker.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Package Scala application as a docker image
+sidebar_position: 5
+---
+
+Use `scala-cli` to package your application into a docker image. ScalaCLI automatically creates executable application and package them into docker image.
```
romanowski:
```suggestion
# Package Scala application as a docker image

ScalaCLI can create an executable application and package it into a docker image.
```

website/docs/cookbooks/scala-docker.md
```diff
@@ -0,0 +1,37 @@
...
+    println("Hello from Docker")
+}
+```
+
+
+Passing `--docker` to the `package` sub-command generates a docker image. The docker image name parameter `--docker-image-repository` is mandatory.
+
+The following command will generate `hello-docker` image with `latest` tag:
+```bash ignore
+scala-cli package --docker HelloWorld.scala --docker-image-repository hello-docker
```
romanowski: Why do we need to ignore this commands?

lwronski: You're right, we should check these code snippets with the docs-checker. I fixed it. 

I only skip checking for building docker image with scala-native, because for now, we support it only on Linux environments.  It would be a pain if docs-checker failed every time on macOS.

examples/scala-docker/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```scala title=HelloDocker.scala
+object HelloDocker extends App {
+    println("Hello from Docker")
+}
+```
+
+Passing `--docker` to the `package` sub-command generates a docker image. The docker image name parameter `--docker-image-repository` is mandatory.
+
+The following command will generate `hello-docker` image with `latest` tag:
```
romanowski, line 20: Could one configure custom tag? If so I think we should add it to the guide

romanowski (approved):

PR #217, opened 2021-10-14
Bloop compile to correct JVM version

tpasternak:


tpasternak:
I'd like to somehow cleaup and organize this code now. I have no idea what's the right way. 

Should I just extend `acceptBloopVersion` field function in `BloopRifleScala` in that way, it checks also bloop's jvm and restarts it if it's invalid? 

Should I download a newer jvm via couriser lib and spawn bloop with it in that case in `Operations.startServer`?

alexarchambault:
> Should I just extend acceptBloopVersion field function in BloopRifleScala in that way, it checks also bloop's jvm and restarts it if it's invalid?

Something along those lines, yes. Maybe by adding a field like `acceptJvm: Option[String => Boolean]`, that would return `true` if the passed JVM version is valid, or `false` to switch JVM (switching to using the java command provided by `BloopRifleConfig.javaPath`, so there should be no new need to initiate downloading a JVM, it's already done in `SharedOptions.bloopRifleConfig()`).

tpasternak:
Still known limitation: no way to enforce JVM 
If you set JVM 11, it is guaranteed that the bloop will be able to compile Java 11 API code, but there's no way to strictly set JVM version/path

modules/bloop-rifle/src/main/scala/scala/build/bloop/BloopServer.scala
```diff
@@ -41,27 +52,57 @@ object BloopServer {
...
+    lazy val bloopInfoJavaPath = bloopInfo.map(i => Paths.get(i.javaHome, "bin", "java").toString())
+
+    val expectedBloopJava =
+      if (bloopJvmIsOk) bloopInfoJavaPath.getOrElse(
+        throw new RuntimeException("Fatal error, could not infer bloop java path")
+      )
+      else config.javaPath
+
+    if (isRunning) {
+      BloopRifle.exit(config, workdir, logger)
```
romanowski: We need to make sure that information that bloop is restarted will be shown to user

modules/bloop-rifle/src/main/scala/scala/build/bloop/VersionOps.scala
```diff
@@ -0,0 +1,24 @@
+package scala.build.bloop
+
+object VersionOps {
+  implicit class VersionOps_(lhs: String) {
+    def toInt(s: String): Option[Int] =
```
romanowski:
```suggestion
    private def toInt(s: String): Option[Int] =
```

modules/bloop-rifle/src/main/scala/scala/build/bloop/VersionOps.scala
```diff
@@ -0,0 +1,24 @@
...
+      try Some(Integer.parseInt(s))
+      catch {
+        case _: NumberFormatException => None
+      }
+
+    def isNewerThan(rhs: String): Boolean = {
```
romanowski: I think we should support a comparison between `1.1.1` and `1.1.2-SNAPSHOT`.  In current implementation `1.1.1` is newer

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
```diff
@@ -188,53 +176,58 @@ object BloopRifle {
 
   def nullInputStream() = new FileInputStream(Util.devNull)
 
-  def extractVersionFromBloopAbout(stdoutFromBloopAbout: String): Option[String] =
-    stdoutFromBloopAbout.split("\n").find(_.startsWith("bloop v")).map(
+  // Probably we should implement an endpoint in Bloop to get this
+  // information in better form. I'm not sure error here should be escalated or ignored.
+  private def extractVersionFromBloopAbout(stdoutFromBloopAbout: String)
+    : Option[BloopServerRuntimeInfo] = {
+    val bloopVersion = stdoutFromBloopAbout.split("\n").find(_.startsWith("bloop v")).map(
```
romanowski:
```suggestion
    val bloopVersion = stdoutFromBloopAbout.linesIterator.find(_.startsWith("bloop v")).map(
```

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
```diff
@@ -188,53 +176,58 @@ object BloopRifle {
...
       _.split(" ")(1).trim().drop(1)
```
romanowski: Could we add a comment with example output from bloop? Also, Woundn't `_.stripPrefix("bloop v").split(" ").head.trim` works?

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
```diff
@@ -188,53 +176,58 @@ object BloopRifle {
...
-    )
+    ).getOrElse(throw new RuntimeException("Could not extract bloop version"))
+    val bloopJvmLine = stdoutFromBloopAbout.split("\n")
+      .find(_.startsWith("Running on Java JDK"))
```
romanowski: Can we migrate to regexes?

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
```diff
@@ -188,53 +176,58 @@ object BloopRifle {
...
-  def extractVersionFromBloopAbout(stdoutFromBloopAbout: String): Option[String] =
-    stdoutFromBloopAbout.split("\n").find(_.startsWith("bloop v")).map(
+  // Probably we should implement an endpoint in Bloop to get this
+  // information in better form. I'm not sure error here should be escalated or ignored.
+  private def extractVersionFromBloopAbout(stdoutFromBloopAbout: String)
+    : Option[BloopServerRuntimeInfo] = {
+    val bloopVersion = stdoutFromBloopAbout.split("\n").find(_.startsWith("bloop v")).map(
       _.split(" ")(1).trim().drop(1)
-    )
+    ).getOrElse(throw new RuntimeException("Could not extract bloop version"))
```
romanowski: We should not throw exceptions when we return an option. We can consider using `Either`

project/deps.sc
```diff
@@ -64,29 +64,29 @@ object Deps {
...
-  def shapeless                  = ivy"com.chuusai::shapeless:2.3.7"
-  def slf4jNop                   = ivy"org.slf4j:slf4j-nop:1.8.0-beta4"
-  def snailgun                   = ivy"me.vican.jorge::snailgun-core:0.4.0"
-  def svm                        = ivy"org.graalvm.nativeimage:svm:$graalVmVersion"
-  def svmSubs                    = ivy"org.scalameta::svm-subs:20.2.0"
-  def swoval                     = ivy"com.swoval:file-tree-views:2.1.5"
-  def testInterface              = ivy"org.scala-sbt:test-interface:1.0"
-  def upickle                    = ivy"com.lihaoyi::upickle:1.3.8"
-  def usingDirectives            = ivy"org.virtuslab:using_directives:0.0.5-a86b6cb-SNAPSHOT"
+  def scala3Compiler(sv: String) = ivy"org.scala-lang::scala3-compiler:$sv".exclude("*" -> "*")
```
romanowski: Could we add a comment why we need those excludes?

romanowski (changes requested):

romanowski (approved):

PR #221, opened 2021-10-15
Add support for running and debugging main classes in Metals

lwronski:
We added support for [Debugging Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/). It allows to run and debug main classes in scala-cli project in VSCode. 

There is still missing support for test classes, it will be added in next PR.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
@@ -10,10 +10,11 @@ import scala.concurrent.{Future, Promise}
 import scala.jdk.CollectionConverters._
 
 class BspServer(
-  bloopServer: b.BuildServer with b.ScalaBuildServer with b.JavaBuildServer,
+  bloopServer: b.BuildServer with b.ScalaBuildServer with b.JavaBuildServer with ScalaDebugServer,
   compile: (() => CompletableFuture[b.CompileResult]) => CompletableFuture[b.CompileResult],
   logger: Logger
 ) extends b.BuildServer with b.ScalaBuildServer with b.JavaBuildServer with BuildServerForwardStubs
+    with ScalaDebugServerForwardStubs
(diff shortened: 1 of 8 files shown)


tpasternak (approved):
Would be nice to have some tests here

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -163,8 +164,9 @@ class BspServer(
       for (target <- res0.getTargets.asScala) {
         val capabilities = target.getCapabilities
         // TODO Re-enable once we support this via BSP
-        capabilities.setCanRun(false)
-        capabilities.setCanTest(false)
+        capabilities.setCanDebug(true)
```
alexarchambault, line 167: Wouldn't just scrapping the `for (target <- res0.getTargets.asScala)` block just work here? That way, that would leave the capabilities returned by Bloop, which handles running / testing / debugging things itself IIUC.

alexarchambault (commented):
Nice. Long term, we might want to handle running or testing things ourselves, rather than delegating things to Bloop, so that there's no discrepancy between what happens via BSP and via the CLI.

alexarchambault:
Merging, my comment can be addressed later if needed.

kpodsiad:
Keep in mind that it needs https://github.com/scalameta/metals/pull/3210 to work correctly.

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
lwronski: Now, that won't work because in `bloop` is not set `canDebug` flag. Users have to explicitly set `capabilities.setCanDebug` to `true`.

kpodsiad: I've created an issue in bloop - https://github.com/scalacenter/bloop/issues/1584. Probably, will do it soon ;)

PR #222, opened 2021-10-15
Introduce defaults to os in install and shell

MarcinAman:
keep in mind that if you would like to test those locally you need to clear values from local storage as docsaurus remembers what you previously have chosen.


alexarchambault (approved):
👍

PR #223, opened 2021-10-15
Pass default dialect to scalafmt

alexarchambault:
So that 'scala fmt .' works fine out-of-the-box for Scala 3 sources.


alexarchambault:
CI is actually green, not sure why the jvm-tests job appears running.

PR #224, opened 2021-10-15
Update internals doc page

alexarchambault:


alexarchambault:
Merging, this is better than the previous page in any case.

PR #225, opened 2021-10-15
Update pretty-stacktraces to 0.0.1

alexarchambault:


alexarchambault:
Seems pretty-stacktraces is built with an unstable Scala version, that `3.0.2` isn't fine with…

alexarchambault:
Closing this for now, but it would still be great to use a non-snapshot version of pretty-stacktraces.

PR #226, opened 2021-10-15
Docs/update

romanowski:
Update and rehaul all of our docs. This PR will require multiple follow ups


modules/build/src/main/scala/scala/build/preprocessing/directives/UsingDependencyDirectiveHandler.scala
```diff
@@ -9,29 +9,35 @@ import scala.build.errors.{BuildException, DependencyFormatError}
...
     directive.values match {
-      case Seq(depStr) if depStr.split(":").count(_.trim.nonEmpty) == 3 =>
-        val res =
-          parseDependency(depStr).map { dep =>
-            BuildOptions(
-              classPathOptions = ClassPathOptions(
-                extraDependencies = Seq(Positioned(Seq(directive.position), dep))
-              )
-            )
+      case Seq("lib" | "libs", depStr @ _*) =>
```
lwronski, line 25: I think it allows to use `lib` and `libs` in two different way. The following example should work and it could be a confusing

`using libs org:name:ver | using lib org:name:ver org2:name2:ver2`

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -88,11 +88,18 @@ object CrossSources {
...
-        .flatMap(_.valueFor(path).toSeq)
-        .foldLeft(BuildRequirements())(_ orElse _)
+    def baseReqs(path: ScopePath): BuildRequirements = {
+      val fromDirectives =
+        scopedRequirementsByRoot
+          .getOrElse(path.root, Nil)
+          .flatMap(_.valueFor(path).toSeq)
+          .foldLeft(BuildRequirements())(_ orElse _)
+
+        // If no scope if defined in file and file ends with `.test.scala` treat it as test
```
lwronski: I think, it should be more descriptive, why this one `if` (`path.path.last.endsWith(".test.scala")`) is needed

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingDependencyDirectiveHandler.scala
romanowski: We are using this pattern across pretty much all other settings - and I like it. 

Is it confusing? I do not thing so

website/docs/17-reference/03-commands.md
```diff
@@ -0,0 +1,253 @@
...
+---
+
+## `about`
+
+Print details about this application
+
+## `bsp`
+
+Accepts options:
+- [bsp](/docs/reference/cli-options#bsp-options)
```
tpasternak: Do we check if these links are correct?

website/docs/14-guides/24-using-directives.md
```diff
@@ -0,0 +1,56 @@
...
+
+Using directives can be defined using special keyword `using` (however it may not be compatible with exisiting tools outside Scala CLI), using them within comments `// using scala 2` or as a special top-level annotation `@using jars libs/model.jar`.
+
+:::info
+For now we recommend using comment-flavor (e.g. `// using scala 3.0.2`) and we will stick it in this guide. 
+
+Until using directives becomes a part of Scala specification, this is the only way that guarantee that your code will work well with IDE, formatters or similar tool.
+:::
+
+Using directives can be declared **before any other scala code**.
```
tpasternak: declared *only before..*

website/docs/14-guides/23-dependencies.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Dependencies
+
+You can add dependencies on the command-line, via `--dependency`:
+```bash
+scala-cli compile Hello.scala --dependency dev.zio::zio:1.0.9
+```
+
+Note that `--dependency` is only meant as a convenience. You should favour
+adding dependencies in the sources themselves via [using directives](/docs/guides/using-directives) or [special imports](/docs/guides/configuration#special-imports).
```
tpasternak: "However, `--dependency` CLI option takes precedence over `using` directives, so it can be used to override version temporarily

website/docs/14-guides/24-using-directives.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Using directives
+---
+
+:::warning
+Using directives are experimental and may change in future versions of Scala CLI
```
tpasternak: How about "Using directives is an experimental language extension that may change ...

website/docs/10-commands/05-compile.md
```diff
@@ -34,6 +33,41 @@ scala-cli compile --watch Hello.scala
...
 
+## Scala version
+
+Scala CLI uses latest stable version of Scala by default (`3.0.2` when this guide was written). You can specify the Scala version you'd like to use with `--scala`:
+
+```bash ignore
+scala-cli compile --scala 2.13.6 Hello.scala
+```
+
+`scala-cli` should work with all major `2.12.x`, `2.13.x`, and `3.x` Scala versions.
```
tpasternak: Do we want to note here, that native is supported only in 2.13?

website/docs/14-guides/21-scripts.md
```diff
@@ -0,0 +1,103 @@
+---
+title: Scripts
+---
+
+# Scripts
+
+`scala-cli` accept Scala scripts, ending in `.sc`. Unlike `.scala` files,
```
tpasternak: Maybe we could explain what's the difference between scripts and scala files (so in scripts it is possible to write top-level expressions)

tpasternak (approved):

website/docs/17-reference/03-commands.md
romanowski: I've rewrited our .md files to link to .md files and Docusaurs checks that

website/docs/10-commands/05-compile.md
romanowski: I've added it to Scala Native section.

PR #231, opened 2021-10-18
Gists cookbook

KacperFKorban:
closes https://github.com/VirtuslabRnD/scala-cli-features/issues/73

Pretty short, but not much to write about here.

### DIFF
--- a/website/docs/cookbooks/gists.md
+++ b/website/docs/cookbooks/gists.md
@@ -1,0 +1,47 @@
+---
+title: Sharing and testing code with Github Gists
+sidebar_position: 6
+---
+
+# Sharing and testing code with Github Gists
+
+## Running code from Gists
+
+`scala-cli` allows running Scala code straight from Github Gists, without the need for manually downloading them first.
+This is done by just passing the link to a Gist as an argument to `scala-cli`.
+
+For example, having a given gist `https://gist.github.com/alexarchambault/7b4ec20c4033690dd750ffd601e540ec`:
+```scala title=Messages.scala
+object Messages {
+  def hello = "Hello"
+}
+```
+```scala title=run.sc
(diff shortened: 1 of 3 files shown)


website/docs/cookbooks/gists.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Sharing and testing code with Github Gists
+
+## Running code from Gists
+
+`scala-cli` allows running Scala code straight from Github Gists, without the need for manually downloading them first.
+This is done by just passing the link to a Gist as an argument to `scala-cli`.
+
+For example, having a given gist `https://gist.github.com/alexarchambault/7b4ec20c4033690dd750ffd601e540ec`:
+```scala
```
MarcinAman: i think you can use something like `scala title=Messages.scala` to indicate the name of the file

MarcinAman (commented):
Maybe it would be nice to be open about the fact that you can't edit this gist but this will be possible in future versions. This could possibly save some confusions

website/docs/cookbooks/gists.md
romanowski: ```scala title=Message.scala```

and ```bash``` are used by our tooling to validate if cookbooks are valid. See more here https://github.com/VirtusLab/scala-cli/tree/master/docs_checker

KacperFKorban: ok

KacperFKorban:
@MarcinAman I think that it is implied that one cannot edit them. And about being able to do os in the future, IMHO it is not a good idea.
It makes this another place that will need to be updated. If one forgets about it, it's not that big a deal. But if it explicitly says that it is not supported, when it already is- that brings some confusion.
WDYT?

romanowski (approved):

PR #234, opened 2021-10-18
Post processors can now report warning.

romanowski:
We report such warnings and skip processing of TASTY files when we face a Scala Version with newer Tasty than we know we support.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -138,6 +138,18 @@ final class BspImpl(
           client.resetBuildExceptionDiagnostics(targetId)
     }
 
+  private val shownGlobalMessages =
+    new java.util.concurrent.ConcurrentHashMap[String, Unit]()
+
+  private def showGlobalWarningOnce(msg: String) =
+    shownGlobalMessages.computeIfAbsent(
+      msg,
+      _ => {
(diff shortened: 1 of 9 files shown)


modules/tasty-lib/src/main/scala/scala/build/tastylib/TastyVersions.scala
```diff
@@ -0,0 +1,32 @@
...
+  object LatestSupportedScala {
+    final val MajorVersion: Int = 3
+    final val MinorVersion: Int = 1
+  }
+
+  def shouldRunPreprocessor(
+    scalaVersion: String,
+    scalaCliVersion: String
+  ): Either[String, Boolean] =
+    if (!scalaVersion.startsWith("3.")) Right(false)
```
tpasternak: Not sure if this will handle `using scala 3`

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -138,6 +138,18 @@ final class BspImpl(
           client.resetBuildExceptionDiagnostics(targetId)
     }
 
+  private val shownGlobalMessages =
+    new java.util.concurrent.ConcurrentHashMap[String, Unit]()
```
tpasternak, line 142: I would prefer to have a token (ex. `case object` containing message) as a key here instead of string, but of course it'll work.

tpasternak (approved):

PR #240, opened 2021-10-19
Add description to help messages

MarcinAman:


modules/cli/src/main/scala/scala/cli/commands/DirectoriesOptions.scala
```diff
@@ -3,6 +3,7 @@ package scala.cli.commands
 import caseapp._
 
 // format: off
+@HelpMessage("Prints directories used by scala-cli")
 final case class DirectoriesOptions(
```
romanowski, line 7: Shouldn't it be hidden?

romanowski (approved):

PR #276, opened 2021-10-24
deps: bump Bloop to 1.4.10.

ckipp01:
This is mainly to avoid someone running a Snapshot of Metals, which now
needs 1.4.10 after they have ran scala-cli once meaning that scala-cli
started Bloop, which is older, they then opened metals, and then were
greeted with a warning and a message offering to re-start Bloop. I just
had this happen to me and on the Metals side we've been told this can be
confusing for users at times.

All that to say, this just bumps Bloop to the newest release


romanowski:
@ckipp01 thank you for your contribution! CI is all green so I am merging the PR.

PR #301, opened 2021-10-26
use more common example libraries

SethTisue:
I think it's better not to get into pure-functional effect systems
in first examples


romanowski:
Thank you @SethTisue for the change. When we were in doc writting mode we stick to one set of exaplamples...

PR #317, opened 2021-10-28
Add command to updating scala-cli

### LINKED ISSUE #247: scala-cli installed from scripts should be able to update its version
When scala-cli is installed using installation scripts we do not offer automatic update mechanism. 

We can update scala-cli automatically or ask user to update if we detect that new version was released.

lwronski:
1.  run `scala-cli update` before `compile`, `run` and `test`
  - If we detect, that `scala-cli` is out-to-date we logging that user should update scala-cli using installation scripts. It's only print log, not update `scala-cli`

2.  if someone run `scala-cli update` from console, it will automatically update scala-cli using installation script.


modules/cli/src/main/scala/scala/cli/commands/Update.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def run(options: UpdateOptions, args: RemainingArgs): Unit = {
+
+    val installDir =
+      options.binDirPath.getOrElse(scala.build.Directories.default().binRepoDir / "scala-cli")
+
+    val scalaCliBinPath = installDir / options.binaryName
+
+    if (!os.exists(scalaCliBinPath)) {
+      System.err.println(
```
romanowski: I think we should also call `which scala-cli` to verify if we are actually using that binary

PR #321, opened 2021-10-31
Fix links in documents

takezoe:


lwronski (approved):
Thank you @takezoe for the fix!

PR #348, opened 2021-11-04
Update nix instructions

kubukoz:
scala-cli was added to Nixpkgs in https://github.com/NixOS/nixpkgs/pull/142850 so there's no need to link to my repository anymore.

After a stable Nixpkgs release (21.11) is out, I'll update these one more time to be just `nix-env -i scala-cli`.


romanowski:
Thank you @kubukoz !

PR #363, opened 2021-11-09
Add cookbook - Setup ScalaCLI project in VSCode

lwronski:


website/docs/cookbooks/vscode.md
```diff
@@ -0,0 +1,57 @@
...
+}
+```
+
+The following command generates configurations files for VSCode support:
+
+```bash
+scala-cli setup-ide .
+```
+
+There is also another way. The first time you run the `run`|`compile`|`test` commands, the configuration files for the VSCode will be also generated.
```
romanowski, line 32: I wonder if we should give setup-ide as an alternative and points to the guide

romanowski (approved):

PR #364, opened 2021-11-09
Using directives improvements

pikinier20:


modules/build/src/main/scala/scala/build/preprocessing/directives/UsingCustomJarDirectiveHandler.scala
```diff
@@ -38,20 +39,27 @@ case object UsingCustomJarDirectiveHandler extends UsingDirectiveHandler {
...
-      // FIXME Handle malformed paths here
-      os.Path(p, root)
+    val errors = extraJars.collect {
+      case Left(error) => error
     }
 
+    val res = if (errors.nonEmpty) Left(CompositeBuildException(errors)) else Right(extraJars.collect {
+      case Right(value) => value
+    })
+
```
pikinier20: For some reason, I couldn't use `partitionMap`. Compiler threw an error that `partitionMap` is not a member of `Seq[Either[A, B]]`

romanowski: It was introduced after 2.12.x

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -399,7 +421,7 @@ class BuildTests extends munit.FunSuite {
           |}
           |""".stripMargin,
       os.rel / "Ignored.scala" ->
-        """using target scala == 2.12
+        """// @using target.scala.== "2.12"
```
romanowski: I had no idea that we support `// @using` syntax

romanowski: Is `// using lib ...` supported?

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -511,23 +533,23 @@ class BuildTests extends munit.FunSuite {
...
-          |}
-          |""".stripMargin
-    )
-    testInputs.withBuild(defaultOptions, buildThreads, bloopConfig) { (_, _, maybeBuild) =>
-      val sources = maybeBuild.toOption.get.successfulOpt.get.sources
-      expect(sources.inMemory.isEmpty)
-      expect(sources.paths.lengthCompare(1) == 0)
-    }
-  }
+//
```
romanowski: Commented out test?

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -572,29 +605,10 @@ class BuildTests extends munit.FunSuite {
     }
   }
 
-  test("ScalaNativeOptions for native-gc with no values") {
```
romanowski, line 574: Removed test?

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -710,22 +724,4 @@ class BuildTests extends munit.FunSuite {
       )
     }
   }
-
-  test("ScalaNativeOptions for native-linking and no value") {
```
romanowski, line 713: Removed test?

romanowski (changes requested):
It seems that `// using lib ... ` sytax is not supported and we should really have it.

Also no documentation is updated in this PR

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
pikinier20: Yes, It just needs one switch in configuration of UsingDirectivesProcessor

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
pikinier20: It's a test that asserts that sourcefile with directives in comments is not modified. Currently, there's no information about that from my library. We can add it or just always cut the directives part from a file

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
pikinier20: This is not valid syntax of using directives and it's gonna break in parser. Idk if we should test such case there or in my library

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
pikinier20: same story as above

modules/build/src/main/scala/scala/build/preprocessing/DirectivesProcessor.scala
```diff
@@ -65,13 +66,16 @@ object DirectivesProcessor {
       .iterator
       .flatMap {
         case (k, v) =>
-          // FIXME using_directives needs to give us positions here
-          val positionOpt = Option.empty[Position]
-          handlersMap.get(k).iterator.map(_(v, cwd, positionOpt))
+          handlersMap.get(k).iterator.map(_(StrictDirective(k, v), path, cwd))
```
romanowski:
```suggestion
          handlersMap.get(k).iterator.map(_.apply(StrictDirective(k, v), path, cwd))
```

romanowski: `_( ... )` syntax is hard to understand

modules/build/src/main/scala/scala/build/preprocessing/DirectivesProcessor.scala
```diff
@@ -44,15 +43,17 @@ object DirectivesProcessor {
...
-  def process(
-    directives: Map[Path, Seq[Value[_]]],
-    handlers: Seq[DirectiveHandler],
+  def process[T: ConfigMonoid](
+    directives: Seq[(Path, Seq[Value[_]])],
+    handlers: Seq[DirectiveHandler[T]],
+    path: Either[String, os.Path],
     cwd: ScopePath
-  ): Either[BuildException, BuildOptions] = {
+  ): Either[BuildException, (T, Seq[Scoped[T]])] = {
```
romanowski: can we turn `(T, Seq[Scoped[T]])` into a case class? Code would be much easier to understand

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -348,13 +342,19 @@ case object ScalaPreprocessor extends Preprocessor {
 
   private def processStrictUsing(
     content: String,
+    path: Either[String, os.Path],
     cwd: ScopePath
-  ): Either[BuildException, Option[(BuildOptions, String)]] = either {
+  ): Either[BuildException, Option[(
```
romanowski: Same here, could we turn 

```(BuildRequirements,
    BuildOptions,
    Seq[Scoped[BuildRequirements]],
    String
  )
  ``` 
  
  into a case class?

File: modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveHandler.scala [hunk 9]
```diff
@@ -14,16 +12,12 @@ trait DirectiveHandler {
...
-    if (keys.isEmpty)
-      sys.error("Cannot happen")
-    else
-      throw new NotImplementedError(
-        "using_directives-based directives need to override handleValues"
-      )
+    directive: StrictDirective,
+    path: Either[String, os.Path],
+    cwd: ScopePath
+  ): Either[BuildException, (Option[T], Seq[Scoped[T]])]
```
romanowski: same as above, case class would help a lot with readablity

modules/build/src/main/scala/scala/build/preprocessing/directives/RequireDirectiveHandler.scala
```diff
@@ -1,12 +1,23 @@
...
+
+  def handleValues(
+    directive: StrictDirective,
+    path: Either[String, os.Path],
+    cwd: ScopePath
+  ): Either[BuildException, (Option[BuildRequirements], Seq[Scoped[BuildRequirements]])] =
+    if (keys.isEmpty)
+      sys.error("Cannot happen")
+    else
+      throw new NotImplementedError(
```
romanowski: Why we have such a strange implementation?

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
romanowski: @alexarchambault @tpasternak @lwronski I have my doubts now if we should support `// @using ...` sytnax. For me e.g. `// @using scala 3` should be a commented using directive

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
romanowski: I think it should not block the code now but I want to avoid source generation as much as possible so I think the library should expose this information as a follow up - since in future we want to warn if `@using` and `// using` style is mixed

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
romanowski: shouldn't we make key optional that default to True?

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
romanowski: If we agree that we do not want `// @ syntax` then we should have a test to make sure it is not recoginized

project/deps.sc
```diff
@@ -86,7 +86,7 @@ object Deps {
   def swoval                   = ivy"com.swoval:file-tree-views:2.1.7"
   def testInterface            = ivy"org.scala-sbt:test-interface:1.0"
   def upickle                  = ivy"com.lihaoyi::upickle:1.4.2"
-  def usingDirectives          = ivy"org.virtuslab:using_directives:0.0.7"
+  def usingDirectives          = ivy"org.virtuslab:using_directives:0.0.7-821b592-SNAPSHOT"
```
romanowski: Before the release we need it to be a stable version

website/docs/guides/using-directives.md
```diff
@@ -12,25 +12,25 @@ The `using` directives mechanism lets you define configuration information withi
...
+using foo "bar", "baz"
 ```
 
 will be interpreted as assigning `bar` and `baz` to the key `foo`.
 
 As shown, `using` directives can be defined using the special keyword `using`. However, this may not be compatible with exisiting tools outside of Scala CLI. Therefore, they can be used in comments:
 
 ```scala
-// using scala 2
+// using scala "2"
```
romanowski, line 23: Do we need quotes around numbers? I was under impression that Strings and primitives are valid values

website/docs/guides/using-directives.md
```diff
@@ -12,25 +12,25 @@ The `using` directives mechanism lets you define configuration information withi
 `using` directives are basically key-value pairs that let you provide multiple values to a single key. For instance, this command:
 
 ```scala
```
romanowski, line 14: This whole section needs to be reworded, since it speaks about 3 ways and now we have only 2. So it should default to comments nad mention top level annotation syntax

romanowski: We should also mention there what can be valid key and how nesting works

romanowski (changes requested):
I think we should remove support for `// @using` syntax include information what syntax is used per directive and allow empty values that defaults to `true`

Plus some rewrites in code style

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -129,32 +129,27 @@ case object ScalaPreprocessor extends Preprocessor {
 
     val (content0, isSheBang) = SheBang.ignoreSheBangLines(content)
 
-    val afterStrictUsing = value(processStrictUsing(content0, scopeRoot))
-    val afterUsing = value {
-      processUsing(path, afterStrictUsing.map(_._2).getOrElse(content0), scopeRoot)
-        .sequence
-    }
+    val afterStrictUsing = value(processStrictUsing(content, scopeRoot))
```
alexarchambault: Why change `content0` to `content` here? That discard the shebang preprocessing stuff.

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -129,32 +129,27 @@ case object ScalaPreprocessor extends Preprocessor {
...
     val afterProcessImports = value {
       processSpecialImports(
-        afterUsing.flatMap(_._4).orElse(afterStrictUsing.map(_._2)).getOrElse(content0),
+        afterStrictUsing.map(_._2).getOrElse(content),
```
alexarchambault: Same as above, shouldn't that be `content0`?

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -129,32 +129,27 @@ case object ScalaPreprocessor extends Preprocessor {
...
         afterProcessImports.map(_._2).toSeq
       val summedOptions = allOptions.foldLeft(BuildOptions())(_ orElse _)
       val lastContentOpt = afterProcessImports
         .map(_._3)
-        .orElse(afterUsing.flatMap(_._4))
         .orElse(afterStrictUsing.map(_._2))
         .orElse(if (isSheBang) Some(content0) else None)
-      val scopedRequirements = afterUsing.map(_._2).getOrElse(Nil)
+
+      val scopedRequirements = Nil
```
alexarchambault: NIT Inline this?

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingCompilerPluginDirectiveHandler.scala
```diff
@@ -45,26 +46,29 @@ case object UsingCompilerPluginDirectiveHandler extends UsingDirectiveHandler {
...
-        .map { dep =>
-          // Really necessary? (might already be handled by the coursier-dependency library)
-          val dep0 = dep.filter(!_.isSpaceChar)
+        .map {
+          case (dep, pos) =>
+            // Really necessary? (might already be handled by the coursier-dependency library)
+            val dep0 = dep.filter(!_.isSpaceChar)
 
-          parseDependency(dep0)
+            parseDependency(dep0).map(_ -> pos)
```
alexarchambault: NIT Build a `Positioned(Seq(pos), _)` right now, so that we don't have to do the same below?

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingCustomJarDirectiveHandler.scala
```diff
@@ -38,20 +39,29 @@ case object UsingCustomJarDirectiveHandler extends UsingDirectiveHandler {
...
 
-    val root = value(Directive.osRoot(cwd, positionOpt))
-    val extraJars = DirectiveUtil.stringValues(values).map { p =>
-      // FIXME Handle malformed paths here
-      os.Path(p, root)
+    val errors = extraJars.collect {
+      case Left(error) => error
     }
 
+    val res =
```
alexarchambault: This is basically
```scala
import scala.build.Ops._
val res = extraJars
  .sequence
```

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingDependencyDirectiveHandler.scala
```diff
@@ -46,27 +47,28 @@ case object UsingDependencyDirectiveHandler extends UsingDirectiveHandler {
...
-        .map { dep =>
-          // Really necessary? (might already be handled by the coursier-dependency library)
-          val dep0 = dep.filter(!_.isSpaceChar)
+        .map {
+          case (dep, pos) =>
+            // Really necessary? (might already be handled by the coursier-dependency library)
+            val dep0 = dep.filter(!_.isSpaceChar)
 
-          parseDependency(dep0)
+            parseDependency(dep0).map(_ -> pos)
```
alexarchambault: Same as above (build a `Positioned` right here?)

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingJavaHomeDirectiveHandler.scala
```diff
@@ -36,18 +38,17 @@ case object UsingJavaHomeDirectiveHandler extends UsingDirectiveHandler {
 
   override def keys = Seq("java-home", "javaHome")
   override def handleValues(
-    values: Seq[Any],
-    cwd: ScopePath,
-    positionOpt: Option[Position]
+    values: Seq[Value[_]],
+    cwd: ScopePath
   ): Either[BuildException, BuildOptions] = either {
     val rawHome = value {
```
alexarchambault: NIT Do a match here, like
```scala
val (rawHome, position) = …
```

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingJavaOptionsDirectiveHandler.scala
```diff
@@ -32,14 +33,13 @@ case object UsingJavaOptionsDirectiveHandler extends UsingDirectiveHandler {
...
-    cwd: ScopePath,
-    positionOpt: Option[Position]
+    values: Seq[Value[_]],
+    cwd: ScopePath
   ): Either[BuildException, BuildOptions] = {
     val javaOpts = DirectiveUtil.stringValues(values)
     val options = BuildOptions(
       javaOptions = JavaOptions(
-        javaOpts = javaOpts.map(o => scala.build.Positioned(positionOpt.toList, o))
+        javaOpts = javaOpts.map(o => scala.build.Positioned(Seq(o._2), o._1))
```
alexarchambault: NIT Do a match? like
```scala
javaOpts.map { case (opt, pos) => scala.build.Positioned(Seq(pos), opt) }
```

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingPlatformDirectiveHandler.scala
```diff
@@ -38,49 +40,49 @@ case object UsingPlatformDirectiveHandler extends UsingDirectiveHandler {
     Platform.parse(Platform.normalize(split(inputs.head)._1)).nonEmpty
 
   private def handle(
-    rawPfStrs: Seq[String],
-    position: Position
+    rawPfStrsWithPos: Seq[(String, Position)]
```
alexarchambault: Couldn't this be a `rawPfStrs: Seq[Positioned(String]]`? (`Positioned` is meant to avoid having `(_, Position)` everywhere)

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingResourcesDirectiveHandler.scala
```diff
@@ -1,8 +1,15 @@
 package scala.build.preprocessing.directives
 
+<<<<<<< HEAD
```
alexarchambault: Shouldn't be here.

modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
```diff
@@ -1,25 +1,16 @@
...
 
-import com.virtuslab.using_directives.custom.model.Value
+import com.virtuslab.using_directives.custom.model.{StringValue, Value}
 
-import scala.jdk.CollectionConverters._
+import scala.build.Position
 
 object DirectiveUtil {
-  def stringValues(values: Seq[Any]): Seq[String] =
+  def stringValues(values: Seq[Value[_]]): Seq[(String, Position)] =
```
alexarchambault: This should return a `Seq[Positioned[String]]`

modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
```diff
@@ -5,12 +5,16 @@ import com.virtuslab.using_directives.custom.model.{StringValue, Value}
 import scala.build.Position
 
 object DirectiveUtil {
-  def stringValues(values: Seq[Value[_]]): Seq[(String, Position)] =
+  def stringValues(
+    values: Seq[Value[_]],
+    path: Either[String, os.Path]
+  ): Seq[(String, Position)] =
```
alexarchambault: This should return a `Seq[Positioned[String]]`

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingJavaOptionsDirectiveHandler.scala
alexarchambault: Maybe not necessary if `DirectiveUtil.stringValues` is changed to return `Seq[Positioned[String]]`

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
alexarchambault: It'd be nice to have a way to preserve those tests.

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -69,44 +69,46 @@ class SourcesTests extends munit.FunSuite {
       val scopedSources = crossSources.scopedSources(BuildOptions()).orThrow
       val sources       = scopedSources.sources(Scope.Main, BuildOptions())
 
-      expect(sources.buildOptions.classPathOptions.extraDependencies.map(_.value) == expectedDeps)
+      expect(sources.buildOptions.classPathOptions.extraDependencies.map(
+        _.value
+      ).toSet == expectedDeps.toSet)
```
alexarchambault: Why is this changed to a set? The order of dependencies matters (it has an impact on the order of JARs in the class path, which can make the resources or classes of certain JARs shadow the same resources or classes in other JARs).

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -69,44 +69,46 @@ class SourcesTests extends munit.FunSuite {
...
-        CrossSources.forInputs(inputs, Sources.defaultPreprocessors(CustomCodeWrapper)).orThrow
-      val scopedSources = crossSources.scopedSources(BuildOptions()).orThrow
-      val sources       = scopedSources.sources(Scope.Main, BuildOptions())
-
-      expect(sources.buildOptions.classPathOptions.extraDependencies.map(_.value) == expectedDeps)
-      expect(sources.paths.isEmpty)
-      expect(sources.inMemory.length == 0)
-    }
-  }
+//  test("dependencies in .scala - using witin tests") {
```
alexarchambault: Does that mean that there's no equivalent to `using target test` syntax?

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -272,33 +274,6 @@ class SourcesTests extends munit.FunSuite {
     }
   }
 
-  test("dependencies in .sc - using") {
```
alexarchambault, line 275: Why is this one removed?

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -9,7 +9,7 @@ class BloopTests extends munit.FunSuite {
   val dummyInputs = TestInputs(
     Seq(
       os.rel / "Test.scala" ->
-        """// using scala 2.13
+        """// @using scala "2.13"
```
alexarchambault: I thought we were not supporting `// @using` (only `// using` or `@using`)

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -555,7 +555,7 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     val inputs = TestInputs(
       Seq(
         os.rel / "Test.scala" ->
-          s"""// using lib com.lihaoyi::pprint:0.0.0.0.0.1
+          s"""// @using lib "com.lihaoyi::pprint:0.0.0.0.0.1"
```
alexarchambault: Same as above (`// @using` syntax)

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -580,9 +580,9 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
         expect(diag.getSeverity == b.DiagnosticSeverity.ERROR)
         expect(diag.getRange.getStart.getLine == 0)
-        expect(diag.getRange.getStart.getCharacter == 3)
+        expect(diag.getRange.getStart.getCharacter == 16)
         expect(diag.getRange.getEnd.getLine == 0)
-        expect(diag.getRange.getEnd.getCharacter == 44)
+        expect(diag.getRange.getEnd.getCharacter == 16)
```
alexarchambault: The end character is a problem here (it should be 46 I guess)

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -49,81 +49,81 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
-            |  }
-            |}
-            |""".stripMargin
-      )
-    )
-    inputs.fromRoot { root =>
-      os.proc(TestUtil.cli, "compile", extraOptions, ".").call(cwd = root)
-    }
-  }
+//  test("test scope") {
```
alexarchambault: Why is this disabled?

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -49,81 +49,81 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
-      expect(res.exitCode == 1)
-      val expectedInOutput =
-        if (actualScalaVersion.startsWith("2."))
-          "not found: value pprint"
-        else
-          "Not found: pprint"
-      expect(res.out.text().contains(expectedInOutput))
-    }
-  }
+//  test("test scope error") {
```
alexarchambault: Why disable this?

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1133,43 +1133,43 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
-            |""".stripMargin
-      )
-    )
-    inputs.fromRoot { root =>
-      val res = os.proc(TestUtil.cli, extraOptions, ".").call(cwd = root)
-      pprint.log(res.out.text())
-      expect(res.out.text().contains("Hello from tests"))
-    }
-  }
+//  test("test scope") {
```
alexarchambault: Why is this disabled?

modules/integration/src/test/scala/scala/cli/integration/TestTestDefinitions.scala
```diff
@@ -437,89 +435,90 @@ abstract class TestTestDefinitions(val scalaVersionOpt: Option[String])
...
-      val expectedCount = 1 + (if (TestUtil.canRunJs) 1 else 0) + (if (supportsNative) 1 else 0)
-      expect(countSubStrings(output, "Hello from shared") == expectedCount)
-      expect(output.contains("Hello from jvm"))
-      if (TestUtil.canRunJs)
-        expect(output.contains("Hello from js"))
-      if (supportsNative)
-        expect(output.contains("Hello from native"))
-    }
-  }
+//  private def countSubStrings(input: String, subString: String): Int = {
```
alexarchambault: Why is this disabled?

modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveHandler.scala
```diff
@@ -10,4 +13,11 @@ trait DirectiveHandler {
 
   // Strict / using_directives-based directives
   def keys: Seq[String]
+
+  def handleValues(
+    directive: StrictDirective,
+    path: Either[String, os.Path],
+    cwd: ScopePath
+  ): Either[BuildException, (Option[T], Seq[Scoped[T]])]
```
alexarchambault: Why do we abstract over `T` here?

modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
```diff
@@ -1,20 +1,44 @@
...
+            Option(v.getScope).map((p: String) => cwd / os.RelPath(p))
+          ))
+      }
+      .flatten
+
+  def numericValues(
+    values: Seq[Value[_]],
+    path: Either[String, os.Path],
+    cwd: ScopePath
+  ): Seq[(String, Position, Option[ScopePath])] =
```
alexarchambault, line 31: We should at least use `Positioned` here, maybe even `Scoped`.

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
```diff
@@ -29,4 +33,53 @@ case object RequirePlatformsDirectiveHandler extends RequireDirectiveHandler {
...
+    cwd: ScopePath
+  ): Either[BuildException, (Option[BuildRequirements], Seq[Scoped[BuildRequirements]])] = {
+    val values          = DirectiveUtil.stringValues(directive.values, path, cwd)
+    val nonscopedValues = values.filter(_._3.isEmpty)
+    val scopedValues    = values.filter(_._3.nonEmpty)
+    val nonScopedPlatforms = Option(nonscopedValues.map(v => Platform.normalize(v._1)))
+      .filter(_.nonEmpty)
+
+    val nonscoped =
+      nonScopedPlatforms.fold[Either[BuildException, Option[BuildRequirements]]](Right(None)) {
```
alexarchambault: A `match` would be clearer than a `fold` here, IMO (no need to write an explicit type with a `match` I think).

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
```diff
@@ -29,4 +33,53 @@ case object RequirePlatformsDirectiveHandler extends RequireDirectiveHandler {
...
     }
+
+  override def handleValues(
+    directive: StrictDirective,
+    path: Either[String, Path],
+    cwd: ScopePath
+  ): Either[BuildException, (Option[BuildRequirements], Seq[Scoped[BuildRequirements]])] = {
+    val values          = DirectiveUtil.stringValues(directive.values, path, cwd)
+    val nonscopedValues = values.filter(_._3.isEmpty)
+    val scopedValues    = values.filter(_._3.nonEmpty)
```
alexarchambault: Change this to
```scala
val scopedValues = values.collect { case (…, Some(path), …) => (…, path, …) }
```
so that we don't have to match a `Some(_)` with no `None` case below, which looks odd?

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
```diff
@@ -29,4 +33,53 @@ case object RequirePlatformsDirectiveHandler extends RequireDirectiveHandler {
...
+            Right(Some(BuildRequirements(
+              platform = Seq(BuildRequirements.PlatformRequirement(p))
+            )))
+          }
+      }
+    val scoped = scopedValues.groupBy(_._3).map {
+      case (Some(scopePath), list) =>
+        val platforms = list.map(_._1).map(Platform.normalize)
+        val parsed    = Platform.parseSpec(platforms)
+        parsed.fold[Either[BuildException, Seq[Scoped[BuildRequirements]]]](
```
alexarchambault: Same as above, a `match` would be clearer.

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
```diff
@@ -29,4 +33,53 @@ case object RequirePlatformsDirectiveHandler extends RequireDirectiveHandler {
...
+          Left(new MalformedPlatformError(platforms.mkString(", ")))
+        ) { p =>
+          Right(Seq(Scoped(
+            scopePath,
+            BuildRequirements(
+              platform = Seq(BuildRequirements.PlatformRequirement(p))
+            )
+          )))
+        }
+    }.foldLeft[Either[BuildException, Seq[Scoped[BuildRequirements]]]](Right(Seq.empty)) {
```
alexarchambault: Using
```scala
  .sequence
  .map(_.flatten)
```

(with `import scala.build.Ops._`) would be better, and would not discard errors (here, only one of the errors is retained)

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
```diff
@@ -29,4 +33,53 @@ case object RequirePlatformsDirectiveHandler extends RequireDirectiveHandler {
...
+      case (Right(seq), Right(v)) => Right(seq ++ v)
+      case (Left(err), _)         => Left(err)
+      case (_, Left(err))         => Left(err)
+    }
+
+    for {
+      ns <- nonscoped
+      s  <- scoped
+    } yield ns -> s
```
alexarchambault: We should use the CPS stuff rather than a for loop. Look for the `either { … }` blocks in the code.

alexarchambault: Note that a `value(nonscoped); value(scoped)` would discard the error in `scoped` if `nonscoped` is errored. To avoid that, you can add a `EitherMap2` in `Ops.scala` (see `EitherMap3` as an example), and call `.traverseN`

modules/build/src/main/scala/scala/build/preprocessing/directives/RequireScalaVersionDirectiveHandler.scala
```diff
@@ -41,4 +50,63 @@ case object RequireScalaVersionDirectiveHandler extends RequireDirectiveHandler
...
+
+  override def handleValues(
+    directive: StrictDirective,
+    path: Either[String, Path],
+    cwd: ScopePath
+  ): Either[BuildException, (Option[BuildRequirements], Seq[Scoped[BuildRequirements]])] = {
+    val values         = DirectiveUtil.stringValues(directive.values, path, cwd)
+    val nonscopedValue = values.find(_._3.isEmpty).map(_._1)
+    val nonscoped =
+      nonscopedValue.fold[Either[BuildException, Option[BuildRequirements]]](Right(None))(v =>
```
alexarchambault: Just like above, a `match` would be clearer than a `fold`.

modules/build/src/main/scala/scala/build/preprocessing/directives/RequireScalaVersionDirectiveHandler.scala
```diff
@@ -41,4 +50,63 @@ case object RequireScalaVersionDirectiveHandler extends RequireDirectiveHandler
...
+        handleVersion(directive.key, v)
+      )
+    val scoped = values.filter(_._3.nonEmpty).map {
+      case (v, _, Some(scopePath)) =>
+        handleVersion(directive.key, v).map(_.map(req => Scoped(scopePath, req)))
+    }.foldLeft[Either[BuildException, Seq[Scoped[BuildRequirements]]]](Right(Seq.empty)) {
```
alexarchambault: Just like above, a `.sequence.left.map(CompositeBuild(_)).map(_.flatten)` would keep all errors rather than just a single one.

modules/build/src/main/scala/scala/build/preprocessing/directives/RequireScalaVersionDirectiveHandler.scala
```diff
@@ -41,4 +50,63 @@ case object RequireScalaVersionDirectiveHandler extends RequireDirectiveHandler
...
+      case (Right(seq), Right(v)) => Right(seq ++ v)
+      case (Left(err), _)         => Left(err)
+      case (_, Left(err))         => Left(err)
+    }
+    for {
```
alexarchambault: Same as above, use the Either CPS stuff, and maybe `.traverseN` too.

alexarchambault (commented):
First round of comments, I didn't review everything yet. I also looked at it commit-by-commit, so feel free to ignore comments that you already addressed in later commits.

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -24,6 +23,31 @@ import scala.jdk.CollectionConverters._
 
 case object ScalaPreprocessor extends Preprocessor {
 
+  case class StrictDirectivesProcessingOutput(
```
alexarchambault: It seems this can be private, isn't it?

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -24,6 +23,31 @@ import scala.jdk.CollectionConverters._
...
+    scopedReqs: Seq[Scoped[BuildRequirements]],
+    strippedContent: Option[String]
+  ) {
+    def isEmpty: Boolean = globalReqs == BuildRequirements.monoid.zero &&
+      globalUsings == BuildOptions.monoid.zero &&
+      scopedReqs.isEmpty &&
+      strippedContent.isEmpty
+  }
+
+  case class SpecialImportsProcessingOutput(
```
alexarchambault: Can be private too, apparently.

modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
```diff
@@ -67,6 +68,10 @@ object ScriptPreprocessor {
 
     val (requirements, scopedRequirements, options, updatedCodeOpt) =
       value(ScalaPreprocessor.process(contentIgnoredSheBangLines, reportingPath, scopePath / os.up))
+        .map {
+          case ProcessingOutput(globalReqs, scopedReqs, opts, updatedContent) =>
```
alexarchambault: Maybe we should keep the `ProcessingOutput` here (and not match above). That case only "unapplies" `ProcessingOutput`.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
pikinier20: Yes, currently my lib returns position of the beginning of the element that you refer to. It needs some work to return correct span. Maybe we can leave it as it is and make it better in some follow-up PR?

modules/build/src/main/scala/scala/build/preprocessing/directives/DirectiveHandler.scala
pikinier20: I've made here the abstraction to be able to reuse DirectiveProcessor.process for require and using directives

modules/build/src/main/scala/scala/build/preprocessing/directives/RequirePlatformsDirectiveHandler.scala
pikinier20: Wow, that is usefeul

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
pikinier20: This was just temporary, in further commits it's not Nil

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingJavaOptionsDirectiveHandler.scala
pikinier20: I've just matched there and left DirectiveUtil as it is

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingResourcesDirectiveHandler.scala
pikinier20: :(

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
pikinier20: You're right. Changed that

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
pikinier20: It's re-enabled in further commits

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
pikinier20: Yes, it's changed in further commits

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
pikinier20: In latest commit all tests are re-enabled and re-added

website/docs/guides/using-directives.md
pikinier20: scala version is usually a string so basically I assumed we accept String there. Numbers are also valid values but for scala version they are not retrieved from directive. It can be changed though

romanowski (approved):

PR #369, opened 2021-11-10
clean .bsp/scala-cli.json when you run clean

ckipp01:
I think it might be best to do this when you run clean. I hit on this a
couple times now where something was out of sync or not working which
caused me to attempt to do a `scala-cli` clean. However, when you clean
it removes the `.scala` dir, but not the entry that was created in
`.bsp/scala-cli.json` which still references the
`.scala/ide-options.json` file which no longer exists after the clean.
Then since the BSP entry still exists, BSP discovery still triggers in
Metals, and then it sends in the initialize request, but scala-cli just
freezes and doesn't respond. This fix just ensures that when you run
`scala-cli` clean, it cleans up both the `.scala/` and also the
`.bsp/scala-json`.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Clean.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Clean.scala
@@ -20,7 +20,8 @@ object Clean extends ScalaCommand[CleanOptions] {
       case Right(i) => i
     }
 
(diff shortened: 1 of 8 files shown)


alexarchambault (approved):
Thanks @ckipp01!

PR #372, opened 2021-11-15
Run native linker only if project was changed

### LINKED ISSUE #281: Native scripts run linker each time
From @tpasternak :

```
#!/usr/bin/env scala-cli
using scala 2.13
using native
println("hello")
```
output:
```
$  ./ls.sc
Linking (1027 ms)
Discovered 972 classes and 7872 methods
Generating intermediate code (1619 ms)
Produced 4 files
Compiling to native code (760 ms)
Linking native code (immix gc, none lto) (152 ms)
Total (3598 ms)
hello
```

as a result, it's slow

From @WojciechMazur :

In the Scala Native 0.4.1 sbt caching mechanism will be used to skip subsequent builds, if you'll manage to implement something similar we will be glad to backport it thus allow to use it for all build tools.
What's important for the SN side we need to track both changes to input files and pass NativeConfig.

lwronski:
closes #281 

To avoid long waiting for executing linker,  we calculate the sha of the project, and only when something has changed we run a linker.
We count sha for the following variable:
-  output (was changed or deleted)
-  native config
-  all files in the project
-  build options
-  ScalaCLI version

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
@@ -1,0 +1,65 @@
+package scala.build.internal
+
+import java.math.BigInteger
+import java.security.MessageDigest
+
+import scala.build.Build
+import scala.scalanative.{build => sn}
+
+case object NativeBuilderHelper {
+
+  private def resolveProjectShaPath(nativeWorkDir: os.Path) = nativeWorkDir / ".project_sha"
+  private def resolveOutputShaPath(nativeWorkDir: os.Path)  = nativeWorkDir / ".output_sha"
+
+  private def fileSha(filePath: os.Path): String = {
+    val md = MessageDigest.getInstance("SHA-1")
+    md.update(os.read.bytes(filePath))
+
+    val digest        = md.digest()
+    val calculatedSum = new BigInteger(1, digest)
+    String.format(s"%040x", calculatedSum)
+  }
+
+  private def projectSha(build: Build.Successful, nativeConfig: sn.NativeConfig) = {
+    val md = MessageDigest.getInstance("SHA-1")
+    md.update(build.inputs.sourceHash().getBytes)
+    md.update(nativeConfig.toString.getBytes)
+    md.update(build.options.hash.getOrElse("").getBytes)
+
+    val digest        = md.digest()
+    val calculatedSum = new BigInteger(1, digest)
+    String.format(s"%040x", calculatedSum)
+  }
+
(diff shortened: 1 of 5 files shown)


modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
```diff
@@ -0,0 +1,65 @@
+package scala.build.internal
+
+import java.math.BigInteger
+import java.security.MessageDigest
+
+import scala.build.Build
+import scala.scalanative.{build => sn}
+
+case object NativeBuilderHelper {
```
alexarchambault: NIT: no need to make that a `case` (`case object` is more for objects that represent some data).

```suggestion
object NativeBuilderHelper {
```

modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
```diff
@@ -0,0 +1,65 @@
...
+    val md = MessageDigest.getInstance("SHA-1")
+    md.update(os.read.bytes(filePath))
+
+    val digest        = md.digest()
+    val calculatedSum = new BigInteger(1, digest)
+    String.format(s"%040x", calculatedSum)
+  }
+
+  private def projectSha(build: Build.Successful, nativeConfig: sn.NativeConfig) = {
+    val md = MessageDigest.getInstance("SHA-1")
```
alexarchambault, line 24: We may want to hash the Scala CLI version somewhere too. Different Scala CLI versions might have different defaults (for options that are not overridden), so we should probably invalidate the cached binary if it changed.

modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
```diff
@@ -0,0 +1,65 @@
...
+    md.update(build.inputs.sourceHash().getBytes)
+    md.update(nativeConfig.toString.getBytes)
```
alexarchambault, line 26: We shouldn't need to hash the native config here, as it originates from the build options (that are hashed right after). I opened https://github.com/VirtusLab/scala-cli/pull/394 to make that more explicit.

lwronski: I think in some cases, it's required to count hash for native config.  For example, if clang path was changed, we do not detect this change in build options, because we discover it after creating buildOptions -> [discover clang path](https://github.com/VirtusLab/scala-cli/blob/master/modules/build/src/main/scala/scala/build/options/ScalaNativeOptions.scala#L39)

alexarchambault (approved):

PR #400, opened 2021-11-18
Add using directive for specifying a (default) main class

### LINKED ISSUE #370: Add using directive for --main-class
Many command-line flags can be set with 'using' directives, but --main-class cannot.
It would be nice to have a default main-class (similar to how a META-INF/manifest.mf with main class is set).
I think the command-line flag --main-class is too technical for the intended users of my small apps (i.e. colleagues). I would like to keep a small number of related 'apps' in a single project, each with different command-line arguments, but where one should be the default.

fourth44:
resolves #370.
All the infrastructure for passing a main class in the build options already existed, so this only adds the directive to fill that option.

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/directives/UsingMainClassDirectiveHandler.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/directives/UsingMainClassDirectiveHandler.scala
@@ -1,0 +1,59 @@
+package scala.build.preprocessing.directives
+import scala.build.Position
+import scala.build.errors.{BuildException, SeveralMainClassesFoundError}
+import scala.build.options.BuildOptions
+import scala.build.preprocessing.ScopePath
+
+case object UsingMainClassDirectiveHandler extends UsingDirectiveHandler {
+
+  def name        = "Main class"
+  def description = "Specify default main class"
+
+  def usage = "using main-class _main class_ | using mainClass _main class_"
+
+  override def usageMd =
+    """`using main-class `_main class_
+      |
+      |`using mainClass `_main class_""".stripMargin
+
+  override def examples = Seq(
+    "using main-class helloWorld"
+  )
+
+  override def handle(
+    directive: Directive,
+    cwd: ScopePath
+  ): Option[Either[BuildException, BuildOptions]] =
+    directive.values match {
+      case Seq("main-class", mainClass) =>
+        val options = BuildOptions(
+          mainClass = Some(mainClass)
+        )
+        Some(Right(options))
+      case _ =>
+        None
+    }
+
+  override def keys = Seq("main-class", "mainClass")
+
+  override def handleValues(
+    values: Seq[Any],
+    cwd: ScopePath,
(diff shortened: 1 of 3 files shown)


modules/build/src/main/scala/scala/build/preprocessing/directives/UsingMainClassDirectiveHandler.scala
```diff
@@ -0,0 +1,51 @@
...
+  override def keys = Seq("main-class", "mainClass")
+
+  override def handleValues(
+    values: Seq[Any],
+    cwd: ScopePath,
+    positionOpt: Option[Position]
+  ): Either[BuildException, BuildOptions] = {
+    val mainClasses = DirectiveUtil.stringValues(values)
+    val options = BuildOptions(
+      mainClass = mainClasses.lastOption
```
lwronski: I think, we should not allowed to passing main class few times. 

You can use `SeveralMainClassesFoundError` which is already defined [here](https://github.com/VirtusLab/scala-cli/blob/master/modules/build/src/main/scala/scala/build/errors/SeveralMainClassesFoundError.scala).

lwronski (commented):
@fourth44  Thank you for your contribution! I have just left one suggestion.

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingMainClassDirectiveHandler.scala
fourth44: I'm happy to implement your suggestion, but I would like to point out that `SeveralMainClassesFoundError` seems to be intended for the number of main classes _found_ by analyzing the classpath, while this situation is about the number of main classes supplied by directives. That said, supplying multiple --main-class command line arguments also results in an error so this should definitely too, and it seems best to reuse this error rather than be pedantic that it's not about _found_ but _supplied_ main classes.

SethTisue:
is there a ticket about `kebab-case` vs `camelCase` for directives? it may seem bike-shed-y, but it's the sort of decision that's hard to change later, too

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingMainClassDirectiveHandler.scala
lwronski: I think too, that is better to reuse `SeveralNoClassdefFoundError` in this case.

Anyway, we started working on an [options validation mechanism](https://github.com/VirtusLab/scala-cli/pull/406), which should detect when a directive is specified multiple times, so for now your proposal I good to merge. Later, we will integrate a new mechanism with your directive.

lwronski (approved):

lwronski:
> is there a ticket about `kebab-case` vs `camelCase` for directives? it may seem bike-shed-y, but it's the sort of decision that's hard to change later, too

Hi @SethTisue, [Here](https://github.com/VirtusLab/scala-cli/issues/414)  is opened the discussion to unify standard for naming convention in directives.

PR #403, opened 2021-11-18
Use the newest scala version if is not specified

### LINKED ISSUE #401: ScalaCLI not use the newest version of scala as default
`@main def run() = println("Hello!")`

```
$ scala-cli Main.scala
Hello
```

Expected:

ScalaCLI should use the newest stable version of `scala`

```
$ scala-cli Main.scala
Hello
```

lwronski:
resolves #401 

Instead of using `Constants.defaultScalaVersion`, we should search the newest stable version of scala.

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
+++ b/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
@@ -2,7 +2,7 @@ package scala.build.internal
 
 import upickle.default.{ReadWriter, macroRW}
 
-case class StableScalaVersion(scalaCliVersion: String, supportedScalaVersion: Seq[String])
+case class StableScalaVersion(scalaCliVersion: String, supportedScalaVersions: Seq[String])
 
 case object StableScalaVersion {
   implicit lazy val jsonCodec: ReadWriter[StableScalaVersion] = macroRW
--- a/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
+++ b/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
@@ -4,6 +4,6 @@ import upickle.default.{ReadWriter, macroRW}
 
 case class StableScalaVersion(scalaCliVersion: String, supportedScalaVersions: Seq[String])
 
-case object StableScalaVersion {
+object StableScalaVersion {
   implicit lazy val jsonCodec: ReadWriter[StableScalaVersion] = macroRW
 }
--- a/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
+++ b/modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
@@ -1,0 +1,9 @@
+package scala.build.internal
+
+import upickle.default.{ReadWriter, macroRW}
+
+case class StableScalaVersion(scalaCliVersion: String, supportedScalaVersion: Seq[String])
+
+case object StableScalaVersion {
+  implicit lazy val jsonCodec: ReadWriter[StableScalaVersion] = macroRW
+}
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -13,20 +13,17 @@ import java.security.MessageDigest
 
 import scala.build.EitherCps.{either, value}
 import scala.build.blooprifle.VersionUtil.parseJavaVersion
-import scala.build.errors.{BuildException, Diagnostic, InvalidBinaryScalaVersionError}
 import scala.build.errors.{
   BuildException,
+  Diagnostic,
   InvalidBinaryScalaVersionError,
   NoValidScalaVersionFoundError,
   UnsupportedScalaVersionError
 }
 import scala.build.internal.Constants._
-import scala.build.internal.{Constants, OsLibc, Util}
+import scala.build.internal.{OsLibc, StableScalaVersion, Util}
 import scala.build.options.validation.BuildOptionsRule
 import scala.build.{Artifacts, Logger, Os, Position, Positioned}
-import scala.util.Properties
-import scala.build.internal.{OsLibc, StableScalaVersion, Util}
-import scala.build.{Artifacts, Logger, Os, Positioned}
 import scala.util.Properties
 
 final case class BuildOptions(
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -167,13 +167,18 @@ final case class BuildOptions(
       }
     }
 
-  def latestSupportedScalaVersion(prefixScalaVersion: Option[String])
-    : Option[coursier.core.Version] = {
-    val supportedScalaVersionUrl =
-      "https://raw.githubusercontent.com/VirtusLab/scala-cli/master/scala-version.scala"
+  def latestSupportedScalaVersion(prefixScalaVersion: Option[String]): coursier.core.Version = {
+
+    // used when downloading is fail
+    lazy val defaultStableScalaVersions = StableScalaVersion(
+      version,
+      Seq(defaultScala212Version, defaultScala213Version, defaultScalaVersion)
+    )
+
+    val supportedScalaVersionsUrl = scalaOptions.scalaVersionsUrl
 
     val task = {
-      val art = Artifact(supportedScalaVersionUrl).withChanging(true)
+      val art = Artifact(supportedScalaVersionsUrl).withChanging(true)
       finalCache.file(art).run.flatMap {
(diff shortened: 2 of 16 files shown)


alexarchambault (approved):

alexarchambault:
(For later) As I heard earlier today IIRC, we might want to restrict the Scala versions (`allVersions`) to those equal to or lower than the versions tested in the Scala CLI CI.

lwronski:
I think we can not use the latest stable version `scala` as default in ScalaCLI. For example, If I dumped locally scala from 3.0.2 to 3.1.0 I got errors from pretty-stacktrace:
```
TASTy signature has wrong version.",
  " expected: {majorVersion: 28, minorVersion: 0}",
  " found   : {majorVersion: 28, minorVersion: 1}",
```
 because this library read TASTy files (we have to wait until this library will be published for 3.1)

As we discussed, I defined some configuration file, which contains the list of supported scala-version per ScalaCli version. I'm not sure it should be hosted on github and if it could be a json file. In this file, we define the highest version that is supported by scala-cli. 

For example, if we defined bellow configuration versions file:

```
{
   "scalaCliVersion":  "0.0.9",
   "supportedScalaVersion": ["3.0.1", "2.13.3", "2.12.14"]
}
```

then ScalaCLI uses bellow Scala version

`scala-cli Hello.scala` -> "3.0.1"
`scala-cli Hello.scala -S 3`  -> "3.0.1"
`scala-cli Hello.scala -S 2.13 `-> "2.13.3"
`scala-cli Hello.scala -S 2.13.7 `-> "2.13.7"
`scala-cli Hello.scala -S 2.12` -> "2.12.14"

romanowski:
I think such a configuration format is a good idea. I don't see why we shouldn't host such file on github. Ah, for sure we need to make sure that without internet or if this file change location Scala-cli still works (by using .json files from its release? )

lwronski:
> I think such a configuration format is a good idea. I don't see why we shouldn't host such file on github. Ah, for sure we need to make sure that without internet or if this file change location Scala-cli still works (by using .json files from its release? )

If there is no internet access, the scala version defaults from the `Deps.sc` file will be used. For these versions, we are sure that scala-cli is working properly. I think this will be very similar solution to keeping a .json configuration
 file in a resource.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -168,6 +169,55 @@ final case class BuildOptions(
...
+      finalCache.file(art).run.flatMap {
+        case Left(e) => Task.fail(e)
+        case Right(f) => Task.delay {
+            val content = os.read.bytes(os.Path(f, Os.pwd))
+            content
+          }
+      }
+    }
+
+    val launchersTask = finalCache.logger.using(Task.gather.gather(Seq(task)))
```
alexarchambault: No need to use `gather` if we only have one task:
```suggestion
    val launchersTask = finalCache.logger.using(task)
```

modules/build/src/main/scala/scala/build/internal/StableScalaVersion.scala
```diff
@@ -0,0 +1,9 @@
+package scala.build.internal
+
+import upickle.default.{ReadWriter, macroRW}
+
+case class StableScalaVersion(scalaCliVersion: String, supportedScalaVersions: Seq[String])
+
+case object StableScalaVersion {
```
alexarchambault: No need to make it a `case object` if the `object` itself doesn't represent some data.
```suggestion
object StableScalaVersion {
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -168,6 +169,55 @@ final case class BuildOptions(
...
+
+    val supportedScalaVersionsUrl = scalaOptions.scalaVersionsUrl
+
+    val task = {
+      val art = Artifact(supportedScalaVersionsUrl).withChanging(true)
+      finalCache.file(art).run.flatMap {
+        case Left(e) => Task.fail(e)
+        case Right(f) => Task.delay {
+            val content = os.read.bytes(os.Path(f, Os.pwd))
+            content
```
alexarchambault: Let's read it as a string, and decode it from here (so that we don't need the `Try` below)
```suggestion
            val content = os.read(os.Path(f, Os.pwd))
            upickle.default.read[Seq[StableScalaVersion]](content)
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -168,6 +169,55 @@ final case class BuildOptions(
       }
     }
 
+  def latestSupportedScalaVersion(prefixScalaVersion: Option[String]): coursier.core.Version = {
+
+    // used when downloading is fail
+    lazy val defaultStableScalaVersions = StableScalaVersion(
+      version,
+      Seq(defaultScala212Version, defaultScala213Version, defaultScalaVersion)
+    )
```
alexarchambault: This could be moved outside of `latestSupportedScalaVersion` (and be made a `def`, as it's cheap to compute and isn't used 100s of times).

modules/build/src/main/scala/scala/build/options/ScalaOptions.scala
```diff
@@ -13,8 +13,14 @@ final case class ScalaOptions(
   extraScalaVersions: Set[String] = Set.empty,
   compilerPlugins: Seq[Positioned[AnyDependency]] = Nil,
   platform: Option[Platform] = None,
-  extraPlatforms: Set[Platform] = Set.empty
+  extraPlatforms: Set[Platform] = Set.empty,
+  supportedScalaVersionsUrl: Option[String] = None
 ) {
+
+  lazy val scalaVersionsUrl = supportedScalaVersionsUrl.getOrElse(
+    "https://raw.githubusercontent.com/VirtusLab/scala-cli/master/scala-version.scala"
```
alexarchambault, line 21: Maybe we should put that file in a separate repo, add a version suffix to it, and give it a `.json` extension, like
```suggestion
    "https://raw.githubusercontent.com/VirtusLab/scala-cli-scala-versions/master/scala-versions-v1.json"
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -168,6 +169,55 @@ final case class BuildOptions(
...
+    val stableScalaVersions =
+      if (stableVersionsForScalaCLI0.isEmpty) defaultStableScalaVersions.supportedScalaVersions
+      else stableVersionsForScalaCLI0
+
+    val prefixScalaVersion0 = prefixScalaVersion.getOrElse("")
+
+    stableScalaVersions
+      .filter(_.startsWith(prefixScalaVersion0))
+      .map(coursier.core.Version(_))
+      .max
```
alexarchambault: The `.max` may fail if called on an empty seq:
```suggestion
    stableScalaVersions
      .map(coursier.core.Version(_))
      .sorted
      .map(_.repr)
      .reverse
```

Note that this returns a `Seq[String]` now, and that `prefixScalaVersion` is unused.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -213,19 +263,28 @@ final case class BuildOptions(
...
-        }
+            Right(matchingVersions.map(Version(_)).filter(_ <= latestSupportedScala).max.repr)
+        case None => Right(allVersions.map(Version(_)).filter(_ <= latestSupportedScala).max.repr)
+      }
     }
+    val maybeSv = scalaVersion match {
+      case None => matchNewestScalaVersion(None)
+      case Some(sv0) =>
+        if (Util.isFullScalaVersion(sv0)) Right(sv0)
+        else matchNewestScalaVersion(Some(sv0))
```
alexarchambault, line 301:
```suggestion
      case None =>
        val versions = matchNewestScalaVersion()
        Right(versions.headOption.getOrElse(Constants.defaultScalaVersion))
      case Some(sv0) =>
        if (Util.isFullScalaVersion(sv0)) Right(sv0)
        else matchNewestScalaVersion().filter(_.startsWith(sv0)).getOrElse(sv0)
```

alexarchambault (commented):
I added quite a few suggestions, that depend on each others… I might just push an extra commit myself to address them.

modules/build/src/main/scala/scala/build/options/ScalaOptions.scala
lwronski: cc @romanowski

romanowski: I am fine with that.

romanowski (changes requested):
We are missing any mention of this feature in the documentation. For sure we have a cookbook, getting started and compile command guides that may require updates.

romanowski (approved):

PR #406, opened 2021-11-19
Options validation

tpasternak:


modules/build/src/main/scala/scala/build/options/ScalaOptions.scala
```diff
@@ -12,18 +12,18 @@ final case class ScalaOptions(
   scalacOptions: Seq[String] = Nil,
   extraScalaVersions: Set[String] = Set.empty,
   compilerPlugins: Seq[Positioned[AnyDependency]] = Nil,
-  platform: Option[Platform] = None,
-  extraPlatforms: Set[Platform] = Set.empty
+  platform: Option[Positioned[Platform]] = None,
+  extraPlatforms: Set[Positioned[Platform]] = Set.empty
```
alexarchambault: I'm not sure it's fine to put `Positioned[…]` in a `Set`. This might result in the same platform, specified both on the command-line and via a using directive say, to end up twice in the set (once per position).

I don't know if this can cause serious bugs down-the-line, but this definitely changes the semantics of the set.

Something like `Map[Platform, Positioned[Unit]]` (instead of `Set[Positioned[Platform]]`) should allow to preserve the semantics I think. We might need to define a `ConfigMonoid` instance for `Unit` and `Map[K, V]` (if `V` has an implicit `ConfigMonoid` too) for it to work, but then, the `ConfigMonoid` stuff should automatically and correctly merge such maps (preserving both positions if a platform is specified twice, etc.).

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
```diff
@@ -0,0 +1,34 @@
+package scala.build.options.validation
+
+import scala.build.Position
+import scala.build.errors.BuildException
+import scala.build.options.BuildOptions
+
+trait BuildOptionsRule {
+  def validate(options: BuildOptions): Either[BuildException, Unit]
```
alexarchambault: Maybe we should return a `ValidationException` here, which is more specific than `BuildException`.

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
```diff
@@ -0,0 +1,34 @@
...
+}
+
+object BuildOptionsRule {
+  def validateAll(options: BuildOptions): Either[BuildException, Unit] =
+    List(JvmOptionsForNonJvmBuild).map(_.validate(options)).partition(_.isLeft) match {
+      case (Nil, _)  => Right(())
+      case (some, _) => some.head
```
alexarchambault: This discard all validation exceptions but the first one. Better keep them all via
```suggestion
      case (some, _) => Left(CompositeBuildException(some.flatMap(_.left.toSeq)))
```

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class ValidationException(
+  message: String,
+  positions: Seq[Position] = Nil
+) extends BuildException(message, positions)
+
+object JvmOptionsForNonJvmBuild extends BuildOptionsRule {
+  def validate(options: BuildOptions): Either[BuildException, Unit] = {
+    val jvmOptions = options.javaOptions.javaOpts.find(p => p.value.nonEmpty)
+    if (jvmOptions.nonEmpty && options.platform.value != scala.build.options.Platform.JVM)
```
alexarchambault, line 21: As I think I heard someone discuss that earlier today, this might be a problem if users just want to give native a try by passing `--native` on the command-line (while some of their sources have using directives adding Java options).

So we may want to not raise an error if the platform originates from the command-line (which can be checked via its position) while the Java options don't.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -251,4 +251,19 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
           )
         )
     }
+
+  test("JVM options only for JVM platform") {
+    val inputs = TestInputs(
+      Seq(os.rel / "Main.scala" -> "// using java-opt -Xss1g")
+    )
+    inputs.fromRoot { root =>
+      val res = os.proc(TestUtil.cli, "compile", extraOptions, "--native", ".").call(
```
alexarchambault: If you agree with my comment above, maybe we should keep this test but check that it doesn't error. And add one specifying the native platform via a using directive, that should error out.

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
tpasternak: I agree. Should we report a warning then, or just continue silently?

tpasternak: Ok, If I understand the next comment correctly, you would prefer a warning. :+1:

tpasternak:
I think I covered the issues. The problem now is that the warnings may be reported twice, as `prepareBuild` method is called more than one time

tpasternak:
Apart from the information about duplicate positions of settings in `Map[Platform, Positioned[Unit]]` is lost, do you think it's a big issue?

modules/build/src/main/scala/scala/build/options/ConfigMonoid.scala
```diff
@@ -81,9 +81,13 @@ object ConfigMonoid {
       main ++ defaults
   }
 
+  implicit def map[K, V]: ConfigMonoid[Map[K, V]] = instance(Map.empty[K, V]) {
+    (main, defaults) =>
+      main ++ defaults
+  }
```
alexarchambault: This discards the values in `main` whose keys are also in `defaults`, which is not what we want. Something like this should do:
```scala
implicit def map[K, V](implicit valueMonoid: ConfigMonoid[V]): ConfigMonoid[Map[K, V]] =
  instance(Map.empty[K, V]) { (main, defaults) =>
    (main.keySet ++ defaults.keySet)
      .iterator
      .map { k =>
        val v = valueMonoid.orElse(main.getOrElse(k, valueMonoid.zero), defaults.getOrElse(k, valueMonoid.zero))
        k -> v
      }
      .toMap
  }
```

modules/build/src/main/scala/scala/build/options/HashedField.scala
```diff
@@ -41,6 +41,17 @@ object HashedField extends LowPriorityHashedField {
...
 
+  implicit def map[K, V](
+    implicit
+    hasherK: Lazy[HashedType[K]],
+    hasherV: Lazy[HashedType[V]],
+    ordering: Ordering[K]
+  ): HashedField[Map[K, V]] = {
+    (name, map0, update) =>
+      for (t <- map0.keys.toVector.sorted(ordering))
+        update(s"$name+=${hasherK.value.hashedValue(t) + hasherV.value.hashedValue(map0(t))}")
```
alexarchambault: NIT (avoids calling `Map.apply`, and builds a single string rather than two)
```suggestion
      for ((k, v) <- map0.toVector.sortBy(_._1)(ordering))
        update(s"$name+=${hasherK.value.hashedValue(k)}${hasherV.value.hashedValue(v)}")
```

modules/build/src/main/scala/scala/build/options/ScalaOptions.scala
```diff
@@ -12,16 +12,16 @@ final case class ScalaOptions(
...
+  extraPlatforms: Map[Platform, Positioned[Unit]] = Map.empty
 ) {
   def normalize: ScalaOptions = {
     var opt = this
     for (sv <- opt.scalaVersion if opt.extraScalaVersions.contains(sv))
       opt = opt.copy(
         extraScalaVersions = opt.extraScalaVersions - sv
       )
-    for (pf <- opt.platform if opt.extraPlatforms.contains(pf))
+    for (pf <- opt.platform.map(_.value) if opt.extraPlatforms.keys.toList.contains(pf))
```
alexarchambault: No need to convert the key set to a list to only call `contains` on it:
```suggestion
    for (pf <- opt.platform.map(_.value) if opt.extraPlatforms.keySet.contains(pf))
```

modules/build/src/main/scala/scala/build/errors/Severity.scala
```diff
@@ -0,0 +1,8 @@
+package scala.build.errors
+
+sealed trait Severity
```
alexarchambault: I'm not sure we need that class (and the additional field in `BuildException`). `BuildException`s that should be considered as warnings should be reported via a different channel than the left side of a `Either[BuildException, …]`. As if there are only warnings, such an either should be `Right(…)`, and whatever was happening can go on.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -293,12 +296,11 @@ final case class BuildOptions(
...
       .extraPlatforms
       .toVector
-      .sorted
-    this +: sortedExtraPlatforms.map { pf =>
+    this +: sortedExtraPlatforms.map { case (pf, pos) =>
       copy(
         scalaOptions = scalaOptions0.copy(
-          platform = Some(pf),
-          extraPlatforms = Set.empty
+          platform = Some(Positioned(List(pos.positions.head), pf)),
```
alexarchambault: Why not this? (it doesn't discard positions this way)
```suggestion
          platform = Some(Positioned(pos.positions, pf)),
```

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
alexarchambault: Yeah, but we don't have a "channel" to report warnings yet (I mean, a way to pass around `BuildException`s that should be considered warnings, up to the user). So no warnings nor errors would be fine for now IMO.

modules/build/src/test/scala/scala/build/tests/TestLogger.scala
```diff
@@ -24,7 +24,7 @@ case class TestLogger(info: Boolean = true, debug: Boolean = false) extends Logg
     if (debug)
       System.err.println(s)
 
-  def log(ex: BuildException): Unit =
+  def log(ex: BuildException, severity: Severity): Unit =
```
alexarchambault: No need to pass a severity here, we should handle warnings later, in a more specific way.

modules/build/src/main/scala/scala/build/errors/Severity.scala
alexarchambault: That also makes invalid states "representable", as one could return a `Left(ex)` with `ex` having only warning severity, which would interrupt the processing while it shouldn't.

modules/build/src/main/scala/scala/build/options/ConfigMonoid.scala
tpasternak: Done

modules/build/src/main/scala/scala/build/errors/Severity.scala
tpasternak: I kept this because I extracted diagnostics logic in a separate class `Diagnostic` that is not derived from `BuildException`

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
tpasternak: It was printing a warning, not an error. I prepared an ad hoc solution to handle warnings. I added a check in the test to ensure the program is not failing and the message is reported with [warn]

modules/build/src/main/scala/scala/build/errors/BuildException.scala
```diff
@@ -2,6 +2,12 @@ package scala.build.errors
 
 import scala.build.Position
 
+case class Diagnostic(
+  val message: String,
+  val positions: Seq[Position] = Nil,
+  val severity: Severity
+)
```
alexarchambault: NIT No need to put `val` in case classes, and it's more convenient to put the fields with default values at the end:
```suggestion
case class Diagnostic(
  message: String,
  severity: Severity,
  positions: Seq[Position] = Nil
)
```

modules/build/src/main/scala/scala/build/errors/Severity.scala
alexarchambault: NIT To define enums in Scala, we usually extends `Product with Serializable`, and have all the sub-types be case classes or case objects (so that they have a nicer to string, behave better with type inference and matches, …). And `abstract class` generates less classes I think (or only used to maybe), and explicitly says this is not really an "interface", more the root of a class hierarchy.
```suggestion
sealed abstract class Severity extends Product with Serializable
```

modules/build/src/main/scala/scala/build/errors/Severity.scala
```diff
@@ -0,0 +1,8 @@
...
+
+object Severity {
+  object ERROR   extends Severity
+  object WARNING extends Severity
```
alexarchambault: NIT See above, and we tend not to use upper case in class names or enums, more camel case.
```suggestion
  case object Error   extends Severity
  case object Warning extends Severity
```

modules/build/src/main/scala/scala/build/options/validation/BuildOptionsRule.scala
```diff
@@ -0,0 +1,29 @@
...
+}
+
+object BuildOptionsRule {
+  def validateAll(options: BuildOptions): Seq[Diagnostic] =
+    List(JvmOptionsForNonJvmBuild).flatMap(_.validate(options))
+}
+
+class ValidationException(
+  diagnostic: Diagnostic
+) extends BuildException(diagnostic.message, diagnostic.positions) {}
```
alexarchambault: NIT
```suggestion
) extends BuildException(diagnostic.message, diagnostic.positions)
```

alexarchambault (commented):
Apart from the NITs, LGTM!

alexarchambault (approved):

tpasternak:
I had to force-push because of conflicts

alexarchambault (approved):

PR #410, opened 2021-11-23
Shebang args

tpasternak:


modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -192,9 +192,18 @@ final case class SharedOptions(
 
   lazy val coursierCache = coursier.coursierCache(logging.logger.coursierLogger)
 
+  def programArgs(args: RemainingArgs) = args.unparsed
+
+  def inputArgs(args: RemainingArgs) = args.remaining
```
alexarchambault: NIT Do we really need those methods? They don't use any of the fields of `SharedOptions`, are single-line, and are only called twice. Couldn't they just be inlined?

modules/cli/src/main/scala/scala/cli/commands/Shebang.scala
```diff
@@ -0,0 +1,26 @@
+package scala.cli.commands
+
+import caseapp._
+
+final case class ShebangOptions(
+  @Recurse
+  runOptions: RunOptions = RunOptions()
+) {}
```
alexarchambault: NIT
```suggestion
)
```

modules/cli/src/main/scala/scala/cli/commands/Shebang.scala
```diff
@@ -0,0 +1,26 @@
+package scala.cli.commands
+
+import caseapp._
+
+final case class ShebangOptions(
```
alexarchambault: NIT Could `ShebangOptions` be moved to its own file (`ShebangOptions.scala`), like we do for all other classes?

website/docs/reference/directives.md
```diff
@@ -71,6 +71,17 @@ Add Java options
...
+
+Specify default main class
+
+`using main-class `_main class_
+
+`using mainClass `_main class_
+
+#### Examples
+`using main-class helloWorld`
+
```
alexarchambault: Why does this get changed in this PR? It seems none of the other changes are related to that.

website/docs/reference/directives.md
```diff
@@ -106,6 +117,12 @@ Manually add a resource directory to the class path
 #### Examples
 `using resource "./resources"`
 
+### Scala Native options
+
+Add Scala Native options
+
+directive using usage
+
```
alexarchambault: Same as above, I wouldn't have expected that to be changed.

website/docs/reference/directives.md
tpasternak: This was probably because I didn't rebase to master

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
tpasternak: :+1: Done

modules/cli/src/main/scala/scala/cli/commands/Shebang.scala
tpasternak: Done

modules/cli/src/main/scala/scala/cli/commands/Shebang.scala
tpasternak: Done

alexarchambault (approved):

PR #412, opened 2021-11-23
Rename resource -> resourceDir

tpasternak:


lwronski (approved):
LGTM, I have just left one suggestion, add very short description in PR why we need this changes.

PR #415, opened 2021-11-24
Accept resources in Github Gists

### LINKED ISSUE #404: Should not compile text files in gist
# Steps

```
> scala-cli compile https://gist.github.com/adpi2/5d2f3d82324146a1ccf2ce938d99284f
https://gist.github.com/adpi2/5d2f3d82324146a1ccf2ce938d99284f/download
  No modified time in response
[warn] 5d2f3d82324146a1ccf2ce938d99284f-master/input:1:13: A pure expression does nothing in statement position; you may be omitting necessary parentheses
[warn] 5d2f3d82324146a1ccf2ce938d99284f-master/input:2:1: A pure expression does nothing in statement position; you may be omitting necessary parentheses
<elided>
```

# Problem

scala-cli tries to compile the `input` text file.

# Expectation

scala-cli shoud not compile text files in gist. It should only compile `.scala` and `.sc` files.

### LINKED ISSUE #405: Cannot run gist with resource file
## Steps

```
> scala-cli compile  https://gist.github.com/adpi2/9eaf82c559a7a24ab658ea7844cf84e3
[error] 9eaf82c559a7a24ab658ea7844cf84e3-master/day1.run.scala:2:4: Can't reference paths from sources from 9eaf82c559a7a24ab658ea7844cf84e3-master/day1.run.scala
```

## Problems

Scala-cli refuses to compile the `advent-of-code-2020-day1.scala` because of the `// using resource "./"` directive.

## Expectation

It should compile the `advent-of-code-2020-day1.scala` file and add the `input` file to its classpath.

## Notes

As an alternative I tried to open the `input` file as a file rather than a resource but there was another problem reported in https://github.com/VirtusLab/scala-cli/issues/404

lwronski:
closes #405

### DIFF
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -264,7 +264,7 @@ object Inputs {
   }
 
   private val githubGistsArchiveRegex: Regex =
-    s""":\\/\\/gist\\.github\\.com\\/[^\\/]*?\\/[^\\/]*$$""".r
+    s"""gist\\.github\\.com\\/[^\\/]*?\\/[^\\/]*$$""".r
 
   private def forNonEmptyArgs(
     args: Seq[String],
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -195,6 +195,8 @@ object Inputs {
       extends Virtual with AnyScalaFile
   final case class VirtualJavaFile(content: Array[Byte], source: String)
       extends Virtual with Compiled
+  final case class VirtualData(content: Array[Byte], source: String)
+      extends Virtual
 
   private def inputsHash(elements: Seq[Element]): String = {
     def bytes(s: String): Array[Byte] = s.getBytes(StandardCharsets.UTF_8)
@@ -226,7 +228,7 @@ object Inputs {
   }
 
   private def forValidatedElems(
-    validElems: Seq[Compiled],
+    validElems: Seq[Element],
     baseProjectName: String,
     directories: Directories
   ): Inputs = {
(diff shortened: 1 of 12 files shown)


romanowski (approved):

tpasternak:
Just a general question - would it be hard to accept any kind of zip file containing scala sources and treat gist as only a special case of it? Then we could write tests with local zips (build on-the-fly during the test) It would be much easier to maintain tests that do not refer to gist addresses.

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingResourcesDirectiveHandler.scala
```diff
@@ -43,12 +47,16 @@ case object UsingResourcesDirectiveHandler extends UsingDirectiveHandler {
...
+    val paths                     = DirectiveUtil.stringValues(values)
+    val paths0                    = rootOpt.map(root => paths.map(os.Path(_, root)))
+    val virtualPaths = virtualRootOpt.map(virtualRoot =>
+      paths.map(path => virtualRoot / os.SubPath(path)).map(_.toString())
+    )
     BuildOptions(
       classPathOptions = ClassPathOptions(
-        extraClassPath = paths0
+        extraClassPath = paths0.toList.flatten,
+        resourceVirtualDir = virtualPaths.toList.flatten
```
alexarchambault, line 59: NIT Why not keeping `os.SubPath`s here? It seems they are parsed again as `os.SubPath` above.

lwronski: I'm not very familiar with shapeless, but  if I use `os.SubPath` in `ClassPathOptions.resourceVirtualDir : Seq[os.SubPath]`, then it throw compilation error:

```
[error] /Users/lwronski/projects/scala-cli/modules/build/src/main/scala/scala/build/options/ClassPathOptions.scala:18:73: could not find implicit value for parameter gen: shapeless.LabelledGeneric.Aux[T,R]
[error]   implicit val hasHashData: HasHashData[ClassPathOptions] = HasHashData.derive
```


Unfortunately, I didn't fix it, so I used `Seq [String]` instead of `Seq[os.SubPaths]`

alexarchambault: It's a matter of having an implicit `HashedType` instance for `os.SubPath` I think. Something like this in the `HashedType` companion should do:
```scala
implicit val subPath: HashedType[os.SubPath] = {
  path => path.toString
}
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -262,7 +264,7 @@ object Inputs {
   }
 
   private val githubGistsArchiveRegex: Regex =
-    s""":\\/\\/gist\\.github\\.com\\/[^\\/]*?\\/[^\\/]*$$""".r
+    s"""gist\\.github\\.com\\/[^\\/]*?\\/[^\\/]*$$""".r
```
alexarchambault, line 267: That means, it also matches local directories with the same pattern, right? Maybe with proper gist-like zip support as input, we could get rid of that (and only match actual gist URLs here). But I guess this can be fixed in another PR.

alexarchambault (approved):

modules/build/src/main/scala/scala/build/Inputs.scala
lwronski: Right, it matches also local directories. I also think it can be done in a separate PR.

PR #417, opened 2021-11-25
Make sure buildServerCapabilities are set when the initialize happens

ckipp01:
I noticed today when looking at the BSP trace that these were empty during the initialize process, which can make clients think that scala-cli as a build server doesn't support anything. I know a lot of stuff is forwarded to Bloop and I _believe_ everything in there is accurate.


romanowski (approved):

romanowski:
@ckipp01, thanks for this PR!

PR #420, opened 2021-11-26
`scala-cli` script for development

tpasternak:
For scala-cli developers. Useful if you need a scala-cli exec that builds and runs code

### DIFF
--- a/scala-cli-src
+++ b/scala-cli-src
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
 ROOT_DIR="$SCRIPT_DIR"
 (
--- a/scala-cli-src
+++ b/scala-cli-src
@@ -1,0 +1,8 @@
+#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
+ROOT_DIR="$SCRIPT_DIR"
+(
+   cd "$ROOT_DIR" || exit
+   ./mill cli.standaloneLauncher
+)
+"$ROOT_DIR/out/cli/standaloneLauncher/dest/launcher" "$@"
--- a/scala-cli-src
+++ b/scala-cli-src
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -euo pipefail
 SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
 ROOT_DIR="$SCRIPT_DIR"
 (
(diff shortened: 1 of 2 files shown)


lwronski (commented):
In my opinion, we should use `standaloneLauncher`, it allows us to share `scala-cli` from the development build. 

We need also any mention about this script in Readme.

tpasternak:
We don't need to merge it, it was just useful for me. I can keep it locally

romanowski:
I like the idea! I would only rename the file to `scala-cli-mill` (or anything that would suggest that it will be run in jvm). Then, later we can add other script that along the lines of @lwronski idea

tpasternak:
I wanted to name it `scala-cli` in order to provide the experience of a released `scala-cli` after just exporting PATH. It's useful for presentations, screenshots etc.

alexarchambault:
We could put it at the root IMO, and name it… `scala-cli-src` for example?

And the current `scala-cli` script at the root of the repo could be removed IMO (unless anyone here uses it?). It's supposed to run the native image built in the sources (but only if it already exists, it doesn't build it itself).

tpasternak:
TBH I didn't notice the old script in root when I was putting this. I preferred it not to bin in a root because the main idea behind this is to  run`export PATH=$PATH:$root/tools/bin` and have the same experience as using the original scala-cli (work example shebangs work then, or you can just copy the exact command from a bug report)

alexarchambault:
It's fine to put it in the root IMO. Unless we have other such scripts lined up for a `tools/bin` directory?

scala-cli-src
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
alexarchambault: NIT
```suggestion
#!/usr/bin/env bash
```
(required on nix IIRC)

alexarchambault: Also, let's add at least `set -e` just in case:
```suggestion
#!/usr/bin/env bash
set -e
```

scala-cli-src
```diff
@@ -0,0 +1,8 @@
...
+SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
+ROOT_DIR="$SCRIPT_DIR"
+(
+   cd "$ROOT_DIR" || exit
+   ./mill cli.standaloneLauncher
+)
+"$ROOT_DIR/out/cli/standaloneLauncher/dest/launcher" "$@"
```
alexarchambault: We can use `exec` here, which will make the `scala-cli-src` invocation exit with the scala-cli exit code.
```suggestion
exec "$ROOT_DIR/out/cli/standaloneLauncher/dest/launcher" "$@"
```

alexarchambault (approved):

PR #425, opened 2021-11-28
Issue: 331 Added directive for native-version, native-compile, native-linking 

### LINKED ISSUE #331: Using directive for Scala Native options
Scala CLI allows to configure following [options for Scala Native:](https://github.com/VirtusLab/scala-cli/blob/master/modules/cli/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala#L17)

- nativeVersion
- nativeGC
- nativeLinking
- nativeCompile

We are missing using directives parser for them similar to [one for JavaOptions](https://github.com/VirtusLab/scala-cli/blob/master/modules/build/src/main/scala/scala/build/preprocessing/directives/UsingJavaOptionsDirectiveHandler.scala).

asjad02:
#331


alexarchambault (approved):
Thanks @asjad02!

PR #433, opened 2021-11-30
Depoly documentation only from stable branch

romanowski:
Release updates stable branch and every time stable branch is changed a PR to master is created ( or updated if exisits)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,6 @@ jobs:
         jvm: 11
     - name: Test documentation
       run: .github/scripts/check_docs.sh
-
     - name: Test gifs
       run: gifs/generate_gifs.sh $(ls gifs/scenarios/)
 
@@ -266,6 +265,11 @@ jobs:
         PGP_PASSWORD: ${{ secrets.PGP_PASSPHRASE }}
         SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
         SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+    - name: Update stable branch
+      uses: ad-m/github-push-action@8407731efefc0d8f72af254c74276b7a90be36e1
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        branch: stable
 
   launchers:
     timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


.github/workflows/website.yaml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  website-nightly:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: true
+    - uses: actions/setup-node@v2.5.0
+      if: env.SHOULD_PUBLISH == 'true'
```
lwronski:
```suggestion
```
I don't think we need this `if` because we always want to deploy docs on this branch.

.github/workflows/website.yaml
```diff
@@ -0,0 +1,31 @@
+name: Website deploy
+on:
+  push:
+    branches:
+    - stable
+
+jobs:
+  website-nightly:
```
lwronski:
```suggestion
  update-website:
```
I guess this CI is not the `nightly` type.

lwronski (approved):

.github/workflows/ci.yml
```diff
@@ -266,6 +265,11 @@ jobs:
         PGP_PASSWORD: ${{ secrets.PGP_PASSPHRASE }}
         SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
         SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+    - name: Update stable branch
+      uses: ad-m/github-push-action@master
```
alexarchambault: Not sure about using `master` for GitHub actions. I'd rather use a tag or a commit, so that we can quickly review what's new before bumping it (dependabot usually links the diff in its update PR), for security mainly.
```suggestion
      uses: ad-m/github-push-action@8407731efefc0d8f72af254c74276b7a90be36e1
```

alexarchambault (commented):
It'd be great if we had a way to push the latest website somewhere (like at https://scala-cli.virtuslab.org/latest/), to have a preview of it. But I guess we can work on that later.

romanowski:
> It'd be great if we had a way to push the latest website somewhere (like at https://scala-cli.virtuslab.org/latest/), to have a preview of it. But I guess we can work on that later.

Agree. I shouldn't be that hard to do I think. I can add it once this is completed and validated.

PR #437, opened 2021-11-30
Added more user-friendly error messages for when

aaronp:
Added a more user-friendly error message in the case where java-home (JAVA_HOME) is not set.

Reproducible by running scala-cli when JAVA_HOME is not set or otherwise specified

### DIFF
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -16,6 +16,7 @@ import scala.build.internal.{Constants, OsLibc, Util}
 import scala.build.options.validation.BuildOptionsRule
 import scala.build.{Artifacts, Logger, Os, Position, Positioned}
 import scala.util.Properties
+import scala.util.control.NonFatal
 
 final case class BuildOptions(
   scalaOptions: ScalaOptions = ScalaOptions(),
@@ -157,7 +158,18 @@ final case class BuildOptions(
           implicit val ec = finalCache.ec
           finalCache.logger.use {
             val path = javaHomeManager.get(jvmId).unsafeRun()
-            os.Path(path)
+            // can throw if path is empty or otherwise invalid
+            try os.Path(path)
+            catch {
+              // os.Path can throw AbsolutePathOutsideRoot or IllegalArgumentException.
+              // rather than be surgical in matching a snapshot-in-time of what might get thrown, this try/catch is
+              // a tight-enough scope where NonFatal should suffice and help our users
+              case NonFatal(err) =>
+                throw new IllegalStateException(
+                  s"JAVA_HOME (java-home) not set. See https://scala-cli.virtuslab.org/docs/reference/directives/#java-home for how to fix this",
+                  err
+                )
+            }
           }
         }
       }


modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -157,7 +158,18 @@ final case class BuildOptions(
           implicit val ec = finalCache.ec
           finalCache.logger.use {
             val path = javaHomeManager.get(jvmId).unsafeRun()
-            os.Path(path)
+            // can throw if path is empty or otherwise invalid
+            try os.Path(path)
```
alexarchambault, line 162: Rather than a try-catch, we can make `os.Path` less likely to throw by calling `os.Path(path, Os.pwd)`.

alexarchambault (commented):
Thanks for opening this!

I'm not sure this should be fixed here or this way… [coursier-jvm](https://github.com/coursier/coursier/tree/master/modules/jvm) is supposed to return either valid paths of its own, or the value of `JAVA_HOME`.

What value of `JAVA_HOME` is a problem in your case? Is it an empty value (could be ignored by coursier-jvm then), or a relative path (can be fixed by us here, see my comment)?

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
aaronp: Thanks @alexarchambault for reviewing the PR.

I don't doubt there's a much better work-around -- the issue in this case was that the 'JAVA_HOME' environment variable wasn't set (nor supplied as a param to scala-cli).

The upshot was a colleague was trying to use my scala-cli script and seeing a very obscure-looking error message:

``` requirement failed:  is not an absolute path```

This PR is an attempt to simply approve on that so other people don't have to pull down the source to figure out why their scripts are mysteriously failing on different environments:


image

Comment (aaronp) [reply modules/build/src/main/scala/scala/build/options/BuildOptions.scala:162]: It's your world - I just live in it (and am a big fan already of scala-cli). I would see this not as an "either/or", but a "both/and".

If any future work-arounds do indeed work, then discussions about the 'catch' clause are moot ... it's just a bit of code noise.

If there *is* a user's environment where that line (which comes from a 3rd party lib) fails, then this just makes life that much nicer for the end users

Comment (aaronp) [reply modules/build/src/main/scala/scala/build/options/BuildOptions.scala:162]: Hello @alexarchambault - Just wanted to see if this could be added based on the above rationale?

**Comment by alexarchambault** [comment]
So I was able to reproduce that locally, with an empty `JAVA_HOME`:
```text
$ JAVA_HOME= scala-cli test examples/tests
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed:  is not an absolute path
	at scala.Predef$.require(Predef.scala:281)
	at os.Path.<init>(Path.scala:430)
	at os.Path$.apply(Path.scala:388)
	at scala.build.options.BuildOptions.$anonfun$javaHomeLocation$2(BuildOptions.scala:173)
	at coursier.cache.CacheLogger.use(CacheLogger.scala:49)
	at coursier.cache.CacheLogger.use$(CacheLogger.scala:47)
	at coursier.cache.loggers.RefreshLogger.use(RefreshLogger.scala:209)
	at scala.build.options.BuildOptions.$anonfun$javaHomeLocation$1(BuildOptions.scala:171)
	at scala.Option.getOrElse(Option.scala:189)
	at scala.build.options.BuildOptions.javaHomeLocation(BuildOptions.scala:169)
	at scala.build.options.BuildOptions.javaCommand0$lzycompute(BuildOptions.scala:135)
	at scala.build.options.BuildOptions.javaCommand0(BuildOptions.scala:134)
	at scala.build.options.BuildOptions.javaHome(BuildOptions.scala:223)
	at scala.cli.commands.SharedOptions.bloopRifleConfig(SharedOptions.scala:182)
	at scala.cli.commands.Test$.run(Test.scala:32)
	at scala.cli.commands.Test$.run(Test.scala:12)
	at caseapp.core.app.CaseApp.main(CaseApp.scala:149)
	at caseapp.core.app.CommandsEntryPoint.main(CommandsEntryPoint.scala:114)
	at scala.cli.ScalaCli$.main(ScalaCli.scala:99)
	at scala.cli.ScalaCli.main(ScalaCli.scala)
```

I'm assuming an empty `JAVA_HOME` is the culprit here (which is different than no `JAVA_HOME`, which can be achieved with `unset JAVA_HOME`).

This needs to be fixed in coursier, I opened https://github.com/coursier/coursier/issues/2292 to not forget.

In the mean time, we can add workarounds for it in Scala CLI. Let me try to pinpoint exactly where we allow coursier to look for the system JVM, in Scala CLI.

alexarchambault:
@aaronp https://github.com/VirtusLab/scala-cli/pull/475 should have a workaround for the root of the issue you're seeing here.

The call you changed here is not supposed to fail, it was only a manifestation of an upstream bug.

PR #439, opened 2021-12-01
Parse JVM 9 version correctly

tpasternak:


modules/bloop-rifle/src/main/scala/scala/build/blooprifle/VersionUtil.scala
```diff
@@ -1,10 +1,27 @@
...
-  }
+
+  /** @param jvmVersion,
+    *   for example '1.8.0.22' or '11.0.2'
+    * @return
+    *   jvm release version (8, 11)
+    */
+  val jvmReleaseRegex = "(1[.])?(\\d+)"
+  def jvmRelease(jvmVersion: String): Option[Int] =
+    jvmReleaseRegex.r.findFirstMatchIn(jvmVersion).map(_.group(2)).map(_.toInt)
```
lwronski:
```suggestion
    jvmReleaseRegex.r.findFirstMatchIn(jvmVersion).flatMap(v => Option(v.group(2))).map(_.toInt)
```

lwronski (approved):

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/VersionUtil.scala
tpasternak: `null` was impossible there, as the group was required by regex, but I agree we can explicitly handle this case

lwronski (approved):

romanowski (approved):

PR #444, opened 2021-12-02
Bloop workspace dir

### LINKED ISSUE #443: Current directory when running via BSP is set to `.scala`
To reproduce:

Run below snippet from within metals and from cli.

```scala
// using lib com.lihaoyi::os-lib:0.7.8

@main
def main =
  println(os.pwd) 
end main
```

lwronski:
Set workspace directory to current dir where scala-cli is running in BSP.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
@@ -1234,4 +1234,21 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
       expect(p.out.text().trim == "hello")
     }
   }
+
+  test("workspace dir") {
+    val inputs = TestInputs(
+      Seq(
+        os.rel / "Hello.scala" ->
+          """|// using lib com.lihaoyi::os-lib:0.7.8
+             |
+             |object Hello extends App {
+             |  println(os.pwd) 
+             |}""".stripMargin
+      )
+    )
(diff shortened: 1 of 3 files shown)


alexarchambault (approved):
If the CI is green, LGTM!

romanowski (approved):

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1234,4 +1234,21 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+        os.rel / "Hello.scala" ->
+          """|// using lib com.lihaoyi::os-lib:0.7.8
+             |
+             |object Hello extends App {
+             |  println(os.pwd) 
+             |}""".stripMargin
+      )
+    )
+    inputs.fromRoot { root =>
+      val p = os.proc(TestUtil.cli, "Hello.scala").call(cwd = root)
```
tpasternak, line 1250: I'm not sure if this test really tests the code that has been change. When you run `scala-cli run ... ` it takes the cwd from `Inputs`, which just takes os.Pwd https://github.com/VirtusLab/scala-cli/blob/75b04cc08d265423cd608f3cebc75d18ba181dd4/modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala#L218

So it passes even without the change. In order to test it, you need to run `scala-cli bsp`

tpasternak:
I would also test if clean works correctly after this change

PR #449, opened 2021-12-03
Move Scala Native linking to runtime fetched scala-native-cli process

### LINKED ISSUE #261: Running with Scala Native 0.4.1 fails
I know, I know... it's just been released :D 

```
❯ cat native.scala
object E extends App {
  println("hello!")
}

❯ scala-cli version
0.0.7

❯ scala-cli clean native.scala && scala-cli run native.scala -S 2.13 --native --native-version 0.4.1
Compiling project (Scala 2.13.6, Scala Native)
Compiled project (Scala 2.13.6, Scala Native)
Exception in thread "main" scala.MatchError: 170 (of class java.lang.Integer)
	at scala.scalanative.nir.serialization.BinaryDeserializer.getInst(BinaryDeserializer.scala:106)
	at scala.scalanative.nir.serialization.BinaryDeserializer.$anonfun$getInsts$1(BinaryDeserializer.scala:103)
	at scala.scalanative.nir.serialization.BinaryDeserializer.$anonfun$getSeq$1(BinaryDeserializer.scala:41)
	at scala.scalanative.nir.serialization.BinaryDeserializer.$anonfun$getSeq$1$adapted(BinaryDeserializer.scala:41)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.immutable.Range.foreach(Range.scala:158)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at scala.scalanative.nir.serialization.BinaryDeserializer.getSeq(BinaryDeserializer.scala:41)
	at scala.scalanative.nir.serialization.BinaryDeserializer.getInsts(BinaryDeserializer.scala:103)
	at scala.scalanative.nir.serialization.BinaryDeserializer.getDefn(BinaryDeserializer.scala:168)
	at scala.scalanative.nir.serialization.BinaryDeserializer.$anonfun$deserialize$1(BinaryDeserializer.scala:35)
```

## Expectation

Works as with 0.4.0:

```
❯ scala-cli clean native.scala && scala-cli run native.scala -S 2.13 --native --native-version 0.4.0
Compiling project (Scala 2.13.6, Scala Native)
Compiled project (Scala 2.13.6, Scala Native)
Linking (1772 ms)
Discovered 974 classes and 7887 methods
Generating intermediate code (1416 ms)
Produced 16 files
Compiling to native code (2167 ms)
Linking native code (immix gc, none lto) (193 ms)
Total (6075 ms)
hello!
```

jchyb:
This is a PR meant to showcase changes to Scala Native linking before they have to be finalized so that no grave mistakes will be made on the scala-native-cli side. The most interesting part is in buildNative in Package.scala, where we run the linking process. Other changes concern redesigning Scala Native logging (we move it to scala-native-cli) and adjusted caching to accommodate the new structure.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Artifacts.scala
+++ b/modules/build/src/main/scala/scala/build/Artifacts.scala
@@ -164,11 +164,10 @@ object Artifacts {
         None
     }
 
-    val scalaNativeCli = fetchedScalaNativeCli match {
-      case Some(fetched) => fetched.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
-          f.toPath
-        }
-      case None => Nil
+    val scalaNativeCli = fetchedScalaNativeCli.toSeq.flatMap { fetched =>
+      fetched.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
+        f.toPath
+      }
     }
 
     val extraStubsJars =
--- a/modules/build/src/main/scala/scala/build/Artifacts.scala
+++ b/modules/build/src/main/scala/scala/build/Artifacts.scala
@@ -25,6 +25,7 @@ final case class Artifacts(
   compilerArtifacts: Seq[(String, Path)],
   compilerPlugins: Seq[(AnyDependency, String, Path)],
   dependencies: Seq[AnyDependency],
+  scalaNativeCli: Seq[Path],
   detailedArtifacts: Seq[(CsDependency, csCore.Publication, csUtil.Artifact, Path)],
   extraClassPath: Seq[Path],
   extraCompileOnlyJars: Seq[Path],
@@ -72,6 +73,7 @@ object Artifacts {
     addJvmTestRunner: Boolean,
     addJsTestBridge: Option[String],
     addJmhDependencies: Option[String],
+    scalaNativeCliVersion: Option[String],
     extraRepositories: Seq[String],
     logger: Logger
   ): Either[BuildException, Artifacts] = either {
@@ -117,6 +119,11 @@ object Artifacts {
         Nil
     }
 
+    val scalaNativeCliDependency =
+      scalaNativeCliVersion.map(version =>
+        Seq(dep"org.scala-native:scala-native-cli_2.12:$version")
+      )
+
     val allExtraRepositories = maybeSnapshotRepo ++ extraRepositories
 
     val updatedDependencies =
@@ -138,6 +145,30 @@ object Artifacts {
         logger,
         classifiersOpt = Some(Set("_") ++ (if (fetchSources) Set("sources") else Set.empty))
       )
+    }
+
+    val fetchedScalaNativeCli = scalaNativeCliDependency match {
+      case Some(dependency) =>
+        Some(
+          value {
+            fetch(
+              Positioned.none(dependency),
+              Nil,
+              params,
+              logger,
+              None
+            )
+          }
+        )
+      case None =>
+        None
+    }
+
+    val scalaNativeCli = fetchedScalaNativeCli match {
+      case Some(fetched) => fetched.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
+          f.toPath
+        }
+      case None => Nil
     }
 
     val extraStubsJars =
@@ -171,6 +202,7 @@ object Artifacts {
       compilerArtifacts,
       compilerPlugins0,
       updatedDependencies.map(_.value),
+      scalaNativeCli,
       fetchRes.fullDetailedArtifacts.collect { case (d, p, a, Some(f)) => (d, p, a, f.toPath) },
       extraClassPath ++ extraStubsJars,
       extraCompileOnlyJars,
--- a/modules/build/src/main/scala/scala/build/errors/ScalaNativeBuildError.scala
+++ b/modules/build/src/main/scala/scala/build/errors/ScalaNativeBuildError.scala
@@ -1,0 +1,4 @@
+package scala.build.errors
+
+final class ScalaNativeBuildError()
+    extends BuildException(s"Error compiling with Scala Native")
--- a/modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
@@ -4,9 +4,10 @@ import java.math.BigInteger
 import java.security.MessageDigest
 
 import scala.build.Build
-import scala.scalanative.{build => sn}
 
 object NativeBuilderHelper {
+
+  case class SNCacheData(val changed: Boolean, val projectSha: String)
 
   private def resolveProjectShaPath(nativeWorkDir: os.Path) = nativeWorkDir / ".project_sha"
   private def resolveOutputShaPath(nativeWorkDir: os.Path)  = nativeWorkDir / ".output_sha"
@@ -20,10 +21,10 @@ object NativeBuilderHelper {
     String.format(s"%040x", calculatedSum)
   }
 
-  private def projectSha(build: Build.Successful, nativeConfig: sn.NativeConfig) = {
+  private def projectSha(build: Build.Successful, config: List[String]) = {
     val md = MessageDigest.getInstance("SHA-1")
     md.update(build.inputs.sourceHash().getBytes)
-    md.update(nativeConfig.toString.getBytes)
+    md.update(config.toString.getBytes)
     md.update(Constants.version.getBytes)
     md.update(build.options.hash.getOrElse("").getBytes)
 
@@ -32,22 +33,29 @@ object NativeBuilderHelper {
     String.format(s"%040x", calculatedSum)
   }
 
-  def updateOutputSha(dest: os.Path, nativeWorkDir: os.Path) = {
+  def updateProjectAndOutputSha(
+    dest: os.Path,
+    nativeWorkDir: os.Path,
+    currentProjectSha: String
+  ) = {
+    val projectShaPath = resolveProjectShaPath(nativeWorkDir)
+    os.write.over(projectShaPath, currentProjectSha, createFolders = true)
+
     val outputShaPath = resolveOutputShaPath(nativeWorkDir)
     val sha           = fileSha(dest)
     os.write.over(outputShaPath, sha)
   }
 
-  def shouldBuildIfChanged(
+  def getCacheData(
     build: Build.Successful,
-    nativeConfig: sn.NativeConfig,
+    config: List[String],
     dest: os.Path,
     nativeWorkDir: os.Path
-  ): Boolean = {
+  ): SNCacheData = {
     val projectShaPath = resolveProjectShaPath(nativeWorkDir)
     val outputShaPath  = resolveOutputShaPath(nativeWorkDir)
 
-    val currentProjectSha = projectSha(build, nativeConfig)
+    val currentProjectSha = projectSha(build, config)
     val currentOutputSha  = if (os.exists(dest)) Some(fileSha(dest)) else None
 
     val previousProjectSha = if (os.exists(projectShaPath)) Some(os.read(projectShaPath)) else None
@@ -58,9 +66,6 @@ object NativeBuilderHelper {
       previousOutputSha != currentOutputSha ||
       !os.exists(dest)
 
-    // update sha in .projectShaPath
-    if (changed) os.write.over(projectShaPath, currentProjectSha, createFolders = true)
-
-    changed
+    SNCacheData(changed, currentProjectSha)
   }
 }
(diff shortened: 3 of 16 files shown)


jchyb:
I would like to confirm the way we handle reporting with the new Scala Native building structure. Previously when building SN programs scala-cli had a direct access to a SN Logger object, similarly as with JVM and JS. When running scala-native-cli (nativeTools cli api) as a separate process this is of course impossible. For now, scala-native-cli simply uses a default SN logger, with added options for individually filtering out log levels (debug (on stderr), info (on stdout), warn (on stdout), error (on stderr)). Scala-cli verbosity levels are mapped to that filtered logger. SN-cli is then run with IO inherited by scala-cli - this means we really have no direct access to errors thrown by SN (although they are being shown of course) and everything is reported with tags (like [info]...). This is the kind of thing that is unimportant in the short run, nor would I personally consider problematic in any way, but if we don't confirm it now I may not be able to change that later, after sn-cli will be published. Are those changes ok?

EDIT 21.12.21: I forgot we always display tags with logs. It should be easy to just catch stdout/stderr and control output manually

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -140,6 +147,30 @@ object Artifacts {
...
+    val fetchedScalaNativeCli = scalaNativeCliDependency match {
+      case Some(dependency) =>
+        Some(
+          value {
+            fetch(
+              Positioned.none(dependency),
+              Nil,
+              params,
+              logger,
+              Some(Set("_") ++ (if (fetchSources) Set("sources") else Set.empty)) // TODO not sure
```
alexarchambault: This isn't a dependency injected in the class path "seen" by users, no need to fetch sources.
```suggestion
              None
```

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -140,6 +147,30 @@ object Artifacts {
...
+      case None =>
+        None
+    }
+
+    val scalaNativeCli = fetchedScalaNativeCli match {
+      case Some(fetched) => fetched.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
+          f.toPath
+        }
+      case None => Nil
+    }
```
alexarchambault: NIT
```suggestion
    val scalaNativeCli = fetchedScalaNativeCli.toSeq.flatMap { fetched =>
      fetched.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
        f.toPath
      }
    }
```

modules/build/src/main/scala/scala/build/internal/NativeBuilderHelper.scala
```diff
@@ -4,10 +4,11 @@ import java.math.BigInteger
 import java.security.MessageDigest
 
 import scala.build.Build
-import scala.scalanative.{build => sn}
 
 object NativeBuilderHelper {
 
+  case class SNCacheData(val changed: Boolean, val projectSha: String)
```
alexarchambault: NIT No need of `val`s in case classes:
```suggestion
  case class SNCacheData(changed: Boolean, projectSha: String)
```

modules/build/src/main/scala/scala/build/options/ScalaNativeOptions.scala
```diff
@@ -79,19 +97,24 @@ final case class ScalaNativeOptions(
...
-      .withClangPP(clangppPath())
-      .withLinkingOptions(linkingOptions)
-      .withCompileOptions(compileOptions)
+  def configCliOptions(): List[String] =
+    gcCliOption()
+      .++(modeCliOption())
+      .++(clangCliOption())
+      .++(clangppCliOption())
+      .++(linkingCliOptions())
+      .++(compileCliOptions())
```
alexarchambault: NIT
```suggestion
    gcCliOption() ++
      modeCliOption() ++
      clangCliOption() ++
      clangppCliOption() ++
      linkingCliOptions() ++
      compileCliOptions()
```

modules/build/src/test/scala/scala/build/tests/NativeBuilderHelperTests.scala
```diff
@@ -43,117 +42,117 @@ class NativeBuilderHelperTests extends munit.FunSuite {
     inputs.withBuild(defaultOptions, buildThreads, bloopConfig) { (root, _, maybeBuild) =>
       val build = maybeBuild.toOption.get.successfulOpt.get
 
-      val nativeConfig  = build.options.scalaNativeOptions.config
+      val config        = build.options.scalaNativeOptions.configCliOptions
```
alexarchambault: NIT
```suggestion
      val config        = build.options.scalaNativeOptions.configCliOptions()
```

modules/build/src/test/scala/scala/build/tests/NativeBuilderHelperTests.scala
```diff
@@ -43,117 +42,117 @@ class NativeBuilderHelperTests extends munit.FunSuite {
...
+      expect(cacheData.changed)
     }
   }
 
   test("should not rebuild the second time") {
     inputs.withBuild(defaultOptions, buildThreads, bloopConfig) { (root, _, maybeBuild) =>
       val build = maybeBuild.toOption.get.successfulOpt.get
 
-      val nativeConfig  = build.options.scalaNativeOptions.config
+      val config        = build.options.scalaNativeOptions.configCliOptions
```
alexarchambault: NIT (same for the others below)
```suggestion
      val config        = build.options.scalaNativeOptions.configCliOptions()
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -537,28 +537,51 @@ object Package extends ScalaCommand[PackageOptions] {
...
+        build,
+        cliOptions,
+        dest,
+        nativeWorkDir
+      ) // TODO add main class
 
-        NativeBuilderHelper.updateOutputSha(dest, nativeWorkDir)
+    if (cacheData.changed)
+      withLibraryJar(build, dest.last.stripSuffix(".jar")) { mainJar =>
+        println(mainJar)
```
alexarchambault: ?
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -537,28 +537,51 @@ object Package extends ScalaCommand[PackageOptions] {
...
+          "--outpath",
+          dest.toString(),
+          "--main",
+          mainClass + "$",
+          "--workdir",
+          nativeWorkDir.toString()
+        ) ++ cliOptions ++ logger.scalaNativeCliInternalLoggerOptions
+
+        args.map(println(_))
+
```
alexarchambault: ?
```suggestion
```

modules/build/src/main/scala/scala/build/errors/ScalaNativeBuildError.scala
```diff
@@ -0,0 +1,5 @@
+package scala.build.errors
+
+final class ScalaNativeBuildError(
+  caughtError: String
+) extends BuildException(s"Error compiling with Scala Native:\n${caughtError}")
```
alexarchambault: Let's not pass an argument here, if we don't really use it yet.
```suggestion
final class ScalaNativeBuildError extends BuildException("Error compiling with Scala Native.")
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -537,28 +537,51 @@ object Package extends ScalaCommand[PackageOptions] {
...
+            build.artifacts.scalaNativeCli.map(_.toFile),
+            "scalanativecli.ScalaNativeCli",
+            args,
+            logger
+          )
+        if (exitCode == 0)
+          NativeBuilderHelper.updateProjectAndOutputSha(dest, nativeWorkDir, cacheData.projectSha)
+        else
+          // TODO Throw caught stderr output
+          throw new ScalaNativeBuildError("TODO")
```
alexarchambault:
```suggestion
          throw new ScalaNativeBuildError
```

modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala
```diff
@@ -156,6 +156,15 @@ class CliLogger(
       def error(msg: String)    = logger.log(msg)
     }
 
+  val scalaNativeCliInternalLoggerOptions: List[String] = {
+    import scala.build.options.ScalaNativeOptions.Logger._
+
+    if (quiet) List(noDebug, noInfo, noWarn)
+    else if (verbosity >= 2) Nil
+    else if (verbosity == 1) List(noDebug)
+    else List(noDebug, noWarn)
```
alexarchambault: Let's only look at `verbosity`, that makes the code less ambiguous (not sure why we pass `quiet` to `CliLogger`):
```suggestion
    if (verbosity < 0) List(noDebug, noInfo, noWarn)
    else if (verbosity == 0) List(noDebug, noWarn)
    else if (verbosity == 1) List(noDebug)
    else Nil
```

alexarchambault (commented):
Nice! I only left NIT comments.

The CI errors may be legit though… Feel free to add a commit replacing all the `maybeBuild.toOption.get` in the tests by `maybeBuild.fold(throw _, identity)`, so that the underlying errors aren't trapped.

alexarchambault:
@jchyb The logging stuff should be fine for now… I guess we can always tweak the Scala Native CLI itself later, to change how they're printed?

jchyb:
@alexarchambault Thank you for the review! About the logging, we are planning to release one scala-native-cli per SN version (so now we are about to release 0.4.0, 0.4.1 and 0.4.2 and finalizing things), I think we can change some things for later versions, but we probably are not going to revisit the 0.4.x ones (then again, I imagine dropping support for older SN versions is not going to be a problem from the scala-cli side).

modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala
jchyb: Actually, could we do it like this? I think we are going to want to keep the log level hierarchy intact on the sn-cli side. I think it's necessary to have the `info` log level as the default, but that only leaves us with the `debug` logs as more verbose.

``` scala
    if (verbosity < 0) List(noDebug, noInfo) // warn
    else if (verbosity == 0) List(noDebug) // info
    else Nil // debug
```
I imagine I could sneak some options to filter out the logs (like it is done in this PR), but I'd like to confirm that it is necessary to do so

alexarchambault: Sure!

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -115,7 +115,9 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
     )
     inputs.fromRoot { root =>
       val output =
-        os.proc(TestUtil.cli, extraOptions, fileName, "--native").call(cwd = root).out.text().trim
+        os.proc(TestUtil.cli, extraOptions, fileName, "--native", "-q")
```
alexarchambault, line 118: Why adding a `-q` here (and below)?

alexarchambault (approved):
Apart from my (minor) question, LGTM!

That said, maybe for a later PR, I wish the discover stuff could be handled by scala-native-cli itself… In Scala CLI, we use "discovered" things in two places:
- to pass options to scala-native-cli
- to write the native config in bloop config files

Bloop actually accepts "empty" values, and can discover things on its own. If scala-native-cli handled discovery, we could get rid of it in Scala CLI (and drop the scala-native-tools dependency). Only `--native-compile-defaults` and `--native-linking-defaults` might be a problem (as these can't be put in a bloop config file), but we might be able to live without that (this part of the bloop config is only used from Metals, when it runs Scala CLI things via BSP, and ultimately, we'd like to handle this ourselves rather than delegating it to Bloop).

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
jchyb: We now print linker [info] logs by default, and so we disable them by the `-q`, therefore getting only the expected program output compared in the test

alexarchambault: Oh ok. (Now I remember you had actually mentioned that to me in other channels, IIRC…)

Issue #259, opened 2021-10-21
Consistent --help and --version flag or commands


olafurpg:
It would be nice if all combination of `help/-help/--help` and `version/-version/--version` were supported. Currently, some of the combinations return error messages.
```
❯ scala-cli version
0.0.7
❯ scala-cli help
help: not found
❯ scala-cli --version
Unrecognized argument: --version
❯ scala-cli --help
Usage: scala-cli <COMMAND> [options]
Compile, run, package Scala code.

Help options:
  --usage                   Print usage and exit
  -h, -help, --help         Print help message and exit
  --help-full, --full-help  Print help message, including hidden options, and exit

Scala options:
  --js                                           Enable Scala.JS
...
```

PR #454, opened 2021-12-06
Unify 'help' and 'version' commands

tpasternak:


tpasternak:
closes #259

tpasternak:
@romanowski should we also support single-hyphen command? `-help` and `-version`?

romanowski:
I think `-help` works right now. Adding support for `-version` may make dangerous precedence and we will need to add single-hypen version to all option so I am leaning towards not adding support for `-version`. @alexarchambault what do you think?

alexarchambault:
> I think `-help` works right now. Adding support for `-version` may make dangerous precedence and we will need to add single-hypen version to all option so I am leaning towards not adding support for `-version`. @alexarchambault what do you think?

I don't really mind… It could be added if we supported `--version` already, which we don't (only `scala-cli version`).

tpasternak:
This is quite interesting, because we support single-hyphen commands only in order to keep compatibility with `scalac` command. But if we added `-version` it would probably return `scala-cli`'s version, not the underlying `scalac`'s

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
 import caseapp._
 import caseapp.core.Error
 
+import scala.build.internal
+//import scala.cli.commands.SharedOptions
+
+case class DefaultOptions(
```
alexarchambault: NIT We could move that to its own file, like the other options.

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
 import caseapp._
 import caseapp.core.Error
 
+import scala.build.internal
+//import scala.cli.commands.SharedOptions
```
alexarchambault: NIT
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
...
+case class DefaultOptions(
+  @Recurse
+  runOptions: RunOptions = RunOptions(),
+  version: Option[Boolean] = None
+)
+
+object DefaultOptions {
+  implicit lazy val parser: Parser[DefaultOptions] = Parser.derive
+  implicit lazy val help: Help[DefaultOptions]     = Help.derive
+//  implicit lazy val jsonCodec: ReadWriter[DefaultOptions] = macroRW
```
alexarchambault:
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
 import caseapp._
 import caseapp.core.Error
 
+import scala.build.internal
+//import scala.cli.commands.SharedOptions
+
+case class DefaultOptions(
+  @Recurse
+  runOptions: RunOptions = RunOptions(),
+  version: Option[Boolean] = None
```
alexarchambault: FYI, I we want to support `-version` (single dash), I think the following would work
```suggestion
  @Name("-version")
    version: Option[Boolean] = None
```

alexarchambault: Why an `Option[Boolean]`? A simple `Boolean` should do (defaulting to `false`).

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
...
     sys.exit(0)
   }
   override def fullHelpAsked(progName: String): Nothing = {
     println(defaultFullHelp)
     sys.exit(0)
   }
-  def run(options: RunOptions, args: RemainingArgs): Unit =
-    if (anyArgs)
+  def run(options: DefaultOptions, args: RemainingArgs): Unit =
+    if (options.version.isDefined) {
```
alexarchambault: If this remains an `Option`, we should check its value like
```suggestion
    if (options.version.contains(true)) {
```
Otherwise, `--version=false` would make it print the version nonetheless.

modules/cli/src/main/scala/scala/cli/commands/DefaultBase.scala
```diff
@@ -3,28 +3,48 @@ package scala.cli.commands
...
+      println(internal.Constants.version)
+      sys.exit(0)
```
alexarchambault: No need of the `sys.exit` it seems.
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/HelpCmd.scala
```diff
@@ -0,0 +1,16 @@
+package scala.cli.commands
+
+import caseapp._
+
+import scala.cli.ScalaCli
+case class HelpOptions()
```
alexarchambault: Just like for the other options, this can be moved to its own file.

modules/cli/src/main/scala/scala/cli/commands/HelpCmd.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object HelpOptions {
+  implicit lazy val parser: Parser[HelpOptions] = Parser.derive
+  implicit lazy val help: Help[HelpOptions]     = Help.derive
+}
+
+object HelpCmd extends ScalaCommand[HelpOptions] {
```
alexarchambault, line 7: Something like `override def names = Seq(Seq("help"))` might be needed here, for `scala-cli help` to work.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -103,10 +103,14 @@ abstract class ScalaCommand[T](implicit parser: Parser[T], help: Help[T])
...
-          if (coursier.paths.Util.useJni()) {
-            val size = coursier.jniutils.WindowsAnsiTerminal.terminalSize()
-            Some(size.getWidth)
-          }
+          if (coursier.paths.Util.useJni())
+            Try(coursier.jniutils.WindowsAnsiTerminal.terminalSize()).toOption.map(
+              _.getWidth
+            ).orElse {
+              val fallback = 120
+              System.err.println(s"Could not get terminal width, falling back to $fallback")
```
alexarchambault: Let's only print that… if a Java option is defined say (just in case, to help debugging), as this may be quite annoying if it's printed for every `scala-cli` run.
```suggestion
              if (java.lang.Boolean.getBoolean("scala.cli.windows-terminal.verbose"))
                System.err.println(s"Could not get terminal width, falling back to $fallback")
```

modules/cli/src/main/scala/scala/cli/commands/HelpOptions.scala
```diff
@@ -0,0 +1,10 @@
+package scala.cli.commands
+
+import caseapp._
+
+case class HelpOptions()
```
alexarchambault, line 6:
```suggestion
@HelpMessage("Print help message")
case class HelpOptions()
```

This should give a description to the `help` command in the help message, like
```text
Other commands:
  bsp                                       Start BSP server
  export                                    Export current project to sbt or Mill
  help                                      Print help message
  install completions, install-completions  Installs completions into your shell
  setup-ide                                 Generate a BSP file that you can import into your IDE
  shebang                                   This command is an equivalent of `run`, but it changes the way how
  update                                    Update scala-cli - it works only for installation script
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1270,4 +1270,31 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
       expect(p.out.text().trim == root.toString)
     }
   }
+
+  test("help command") {
+    for { helpOption <- Seq("help", "-help", "--help") } {
+      pprint.log(os.proc(TestUtil.cli, helpOption).call(
+        check = false,
+        mergeErrIntoOut = true
+      ).out.text())
```
alexarchambault: Isn't this just for debugging?
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1270,4 +1270,31 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+      assert(help.exitCode == 0, clues(helpOption, help.out.text(), help.err.text(), help.exitCode))
+      expect(help.out.text().contains("Usage:"))
+    }
+  }
+
+  test("version command") {
+    // tests if the format is correct instead of comparing to a version passed via Constants
+    // in order to catch errors in Mill configuration, too
+    val versionRegex = ".*\\d+[.]\\d+[.]\\d+.*".r
+    for { versionOption <- Seq("version", "--version") } {
```
alexarchambault: NIT
```suggestion
    for (versionOption <- Seq("version", "--version")) {
```

alexarchambault (approved):

PR #458, opened 2021-12-06
Add projects gif

lwronski:


tpasternak (approved):
Personally, I'd prefer to run the server in background and curl it,but this is OK, too

PR #460, opened 2021-12-07
Accept zip as a input

lwronski:


romanowski (approved):

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -286,43 +321,16 @@ object Inputs {
         if (isStdin) Right(Seq(VirtualScalaFile(stdinOpt0.get, "<stdin>")))
         else if ((arg == "-" || arg == "-.sc" || arg == "_.sc") && stdinOpt0.nonEmpty)
           Right(Seq(VirtualScript(stdinOpt0.get, "stdin", os.sub / "stdin.sc")))
+        else if (arg.contains(".zip") && os.exists(os.Path(arg, cwd))) {
```
alexarchambault: `endsWith` should be enough (`contains` may match stuff like `foo/aa.zipp/b/c.scala`):
```suggestion
        else if (arg.endsWith(".zip") && os.exists(os.Path(arg, cwd))) {
```
(we may even want it to be case-insensitive, but that might an overkill…)

alexarchambault (commented):
Apart from the comment I left, LGTM!

alexarchambault (approved):

PR #461, opened 2021-12-07
Print stacktraces to files, instead of stdout

tpasternak:


alexarchambault:
I pushed a few commits to:
- not mess with today's `ScalaCli.scala` much in the diff
- try to use a sensible workspace value when possible (only with the `run` command, the other commands need to be updated too)
- keep the exception causes in the log (quite important, many times, the relevant exception is in a cause)
- take into account verbosity (keep throwing in the current console if verbosity >= 2, only with the `run` command too, the other commands need to be updated there too)

Feel to squash them / push force on this branch.

alexarchambault (commented):
I pushed a few commits to:
- not mess with today's `ScalaCli.scala` much in the diff
- try to use a sensible workspace value when possible (only with the `run` command, the other commands need to be updated too)
- keep the exception causes in the log (quite important, many times, the relevant exception is in a cause)
- take into account verbosity (keep throwing in the current console if verbosity >= 2, only with the `run` command too, the other commands need to be updated there too)

Feel to squash them / push force on this branch.

tpasternak:
@alexarchambault so should I squash and merge it as is?

alexarchambault:
> so should I squash and merge it as is?

Just pushed one more commit. ~I guess it can be squashed-and-merged, yes.~

alexarchambault:
We need to report the changes in `Run.scala` (setting things in `CurrentParams`) to the other commands, prior to merging this.

tpasternak:
I had to rebase due to conflicts

alexarchambault:
@tpasternak Just pushed a commit to factor `-v` out of `LoggingOptions`, so that it can be added to commands without adding all the other logging options (which would be unused).

Good to merge for me, if that's fine for you (and the CI is green).

alexarchambault (approved):

alexarchambault:
@tpasternak It seems you discarded one of the commits I had pushed, factoring `-v` / `--verbose`, so that we don't add all the logging options when they're not needed.

I just pushed my version of this branch [here](https://github.com/alexarchambault/scala-cli/tree/hide-non-recoverable-exceptions), its last commit is missing in particular (and makes sense for us here IMO).

tpasternak:
Right, I rebased and squashed, as the history became quite complex

modules/cli/src/main/scala/scala/cli/commands/InstallHomeOptions.scala
```diff
@@ -5,6 +5,8 @@ import caseapp._
 // format: off
 @HelpMessage("Install `scala-cli` in a sub-directory of the home directory")
 final case class InstallHomeOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
```
alexarchambault: This one should be changed to `VerbosityOptions` I think.

modules/cli/src/main/scala/scala/cli/commands/Metabrowse.scala
```diff
@@ -19,8 +20,9 @@ object Metabrowse extends ScalaCommand[MetabrowseOptions] {
   override def sharedOptions(options: MetabrowseOptions) = Some(options.shared)
 
   def run(options: MetabrowseOptions, args: RemainingArgs): Unit = {
-
+    CurrentParams.verbosity = options.shared.logging.verbosity
     val inputs = options.shared.inputsOrExit(args)
+    CurrentParams.verbosity = options.shared.logging.verbosity
```
alexarchambault: Seems we want to set the workspace on that line.

alexarchambault (approved):

PR #465, opened 2021-12-07
Handle errors from using directives

### LINKED ISSUE #459: No diagnostics via BSP in case of wrong using directives
Recently `resource` was renamed to `resourceDir` and now when for example using a snippet:
```scala
// using resource "./resources"
// using lib com.lihaoyi::pprint:0.6.6
// using lib org.scala-lang.modules::scala-parallel-collections:1.0.4

@main
def main = println("Hello world!")
```

we get in Metals:
```
2021.12.07 11:18:38 ERROR Timeout waiting for 'build/initialize' response
2021.12.07 11:18:38 WARN  Retrying connection to the build server scala-cli
2021.12.07 11:18:38 INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at /home/tgodzik/.cache/metals/bsp.trace.json
```

It's very confusing for users since there is nothing to indicate what the error is.

tpasternak:
Fixes #459

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -57,7 +57,7 @@ final class BspImpl(
         Sources.defaultPreprocessors(
           buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper)
         )
-      ).swap.map((_, Scope.Main)).swap
+      ).left.map((_, Scope.Main))
     }
 
     if (verbosity >= 3)
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -57,7 +57,7 @@ final class BspImpl(
         Sources.defaultPreprocessors(
           buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper)
         )
-      )
+      ).swap.map((_, Scope.Main)).swap
     }
 
     if (verbosity >= 3)
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -273,7 +273,8 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
     withBsp(inputs, Seq(".")) { (root, _, remoteServer) =>
       async {
-        val buildTargetsResp = await(remoteServer.workspaceBuildTargets().asScala)
+        val buildTargetsResp =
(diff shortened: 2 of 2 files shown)


tpasternak:
Btw we should fix it in EtiherCps ,but I didn't find how to do it yet

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -57,7 +57,7 @@ final class BspImpl(
         Sources.defaultPreprocessors(
           buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper)
         )
-      )
+      ).swap.map((_, Scope.Main)).swap
```
alexarchambault: ?
```suggestion
      ).left.map((_, Scope.Main))
```

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -273,7 +273,8 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
     withBsp(inputs, Seq(".")) { (root, _, remoteServer) =>
       async {
-        val buildTargetsResp = await(remoteServer.workspaceBuildTargets().asScala)
+        val buildTargetsResp =
+          Await.result(remoteServer.workspaceBuildTargets().asScala, 60.seconds)
```
alexarchambault: Did you change this to make it timeout earlier? Aren't the timeouts in `withBsp` enough?

If you really want to make just this future timeout earlier, [this](https://stackoverflow.com/a/45272591/3714539) can be a clean way of doing that.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -518,6 +519,39 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+
+        val diag = diagnosticsParams.getDiagnostics.asScala.toSeq.head
+
+        expect(
+          diag.getSeverity == b.DiagnosticSeverity.ERROR,
+          diag.getRange.getStart.getLine == 0,
+          diag.getRange.getStart.getCharacter == 3,
+          diag.getRange.getEnd.getLine == 0,
+          diag.getRange.getEnd.getCharacter == 29,
+          diag.getMessage.contains(s"Unrecognized directive: using resource ./resources")
```
alexarchambault: NIT
```suggestion
          diag.getMessage.contains("Unrecognized directive: using resource ./resources")
```

alexarchambault (commented):
So the either CPS stuff silently accepted the wrong left type? (and discarded its value at runtime?) That's worrying :|

alexarchambault (approved):

PR #469, opened 2021-12-08
Use Bloop fork, default to 100% domain sockets for Bloop communication

alexarchambault:


SethTisue:
`io.github.alexarchambault.bleep` 👀

sjrd:
How come this is necessary? We have control over bloop. We can upstream the necessary changes there, if required.
/cc @tgodzik

SethTisue:
https://github.com/oyvindberg/bleep , but I don't think anyone's using blaap or bluup

alexarchambault:
FYI, the fork of Bloop (called bleep for now) lives [here](https://github.com/scala-cli/bleep). The name collision with https://github.com/oyvindberg/bleep is unfortunate, but scala-cli/bleep might be temporary.

Some of the motivations to work from it rather than from upstream Bloop are
- making the build bearable, by removing what makes the sbt build of Bloop unnecessarily and outrageously complex (mainly the shading and meta-meta-meta-meta build stuff)
- actively investigating and trying to address the numerous transient errors in the tests when they happen

It's 100% my own personal initiative for now. We (VirtusLab) might soon come up with plans to ease the maintenance of Bloop, and hopefully make this fork unnecessary.

Issue #452, opened 2021-12-06
Too high max inline level shows no messages


tgodzik:
While toying around with advent of code I created a bit ridiculous snippet:
```scala
// using resource "./resources"
// using lib com.lihaoyi::pprint:0.6.6
// using lib org.scala-lang.modules::scala-parallel-collections:1.0.4
// using option -Xmax-inlines 256

package day6

import java.nio.file.Files
import java.nio.file.Paths
import scala.jdk.CollectionConverters.*
import scala.collection.mutable
import scala.annotation.tailrec
import scala.collection.parallel.CollectionConverters.*
import java.util.concurrent.atomic.AtomicInteger

extension (s: String)
  def input: List[Int] =
    val uri = this.getClass().getClassLoader().getResource(s).toURI()
    val path = Paths.get(uri)
    Files.readAllLines(path).asScala.head.split(",").toList.map(_.toInt)

inline def fishIteration(daysLeft: Int): Long =
  if daysLeft > 0 then
    1 + fishIteration(daysLeft - 9) + fishIteration(daysLeft - 7)
  else 0

def howManyFish(days: Int) =
  val fish = "input1".input
  val size = fish.size
  val currentCalculated = new AtomicInteger(0)
  fish.zipWithIndex.par.map { case (f, i) =>
    val desendants = 1 + fishIteration(days - f)
    println(s"${100 * currentCalculated.incrementAndGet / size.toDouble}%")
    desendants
  }.sum
end howManyFish

@main
def main =
  val part1 = howManyFish(18)
  pprint.log(part1)

end main
```

The compilation should end saying that the max-inline level is too low, but instead it hangs. When running verbose we get:
```
nailgun debug: Received action Print([B@963dd01) from Nailgun server
java.lang.StackOverflowError while compiling /home/tgodzik/Documents/advent-of-code/day6/main.scalanailgun debug: Received action Print([B@407093ad) from Nailgun server

nailgun debug: Received action Print([B@63f13573) from Nailgun server
java.lang.StackOverflowErrornailgun debug: Received action Print([B@2c9792d0) from Nailgun server

nailgun debug: Received action Print([B@711da86d) from Nailgun server
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)nailgun debug: Received action Print([B@1b1a6676) from Nailgun server

nailgun debug: Received action Print([B@3f8d299b) from Nailgun server
[...]

PR #471, opened 2021-12-08
Always forward Bloop stderr

tpasternak:
Usually, Bloop prints nothing to stderr, so there's no difference, but
this could help a lot in cases when it gets lost, for example #452


alexarchambault (approved):
Let's try… I hope this won't print unintended stuff.

In the current state of https://github.com/VirtusLab/scala-cli/pull/469, the Bloop server should print much more log, so I guess I'll have to make it write most of its output to files instead (needs to be changed on the Bloop / [bleep](https://github.com/scala-cli/bleep) side).

PR #474, opened 2021-12-09
Native image: do not filter-out JVM options passed to scala-cli app

tpasternak:
Graal's native image bulids a native application that has embedded
runtime named SubstrateVM. SubstrateVM processes and filters command line
arguments before passing them to the actual application. For example
it catches all `-X` arguments and uses them to configure itself. In
scala-cli we sometimes need this kind of arguments, for example when
we call `scala-cli --java-opt "-Xmx1g"`.

This commit disables the parsing by setting `-H:-ParseRuntimeOptions`
build option

The graal's parsing logic is defined here: https://github.com/oracle/graal/blob/6773ebc7dcd59ec7185ba5f5db70b714aa43e108/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/option/RuntimeOptionParser.java#L81

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -79,7 +79,9 @@ object ScalaCli extends CommandsEntryPoint {
     systemProps.map(_.stripPrefix("-D")).foreach { prop =>
       prop.split("=", 2) match {
         case Array(key, value) =>
-          System.setProperty(key, value.stripPrefix("=")) // todo log
+          System.setProperty(key, value)
+        case Array(key) =>
+          System.setProperty(key, "")
       }
     }
   }
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -70,7 +70,23 @@ object ScalaCli extends CommandsEntryPoint {
     }
   }
 
+  private def partitionArgs(args: Array[String]): (Array[String], Array[String]) = {
+    val systemProps = args.takeWhile(_.startsWith("-D"))
+    (systemProps, args.drop(systemProps.size))
+  }
+
+  private def setSystemProps(systemProps: Array[String]): Unit = {
+    systemProps.map(_.stripPrefix("-D")).foreach { prop =>
+      prop.split("=", 2) match {
+        case Array(key, value) =>
+          System.setProperty(key, value.stripPrefix("=")) // todo log
+      }
+    }
+  }
+
   override def main(args: Array[String]): Unit = {
+    val (systemProps, scalaCliArgs) = partitionArgs(args)
+    setSystemProps(systemProps)
 
     if (Properties.isWin && isGraalvmNativeImage)
       // The DLL loaded by LoadWindowsLibrary is statically linked in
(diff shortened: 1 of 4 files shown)


tpasternak:
There's no test for -D arguments parsing. I have no idea for this except just adding `println(System.getProperty("print.hello"))` in `Run,scala` and check if it takes value from `scala-cli -Dprint.hello=hello`. Should I do it?

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -70,7 +70,23 @@ object ScalaCli extends CommandsEntryPoint {
...
+  private def partitionArgs(args: Array[String]): (Array[String], Array[String]) = {
+    val systemProps = args.takeWhile(_.startsWith("-D"))
+    (systemProps, args.drop(systemProps.size))
+  }
+
+  private def setSystemProps(systemProps: Array[String]): Unit = {
+    systemProps.map(_.stripPrefix("-D")).foreach { prop =>
+      prop.split("=", 2) match {
+        case Array(key, value) =>
+          System.setProperty(key, value.stripPrefix("=")) // todo log
```
alexarchambault: We should handle the `Array(key)` case too, corresponding to arguments like `-Dfoo` (which are legit). Also, we don't need the `stripPrefix("=")`, `split` should already have removed it.
```suggestion
        case Array(key, value) =>
          System.setProperty(key, value)
        case Array(key) =>
          System.setProperty(key, "")
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1293,4 +1293,41 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  test("-D.. options passed to the child app") {
+    val inputs = TestInputs(
+      Seq(
+        os.rel / "Hello.scala" -> """
+                                    |import java.lang.System
```
alexarchambault: No need to import this, `java.lang.*` is automatically imported in Scala sources.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1293,4 +1293,41 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  test("-D.. options passed to the child app") {
```
alexarchambault, line 1296: We should probably only run those if `TestUtil.isNativeCli` is true. (They'll fail when run with a JVM launcher else, I think)

project/settings.sc
```diff
@@ -210,6 +210,7 @@ trait CliLaunchers extends SbtModule { self =>
         else staticLibDir().path.toString
       Seq(
         s"-H:IncludeResources=$localRepoResourcePath",
+        s"-H:-ParseRuntimeOptions",
```
alexarchambault: NIT
```suggestion
        "-H:-ParseRuntimeOptions",
```

website/docs/under-the-hood.md
```diff
@@ -37,3 +37,12 @@ The main point to know is that Scala CLI takes care of fetching and starting Blo
 Scala CLI uses [Coursier](https://get-coursier.io/) to manage dependencies.
 It automatically downloads and unpacks a JVM if none is installed on your system, so that all its commands work fine even if a JVM isn't already installed.
 Scala CLI shares Coursier caches with other tools like [sbt](https://www.scala-sbt.org/), [Mill](https://github.com/com-lihaoyi/mill), or [Metals](https://scalameta.org/metals/).
+
+### scala-cli's JVM
+Scala CLI is a JVM application. Although by default it is distributed as a native image, it is still possible to tweak it with java properties.
```
alexarchambault:
```suggestion
Scala CLI is a JVM application. Although by default it is distributed as a GraalVM native image, it is still possible to set Java properties.
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
alexarchambault: Actually, they seem to run fine there too…

tpasternak: Yep, and imo they could be kept, as this PR also affects JVM runners.

alexarchambault (approved):

PR #489, opened 2021-12-13
Add help parameter to print specific options for ScalaNative and Scal…

### LINKED ISSUE #286: Rework help messages
Currently, our help message is quite too long with many misleading options dedicated for specific cases. The idea is to make it more hierarchical, so the main page shows only one line for all Scala.js options and allows users to expand more details. It can be achieved by dedicated options (e.g. --help-js) or by a combination of existing options (e.g. --help --js) or some other way - this investigation is a part of this task.

We should also ensure each help message is no longer than ~80 lines.

lwronski:
…aJs groups

By default, options for ScalaJs and ScalaNative will be hidden. These options will be displayed if someone explicitly pass `--help-js` or` --help-native` to options.

I added a new `hiddenGroups` parameter in `HelpFormat` to hide all parameters from selected groups by default:
```
ScalaComand.scala
...
.withHiddenGroups(Some(Seq(
         "Scala.JS",
         "Scala Native"
       )))
```

and then, to display for example options for ScalaJs, it is required to specify `--help-js`. 
Printing options for ScalaJS is handled in `HelpGroupOptions.scala`. Before printing help, I will remove the `ScalaJS` from the sequence of hidden groups in HelpFormat.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -1,7 +1,7 @@
 package scala.cli
 
 import caseapp.core.app.CommandsEntryPoint
-import caseapp.core.help.RuntimeCommandsHelp
+import caseapp.core.help.{Help, RuntimeCommandsHelp}
 
 import java.io.{ByteArrayOutputStream, PrintStream}
 import java.nio.charset.StandardCharsets
@@ -40,16 +40,17 @@ object ScalaCli extends CommandsEntryPoint {
     Version
   )
 
-  lazy val progName                 = (new Argv0).get("scala-cli")
-  override def description          = "Compile, run, package Scala code."
+  lazy val progName = (new Argv0).get("scala-cli")
+  override def description =
+    "Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code."
+  override def summaryDesc =
+    "See 'scala-cli <command> --help' to read about a specific subcommand or concept. To see full help run 'scala-cli <command> --help-full'."
   final override def defaultCommand = Some(actualDefaultCommand)
 
   // FIXME Report this in case-app default NameFormatter
   override lazy val help: RuntimeCommandsHelp = {
     val parent = super.help
-    parent.withDefaultHelp(
-      parent.defaultHelp.withNameFormatter(actualDefaultCommand.nameFormatter)
-    )
+    parent.withDefaultHelp(Help[Unit]())
   }
 
   override def enableCompleteCommand    = true
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -44,7 +44,7 @@ object ScalaCli extends CommandsEntryPoint {
   override def description =
     "Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code."
   override def summaryDesc =
-    "See 'scala-cli <command> --help' to read about a specific subcommand or concept. To see full help run 'scala-cli <command> --help-full'."
+    "See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'."
   final override def defaultCommand = Some(actualDefaultCommand)
 
   // FIXME Report this in case-app default NameFormatter
--- a/modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
@@ -11,6 +11,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
 object Bsp extends ScalaCommand[BspOptions] {
+  override def hidden = true
   def run(options: BspOptions, args: RemainingArgs): Unit = {
     CurrentParams.verbosity = options.shared.logging.verbosity
     if (options.shared.logging.verbosity >= 3)
--- a/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
@@ -11,7 +11,7 @@ object Compile extends ScalaCommand[CompileOptions] {
   override def group                                  = "Main"
   override def sharedOptions(options: CompileOptions) = Some(options.shared)
   def run(options: CompileOptions, args: RemainingArgs): Unit = {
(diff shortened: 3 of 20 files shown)


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -38,16 +38,17 @@ object ScalaCli extends CommandsEntryPoint {
     Version
   )
 
-  lazy val progName                 = (new Argv0).get("scala-cli")
-  override def description          = "Compile, run, package Scala code."
+  lazy val progName = (new Argv0).get("scala-cli")
+  override def description =
+    "Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code."
+  override def summaryDesc =
+    "See 'scala-cli <command> --help' to read about a specific subcommand or concept. To see full help run 'scala-cli <command> --help-full'."
```
alexarchambault: Why "concept"? (Keeping only "subcommand" would be fine IMO.)

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -38,16 +38,17 @@ object ScalaCli extends CommandsEntryPoint {
...
   final override def defaultCommand = Some(actualDefaultCommand)
 
   // FIXME Report this in case-app default NameFormatter
   override lazy val help: RuntimeCommandsHelp = {
     val parent = super.help
-    parent.withDefaultHelp(
-      parent.defaultHelp.withNameFormatter(actualDefaultCommand.nameFormatter)
-    )
+    parent.withDefaultHelp(Help[Unit]())
```
alexarchambault, line 53: Why does this need to be changed? This was mainly ensuring options with a single dash (like `-help`) are printed correctly IIRC.

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
```diff
@@ -10,6 +10,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
 object Bsp extends ScalaCommand[BspOptions] {
+  override def hidden = true
```
alexarchambault, line 14: 👍

modules/cli/src/main/scala/scala/cli/commands/HelpGroupOptions.scala
```diff
@@ -0,0 +1,33 @@
...
+) {
+
+  private def printHelpWithGroup(help: Help[_], group: String) = {
+    println(help.help(ScalaCli.helpFormat.withHiddenGroups(
+      ScalaCli.helpFormat.hiddenGroups.map(_.filterNot(_ == group))
+    )))
+    sys.exit(0)
+  }
+
+  def printHelp(help: Help[_]): Unit = {
```
alexarchambault: Could this be renamed to `maybePrintGroupHelp`? Seeing just `printHelp` in the sub-commands' `run` methods is confusing (it doesn't print anything if no `--help-*` option is specified)

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
```diff
@@ -18,7 +18,9 @@ final case class CompileOptions(
   @Name("p")
   @Name("classpath")
   @HelpMessage("Print the resulting class path")
-    classPath: Boolean = false
+    classPath: Boolean = false,
+  @Recurse
+    helpGroups: HelpGroupOptions = HelpGroupOptions()
```
alexarchambault: Shouldn't this be moved to `SharedOptions`? This is where we pull `ScalaJsOptions` and `ScalaNativeOptions`, and `HelpGroupOptions` only makes sense alongside those.

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -10,6 +10,7 @@ object Compile extends ScalaCommand[CompileOptions] {
   override def group                                  = "Main"
   override def sharedOptions(options: CompileOptions) = Some(options.shared)
   def run(options: CompileOptions, args: RemainingArgs): Unit = {
+    options.helpGroups.printHelp(RunOptions.help)
```
alexarchambault: If these options are pulled via `SharedOptions` (see my comment below), maybe a helper method could be added to `ScalaCommand`, like
```scala
def maybePrintGroupHelp(options: T): Unit =
  for (shared <- sharedOptions(options))
    shared.helpGroups.maybePrintGroupHelp(help)
```
(`ScalaCommand` already has the `Help[_]` instance, and has a method to extract the shared options from the sub-command ones.)

Then, sub-commands would just have to call
```scala
maybePrintGroupHelp(options)
```

lwronski: Good point, thanks for review.

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
lwronski: As @romanowski pointed out in #286, scala-cli help _is quite long with many misleading options dedicated for specific cases_.  I wanted to reduce number of lines in help message and now it looks like this:

```
$ scala-cli help
Usage: scala-cli <COMMAND>
Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code.

Main commands:
  clean                  Clean the workspace
  compile                Compile Scala code
  fmt, format, scalafmt  Format Scala code
  console, repl          Fire-up a Scala REPL
  package                Compile and package Scala code
  run                    Compile and run Scala code.
  test                   Compile and test Scala code

Miscellaneous commands:
  version  Print `scala-cli` version

Other commands:
  export                                    Export current project to sbt or Mill
  help                                      Print help message
  install completions, install-completions  Installs completions into your shell
  setup-ide                                 Generate a BSP file that you can import into your IDE
  shebang                                   This command is an equivalent of `run`, but it changes the way how
  update                                    Update scala-cli - it works only for installation script

See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'.
```

if you want to see more options for a specific command, you have to run `scala-cli <command> --help`, and then it prints all options without `ScalaJs` and `ScalaNative` group.

lwronski: Help message I attached above. It is prints if you run one of three commands:
```
scala-cli help
scala-cli -help
scala-cli --help
```

alexarchambault: Oh ok. Makes sense.

alexarchambault (approved):

PR #492, opened 2021-12-14
Js using directives

tpasternak:


modules/build/src/main/scala/scala/build/preprocessing/directives/UsingScalaJsOptionsDirectiveHandler.scala
```diff
@@ -0,0 +1,102 @@
...
+      |`// using jsVersion `_value_
+      |`// using jsMode `_value_
+      |`// using jsModuleKind `_value_
+      |`// using jsCheckIr true|false`
+      |`// using jsEmitSourceMaps true|false``
+      |`// using jsDom true|false``
+      |""".stripMargin
+
+  override def examples: Seq[String] = Seq(
+    "// using jsDom \"true\""
```
sjrd: The option the most likely to be used, and which makes the most sense to me here, is `moduleKind`.

alexarchambault (commented):
Apart from the point that @sjrd raises, LGTM.

PR #496, opened 2021-12-14
Don't write stacktraces to files on CIs

alexarchambault:
Users usually can't easily read the stacktrace file there.

Plus this was also hiding stack traces in the Scala CLI integration tests, which we don't need.


tpasternak (approved):
Wouldn't it be better if stacktraces were printed always on CI?

alexarchambault:
> Wouldn't it be better if stacktraces were printed always on CI?

This is what this PR does. When throwing from a `main`method, the JVM prints the exception (and its causes) itself.

PR #499, opened 2021-12-15
Do not run TastyPostProcessor for snapshot compilers

### LINKED ISSUE #232: TASTy version should depend on the Scala version
Scala 3.1.0 has a new TASTy version, however this makes using scala-cli unusable for 3.1.0 as it is hard coded to expect the same tasty version as Scala 3.0.0 (i.e. 28.0-0)

here is a good test file: (run with `scala-cli -S 3.1.0 tastyVersion.sc`)
```scala
import dotty.tools.tasty.TastyFormat.*

val isPreview =
  if ExperimentalVersion > 0 then
    s" (preview $ExperimentalVersion)"
  else
    ""

val tastyVersion = s"$MajorVersion.$MinorVersion$isPreview"

println(s"TASTy version is $tastyVersion")

```

currently the renaming step causes a crash due to unexpected TASTy:
```scala
Exception in thread "main" scala.build.tastylib.UnpickleException: TASTy signature has wrong version.
 expected: {majorVersion: 28, minorVersion: 0}
 found   : {majorVersion: 28, minorVersion: 1}

This TASTy file was produced by a more recent, forwards incompatible release.
To read this TASTy file, please upgrade your tooling.
The TASTy file was produced by Scala 3.1.0.
```

tpasternak:
Fixes #232

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -752,10 +752,12 @@ object Build {
       }
       .toMap
 
+    val fixTastyPaths = !scalaVersion.endsWith("-SNAPSHOT")
+
     val postProcessors =
       Seq(ByteCodePostProcessor) ++
         (if (updateSemanticDbs) Seq(SemanticDbPostProcessor) else Nil) ++
-        Seq(TastyPostProcessor)
+        (if (fixTastyPaths) Seq(TastyPostProcessor) else Nil)
 
     val failures = postProcessors.flatMap(
       _.postProcess(generatedSources, mappings, workspace, classesDir, logger, scalaVersion)


alexarchambault:
@tpasternak What's the problem with tasty post-processing and snapshot compilers?

tpasternak:
In general builds just fail for snapshot compilers now, because scalac experimental version number is always wrong. (in TastyFormat.scala file). It's the issue described in #232

We could ignore this check for snapshot builds, but IMO it would be better to just disable the postprocessor in this case, because it's not that uncommon to change tasty format if changing the compiler code.

If you think it would be better to just ignore the check, then I'm ok with it.

alexarchambault:
Maybe we could just add options to disable some post-processing steps then, as you suggested elsewhere. (And tasty post-processing could be disabled by default for snapshot compilers via these options' defaults.)

These options could be useful even for non-snapshot future Scala 3 versions for example.

alexarchambault:
About disabling the tasty version check, I've been wanting to do something along those lines some time ago IIRC.

Actually, I just found the local branch where I've been doing that, simplifying the `tasty-lib` module along the way. I'm going to try to open a PR with that. But options to disable some post-processing steps would definitely be useful too.

PR #502, opened 2021-12-16
Better musl handling

alexarchambault:
~This should allow to spot musl-related bugs more easily (those in the Linux "static" launcher), by not defaulting to glibc from the static launcher, even when it's available when we're running (when we pick a JVM, when we download scalafmt or metabrowse external binaries, …)~


alexarchambault:
When using the static launcher, the last commit should make us default to a Java 8 liberica on alpine in particular, rather than a latest-Java-version liberica (currently, 17).

PR #505, opened 2021-12-16
Add log if someone try to run test without zio-test-sbt

### LINKED ISSUE #426: Can't use zio-test with scala-cli
I created a[ scala-cli gist](https://gist.github.com/bbarker/ccfc54cc7afce875703183fb9d6d7d2c/49424a63b698328cf51df0a78fb054bcdbfd4fda), the Main file runs fine, but the test, though it compiles, does not run. I noticed an munit example worked fine when I tried it, so I'm guessing scala-cli either needs to support zio-test for this to work, or alternatively, there is some problem with zio(-test).

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -162,21 +165,47 @@ object Test extends ScalaCommand[TestOptions] {
...
           Constants.testRunnerMainClass,
           extraArgs,
           logger,
           allowExecve = allowExecve
         )
     }
   }
+
+  def findTestFramework(classPath: Seq[Path], logger: Logger): Option[String] = {
+    val parentInspector = new AsmTestRunner.ParentInspector(classPath)
```
alexarchambault: NIT This can be inlined right below when we call `Runner.frameworkName`, or moved to the first branch of the `if` (it's only used there).

modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -162,21 +165,47 @@ object Test extends ScalaCommand[TestOptions] {
           }
         }
       case Platform.JVM =>
+        val classPath = build.fullClassPath
+
+        val testFrameworkOpt0 = testFrameworkOpt match {
+          case Some(fw) => Some(fw)
+          case None     => findTestFramework(classPath, logger)
+        }
```
alexarchambault: NIT
```suggestion
        val testFrameworkOpt0 = testFrameworkOpt.orElse {
          findTestFramework(classPath, logger)
        }
```

modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -162,21 +165,47 @@ object Test extends ScalaCommand[TestOptions] {
...
+    val parentInspector = new AsmTestRunner.ParentInspector(classPath)
+    val classPath0      = classPath.map(_.toString)
+
+    // https://github.com/VirtusLab/scala-cli/issues/426
+    if (classPath0.exists(_.contains("zio-test")) && !classPath0.exists(_.contains("zio-test-sbt")))
+      Runner.frameworkName(classPath, parentInspector) match {
+        case Right(f) => Some(f)
+        case Left(_) =>
+          logger.message(
+            "ScalaCLI detects that you use zio-test, please import zio-test-sbt to run zio tests using scala-cli."
```
alexarchambault:
```suggestion
            "zio-test found in the class path, zio-test-sbt should be added to run zio tests with Scala CLI."
```

alexarchambault (approved):

PR #507, opened 2021-12-17
Update GraalVM to 21.3.0, use Java 17 GraalVM

alexarchambault:


alexarchambault:
Weird dead lock error during image building on Windows. I built it fine on my local Windows machine though…

PR #508, opened 2021-12-17
Start Bloop with Java 17

alexarchambault:


.github/scripts/docker/ScalaCliSlimDockerFile
```diff
@@ -1,3 +1,6 @@
+FROM debian:stable-slim AS build-env
+
 FROM gcr.io/distroless/base-debian10
 ADD scala-cli /usr/local/bin/scala-cli
+COPY --from=build-env /lib/x86_64-linux-gnu/libz.so.1 /lib/x86_64-linux-gnu/libz.so.1
```
tpasternak, line 5: I'm not sure, but we probably should copy distroless license as well

alexarchambault: This file originates from the `debian:stable-slim` image, not a distroless one. I guess this should be fine…

PR #511, opened 2021-12-20
Bump coursier to 2.1.0-M2

ckipp01:
This is mainly to pull in https://github.com/coursier/coursier/pull/2242
which allows for a user to be able to set `COURSIER_CREDENTIALS` and get it
correctly pulled in to resolve deps. (trying to use this for some works scripts
and currently hitting on this).


alexarchambault (approved):
Thanks @ckipp01!

PR #518, opened 2021-12-23
Use 'ng-stop' instead of 'exit' to quit bloop

tpasternak:
Seems that `ng-stop` has some special behavior in snailgun and returns different
exit code

https://github.com/jvican/snailgun/blob/1c6ba5cfbd948aaf055a729cc481d6292c661f7a/core/src/main/scala/snailgun/protocol/Protocol.scala#L113

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
@@ -106,12 +106,11 @@ class BloopTests extends munit.FunSuite {
       BloopUtil.killBloop()
       TestUtil.retry()(assert(!bloopRunning()))
 
-      val res = runScalaCli("bloop", "start").call(check=false)
+      val res = runScalaCli("bloop", "start").call(check = false)
       assert(res.exitCode == 0, clues(res.out.text()))
       assert(bloopRunning(), clues(res.out.text()))
 
-
-      val resExit = runScalaCli("bloop", "exit").call(check=false)
+      val resExit = runScalaCli("bloop", "exit").call(check = false)
       assert(resExit.exitCode == 0, clues(resExit.out.text()))
       assert(!bloopRunning())
     }
--- a/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
@@ -98,7 +98,7 @@ class BloopTests extends munit.FunSuite {
   test("bloop exit works") {
     def assertBloopNotRunning(): Unit = {
       val javaProcesses = os.proc("jps", "-l").call().out.text()
-      expect(!javaProcesses.contains("bloop.Server"))
+      expect(!javaProcesses.contains("bloop.Bloop"))
     }
 
     val inputs = TestInputs(Seq.empty)
--- a/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
@@ -11,9 +11,6 @@ class BloopTests extends munit.FunSuite {
   private lazy val bloopDaemonDir =
     BloopUtil.bloopDaemonDir(runScalaCli("directories").call().out.text())
 
-  // temporary, bleep exit does exit, but is having issues later on…
-  private def exitCheck = false
-
   val dummyInputs = TestInputs(
     Seq(
       os.rel / "Test.scala" ->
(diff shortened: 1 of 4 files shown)


modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -31,9 +28,9 @@ class BloopTests extends munit.FunSuite {
     shouldRestart: Boolean
   ): Unit = TestUtil.retryOnCi() {
     dummyInputs.fromRoot { root =>
+      BloopUtil.killBloop()
 
       val bloop = BloopUtil.bloop(currentBloopVersion, bloopDaemonDir)
-      bloop("exit").call(cwd = root, stdout = os.Inherit, check = exitCheck)
```
tpasternak, line 36: bloop("exit") still does not work, replaced with `BloopUtil.killBloop` cc @alexarchambault

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -98,4 +95,23 @@ class BloopTests extends munit.FunSuite {
...
+    }
+
+    val inputs = TestInputs(Seq.empty)
+    inputs.fromRoot { _ =>
+      BloopUtil.killBloop()
+      TestUtil.retry()(assertBloopNotRunning())
+
+      val res = runScalaCli("bloop", "start").call()
+      assert(res.exitCode == 0, clues(res.out.text()))
+
```
alexarchambault, line 112: NIT I'd check if bloop is running here (same kind of check as `assertBloopNotRunning` say, but negated).

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -98,4 +95,23 @@ class BloopTests extends munit.FunSuite {
...
+      expect(!javaProcesses.contains("bloop.Bloop"))
+    }
+
+    val inputs = TestInputs(Seq.empty)
+    inputs.fromRoot { _ =>
+      BloopUtil.killBloop()
+      TestUtil.retry()(assertBloopNotRunning())
+
+      val res = runScalaCli("bloop", "start").call()
+      assert(res.exitCode == 0, clues(res.out.text()))
```
alexarchambault, line 110: No need to check `res.exitCode`, os-lib checks it by default (unless we pass `check = false` to `call()`).

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -98,4 +95,23 @@ class BloopTests extends munit.FunSuite {
...
+
+      val resExit = runScalaCli("bloop", "exit").call()
+      assert(resExit.exitCode == 0, clues(resExit.out.text()))
```
alexarchambault, line 114: Same as above (no need to check `resExit.exitCode`).

modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
```diff
@@ -102,10 +102,21 @@ object TestUtil {
             helper(count + 1)
           }
       }
-    helper(if (isCI) 1 else maxAttempts)
+    helper(1)
   }
 
+  def retryOnCi[T](maxAttempts: Int = 3, waitDuration: FiniteDuration = 5.seconds)(
+    run: => T
+  ) = retry((if (isCI) 1 else maxAttempts), waitDuration)(run)
```
alexarchambault: I think it's the opposite here
```suggestion
  ) = retry((if (isCI) maxAttempts else 1), waitDuration)(run)
```

modules/integration/src/test/scala/scala/cli/integration/util/BloopUtil.scala
```diff
@@ -71,5 +71,15 @@ object BloopUtil {
           args
         )
     }
-
+  def killBloop() = {
+    val javaProcesses = os.proc("jps", "-l").call().out.text().linesIterator
```
alexarchambault, line 75: For later: instead of using `jps`, we could rely on some libdaemon-jvm stuff, that [writes a file with the daemon PID](https://github.com/scala-cli/libdaemon-jvm/blob/965ffeb0328191a453f1a8744038cff04f2265a6/library/src/libdaemonjvm/client/Connect.scala#L25-L31). But its API doesn't provide an easy way to read this PID for now…

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
tpasternak: Right, my mistake is I forgot check=false, but I'd like to keep separate assertions in order to get `clues` if it fails on CI

alexarchambault (approved):

PR #523, opened 2022-01-03
Include scope test in export to sbt

### LINKED ISSUE #353: Export to sbt ignores test target
When you have files in your sources that have something like

`// using target test`

Their dependencies will not be added to the exported project and the sources won't be visible in the output either.

There's also the comment `// Scala CLI doesn't distinguish main and test sources for now.` which I don't think is up to date anymore :)

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Export.scala
```diff
@@ -76,21 +79,47 @@ object Export extends ScalaCommand[ExportOptions] {
...
+      prepareBuild(inputs, baseOptions, logger, options.shared.logging.verbosity, Scope.Test)
+        .orExit(logger)
+
+    for {
+      svMain <- optionsMain0.scalaOptions.scalaVersion
+      svTest <- optionsTest0.scalaOptions.scalaVersion
+    } if (svMain != svTest) {
+      System.err.println(
+        s"""Detect different version of scala in scope test and main. 
+           |Please set the scala versions explicitly in scope test and main with using directives or pass -S, --scala-version as parameter""".stripMargin
```
alexarchambault: NIT
```suggestion
        s"""Detected different Scala versions in main and test scopes. 
           |Please set the Scala version explicitly in the main and test scope with using directives or pass -S, --scala-version as parameter""".stripMargin
```

modules/cli/src/main/scala/scala/cli/commands/Export.scala
```diff
@@ -76,21 +79,47 @@ object Export extends ScalaCommand[ExportOptions] {
...
+      )
+      sys.exit(1)
+    }
+
+    if (
+      optionsMain0.scalaOptions.scalaVersion.isEmpty && optionsTest0.scalaOptions.scalaVersion.nonEmpty
+    ) {
+      System.err.println(
+        s"""Detect that scala version was set only in scope test. 
+           |Please set the scala versions explicitly in scope test and main with using directives or pass -S, --scala-version as parameter""".stripMargin
```
alexarchambault: NIT
```suggestion
        s"""Detected that the Scala version is only set in test scope. 
           |Please set the Scala version explicitly in the main and test scopes with using directives or pass -S, --scala-version as parameter""".stripMargin
```

modules/cli/src/main/scala/scala/cli/export/Mill.scala
```diff
@@ -111,7 +137,19 @@ final case class Mill(
...
+      case Right(artifacts) => artifacts.classPath
+      case Left(exception) =>
+        logger.debug(exception.message)
+        Seq.empty
+    }
+    val parentInspector = new AsmTestRunner.ParentInspector(testClassPath)
+    val frameworkName0 = options.testOptions.frameworkOpt match {
+      case Some(fw) => Some(fw)
+      case None     => frameworkName(testClassPath, parentInspector).toOption
+    }
```
alexarchambault: NIT
```suggestion
    val frameworkName0 = options.testOptions.frameworkOpt.orElse {
      frameworkName(testClassPath, parentInspector).toOption
    }
```

alexarchambault (commented):
Apart from the NITs, LGTM!

PR #527, opened 2022-01-05
Prepare scala-cli to handle Scala 3 Native correctly

jchyb:
This includes a couple of changes:
* Similarly to a regular Scala or Scala.js, a scala3-library is fetched if Scala 3 is used (here called `scala3lib` for Native)
* Scala Native compatibility checks were updated - now everything is checked against both Scala language version and Scala Native version. The error message was also adjusted.
* If a snapshot version of scala-native-cli is used, now a Sonatype snapshot repository should be also checked for it. This is something I did not take into consideration before, and while up to this point no snapshot version of scala-native-cli was released, this is more of a future-proof than necessary change.

I was testing it with a locally published scala-native-cli 0.4.3-SNAPSHOT and scala-native-cli 0.4.3-RC2 should be released soon. On that release I'll add additional Scala 3 Native integration tests to the CI (either in this PR or separately, however will be more convenient)

### DIFF
--- a/modules/build/src/main/scala/scala/build/Artifacts.scala
+++ b/modules/build/src/main/scala/scala/build/Artifacts.scala
@@ -112,7 +112,7 @@ object Artifacts {
     val maybeSnapshotRepo = {
       val hasSnapshots = (jvmRunnerDependencies ++ jvmTestRunnerDependencies)
         .exists(_.version.endsWith("SNAPSHOT")) ||
-        scalaNativeCliVersion.map(_.endsWith("SNAPSHOT")).getOrElse(false)
+        scalaNativeCliVersion.exists(_.endsWith("SNAPSHOT"))
       val runnerNeedsSonatypeSnapshots = Constants.runnerNeedsSonatypeSnapshots(params.scalaVersion)
       if (hasSnapshots || runnerNeedsSonatypeSnapshots)
         Seq(coursier.Repositories.sonatype("snapshots").root)
--- a/modules/build/src/main/scala/scala/build/Artifacts.scala
+++ b/modules/build/src/main/scala/scala/build/Artifacts.scala
@@ -111,7 +111,8 @@ object Artifacts {
 
     val maybeSnapshotRepo = {
       val hasSnapshots = (jvmRunnerDependencies ++ jvmTestRunnerDependencies)
-        .exists(_.version.endsWith("SNAPSHOT"))
+        .exists(_.version.endsWith("SNAPSHOT")) ||
+        scalaNativeCliVersion.map(_.endsWith("SNAPSHOT")).getOrElse(false)
       val runnerNeedsSonatypeSnapshots = Constants.runnerNeedsSonatypeSnapshots(params.scalaVersion)
       if (hasSnapshots || runnerNeedsSonatypeSnapshots)
         Seq(coursier.Repositories.sonatype("snapshots").root)
@@ -153,7 +154,7 @@ object Artifacts {
           value {
             fetch(
               Positioned.none(dependency),
-              Nil,
+              allExtraRepositories,
               params,
               logger,
               None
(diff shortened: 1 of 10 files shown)


modules/build/src/main/scala/scala/build/options/SNNumeralVersion.scala
```diff
@@ -0,0 +1,38 @@
+package scala.build.options
+
+case class SNNumeralVersion(major: Int, minor: Int, patch: Int) {
+
+  override def equals(other: Any): Boolean =
```
WojciechMazur: Since SNNumberalVersion is a case class it contains the same implementation of equals by default

modules/build/src/main/scala/scala/build/options/ScalaNativeOptions.scala
```diff
@@ -68,9 +70,18 @@ final case class ScalaNativeOptions(
...
-      .map(name => dep"org.scala-native::$name::$finalVersion")
+
+  private def nativeScalalibDependency(scalaVersion: String): AnyDependency =
+    if (scalaVersion.startsWith("2."))
+      dep"org.scala-native::scalalib::$finalVersion"
+    else
+      dep"org.scala-native::scala3lib::$finalVersion"
+
+  def nativeDependencies(scalaVersion: String): Seq[AnyDependency] =
+    Seq("nativelib", "javalib", "auxlib")
```
WojciechMazur: In theory if scala-cli would also handle all the transitive dependencies on it's own we can specify only `scalalib` or `scala3lib` (when using Scala3), all other libraries like nativelib, javalib, auxlib, and not listed here clib, posixlib and windowslib are dependencies of scalalib/scala3lib

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -314,15 +314,26 @@ object Build {
...
+          false
+        else if (scalaVersion.startsWith("3"))
+          snNumeralVer >= SNNumeralVersion(0, 4, 3)
+        else if (scalaVersion.startsWith("2.13"))
+          true
+        else if (scalaVersion.startsWith("2.12"))
+          !inputs.sourceFiles().exists {
+            case _: Inputs.AnyScript => true
+            case _                   => false
+          }
```
WojciechMazur: Also, this looks like a candidate for refactoring, no need for negation if we can simply set it to correct value in the first place, becouse `!exists` can be replaced with `forall`. We can simplify it even further to 
```suggestion
          inputs.sourceFiles().forall(!_.isInstanceOf[Input.AnyScript])
```

jchyb: In a previous PR I saw that using pattern matching instead of isInstanceOf may have benefits here - we'll get a compiler error if the type of sourceFile changes (https://github.com/VirtusLab/scala-cli/pull/178#discussion_r723368690). I will change to forall though

jchyb:
I want to add some tests, I will add them and mark this PR as ready after the scala-native 0.4.2-RC2 release

WojciechMazur:
Scala Native CLI artifacts for 0.4.3-RC2 were published today

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -111,7 +111,8 @@ object Artifacts {
 
     val maybeSnapshotRepo = {
       val hasSnapshots = (jvmRunnerDependencies ++ jvmTestRunnerDependencies)
-        .exists(_.version.endsWith("SNAPSHOT"))
+        .exists(_.version.endsWith("SNAPSHOT")) ||
+        scalaNativeCliVersion.map(_.endsWith("SNAPSHOT")).getOrElse(false)
```
romanowski:
```suggestion
        scalaNativeCliVersion.contains(_.endsWith("SNAPSHOT"))
```

romanowski (changes requested):
There is one nit I found but this is not that important. 

I think the documentation needs to be updated since e.g. our [Scala Native guide](https://scala-cli.virtuslab.org/docs/guides/scala-native) mentions Scala versions.

romanowski (approved):

PR #535, opened 2022-01-09
Use scala-cli for local compilation

romanowski:


alexarchambault:
Mmmh, turns out my changes need some more work (issues with coursier bootstrap class path leaking in the failing Windows job here). Just opened https://github.com/VirtusLab/scala-cli/pull/537 with only the first commit (rebased), that we can merge as soon as the CI is green.

romanowski:
Let's close this one for now not to confuse reviewers and we can re-review it once it is ready.

PR #536, opened 2022-01-10
Fix #525: Spuriously duplicate options

### LINKED ISSUE #525: Setting scalacOptions twice causing warnings
Give an minimal snippet like this:

```scala
// using scala 3.1.0

def foo = "bar"
```
Do a compile and get no warning. Check the `.scala/.bloop/<build-target>.json` and see that there are no options containing `-feature`, `-deprecation`, or `-explain`.

Change you script to:

```scala
// using scala 3.1.0
// using options -deprecation -feature -explain

def foo = "bar"
```

Do a `scala-cli clean .` just to be safe and then follow it up with a `scala-cli compile .` and you see the following:

```
Warning: Flag -deprecation set repeatedly
Warning: Flag -feature set repeatedly
Warning: Flag -explain set repeatedly
```

If you now look in the bloop json file again you see that they are duplicated in `options`. I'd expect this not to happen so the user doesn't see these errors.

jchyb:
This PR fixes the issue and adds a unit test based on the original issue snippet. It is worth noting that I decided to fix only the _spuriously_ duplicate options - users still can duplicate them on will f.e. by providing multiple of the same using directives in different files or by adding an already defined option via cli. On that occasion they will most likely still receive the same warning. 

I did not want to limit users in defining duplicate options as that would require a more strict control of a scalac api, f.e. perhaps one day scalac will have an option that will be able to be defined multiple times (like our verbose) and then our hypothetical assumption would become incorrect.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -159,12 +159,13 @@ object Build {
         overrideOptions.orElse(options) // update hash in inputs with options coming from the CLI or cross-building, not from the sources
       )
 
-      val baseOptions = overrideOptions.orElse(sharedOptions)
+      val overridenWithSharedOptions = overrideOptions.orElse(sharedOptions)
+      val overridenOptions           = overrideOptions.orElse(options)
 
-      val crossSources0 = crossSources.withVirtualDir(inputs0, scope, baseOptions)
+      val crossSources0 = crossSources.withVirtualDir(inputs0, scope, overridenWithSharedOptions)
 
-      val sources = value(crossSources0.scopedSources(baseOptions))
-        .sources(scope, baseOptions)
+      val sources = value(crossSources0.scopedSources(overridenWithSharedOptions))
+        .sources(scope, overridenOptions)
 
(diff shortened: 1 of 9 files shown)


romanowski (approved):

alexarchambault:
Just tried something slightly different in the commits I just pushed, as the CI errors looked legit… Feel free to revert them, if ever my approach doesn't pan out well.

alexarchambault:
The idea of my approach is that `CrossSources` basically splits its `buildOptions: Seq[HasBuildRequirements[BuildOptions]]` field in two: some options go in `def sharedOptions`, the others go into the `ScopedSources` returned by `def scopedSources`. My changes make sure options go only in one of those, not both (as was the case before it seems).

jchyb:
Thank you for the fixes, I must have messed something up with the scopes before. I'll be more careful next time

alexarchambault:
> Thank you for the fixes, I must have messed something up with the scopes before. I'll be more careful next time

No problem, that's a super hairy area of the code base. As the commit history shows, I definitely didn't address that in the first try 😅

PR #539, opened 2022-01-11
Add more ScalaJs using directives

lwronski:
Add support for configure bellow ScalaJs options in using directives:

- jsHeader
- jsAllowBigIntsForLongs
- jsAvoidClasses
- jsAvoidLetsAndConsts
- jsModuleSplitStyle
- jsEsVersion

### DIFF
--- a/modules/build/src/main/scala/scala/build/options/ScalaJsOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/ScalaJsOptions.scala
@@ -2,7 +2,14 @@ package scala.build.options
 
 import bloop.config.{Config => BloopConfig}
 import dependency._
-import org.scalajs.linker.interface.{ESFeatures, ModuleKind, Semantics, StandardConfig}
+import org.scalajs.linker.interface.{
+  ESFeatures,
+  ESVersion,
+  ModuleKind,
+  ModuleSplitStyle,
+  Semantics,
+  StandardConfig
+}
 
 import java.util.Locale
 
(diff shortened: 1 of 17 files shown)


modules/build/src/main/scala/scala/build/options/ScalaJsOptions.scala
```diff
@@ -37,6 +50,26 @@ final case class ScalaJsOptions(
       case _                     => ModuleKind.NoModule
     }
 
+  def moduleSplitStyle: ModuleSplitStyle =
+    moduleSplitStyleStr.map(_.trim.toLowerCase(Locale.ROOT)).getOrElse("") match {
+      case "fewestmodules"   => ModuleSplitStyle.FewestModules
+      case "smallestmodules" => ModuleSplitStyle.SmallestModules
+      case _                 => ModuleSplitStyle.FewestModules
```
romanowski: Shouldn't it be an error?

modules/build/src/main/scala/scala/build/options/ScalaJsOptions.scala
```diff
@@ -37,6 +50,26 @@ final case class ScalaJsOptions(
...
+    esVersionStr.map(_.trim.toLowerCase(Locale.ROOT)).getOrElse("") match {
+      case "es5_1"  => ESVersion.ES5_1
+      case "es2015" => ESVersion.ES2015
+      case "es2016" => ESVersion.ES2016
+      case "es2017" => ESVersion.ES2017
+      case "es2018" => ESVersion.ES2018
+      case "es2019" => ESVersion.ES2019
+      case "es2020" => ESVersion.ES2020
+      case "es2021" => ESVersion.ES2021
+      case _        => ESFeatures.Defaults.esVersion
```
romanowski: Shouldn't it be an error?

romanowski (changes requested):
We should update JS guide as well.

modules/build/src/main/scala/scala/build/options/ScalaJsOptions.scala
lwronski: Previously, we used the default version for `moduleKind`, when someone pass incorrect value. 

and now I added a warning logging, is it good enough?

romanowski (approved):

PR #545, opened 2022-01-14
Fix #528: Non-default scalafmt version leading to errors

### LINKED ISSUE #241: Fmt command should be aware of scala-fmt version from the configuration
- scala-cli should read the scalafmt version from the configuration file if it is not provided manually. If we are not doing it then the user is "greeted" with:
```
error: invalid Scalafmt version.

This Scalafmt installation has version '3.0.3' and the version configured in '/Users/maman/Documents/openaq-streamer/.scalafmt.conf' is '2.6.4'.
To fix this problem, add the following line to .scalafmt.conf:
```
version = '3.0.3'
```

NOTE: this error happens only when running a native Scalafmt binary.
Scalafmt automatically installs and invokes the correct version of Scalafmt when running on the JVM.
```
- Fmt command should have more scalafmt tags to choose from (eg. more tags pushed here: https://github.com/alexarchambault/scalafmt-native-image/releases)
- `--check` doesn't work because afaik it is not used anywhere

### LINKED ISSUE #528: `scala-cli format --check` won't work unless you have the exact scalafmt version
I feel like it will be a common workflow in CI to check formatting. Currently with `scala-cli` if you try this and your version of scalafmt doesn't exactly match the version that scala-cli has you get this:

```
❯ scala-cli format --check .
error: invalid Scalafmt version.

This Scalafmt installation has version '3.0.3' and the version configured in '/Users/ckipp/Documents/scala-workspace/tooling-talks/src/.scalafmt.conf' is '3.2.1'.
To fix this problem, add the following line to .scalafmt.conf:

version = '3.0.3'


NOTE: this error happens only when running a native Scalafmt binary.
Scalafmt automatically installs and invokes the correct version of Scalafmt when running on the JVM.
```

jchyb:


modules/integration/src/test/scala/scala/cli/integration/FmtTests.scala
```diff
@@ -4,11 +4,12 @@ import com.eed3si9n.expecty.Expecty.expect
 
 class FmtTests extends munit.FunSuite {
 
-  val simpleInputs = TestInputs(
+  def simpleInputs(additionalConfInfo: String) = TestInputs(
```
lwronski:
```suggestion
  def simpleInputs(confInfo: String = "") = TestInputs(
```

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -15,8 +16,64 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
     List("scalafmt")
   )
+
+  private def getGitRoot(workspace: os.Path, logger: Logger): Option[String] =
+    try {
+      val result = os.proc("git", "rev-parse", "--show-toplevel").call(cwd = workspace).out.trim
+      Option(result)
+    }
+    catch {
+      case NonFatal(e) =>
```
lwronski, line 26: I think, that we should ignore any exception from above process and use default scalafmt versions. Add only logging what happened.

modules/integration/src/test/scala/scala/cli/integration/FmtTests.scala
```diff
@@ -31,19 +32,27 @@ class FmtTests extends munit.FunSuite {
...
   test("no inputs") {
-    simpleInputs.fromRoot { root =>
+    simpleInputs("").fromRoot { root =>
       os.proc(TestUtil.cli, "fmt").call(cwd = root)
       val updatedContent = noCrLf(os.read(root / "Foo.scala"))
       expect(updatedContent == expectedSimpleInputsFormattedContent)
     }
   }
 
+  test("using non-default scalafmt version") {
```
lwronski: I'm not sure, but maybe just writing unit tests would be enough. This test is very similar to "simple", and version determination can only be a unit test.

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -15,8 +16,64 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
+        val gitRootMaybe       = getGitRoot(workspace, logger)
+        val confInGitRootMaybe = gitRootMaybe.map(os.Path(_) / confName)
+        confInGitRootMaybe.find(os.exists(_))
+      }
+    }
+
+    val confContentMaybe = pathMaybe.flatMap { path =>
+      metaconfig.Hocon.parseInput[ScalafmtVersionConfig](
+        metaconfig.Input.File(path.toNIO)
+      ).toEither.toOption
```
lwronski: We can print exceptions as warning.

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
jchyb: The catch block here is only used for logging. I have adjusted the phrasing to make this come across better - it no longer says "Failed" to not confuse users.

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -15,8 +16,71 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
+        val gitRootMaybe       = getGitRoot(workspace, logger)
+        val confInGitRootMaybe = gitRootMaybe.map(os.Path(_) / confName)
+        confInGitRootMaybe.find(os.exists(_))
+      }
+    }
+
+    val confContentMaybe = pathMaybe.flatMap { path =>
+      val either = metaconfig.Hocon.parseInput[ScalafmtVersionConfig](
+        metaconfig.Input.File(path.toNIO)
+      ).toEither
```
lwronski, line 64:
```suggestion
      ).toEither.left.foreach(logger.log(_))
```

lwronski (approved):

romanowski:
fixes #528

romanowski (approved):

PR #546, opened 2022-01-15
New using directive syntax

romanowski:
This PR changes the syntax of using directives. Following syntax is supported:
- `using foo bar` with `using` keyword
- `//> using scala "3"` or `/*> using scala "#"` `using` keyword within a comment with special indicator `>`

Following syntax is deprecated:

- `@using scala 3`, `// @using scala 3` -  with`@using` global annotation
- `// using scala "3"`, `/* using scala "3"*/` - within a comment within special indicator.

Deprecated syntaxes result in a warning and will be supported within the next minor release: `0.1.x` and results in error within `0.2.x`.

Warnings are also working within BSP.

cc @julienrf


project/deps.sc
```diff
@@ -113,3 +113,10 @@ object Docker {
   def alpineTestImage =
     "alpine@sha256:234cb88d3020898631af0ccbbcca9a66ae7306ecd30c9720690858c1b007d2a0"
 }
+
+def customRepositories =
+  Seq(
+    coursier.Repositories.sonatype("snapshots"),
+    // Uncomment for local development
+    coursier.MavenRepository(s"file://${System.getProperty("user.home")}/.m2/repository")
```
alexarchambault: Should be commented out?

scala-cli
```diff
@@ -17,7 +17,7 @@ if [ ! -x "$LAUNCHER_DIR/scala" ]; then
   echo "native-image launcher not built yet." 1>&2
   echo "In order to build it, go in $DIR, and run" 1>&2
   echo 1>&2
-  echo "  ./mill cli.nativeImage" 1>&2
+  echo "  ./mill copyTo cli.nativeImage out/cli/nativeImage/dest/scala" 1>&2
```
alexarchambault: Just `./mill cli.nativeImage` should be enough. We should change `LAUNCHER_DIR` if the native image doesn't land there.

website/docs/guides/using-directives.md
```diff
@@ -17,33 +17,70 @@ using foo "bar", "baz"
...
 ```
 
 :::info
-For now we recommend using the comment-flavor (`// using scala "3.0.2"`), and we will use that syntax in this guide.
+For now we recommend using the special comment (`//> using scala "3.0.2"`), and we will use that syntax in this guide.
 
 Until `using` directives becomes a part of the Scala specification, this is the only way that guarantees that your code will work well with IDEs, code formatters, and other tools.
 :::
 
+Within one file, only one flavor of using directives can be used. The keyword-base syntax (`using scala "3"`) has precedence over special comments (`//> using scala "3"`) and deprecated, plain comments (`// using scala "3"`) has lowest priority.
```
alexarchambault:
```suggestion
Within one file, only one flavor of using directives can be used. The keyword-based syntax (`using scala "3"`) has precedence over special comments (`//> using scala "3"`) and deprecated, plain comments (`// using scala "3"`) has lowest priority.
```

website/docs/guides/using-directives.md
```diff
@@ -17,33 +17,70 @@ using foo "bar", "baz"
...
+
+For now `using` and `@using` can be mixed within given syntax however we strongly suggest to does not use deprecated `@using`.
```
alexarchambault: ?
```suggestion
For now `using` and `@using` can be mixed within a given syntax however we strongly suggest not to use deprecated `@using`. 
```

website/docs/guides/using-directives.md
```diff
@@ -17,33 +17,70 @@ using foo "bar", "baz"
...
+[warn] ./.pg/a.scala:2:1: This using directive is ignored. File contains directives outside comments and those has higher precedence.
+[warn] // using scala "2.13.8"
+[warn] ^^^
+[warn] ./.pg/a.scala:3:1: This using directive is ignored. File contains directives outside comments and those has higher precedence.
+[warn] //> using scala "2.12.11"
+[warn] ^^^^
+```
+## Deprecated syntax
+
+As a part of `0.0.x` series we experimented with different syntax for using directives. Based on the feedback and discussions with Scala compiler teams we decided to deprecated `@using` (using annotations) and `// using` (using within plain import). Those syntaxes will work as a part of of `0.1.x` series and we will result with an error since `0.2.x`.
```
alexarchambault: ?
```suggestion
As a part of `0.0.x` series we experimented with different syntaxes for using directives. Based on feedback and discussions with the Scala compiler team, we decided to deprecate `@using` (using annotations) and `// using` (using within plain comment). Those syntaxes will keep working in the `0.1.x` series and will result in an error starting from `0.2.x`.
```

website/docs/guides/using-directives.md
```diff
@@ -17,33 +17,70 @@ using foo "bar", "baz"
...
+
+Scala CLI produces a warnings if any of the syntaxes above is used:
```
alexarchambault:
```suggestion
Scala CLI produces warnings if any of the syntaxes above is used:
```

alexarchambault (commented):
I didn't review the core of the changes yet, but I wish the git history was clearer: as of now, this mixes various unrelated changes (refactorings, new using directive syntax, changes in the build), so I'd rather not squash-and-merge, but I wouldn't want to merge with a merge commit either to avoid "reformating" or "address CI failures" commits (which don't really make sense on their own, when later bisecting issues for example).

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -135,40 +143,42 @@ final class BspImpl(
       buildChangedTest
     )
 
-    (mainScope, testScope)
+    // ala a a
```
alexarchambault: ?
```suggestion
```

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -135,40 +143,42 @@ final class BspImpl(
...
   ): Either[(BuildException, Scope), Unit] = either {
-    val (preBuildDataMain, preBuildDataTest) =
-      value(prepareBuild(actualLocalServer))
-    if (notifyChanges && (preBuildDataMain.buildChanged || preBuildDataTest.buildChanged))
+    val preBuild = value(prepareBuild(actualLocalServer))
+
+    if (notifyChanges && (preBuild.mainScope.buildChanged || preBuild.testScope.buildChanged))
       notifyBuildChange(actualLocalServer)
+
     Build.buildOnce(
```
alexarchambault: Independently of the changes of this PR, it seems both `Build.buildOnce` calls should have been wrapped in a `value { … }` :/

alexarchambault: Checking how it goes in https://github.com/VirtusLab/scala-cli/pull/557.

modules/build/src/main/scala/scala/build/errors/BuildException.scala
```diff
@@ -3,7 +3,9 @@ package scala.build.errors
...
 
 abstract class BuildException(
-  val message: String,
-  val positions: Seq[Position] = Nil,
+  override val message: String,
+  override val positions: Seq[Position] = Nil,
   cause: Throwable = null
-) extends Exception(message, cause)
+) extends Exception(message, cause) with Diagnostic {
+  override def severity: Severity = Severity.Error
```
alexarchambault: We shouldn't need to put a severity here… This might lead to methods returning `Either[BuildException, T]` to return a `Left(ex)` with `ex.severity == Severity.Warning`, which doesn't make sense (exceptions on the left here are implicitly assumed to be plain errors). The severity field makes "invalid states representable".

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -1,6 +1,13 @@
 package scala.build.preprocessing
 
+import _root_.com.virtuslab.using_directives.custom.utils.ast.UsingDefs
```
alexarchambault: NIT
```suggestion
import com.virtuslab.using_directives.custom.utils.ast.UsingDefs
```

modules/build/src/test/scala/scala/build/tests/ScalaPreprocessorTests.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  val path = os.pwd
+
+  trait Check {
+    def test(diags: Seq[Diagnostic]): Boolean
+  }
+  case class Warn(messagePattern: String, line: Int, index: Int) extends Check {
+    def test(diags: Seq[Diagnostic]): Boolean = {
+      val matching = diags.filter(_.positions.collect {
+        case Position.File(_, (l, c), _) if l == line && index == c => l
```
alexarchambault: NIT
```suggestion
        case Position.File(_, (`line`, `index`), _) => line
```

(That said, I often forget about matching with backticks myself, so there might be other cases in the code base where we can do the same…)

modules/build/src/test/scala/scala/build/tests/ScalaPreprocessorTests.scala
```diff
@@ -0,0 +1,83 @@
...
+      }.nonEmpty)
```
alexarchambault: Maybe the `.collect { … }.nonEmpty` can be changed to `.exists { case Position.File(…) => …; case _ => false }`?

modules/build/src/test/scala/scala/build/tests/ScalaPreprocessorTests.scala
```diff
@@ -0,0 +1,83 @@
...
+      val Regex = s".*$messagePattern.*".r
+      matching.map(_.message).collect {
+        case Regex() => true
+      }.nonEmpty
```
alexarchambault: Same as above
```suggestion
      matching.map(_.message).exists {
        case Regex() => true
        case _ => false
      }
```

alexarchambault (commented):
(It seems some `*.semanticdb` are committed too.)

modules/build/src/main/scala/scala/build/errors/BuildException.scala
romanowski: I've put severity here just to be able to process `BuildException` and diagnostics using the same code and hence I need to fix severity here. I can change it to `final def` to make sure nobody will treat Exception as warnings.

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
romanowski: I can wrap it but let me know if `value { }` works

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
romanowski: ah,I needed to force recompilation :/

project/deps.sc
romanowski: yes, my bad

romanowski:
@alexarchambault I updated the commit structure so it should be easier to review. I can also separate 202f37aee247ba18baf0f49400e6f9a6f6d8e69f to different PR since all those changes are basically replacing `// using` with `//> using`

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -135,40 +143,35 @@ final class BspImpl(
...
+        inputs.generatedSrcRoot(scope),
+        data.generatedSources,
+        data.buildOptions,
+        scope,
+        logger,
+        actualLocalClient,
+        bloopServer
+      ).left.map(_ -> scope)
+
+    value(doBuildOnce(preBuild.mainScope, Scope.Main))
```
tpasternak: Please use for-comprehensions as long as we have problems with left-types mismatch in EitherCps

romanowski: sure

project/deps.sc
```diff
@@ -113,3 +113,13 @@ object Docker {
...
 }
+
+def customRepositories =
+  Seq(
+    coursier.Repositories.sonatype("snapshots")
+    // Uncomment for local development
+    // coursier.MavenRepository(
+    //   (os.Path(System.getProperty("user.home")) / ".m2" / "repository")
+    //     .toNIO.toUri.toASCIIString
+    // )
```
pikinier20, line 124: IIRC you can add MavenLocal repo with: `coursier.LocalRepositories.Dangerous.maven2Local`

website/docs/guides/using-directives.md
```diff
@@ -17,33 +17,70 @@ using foo "bar", "baz"
...
 ```
 
 :::info
-For now we recommend using the comment-flavor (`// using scala "3.0.2"`), and we will use that syntax in this guide.
+For now we recommend using the special comment (`//> using scala "3.0.2"`), and we will use that syntax in this guide.
 
 Until `using` directives becomes a part of the Scala specification, this is the only way that guarantees that your code will work well with IDEs, code formatters, and other tools.
 :::
 
+Within one file, only one flavor of using directives can be used. The keyword-based syntax (`using scala "3"`) has precedence over special comments (`//> using scala "3"`) and deprecated, plain comments (`// using scala "3"`) has lowest priority.
```
pikinier20, line 34:
```suggestion
Within one file, only one flavor of using directives can be used. The keyword-based syntax (`using scala "3"`) has precedence over special comments (`//> using scala "3"`). The deprecated, plain comments (`// using scala "3"`) have lowest priority.
```

examples/cross-build/Hello.js.scala
```diff
@@ -1,5 +1,5 @@
-// using scala 2.13
-// using target scala-js
+//> using scala 2.13
+//> using target scala-js
```
pikinier20, line 2: Did I omit a moment when we allowed strings without quotes?

examples/cross-build/Hello.native.scala
```diff
@@ -1,5 +1,5 @@
-// using scala 2.13
-// using target scala-native
+//> using scala 2.13
+//> using target scala-native
```
pikinier20, line 2: As above

examples/scala-versions/README.md
```diff
@@ -95,7 +95,7 @@ Using directives syntax is still experimental and may change in future versions
 So when we will have:
 
 ```scala title=version.scala
-// using scala 2.12.5
+//> using scala 2.12.5
```
pikinier20, line 98: As above

examples/scala-versions/version.scala
```diff
@@ -1,6 +1,6 @@
 // File was generated from based on docs/cookbooks/scala-versions.md, do not edit manually!
 
-// using scala 2.12.5
+//> using scala 2.12.5
```
pikinier20, line 3: As above

examples/cross-build/Hello.js.scala
romanowski: I think we do not check examples if compile file, a good catch, I've open a new issue for that: https://github.com/VirtusLab/scala-cli/issues/575

I will also include fixes in a followup PR.

PR #548, opened 2022-01-16
Allow to run scala-cli snapshot

lwronski:
`scala-cli` can also run the other version of `scala-cli`, it may be helpful to test unreleased scala-cli functionality.

```bash ignore
scala-cli about
# Scala CLI version 0.0.9 (v0.0.9)
```

```bash ignore
scala-cli about --scala-cli-version 0.0.9+131-gf0ab5c40-SNAPSHOT
# Scala CLI version 0.0.9+131-gf0ab5c40-SNAPSHOT (v0.0.9-131-f0ab5c)
```


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -138,6 +139,7 @@ object ScalaCli extends CommandsEntryPoint {
   }
 
   private def main0(args: Array[String]): Unit = {
+    LauncherCLI.main(args)
```
romanowski: I do not like situation when calling a random method can side-effect in rest of the code not being run (`sys.exit`) and here this is a base of the design. 

I think we need at least a comment but I would love if this will be more obvious from the code.

website/docs/commands/basics.md
```diff
@@ -181,6 +181,20 @@ echo 'println("Hello")' | scala-cli -
 # Hello
 ```
 
+## ScalaCLI version
+
+`scala-cli` can also run the other version of `scala-cli`, it may be helpful to test unreleased scala-cli functionality.
+
+```bash ignore
```
romanowski: Can we use `ChainedSnippets` and use expected outputs as in https://github.com/VirtusLab/scala-cli/blob/master/website/docs/getting_started.md ?

romanowski: Also why we do need to use `ignore`?

modules/cli/src/main/scala/scala/cli/launcher/LauncherOptions.scala
```diff
@@ -0,0 +1,21 @@
+package scala.cli.launcher
+
+import caseapp._
+import upickle.default.{ReadWriter, macroRW}
+
+@HelpMessage("Set the ScalaCLI version")
+final case class LauncherOptions(
+  @Group("Launcher")
+  scalaCliVersion: Option[String] = None,
```
romanowski: I think we need a help message

romanowski (changes requested):

modules/cli/src/main/scala/scala/cli/launcher/LauncherOptions.scala
```diff
@@ -0,0 +1,21 @@
...
+  @HelpMessage("The version of Scala on which scalaCLI was published")
+  @ValueDescription("2.12|2.13|3")
+  publishedScalaVersion: Option[String] = None
+)
+
+object LauncherOptions {
+  lazy val parser: Parser[LauncherOptions]                           = Parser.derive
+  implicit lazy val parserAux: Parser.Aux[LauncherOptions, parser.D] = parser
+  implicit lazy val help: Help[LauncherOptions]                      = Help.derive
+  implicit lazy val jsonCodec: ReadWriter[LauncherOptions]           = macroRW
```
alexarchambault: We shouldn't need this I think
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -101,7 +102,9 @@ final case class SharedOptions(
   @Hidden
     forbid: List[String] = Nil,
   @Recurse
-  helpGroups: HelpGroupOptions = HelpGroupOptions()
+    helpGroups: HelpGroupOptions = HelpGroupOptions(),
+  @Recurse
+    launcherOptions: LauncherOptions = LauncherOptions()
```
alexarchambault: Maybe we shouldn't put those here, but rather manually parse those, requiring those options to appear first. Something along those lines in `ScalaCli.main` should work:
```scala
val remainingArgs = LauncherOptions.parser.stopAtFirstUnrecognized.parse(args) match {
  case Left(error) =>
    ??? // print error and exit
  case Right((launcherOpts, args0)) =>
    launcherOpts.scalaCliVersion.map(_.trim).filter(_.nonEmpty) match {
      case Some(ver) =>
        LauncherCli.runAndExit(ver, args0)
      case None =>
        args0
    }
}
```

modules/cli/src/main/scala/scala/cli/launcher/LauncherCLI.scala
```diff
@@ -0,0 +1,65 @@
...
+          System.err.println(value.message)
+          sys.exit(1)
+      }
+
+    val scalaCli =
+      fetchedScalaCli.fullDetailedArtifacts.collect { case (_, _, _, Some(f)) =>
+        f.toPath.toFile
+      }
+
+    val buildOptions = BuildOptions()
```
alexarchambault: Ideally, we should ensure we use at least Java 16 to run Scala CLI. So the default system JVM might not work. I guess in a first time, we could just use a JVM of our own, like `OsLib.baseDefaultJvm(OsLibc.jvmIndexOs, 17)`, and put that in `buildOptions.javaOptions.jvmIdOpt`.

modules/cli/src/main/scala/scala/cli/launcher/LauncherCLI.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    val exitCode =
+      Runner.runJvm(
+        buildOptions.javaHome().value.javaCommand,
+        buildOptions.javaOptions.javaOpts.map(_.value),
+        scalaCli,
+        "scala.cli.ScalaCli",
+        remainingArgs.remaining ++ Seq("--") ++ remainingArgs.unparsed,
+        logger
```
alexarchambault: You can pass `allowExecve = true` here. On Linux and macOS, when run from a native image, this will make the sub-process "take over" the current one, just like how `exec something` would do in a bash script.

modules/cli/src/main/scala/scala/cli/launcher/LauncherCLI.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import caseapp.core.RemainingArgs
+import dependency._
+
+import scala.build.internal.Runner
+import scala.build.options.BuildOptions
+import scala.build.{Artifacts, Positioned}
+import scala.cli.commands.{LoggingOptions, ScalaCommand}
+
+object LauncherCLI extends ScalaCommand[LauncherOptions] {
```
alexarchambault: We probably don't need this to be a `ScalaCommand` or a case-app thing. A simple method, like `def runAndExit(version: String, args: Seq[String]): Nothing` should work (see my comment above).

website/docs/commands/basics.md
```diff
@@ -181,6 +181,20 @@ echo 'println("Hello")' | scala-cli -
 # Hello
 ```
 
+## ScalaCLI version
```
alexarchambault:
```suggestion
## Scala CLI version
```

website/docs/commands/basics.md
```diff
@@ -181,6 +181,20 @@ echo 'println("Hello")' | scala-cli -
...
+
+`scala-cli` can also run the other version of `scala-cli`, it may be helpful to test unreleased scala-cli functionality.
```
alexarchambault:
```suggestion
`scala-cli` can also run another Scala CLI version, which can be helpful to test unreleased yet Scala CLI functionality.
```

website/docs/commands/basics.md
```diff
@@ -181,6 +181,20 @@ echo 'println("Hello")' | scala-cli -
...
+
+```bash ignore
+scala-cli about
+# Scala CLI version 0.0.9 (v0.0.9)
+```
+
+```bash ignore
+scala-cli about --scala-cli-version 0.0.9+131-gf0ab5c40-SNAPSHOT
```
alexarchambault: With my suggestion above, that would be
```suggestion
scala-cli --scala-cli-version 0.0.9+131-gf0ab5c40-SNAPSHOT about
```

website/docs/commands/basics.md
alexarchambault: I would also warn users that this runs a JVM-based Scala CLI, not a native one, so that this might be a bit slower.

modules/cli/src/main/scala/scala/cli/launcher/LauncherOptions.scala
alexarchambault: Maybe a shorter option would be better, like
```suggestion
  cliVersion: Option[String] = None,
```

modules/cli/src/main/scala/scala/cli/launcher/LauncherOptions.scala
```diff
@@ -0,0 +1,21 @@
...
+import upickle.default.{ReadWriter, macroRW}
+
+@HelpMessage("Set the ScalaCLI version")
+final case class LauncherOptions(
+  @Group("Launcher")
+  scalaCliVersion: Option[String] = None,
+  @Group("Launcher")
+  @HelpMessage("The version of Scala on which scalaCLI was published")
+  @ValueDescription("2.12|2.13|3")
+  publishedScalaVersion: Option[String] = None
```
alexarchambault: Maybe we could use something shorter here too?
```suggestion
  cliScalaVersion: Option[String] = None
```

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -45,7 +46,8 @@ object ScalaCli extends CommandsEntryPoint {
   override def description =
     "Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code."
   override def summaryDesc =
-    "See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'."
+    """|To run another Scala CLI version, pass '--cli-version' at the beginning of the argument list. 'scala-cli --cli-version <versions> args'.
+       |See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'.""".stripMargin
```
alexarchambault: I would have left the former message first here, as it should matter to more users. The `--cli-version` stuff should only be relevant for rather advanced users. (I'd be almost reluctant to mention it here actually…)
```suggestion
    """|See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'.
       |To run another Scala CLI version, specify it with '--cli-version' before any other argument, like 'scala-cli --cli-version <version> args'.""".stripMargin
```

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -138,6 +140,17 @@ object ScalaCli extends CommandsEntryPoint {
...
   private def main0(args: Array[String]): Unit = {
+    LauncherOptions.parser.stopAtFirstUnrecognized.parse(args) match {
+      case Left(e) =>
+        System.err.println(e.message)
+        sys.exit(1)
+      case Right((launcherOpts, args0)) =>
+        launcherOpts.cliVersion.map(_.trim).filter(_.nonEmpty) match {
+          case Some(ver) =>
+            LauncherCLI.runAndExit(ver, launcherOpts, args0)
+          case None =>
```
alexarchambault: I think we should really ensure we use `args0` down below. Users might inadvertently specify `--cli-scala-version …` for example, and we should ignore it if that's the case.

modules/cli/src/main/scala/scala/cli/launcher/LauncherCLI.scala
```diff
@@ -0,0 +1,59 @@
+package scala.cli.launcher
+import dependency._
+
+import scala.build.internal.{OsLibc, Runner}
+import scala.build.options.{BuildOptions, JavaOptions}
+import scala.build.{Artifacts, Positioned}
+import scala.cli.commands.LoggingOptions
+
+object LauncherCLI {
```
alexarchambault: NIT NIT NIT Could we rename this to `LauncherCli`? That's the case convention we've been following in the Scala CLI code base, when "cli" is in a class name.

modules/cli/src/main/scala/scala/cli/launcher/LauncherCLI.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def runAndExit(version: String, options: LauncherOptions, remainingArgs: Seq[String]) = {
```
alexarchambault: NIT This can explicitly be typed as returning `Nothing` (that makes it clearer that this method doesn't really return)
```suggestion
  def runAndExit(version: String, options: LauncherOptions, remainingArgs: Seq[String]): Nothing = {
```

romanowski (approved):

PR #557, opened 2022-01-18
Don't discard errors in BSP server

alexarchambault:


tpasternak:
Not sure, but I remember we have had a bug in EitherCps that made `value` ignoring errors if the `Left` type is incompatible

tpasternak:
Exactly https://github.com/VirtusLab/scala-cli/pull/465/files

Would be good to write a test that checks whether the errors are caught here, or use for comprehensions. Otherwise, there's a high risk of regression here.

romanowski:
I've included this changes in [my PR](https://github.com/VirtusLab/scala-cli/pull/546)

PR #563, opened 2022-01-20
For -S 3.x input use latest release from 3.x versions

lwronski:
For now, If someone pass -S 3.1 parameter to scala-cli, it throws error:
```
VL-D-0317:scala-cli lwronski$ scala-cli hello.sc -S 3.1
Unsupported Scala version: 3.1
```
My expectation is, that scala-cli use latest version from 3.1.

There are two ways to solve this problem:
- If someone pass 3.x, we use latest scala version for 3.x (I have already done it in this PR).
- we have to add entries to [scala-versions-v1.json](https://raw.githubusercontent.com/VirtuslabRnD/scala-cli-scala-versions/master/scala-versions-v1.json) such as: { "3.0.2", "3.1.0", "3.2.x", ..., "3.x.x"}. Mechanism search the latest supported version for prefix "3.1", but it doesn't exists in scala-versions-v1.json, therefore the error was throw. 

Is the first option good enough?

### DIFF
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -284,15 +284,17 @@ final case class BuildOptions(
       sv match {
         case Some(sv0) =>
           val prefix           = if (sv0.endsWith(".")) sv0 else sv0 + "."
-          val matchingVersions = allVersions.filter(_.startsWith(prefix))
+          val matchingVersions = allVersions.filter(_.startsWith(prefix)).map(Version(_))
           if (matchingVersions.isEmpty)
             Left(new InvalidBinaryScalaVersionError(sv0))
           else {
             val validMaxVersions = maxSupportedScalaVersions
               .filter(_.repr.startsWith(prefix))
-            val validMatchingVersions = matchingVersions
-              .map(Version(_))
-              .filter(v => validMaxVersions.exists(v <= _))
+            val validMatchingVersions = {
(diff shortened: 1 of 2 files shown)


modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -284,15 +284,17 @@ final case class BuildOptions(
...
             val validMaxVersions = maxSupportedScalaVersions
               .filter(_.repr.startsWith(prefix))
-            val validMatchingVersions = matchingVersions
-              .map(Version(_))
-              .filter(v => validMaxVersions.exists(v <= _))
+            val validMatchingVersions =
+              if (prefix.startsWith("3.") && prefix.length > 2) matchingVersions
+              else
+                matchingVersions
+                  .filter(v => validMaxVersions.exists(v <= _))
```
alexarchambault: I would have done that slightly differently here: if some valid versions match the user input, only consider those. If no valid versions match, then look at all available Scala versions. Something like
```suggestion
            val validMatchingVersions = {
              val filtered = matchingVersions.filter(v => validMaxVersions.exists(v <= _))
              if (filtered.isEmpty) matchingVersions
              else filtered
            }
```

alexarchambault: That would correspond to a "best effort" to use the valid versions, while still allowing non-valid ones when these can't work.

lwronski: Good point, your idea is much more universal.

romanowski (approved):

PR #567, opened 2022-01-21
Generate semantic DBs for Java sources, pass wrapped sources details over BSP

alexarchambault:


alexarchambault:
Looks like there's a stale Java 8 Bloop lurking around in the tests…

alexarchambault:
Generating semantic DBs for Java stuff seems to be required for recent versions of Metals not to complain.

PR #570, opened 2022-01-24
Update Scala to 2.13.8 and 3.1.1

### LINKED ISSUE #555: Inconvenient default scala version for native projects/scripts
When compiling Scala Native via scala-cli and not specifying scala version (any main.sc script will provide reproduction):
```bash
scala-cli --native main.sc
```
we get:
```
Used Scala Native version is incompatible the passed options.
Please try one of the following combinations:
  Scala Native version >= 0.4.3 for Scala 3.1 (*.sc & *.scala files)
  Scala Native version >= 0.4.0 for Scala 2.13 (*.sc & *.scala files)
  Scala Native version >= 0.4.0 for Scala 2.12 (*.scala files)
Windows is supported since Scala Native 0.4.1.
```
This is correct, but inconvenient. Scala-cli uses scala 3.0.2 by default, but Scala Native is incompatible for scala 3.0

alexarchambault:


alexarchambault:
This is still awaiting some of the latest Ammonite builds to hit Central…

jchyb:
Oh, looking at the CI I think I might need to fix some of the SN tests

jchyb:
If we change the default nativeVersion (which I think we should do either way) to 0.4.3 in Deps.sc the native tests pass, at least locally (the previous 0.4.2 was still incompatible with Scala 3), the 'scala.cli.integration.RunTestsDefault.stack traces' seems to be unrelated (probably some intended changes to the stack traces there I guess)

alexarchambault:
Going to give Scala Native `0.4.3` a try, yes.

lolgab:
The linking error in Scala Native is due to a change that was fixed in Mill `0.10.0`. I suggest to update to a even newer version (like `0.10.0-21-c4247b`) since there was another problem with Scala 3 compiler plugins that was generating a lot of warnings.
I tried to update Mill myself but I saw that a plugin you maintain was not updated yet.
I opened a PR that is a step forward to scala-cli on Scala Native for Scala 3 :)
https://github.com/alexarchambault/mill-native-image/pull/5

EDIT: Also opened a PR to update the build that depends on the mill-native-image PR: https://github.com/VirtusLab/scala-cli/pull/581

alexarchambault:
> The linking error in Scala Native is due to a change that was fixed in Mill `0.10.0`.

@lolgab Which linking error? We don't use the Scala Native capabilities of Mill in the Scala CLI build. I might have missed a linking error in the CI logs, but it should mean that Scala CLI should apply to itself the same changes as Mill did then.

lolgab:
Oh gotcha! You use Bloop which is broken probably and can't handle the fact that now Scala Native generates a static main method in the class instead of a method in the companion object.
[This](https://github.com/com-lihaoyi/mill/blob/e9685e58d50848694f687e84e141bce78a53f049/scalanativelib/worker/0.4/src/ScalaNativeWorkerImpl.scala#L56) is what I needed to do to fix Mill and it will be needed in Bloop too.

This is the error I'm talking about:
```
Error:  Found 1 missing definitions while linking
Error:  Not found Member(Top(Test$),D4mainLAL16java.lang.String_uEo)
```

lolgab:
Opened a PR to Bloop to fix the `mainClass` problem in Scala Native `0.4.3+` https://github.com/scalacenter/bloop/pull/1664

alexarchambault:
@lolgab Seems [it was Mill after all](https://github.com/VirtusLab/scala-cli/runs/4944294974?check_suite_focus=true#step:8:4017). Not the Mill that builds Scala CLI though, but rather the one launched during the tests of the "export" feature of Scala CLI (tests of export to Mill). I bumped the Mill version used when exporting to the one of https://github.com/VirtusLab/scala-cli/pull/581, that seems to have fixed the issue.

Issue #343, opened 2021-11-03
Cannot run Scala-cli with `sudo`


WojciechMazur:
In the case of trying to run `sudo scala-cli MyApp.scala` we get a large list of warnings and the build fails to connect with Bloop. On subsequent runs, we get different outputs. 
From this point, next runs without `sudo` privilege would also fail until we would run `scala-cli clean MyApp.scala`

<details>
  <summary>Click to show output on first run!</summary>
  
```
sudo scala-cli sRAPL.scala
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/3.0.1/scala3-compiler_3-3.0.1.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-interfaces/3.0.1/scala3-interfaces-3.0.1.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_3/3.0.1/scala3-library_3-3.0.1.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/tasty-core_3/3.0.1/tasty-core_3-3.0.1.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-asm/9.1.0-scala-1/scala-asm-9.1.0-scala-1.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/compiler-interface/1.3.5/compiler-interface-1.3.5.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-reader/3.19.0/jline-reader-3.19.0.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal/3.19.0/jline-terminal-3.19.0.jar doesn't exist!
Warning: Scala instance jar /root/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal-jna/3.19.0/jline-terminal-jna-3.19.0.jar doesn't exist!
[...]

PR #573, opened 2022-01-24
More strict workspace dir rights checks, rename it from .scala to .scala-build

alexarchambault:


alexarchambault:
I need to check, but that should fix https://github.com/VirtusLab/scala-cli/issues/343 (via the more strict rights checks)

romanowski (approved):

PR #576, opened 2022-01-24
Test examples on CI

### LINKED ISSUE #575: Check if code in examples is working.
Currently, our example does not work with lates scala-cli. We should fix it and create mechanism to check all those examples as a part of a CI

romanowski:


lwronski (commented):
- Maybe, instead of scripting in bash, we should write Scala Script.
- I don't think, that you should add `.scala.semanticdb` to git repository.

.github/workflows/ci.yml
```diff
@@ -202,6 +202,10 @@ jobs:
       run: .github/scripts/check_docs.sh
     - name: Test gifs
       run: gifs/generate_gifs.sh $(ls gifs/scenarios/)
+    - name: Build binary
+      run: ./mill copyTo cli.nativeImage ./scala-cli
```
alexarchambault: Better use the JVM launcher, it should be much faster to generate:
```suggestion
      run: ./mill copyTo cli.launcher ./scala-cli 
```

.github/workflows/ci.yml
```diff
@@ -202,6 +202,10 @@ jobs:
       run: .github/scripts/check_docs.sh
     - name: Test gifs
       run: gifs/generate_gifs.sh $(ls gifs/scenarios/)
+    - name: Build binary
+      run: ./mill copyTo cli.launcher ./scala-cli 
+    - name: Check examples
+      run: ./scala-cli --jvm temurin:17  .github/scripts/check_examples.sc
```
alexarchambault: This ought to work I think:
```suggestion
      run: bash ./scala-cli --jvm temurin:17  .github/scripts/check_examples.sc
```

PR #580, opened 2022-01-25
Hide some commands when called as 'scala'

alexarchambault:


alexarchambault:
Now realizing this won't work as is for Windows…

PR #581, opened 2022-01-25
Migrate build to Mill 0.10 compatible API and libs

lolgab:
## Motivation

Scala Native `0.4.3+` is supported only on Mill `0.10`. To have that supported Mill needs to be updated.

## Blockers

- [x] mill-nativeimage PR here: https://github.com/alexarchambault/mill-native-image/pull/5


alexarchambault:
Seems we're hitting a regression in recent coursier versions, with "Maven snapshot versioning" (the dates in file names [here](https://oss.sonatype.org/content/repositories/snapshots/org/virtuslab/using_directives/0.0.7-277bd4a-SNAPSHOT/) for example)…

lolgab:
Since there is a problem we can merge the source compatible changes first and then when the problem is solved just update `.mill-version`

alexarchambault (approved):
Merging, thanks @lolgab!

PR #582, opened 2022-01-25
Replace .scala remining .scala directories

romanowski:


modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -2,6 +2,7 @@ package scala.cli.integration
 
 import com.eed3si9n.expecty.Expecty.expect
 
+import scala.cli.integration.Constants
```
alexarchambault: Shouldn't be needed (we're already in the `scala.cli.integration` package).

alexarchambault (commented):
Apart from the (minor) comment, LGTM!

BTW, this originates from some of my PRs, merged independently around the same time, that clashed once merged (https://github.com/VirtusLab/scala-cli/pull/567 and https://github.com/VirtusLab/scala-cli/pull/573).

PR #584, opened 2022-01-26
Update GraalVM to 22.0.0

alexarchambault:


alexarchambault:
Not sure what's going on here… A `cl.exe` invocation of native-image [complains](https://github.com/VirtusLab/scala-cli/runs/4964944962?check_suite_focus=true#step:5:2701) that this file is not found:
```text
C:\Users\runneradmin\AppData\Local\Coursier\cache\arc\https\github.com\graalvm\graalvm-ce-builds\releases\download\vm-22.0.0.2\graalvm-ce-java17-windows-amd64-22.0.0.2.zip\graalvm-ce-java17-22.0.0.2\lib\svm\clibraries\windows-amd64\include\aarch64cpufeatures.h
```
But it exists [right before](https://github.com/VirtusLab/scala-cli/runs/4964944962?check_suite_focus=true#step:5:2633) and [right after](https://github.com/VirtusLab/scala-cli/runs/4964944962?check_suite_focus=true#step:5:2732) calling native-image… 🤔

alexarchambault:
Switching to GraalVM 22 on Windows looks really worthwhile: locally, it only takes ~3 minutes to generate a native image, and says it only needs ~5 GB of memory (while it used to sometimes fail with only 7 GB of memory).

alexarchambault:
Maybe the path of the file above has too many characters… (260)

PR #590, opened 2022-01-28
Use HttpURLConnection to download files

lwronski:


romanowski (approved):

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -29,4 +31,12 @@ object ProcUtil {
     usesSh
   }
 
+  def downloadFile(url: String): String = {
+    val connection = new URL(url).openConnection.asInstanceOf[HttpURLConnection]
+    connection.setRequestMethod("GET")
+    val inputStream = connection.getInputStream
+    val br          = new BufferedReader(new InputStreamReader(inputStream))
+    Stream.continually(br.readLine()).takeWhile(_ != null).mkString("\n")
```
alexarchambault: Not sure we should use a `BufferedReader` here… Let's just read it all in a `ByteArrayOutputStream`, and convert to a string in one go. Also, the connection isn't closed it seems. And we shouldn't need to set the request method to GET (that should be the default, AFAIK). Something like this should do:
```suggestion
    val buffer = new ByteArrayOutputStream
    val data   = Array.ofDim[Byte](16384)
    var inputStream: InputStream = null
    try {
      inputStream = new URL(url).openStream()
      var read = 0
      while ({
        read = inputStream.read(data, 0, data.length)
        read != -1
      })
        buffer.write(data, 0, read)
    }
    finally {
      if (inputStream != null)
        inputStream.close()
    }
    new String(buffer.toByteArray, StandardCharsets.UTF_8)
```

alexarchambault: (Recent JVMs might even have helpers to convert an `InputStream` to an `Array[Byte]`…)

alexarchambault: Yeah
```suggestion
    var inputStream: InputStream = null
    val data =
      try {
        inputStream = new URL(url).openStream()
        inputStream. readAllBytes()
      }
      finally {
        if (inputStream != null)
          inputStream.close()
      }
    new String(data, StandardCharsets.UTF_8)
```

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -29,4 +31,18 @@ object ProcUtil {
     usesSh
   }
 
+  def downloadFile(url: String): String = {
+    val connection = new URL(url).openConnection.asInstanceOf[HttpURLConnection]
+    connection.setRequestMethod("GET")
```
alexarchambault: ?
```suggestion
```

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -1,5 +1,7 @@
 package scala.cli.internal
 
+import java.io.InputStream
+import java.net.{HttpURLConnection, URL}
```
alexarchambault: With the suggestion below, that should become 
```suggestion
import java.net.URL
```

alexarchambault (approved):

PR #594, opened 2022-01-30
Quotation marks added to configuration example

Quafadas:
I believe the scala version must be enclosed in quotation marks.

This was a head scratcher for me - hoping it helps someone else.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,24 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Version(s)**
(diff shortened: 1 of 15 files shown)


lwronski:
@Quafadas Thanks for contributing, can you change target branch to `stable`? Because documentation is generated from the `stable` branch.

Instead of specify url to png, can you attach png image to repo files?

Quafadas:
Yes. Thankyou for the hint - bear with me on the PNG.

lwronski:
and I have one suggestion, can you change branch base from `master` to `stable`? It would be great if you add only one your commit to `stable` branch.

Changes from the `master` are merged to `stable` only after release, so unreleased functionalities are not visible in the documentation.

I think, we need to create some contributing guide.

Quafadas:
brb

Issue #294, opened 2021-10-25
Scala-cli should manage Scala Native dependencies.


romanowski:
Currently, Scala CLI requires cland (llvm) and optionally other native packaged installed manually. We can investigate ways how scala-cli can manage such dependencies by itself.

from @alexarchambault 

[conda](https://docs.conda.io/en/latest/miniconda.html) allows Python developers to install LLVM under specific directories, on Linux / macOS / Windows.

Installing conda ourselves could be cumbersome, but [micromamba](https://github.com/mamba-org/mamba/tree/fe4ecc5061a49c5b400fa7e7390b679e983e8456#micromamba) might help. micromamba is a dependency-free binary that can install conda packages. We could just download micromamba, and call it to install llvm under a specific directory.

jaroslawk:
The initial step was done here https://github.com/VirtusLab/scala-cli/pull/611
There is a need for follow-up PR connecting the environment properly.

PR #611, opened 2022-02-03
introduce managed clang environment

jaroslawk:
Automation attempting to install llvm based on mamba


modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -83,6 +85,25 @@ object Artifacts {
     logger: Logger
   ): Either[BuildException, Artifacts] = either {
 
+    addClang.map { _ =>
```
alexarchambault: That dangling call to `map` looks fishy:
```suggestion
    if (addClang.getOrElse(false)) {
```

alexarchambault: But when do we use that launcher? We should either fetch the launcher right before we use it, or keep the path to the fetched launcher in a field of `Artifacts`.

alexarchambault: (But maybe you just planned to do that in the final version of the PR?)

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -83,6 +85,25 @@ object Artifacts {
...
+      val cache = FileCache()
+      val task = {
+
+        val artifact = Artifact(url = CLangInstaller.mambaBinaryUrl
+        // checksumUrls TODO: verify check sum of archive
+        ).withChanging(true)
+        cache.file(artifact).run.flatMap {
+          case Left(e: Exception) => Task.fail(e)
```
alexarchambault: NIT Just in case, so that we keep this line in the stack traces we would get. We also don't need the `: Exception` annotation (I think it's actually an `ArtifactError`, but better not put type annotations like this in matches, when the match doesn't need to do an `asInstanceOf` under-the-hood)
```suggestion
          case Left(e) => Task.fail(new Exception(e))
```

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -83,6 +85,25 @@ object Artifacts {
...
+          case Right(archive: File) =>
+            Task.delay {
+              Task.fromEither(CLangInstaller.install(os.Path(archive.toPath), logger))
+            }
```
alexarchambault: The `Task.delay(Task.fromEither(…))` looks fishy… You might need a `.flatMap`, like `Task.delay(Task.fromEither(…)).flatMap(identity)`

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,48 @@
...
+
+final class CLangInstallException extends BuildException("Failed to Install CLang")
+
+object CLangInstaller {
+  /*
+  Installer based on micromamba attempts to install mamba and llvm based on that
+  Following https://gist.github.com/wolfv/fe1ea521979973ab1d016d95a589dcde
+   */
+  val mambaApiUrl = "https://micromamba.snakepit.net/api/micromamba"
+  def mambaBinaryUrl = Properties.osName match {
```
alexarchambault: We might want to make this return an `Either[BuildException, String]`, and check that the architecture (`sys.props("os.arch")` is valid too, see [here](https://github.com/VirtusLab/scala-cli/blob/f999de23f6b1aaead9078b0b9d1ae0cde55c3c84/modules/cli/src/main/scala/scala/cli/internal/FetchExternalBinary.scala#L112-L115) for example).

alexarchambault: The match doesn't look exhaustive too, so we might need an extra
```scala
    case unknownOs => …
```

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def install(microMambaArchive: Path, logger: Logger): Either[BuildException, Path] =
+    install(microMambaArchive, os.Path("/tmp/clang_output"), logger = logger)
+
+  def install(
+    microMambaArchive: Path,
+    outputPath: Path,
+    logger: Logger
+  ): Either[BuildException, Path] =
+    TarArchive.decompress(os.read.inputStream(microMambaArchive), outputPath)
```
alexarchambault: If you need to uncompress an archive, you can use `coursier.cache.ArchiveCache`. One is created [here](https://github.com/VirtusLab/scala-cli/blob/f999de23f6b1aaead9078b0b9d1ae0cde55c3c84/modules/build/src/main/scala/scala/build/options/BuildOptions.scala#L268) for example. Then its `get` method will give you a directory where the archive is extracted.

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,48 @@
...
+      .left.map(_ => new CLangInstallException)
+      .right.map { miniMambaPath: Path =>
+        val mambaPath = miniMambaPath / 'bin / "micromamba"
+
+        if (!Properties.isWin)
+          os.perms.set(mambaPath, "rwxr-xr-x")
```
alexarchambault: Do we really need that? Doesn't the unarchiving take care of this?

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,48 @@
...
+
+        /*
+        TODO: we need to have mamba initialized
+        ./micromamba shell init -s bash -p ~/micromamba
```
alexarchambault: Isn't there a way to use it without this?

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,48 @@
...
+        source ~/.bashrc
+        */
+        val installLLVM = s"$mambaPath install llvm -n base -c conda-forge -y".split(" ").toSeq
```
alexarchambault: We'll probably want to install things in a directory managed by Scala CLI. It can be added / managed via [`Directories`](https://github.com/VirtusLab/scala-cli/blob/f999de23f6b1aaead9078b0b9d1ae0cde55c3c84/modules/build/src/main/scala/scala/build/Directories.scala).

modules/build/src/main/scala/scala/build/TarArchive.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.commons.compress.archivers.tar.{TarArchiveEntry, TarArchiveInputStream}
+import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream
+import org.apache.commons.compress.utils.IOUtils
+import os.{Path, RelPath}
+
+import java.io.{FileOutputStream, InputStream}
+import java.nio.file.Files
+import scala.util.Try
+
+object TarArchive {
```
alexarchambault: We shouldn't need that if we use `coursier.cache.ArchiveCache` (see above).

modules/build/src/main/scala/scala/build/preprocessing/directives/UsingScalaNativeOptionsDirectiveHandler.scala
```diff
@@ -33,7 +33,8 @@ case object UsingScalaNativeOptionsDirectiveHandler extends UsingDirectiveHandle
       "nativeGc",
       "nativeVersion",
       "nativeCompile",
-      "nativeLinking"
+      "nativeLinking",
+      "installClang"
```
alexarchambault: I'd rather use a key like `nativeClang`, and accept values such as `"system"` or `"managed"` for it (with the latter corresponding to the ones managed by the changes in this PR).

modules/build/src/main/scala/scala/build/CLangInstaller.scala
alexarchambault: Seems other architectures might work: https://anaconda.org/conda-forge/micromamba/files

alexarchambault:
It seems the following installs things under a custom prefix, without the need to initialize anything:
```text
$ micromamba create -p /path/to/custom/prefix jupyter -c conda-forge
```

Then, in order to call a Scala Native sub-process, we should probably create a shell script calling the Scala Native command, but with the output of the following command at the top:
```text
$ micromamba shell -p /path/to/custom/prefix -s bash --no-rc --dry-run activate
```

This might also work on Windows, by replacing `-s bash` with `-s cmd.exe`, and reading the content of the file it prints the path of.

modules/build/src/main/scala/scala/build/CLangInstaller.scala
jaroslawk: nice 👍

modules/build/src/main/scala/scala/build/CLangInstaller.scala
jaroslawk: we do - under archive, there is no +x and I need to run it ...

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+
+import java.net.{HttpURLConnection, URL}
+import java.util.Locale
+
+import scala.build.EitherCps.either
+import scala.build.errors.BuildException
+import scala.build.internal.Runner
+import scala.util.{Properties, Try}
+
+final class CLangInstallException(message: String)
```
alexarchambault: NIT Can this be put in a separate file? (in order to have only one class per file)

alexarchambault:
`.tar.bz2` support should land in the upcoming coursier release, thanks to https://github.com/coursier/coursier/pull/2364

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+    // parsing os.arch example =>  https://github.com/trustin/os-maven-plugin/blob/master/src/main/java/kr/motd/maven/os/Detector.java
+    val arch = sys.props("os.arch").toLowerCase(Locale.ROOT)
+    (os, arch) match {
+      case ("win", "x86_64")    => Right("win-64")
+      case ("osx", "x86_64")    => Right("osx-64")
+      case ("osx", "arm64")     => Right("osx-arm64")
+      case ("linux", "x86_64")  => Right("linux-64")
+      case ("linux", "aarch64") => Right("linux-aarch64")
+      case ("linux", "ppc64le") => Right("linux-ppc64le")
+      case _ => Left(new CLangInstallException("Unsupported architecture/system combination."))
```
alexarchambault: Can `os` and `arch` be put in the error message?

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+        case Right(f) => Task.point(os.Path(f, os.pwd))
+      }.unsafeRun()(cache.ec)
+    }
+  }.left.map(_ => new CLangInstallException("Failed to fetch Mamba binary."))
+
+  def doInstall(
+    microMambaArchive: Path,
+    logger: Logger
+  ): Either[BuildException, Path] =
+    // TODO: we should relay on coursier to give us decompressed archive file but there is no bzip2 support yet
```
alexarchambault: It should be there now, in coursier `2.1.0-M5`.

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+    // for ArchiveCache => https://github.com/coursier/coursier/blob/master/modules/cache/jvm/src/main/scala/coursier/cache/ArchiveCache.scala#L151
+
+    TarArchive.decompress(os.read.inputStream(microMambaArchive), directories.mambaBaseDir)
+      .left.map(_ => new CLangInstallException("Failed to decompress minimamba archive."))
+      .right.map { miniMambaPath: Path =>
```
alexarchambault: NIT
```suggestion
      .map { miniMambaPath =>
```

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+          s"""
+             |#!/bin/sh
+             |$mambaPath -r $baseDir shell -s posix activate > $baseDir/activate_env
+             |. $baseDir/activate_env
+             |$mambaPath install compilers -r $baseDir -n base -c conda-forge -y
+             |""".stripMargin
+        os.write.over(installScript, activateAndInstall)
+        Runner.run(
+          "install llvm toolchain",
+          s"sh $installScript".split(" ").toSeq,
```
alexarchambault: Why split the string, when we can build the correct seq upfront?
```suggestion
          Seq("sh", installScript.toString),
```

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+          throw new CLangInstallException("Not Implemented yet.")
+
+        os.perms.set(mambaPath, "rwxr-xr-x")
+
+        val activateAndInstall =
+          s"""
+             |#!/bin/sh
+             |$mambaPath -r $baseDir shell -s posix activate > $baseDir/activate_env
+             |. $baseDir/activate_env
+             |$mambaPath install compilers -r $baseDir -n base -c conda-forge -y
```
alexarchambault: Some of these paths might have spaces in them. They should be quoted IMO.
```suggestion
             |"$mambaPath" -r "$baseDir" shell -s posix activate > "$baseDir/activate_env"
             |. "$baseDir/activate_env"
             |"$mambaPath" install compilers -r "$baseDir" -n base -c conda-forge -y
```

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+             |""".stripMargin
+        os.write.over(installScript, activateAndInstall)
+        Runner.run(
+          "install llvm toolchain",
```
alexarchambault: This string is unused (it's only used when `allowExecve = true` is passed), better make that explicit
```suggestion
          "unused",
```

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -678,6 +678,20 @@ class BuildTests extends munit.FunSuite {
...
+    val inputs = TestInputs(
+      os.rel / "p.sc" ->
+        """//> using `native-version` "0.4.0"
+          |//> using nativeClangManaged
+          |""".stripMargin
+    )
+    val buildOptions =
+      defaultOptions.copy(internal = defaultOptions.internal.copy(keepDiagnostics = true))
+    inputs.withBuild(buildOptions, buildThreads, bloopConfig) { (_, _, maybeBuild) =>
+      assert(maybeBuild.toOption.get.options.scalaNativeOptions.clangManaged.get == true)
```
alexarchambault:
```suggestion
      assert(maybeBuild.toOption.get.options.scalaNativeOptions.clangManaged.contains(true))
```
(shoudn't throw before the assertion is checked, if ever `clangManaged` is empty)

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+      case ("osx", "arm64")     => Right("osx-arm64")
+      case ("linux", "x86_64")  => Right("linux-64")
+      case ("linux", "aarch64") => Right("linux-aarch64")
+      case ("linux", "ppc64le") => Right("linux-ppc64le")
+      case _ => Left(new CLangInstallException("Unsupported architecture/system combination."))
+    }
+  }
+
+  def resolveMicroMambaBinaryUrl(platform: String) = Try {
+    val con: HttpURLConnection = new URL(s"$mambaApiUrl/$platform/latest")
```
alexarchambault, line 52: It seems this is going to run every time users rely on micromamba-managed stuff. That means users won't be able to use micromamba-managed directory if they don't have an internet connection, even if the right micromamba environment is there…

IMO, we should just hardcode the URLs this provides, unless these look too unstable, but they don't (I'm getting `https://api.anaconda.org/download/conda-forge/micromamba/0.21.0/osx-64/micromamba-0.21.0-0.tar.bz2` for example).

We should also hardcode the micromamba version, via `project/deps.sc` and `Constants.scala` (generated in `build.sc`), so that there's no need to check the latest version.

modules/build/src/main/scala/scala/build/CLangInstaller.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  def fetchFile(url: String, logger: Logger) = either {
+    val cache = FileCache().withLogger(logger.coursierLogger)
+    cache.logger.use {
+      cache.file(Artifact(url).withChanging(true)).run.flatMap {
+        case Left(e)  => Task.fail(new Exception(e))
+        case Right(f) => Task.point(os.Path(f, os.pwd))
+      }.unsafeRun()(cache.ec)
+    }
+  }.left.map(_ => new CLangInstallException("Failed to fetch Mamba binary."))
```
alexarchambault: This discard the original exception. Better change `CLangInstallException` so that it accepts a cause:
```suggestion
  }.left.map(e => new CLangInstallException("Failed to fetch Mamba binary", e))
```

alexarchambault (commented):
It seems we don't use `directories.mambaBaseDir` later on, when compiling Scala Native code.

Maybe you should add a docker-based test of this, trying to run a Scala Native script via Scala CLI, with no clang installed beforehand. And the test would check that Scala CLI does indeed fetch and uncompress clang.

For that to work, we might need to add the right mamba `bin` dir in `PATH` before calling scala-native-cli around [here](https://github.com/VirtusLab/scala-cli/blob/956e0c132ea5a2ee791edcf502eace0652ad1f5b/modules/cli/src/main/scala/scala/cli/commands/Package.scala#L576-L583).

modules/build/src/main/scala/scala/build/CLangInstaller.scala
jaroslawk: 👍

modules/build/src/main/scala/scala/build/CLangInstaller.scala
jaroslawk: 👍

Gedochao:
It seems this PR has gotten stale, there's been no activity for 5 months as of this comment.
I'm closing it for now, any future progress should be tracked in #294

PR #616, opened 2022-02-03
Fix generate help for shebang

### LINKED ISSUE #601: Truncated shebang command help description
**Version(s)**

0.1.0

**Describe the bug**

Command-line help for *shebang* command currently states:

```
This command is an equivalent of `run`, but it changes the way how
```

Apparently the description is incomplete and seems to have been truncated.

**To Reproduce**

Run:
```
scala-cli --help
```

Look for the line mentioning the *shebang* command.

**Expected behaviour**

The command description is complete.

lwronski:
Add a shorter description for shebang command in `scala-cli --help`

Replace from 
``` 
shebang    This command is an equivalent of `run`, but it changes the way how
```
to
```
shebang   Like 'run', but more handy from shebang scripts
```

Closes #601

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
@@ -3,8 +3,10 @@ package scala.cli.commands
 import caseapp._
 
 @HelpMessage(
-  """|This command is an equivalent of `run`, but it changes the way how
-     |`scala-cli` parses it's command-line arguments in order to be compatibility
+  """|Like 'run', but more handy from shebang scripts
+     |
+     |This command is equivalent to `run`, but it changes the way
+     |`scala-cli` parses its command-line arguments in order to be compatible
      |with shebang scripts.
      |
      |Normally, inputs and scala-cli options can be mixed. Program have to be specified after `--`
--- a/website/docs/reference/commands.md
+++ b/website/docs/reference/commands.md
@@ -200,8 +200,10 @@ Accepts options:
 
 ## `shebang`
 
-This command is an equivalent of `run`, but it changes the way how
-`scala-cli` parses it's command-line arguments in order to be compatibility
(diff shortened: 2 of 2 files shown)


modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
```diff
@@ -3,7 +3,9 @@ package scala.cli.commands
 import caseapp._
 
 @HelpMessage(
-  """|This command is an equivalent of `run`, but it changes the way how
+  """|Access to command line arguments in shebang scripts
```
alexarchambault: I'm not sure about that exact wording… Maybe:
```suggestion
  """|Like 'run', but more handy from shebang scripts
```

modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
```diff
@@ -3,7 +3,9 @@ package scala.cli.commands
...
+     |
+     |This command is an equivalent of `run`, but it changes the way how
```
alexarchambault:
```suggestion
     |This command is equivalent to `run`, but it changes the way
```
(no need of "how" here, I think, and not 100% sure, but "equivalent to" sounds better to me)

modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
```diff
@@ -3,7 +3,9 @@ package scala.cli.commands
...
      |`scala-cli` parses it's command-line arguments in order to be compatibility
```
alexarchambault:
```suggestion
     |`scala-cli` parses its command-line arguments in order to be compatible
```
(typos)

modules/cli/src/main/scala/scala/cli/commands/ShebangOptions.scala
lwronski: It sounds better.

PR #618, opened 2022-02-04
Provide proper logger for using directives plus Tweaking 

alexarchambault:
These are small adjustments / refactorings, that I made while working on adding a `publish` command.


build.sc
```diff
@@ -613,6 +616,14 @@ class TastyLib(val crossScalaVersion: String) extends ScalaCliCrossSbtModule
 }
 
 object `local-repo` extends LocalRepo {
+
+  /*
+   * If you are developing locally on any of the stub modules (stubs, runner, test-runner),
+   * set this to true, so that Mill's watch mode takes into account changes in those modules
+   * when embedding their JARs in the Scala CLI launcher.
+   */
+  def developingOnStubModules = false
```
romanowski, line 625: Should we add similar comment to ReadME.md?

modules/build/src/main/scala/scala/build/errors/NoValueProvidedError.scala
```diff
@@ -0,0 +1,12 @@
+package scala.build.errors
+
+import scala.build.preprocessing.directives.StrictDirective
+
+final class NoValueProvidedError(
+  val directive: StrictDirective
+) extends BuildException(
+      s"Expected a value for directive ${directive.key}",
+      positions = Nil // I wish using_directives provided the key position…
```
romanowski, line 9: Directives provide positions so it is a matter of wiring it up

romanowski (commented):
We can apply my comments in the follow up. The only problem I see is that it will conflict with #639

PR #638, opened 2022-02-08
Avoid compiling scope test for package/run/ commands

### LINKED ISSUE #589: Don't compile test target when you package
I'd like to be able to use Scala 3 and produce a native image via Scala Native. While my code can do that, the testing framework I'm using isn't published for Scala Native. When trying to publish my image it won't publish because of a test dependency.

I'm not sure this behavior is expected, and I'm not 100% sure if this even needs to be, but I think it'd be nice to not have the package command rely on the test compilation. This would unlock more code bases to be able to produce a native image for example like I'm trying to.

Move my code into different directories and try to only point the package command towards them.

lwronski:


modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -212,41 +213,47 @@ object Build {
...
+            }
+          }
+          Some(testBuild)
         }
-      }
+        else None
 
       doPostProcess(mainBuild, inputs0, Scope.Main)
-      doPostProcess(testBuild, inputs0, Scope.Test)
+      if (testBuildOpt.nonEmpty) doPostProcess(testBuildOpt.get, inputs0, Scope.Test)
```
alexarchambault: More idiomatic way of doing that (without a `.get` in particular):
```suggestion
      for (testBuild <- testBuildOpt)
        doPostProcess(testBuild, inputs0, Scope.Test)
```

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -259,20 +266,20 @@ object Build {
...
 
         (mainBuild, extraMainBuilds, testBuild, extraTestBuilds)
       }
 
-    val (mainBuild, extraBuilds, testBuild, extraTestBuilds) = value(buildScopes())
+    val (mainBuild, extraBuilds, testBuildOpt, extraTestBuilds) = value(buildScopes())
 
     copyResourceToClassesDir(mainBuild)
-    copyResourceToClassesDir(testBuild)
+    if (testBuildOpt.nonEmpty) copyResourceToClassesDir(testBuildOpt.get)
```
alexarchambault: Same as above (we can avoid the `.get`)

modules/cli/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -110,7 +110,8 @@ final case class SharedOptions(
   def buildOptions(
     enableJmh: Boolean,
     jmhVersion: Option[String],
-    ignoreErrors: Boolean = false
+    ignoreErrors: Boolean = false,
+    shouldCompileTest: Boolean = false
```
alexarchambault: Maybe this should be an `Option`, defaulting to `None`, so that nothing is changed in `TestOptions` by default.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -139,7 +139,8 @@ object Build {
     logger: Logger,
     buildClient: BloopBuildClient,
     bloopServer: bloop.BloopServer,
-    crossBuilds: Boolean
+    crossBuilds: Boolean,
+    isTest: Boolean
```
alexarchambault: NIT Maybe `buildTests` would be a better name, that's what we really mean.

alexarchambault (commented):
We may want to ensure that `--class-path` makes `compile` print the full test scope class path, rather than the main one, when `--test` is passed.

For later: maybe the new CLI option could be moved to `CompileCrossOptions`. That would make most commands accept it, but we'd have to think about how each of them should use it (`repl` could put test classes in the class path, `publish` would publish test JARs alongside the main ones, `test` could even only look for tests in the main scope if `--test=false` is passed, etc.)

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -52,7 +53,13 @@ object Compile extends ScalaCommand[CompileOptions] {
       }
       else if (options.classPath)
         for (s <- builds.main.successfulOpt) {
-          val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
+          val testBuildOpt = builds.get(Scope.Test)
+          val testFullClassPath =
+            testBuildOpt.flatMap(_.successfulOpt).map(_.fullClassPath).getOrElse(Nil)
+          val fullClassPath = s.fullClassPath.union(testFullClassPath).distinct
```
alexarchambault: I'm not sure about the `union`… Keeping either the main class path or the test one should be enough. Something like this a few lines above should do:
```scala
for (s <- builds.get(Scope.Test).orElse(builds.main.successfulOpt))
```

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -70,7 +71,7 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
             |  }
             |}
             |""".stripMargin,
-        os.rel / "Tests.scala" ->
+        os.rel / "Tests.test.scala" ->
```
alexarchambault: Isn't the `using target.scope "test"` just below enough?

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -87,7 +88,17 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
+      val output =
+        os.proc(TestUtil.cli, "compile", "--test", "--class-path", extraOptions, ".").call(cwd =
+          root).out.text().trim
+      val escapeBackslash = if (Properties.isWin) "\\\\/" else "\\/"
+      val testPathRegex =
+        s":$root/${Constants.workspaceDirName}/project_.*?/classes/test".replace(
+          "/",
+          escapeBackslash
+        ).r
+      val isDefinedTestPathInClassPath = testPathRegex.findFirstMatchIn(output).isDefined
```
alexarchambault: Split the lines of `output` (`output.linesIterator`), and look for one starting with `(root / Constants.workspaceDirName).toString` and ending with `Seq("classes", "test").mkString(File.separator)`, rather than using a regex? (regexes can be hard to get right sometimes…)

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -101,7 +112,7 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
             |    println(message)
             |}
             |""".stripMargin,
-        os.rel / "Tests.scala" ->
+        os.rel / "Tests.test.scala" ->
```
alexarchambault: Same as above, do we really need that as we already have `using target.scope "test"`?

lwronski: Yes, we don't need change name of file to `*.test.scala` if we have `using target.scope "test"`

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
lwronski: Yes, good point. So I will avoid using regex.

alexarchambault (approved):

PR #639, opened 2022-02-08
Add support for using directive in java files

### LINKED ISSUE #533: Support "using lib" for java scripts
It would be really great if the `using lib` directive was also available for java based scripts.
For example : https://gist.github.com/dacr/e4d928a85acd580b4a756fd7ff36a735
```java
// using lib "org.apache.commons:commons-lang3:3.12.0"

import org.apache.commons.lang3.StringUtils;

public class Issue533UsingLibForJava {
  public static void main(String[] args) {
    var message = "hello world";
    System.out.println(StringUtils.capitalize(message));
  }
}
```

Right now it prints : 
```
$ scl https://gist.github.com/dacr/e4d928a85acd580b4a756fd7ff36a735
[error] e4d928a85acd580b4a756fd7ff36a735-master/Issue533UsingLibForJava.java:13:1: package org.apache.commons.lang3 does not exist
Error compiling project (Scala 3.0.2, JVM)
Compilation failed

```

ghostbuster91:
This closes #533 

Joint work with @mtk, @romanowski in the framework of the issue spree.

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
@@ -1,0 +1,120 @@
+package scala.build.preprocessing
+
+import com.virtuslab.using_directives.config.Settings
+import com.virtuslab.using_directives.custom.model.{
+  UsingDirectiveKind,
+  UsingDirectiveSyntax,
+  UsingDirectives
+}
+import com.virtuslab.using_directives.custom.utils.ast.{UsingDef, UsingDefs}
+import com.virtuslab.using_directives.{Context, UsingDirectivesProcessor}
+
+import scala.build.errors._
+import scala.build.preprocessing.directives.StrictDirective
+import scala.build.{Logger, Position}
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
+case class ExtractedDirectives(
+  offset: Int,
+  directives: Seq[StrictDirective],
+  kind: UsingDirectiveKind
+)
+
+object ExtractedDirectives {
+
+  val changeToSpecialCommentMsg =
+    "Using directive using plain comments are deprecated. Please use a special comment syntax: '//> ...' or '/*> ... */'"
+
+  def from(
+    contentChars: Array[Char],
+    path: Either[String, os.Path],
+    logger: Logger
+  ): Either[BuildException, ExtractedDirectives] = {
+    val errors = new mutable.ListBuffer[Diagnostic]
+    val reporter = CustomDirectivesReporter.create(path) { diag =>
+      if (diag.severity == Severity.Warning)
+        logger.log(Seq(diag))
+      else
+        errors += diag
+    }
+    val processor = {
+      val settings = new Settings
+      settings.setAllowStartWithoutAt(true)
+      settings.setAllowRequire(false)
+      val context = new Context(reporter, settings)
+      new UsingDirectivesProcessor(context)
+    }
+    val all = processor.extract(contentChars, true, true).asScala
+    if (errors.isEmpty) {
+
+      def byKind(kind: UsingDirectiveKind) = all.find(_.getKind == kind).get
+
+      def getDirectives(directives: UsingDirectives) =
+        directives.getAst() match {
+          case ud: UsingDefs =>
+            ud.getUsingDefs().asScala
+          case _ =>
+            Nil
+        }
+
+      val codeDirectives           = byKind(UsingDirectiveKind.Code)
+      val specialCommentDirectives = byKind(UsingDirectiveKind.SpecialComment)
+      val plainCommentDirectives   = byKind(UsingDirectiveKind.PlainComment)
+
+      def reportWarning(msg: String, values: Seq[UsingDef], before: Boolean = true): Unit =
+        values.foreach { v =>
+          val astPos = v.getPosition()
+          val (start, end) =
+            if (before) (0, astPos.getColumn())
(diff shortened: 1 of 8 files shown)


modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
```diff
@@ -11,9 +16,31 @@ case object JavaPreprocessor extends Preprocessor {
...
   ): Option[Either[BuildException, Seq[PreprocessedSource]]] =
     input match {
-      case j: Inputs.JavaFile =>
-        Some(Right(Seq(PreprocessedSource.OnDisk(j.path, None, None, Nil, None))))
-
+      case j: Inputs.JavaFile => Some(either {
+          val content   = value(PreprocessingUtil.maybeRead(j.path))
+          val scopePath = ScopePath.fromPath(j.path)
+          val ExtractedDirectives(_, directives0) =
+            value(from(
```
ghostbuster91, line 23: It would be nice if we could enrich the `DirectiveError` here with the information that it happened in the java context but the `DirectiveError` is an opaque class.

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -221,6 +221,63 @@ class SourcesTests extends munit.FunSuite {
...
+      expect(
+        sources.buildOptions.classPathOptions.extraDependencies.toSeq.map(_.value) == expectedDeps
+      )
+      expect(sources.paths.length == 1)
+      expect(sources.paths.map(_._2) == Seq(os.rel / "Something.java"))
+      expect(sources.inMemory.isEmpty)
+    }
+  }
+
+  test("should fail dependencies in .java - //> using") {
```
romanowski:
```suggestion
  test("should fail dependencies in .java  with using keyword") {
```

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -0,0 +1,107 @@
...
+      values.foreach { v =>
+        val astPos = v.getPosition()
+        val (start, end) =
+          if (before) (0, astPos.getColumn())
+          else (astPos.getColumn(), astPos.getColumn() + v.getSyntax.getKeyword.size)
+        val position = Position.File(path, (astPos.getLine(), start), (astPos.getLine(), end))
+        logger.diagnostic(msg, positions = Seq(position))
+      }
+
+    val usedDirectives =
```
romanowski: In a situation when Java file starts with following code:

```
using foo 1
//> using bar 2
```

User will get a warning that bar is ignored and error that  foo is unsupported what is a bit confusing.

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -0,0 +1,107 @@
...
+    val strictDirectives =
+      flattened.map { case (k, l) => StrictDirective(k.getPath.asScala.mkString("."), l.asScala) }
+
+    val offset =
+      if (usedDirectives.getKind() != UsingDirectiveKind.Code) 0
+      else usedDirectives.getCodeOffset()
+    if (supportedDirectives.contains(usedDirectives.getKind()))
+      Right(ExtractedDirectives(offset, strictDirectives))
+    else
+      Left(new DirectiveErrors(::(s"Unsupported using directive kind ${usedDirectives.getKind}", Nil)))
```
romanowski: The message may be confusing since in general keyword using directives are supported but not in the context of Java files and message should reflect that.

romanowski: Also, I've just realized that we do not provide a location where are the problematic cases. I think we should use something similar `UnusedDirectiveError`

romanowski (changes requested):
The code looks good, the UX parts needs to be polished. We can also do it as a follow up PR, especially that we need to update documentation etc.

romanowski:
@ghostbuster91 we are cutting a release and we needed #618 that introduce conflicts with your PR. I am happy to fix the merge conflicts for you, just let me know if I should do it.

ghostbuster91:
> @ghostbuster91 we are cutting a release and we needed #618 that introduce conflicts with your PR. I am happy to fix the merge conflicts for you, just let me know if I should do it.

Feel free, I don't mind doing it either way.

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
ghostbuster91: > The message may be confusing since in general keyword using directives are supported but not in the context of Java files and message should reflect that.

Agree.

The problem is that higher layers cannot refine error messages from lower layers as they are opaque in the current error design. Here we call same function `ExtractedDirectives.from` from two different context: scala and java. Each of these context should enrich the message. If only `DirectiveErrors` was a case class...


I can change that but I am afraid that it will makes this PR much bigger.

ghostbuster91:
When I tried running `./mill -i scala https://gist.github.com/dacr/e4d928a85acd580b4a756fd7ff36a735` I got the same error as in the original issue. However when I run the script locally (`./mill -i scala Issue533UsingLibForJava.java`) it doesn't fail but helloWorlds doesn't get printed either. Any idea what is the reason behind such behavior?

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
ghostbuster91: Would you like to resolve priority only against supported directives?

PR #641, opened 2022-02-09
Fix Scala Native directory project caching

jchyb:
Previously if `scala-cli run srcDir --native` was run, only the initial project would be cached and used, with following changes to the codebase being ignored.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -26,26 +26,10 @@ final case class Inputs(
 
   def singleFiles(): Seq[Inputs.SingleFile] =
     elements.flatMap {
-      case f: Inputs.SingleFile => Seq(f)
-      case d: Inputs.Directory =>
-        os.walk.stream(d.path)
-          .filter { p =>
-            !p.relativeTo(d.path).segments.exists(_.startsWith("."))
-          }
-          .filter(os.isFile(_))
-          .collect {
-            case p if p.last.endsWith(".java") =>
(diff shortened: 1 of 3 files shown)


alexarchambault (commented):
👍 

It would be neat to have a non-regression test for that.

romanowski (approved):

jchyb:
Sure! I'll add some tests

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -137,11 +105,16 @@ final case class Inputs(
...
+            Seq("dir:") ++ singleFilesFromDirectory(dirInput)
+              .map(file => s"${file.path}:" + os.read(file.path))
+          case resDirInput: Inputs.ResourceDirectory =>
+            // Resource changes for SN require relinking, so they should also be hashed
+            Seq("resource-dir:") ++ os.walk(resDirInput.path)
+              .map(filePath => s"$filePath:" + os.read(filePath))
+          case _ => Seq(os.read(elem.path))
         }
-        Iterator(elem.path.toString, content, "\n").map(bytes)
+        Iterator((elem.path.toString +: content :+ "\n"): _*).map(bytes)
```
alexarchambault: We don't need to build a new string here, only to convert it to bytes from the iterator later on. This should have worked:
```suggestion
        (Iterator(elem.path.toString) ++ content.iterator ++ Iterator("\n")).map(bytes)
```

alexarchambault (approved):
We can address the point in my comment in another PR…

PR #648, opened 2022-02-10
Don't compute "hash data" for options that are not used by Bloop

alexarchambault:
This PR makes that more explicit via `notForBloopOptions: PostBuildOptions` in `BuildOptions`.

This hash is only used to make Bloop directories unique, so it doesn't need to take package or repl options into account.

### DIFF
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -38,8 +38,7 @@ final case class BuildOptions(
   internal: InternalOptions = InternalOptions(),
   mainClass: Option[String] = None,
   testOptions: TestOptions = TestOptions(),
-  packageOptions: PackageOptions = PackageOptions(),
-  replOptions: ReplOptions = ReplOptions()
+  notForBloopOptions: PostBuildOptions = PostBuildOptions()
 ) {
 
   lazy val platform: Positioned[Platform] =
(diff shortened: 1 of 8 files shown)


modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -38,8 +38,7 @@ final case class BuildOptions(
   internal: InternalOptions = InternalOptions(),
   mainClass: Option[String] = None,
   testOptions: TestOptions = TestOptions(),
-  packageOptions: PackageOptions = PackageOptions(),
-  replOptions: ReplOptions = ReplOptions()
+  notForBloopOptions: PostBuildOptions = PostBuildOptions()
```
romanowski, line 41: I think we should call it `postBuildOptions`

romanowski (approved):

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
alexarchambault: Or maybe rename `PostBuildOptions` to `NotForBloopOptions `? I wanted to make it clear that it shouldn't be used when generating the Bloop config for example.

romanowski: It is strange that BuildOptions contains PostBuild / NotForBloop. maybe we should Refactor it at some point so we have an Option object that contains BuildOptions (ones that bloop cares for), runtime options (like main class, tests, repl things), package options etc.

alexarchambault: That's what I wanted to do initially, something like
```scala
final case class AllBuildOptions(
  buildOptions: BuildOptions = BuildOptions(),
  postBuildOption: PostBuildOptions = PostBuildOptions()
)
```

I just fear it'll touch many-many files in the codebase.

romanowski: ok, we can live now with `notForBloopOptions` and refactor option to `AlLBuildOption` at some point

PR #649, opened 2022-02-10
Build CLI with Scala 2.13

alexarchambault:


romanowski (approved):

lwronski:
I think, it should be also updated `scala` version in `scala-cli` dependency - [here](https://github.com/VirtusLab/scala-cli/blob/6de98417c8f98e9a9377391bdcb59797e5f8f2f6/modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala#L14)

alexarchambault:
(The CI error on Windows looks legit… I'd like to debug it locally on a Windows machine.)

alexarchambault:
Includes https://github.com/VirtusLab/scala-cli/pull/689 as of now (should go away after a rebase once https://github.com/VirtusLab/scala-cli/pull/689 is merged).

alexarchambault:
Still having one issue in 2.13: when verbosity is increased, we sometimes use [PPrint](https://github.com/com-lihaoyi/PPrint) to print stuff. In 2.13, PPrint [uses reflection](https://github.com/com-lihaoyi/PPrint/blob/3abeed80a4282f2c08768528eff7e24c07d05603/pprint/src-2.13/pprint/StringPrefix.scala#L4) to call `Iterable.collectionClassName`. From native images, this would require concrete `Iterable` classes to be added to the native-image reflection config to work fine, which is kind of a hassle. https://github.com/com-lihaoyi/PPrint/pull/80 should make PPrint not rely on reflection for that.

PR #650, opened 2022-02-11
add support for 2.nightly & 3.nightly and not specifying the repository in getting 2 nightly versions

### LINKED ISSUE #571: `using scala 3-nightly`
Scala-CLI allows defining the following using directive:

```scala
using scala "3"
```

which sets the Scala version to the latest stable version.

It would be a great improvement to be able to define a using directive like:

```scala
using scala 3-nightly
```

which would set the Scala version to the latest nightly.

ghost:
it resolves #571

### DIFF
--- a/modules/build/src/main/scala/scala/build/Artifacts.scala
+++ b/modules/build/src/main/scala/scala/build/Artifacts.scala
@@ -129,12 +129,23 @@ object Artifacts {
         Nil
     }
 
+    val scala2NightlyRepo = Seq(coursier.Repositories.scalaIntegration.root)
+
     val scalaNativeCliDependency =
       scalaNativeCliVersion.map(version =>
         Seq(dep"org.scala-native:scala-native-cli_2.12:$version")
       )
 
-    val allExtraRepositories = maybeSnapshotRepo ++ extraRepositories
+    val scala2NightlyRegex = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
+
+    val isScala2NightlyRequested = scala2NightlyRegex.unapplySeq(params.scalaVersion).isDefined
+
+    val allExtraRepositories = {
+      val baseRepos =
+        maybeSnapshotRepo ++ extraRepositories
+      if (isScala2NightlyRequested) baseRepos ++ scala2NightlyRepo
+      else baseRepos
+    }
 
     val updatedDependencies =
       dependencies ++
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -369,7 +369,7 @@ final case class BuildOptions(
           Versions()
             .withModule(scala3)
             .result()
-            .unsafeRun()(ec)
+            .unsafeRun()(finalCache.ec)
         }
         res.versions.latest(Latest.Release)
       }
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -274,9 +274,16 @@ final case class BuildOptions(
   def finalRepositories: Seq[String] =
     classPathOptions.extraRepositories ++ internal.localRepository.toSeq
 
-  private def computeScalaVersions(
-    scalaVersion: Option[String],
-    scalaBinaryVersion: Option[String]
+  /** @param scalaVersionArg
+    *   the command line, using directive, or default argument passed as scala version
+    * @param scalaBinaryVersionArg
+    *   the command line, using directive, or default argument passed as scala Binary version
+    * @return
+    *   Either a BuildException or the calculated (ScalaVersion, ScalaBinaryVersion) tuple
+    */
+  private def turnScalaVersionArgToScalaVersions(
+    scalaVersionArg: Option[String],
+    scalaBinaryVersionArg: Option[String]
   ): Either[BuildException, (String, String)] = either {
     lazy val allVersions = {
       import coursier._
@@ -285,8 +292,9 @@ final case class BuildOptions(
         def scala2 = mod"org.scala-lang:scala-library"
         // No unstable, that *ought* not to be a problem down-the-line…?
         def scala3 = mod"org.scala-lang:scala3-library_3"
-        if (scalaVersion.contains("2") || scalaVersion.exists(_.startsWith("2."))) Seq(scala2)
-        else if (scalaVersion.contains("3") || scalaVersion.exists(_.startsWith("3."))) Seq(scala3)
+        if (scalaVersionArg.contains("2") || scalaVersionArg.exists(_.startsWith("2."))) Seq(scala2)
+        else if (scalaVersionArg.contains("3") || scalaVersionArg.exists(_.startsWith("3.")))
+          Seq(scala3)
         else Seq(scala2, scala3)
       }
       def isStable(v: String): Boolean =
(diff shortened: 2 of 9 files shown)


romanowski (approved):

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -334,21 +342,85 @@ final case class BuildOptions(
...
+        def scala3 = mod"org.scala-lang:scala3-library_3"
+        val res = finalCache.logger.use {
+          Versions()
+            .withModule(scala3)
+            .result()
+            .unsafeRun()(ec)
+        }
+        res.versions.latest(Latest.Release)
+      }
+      val scalaVersion       = moduleVersion.get
```
alexarchambault: We shouldn't call `Option.get` here. Better do something like
```suggestion
      val scalaVersion       = value(moduleVersion.toRight("No latest version found for $scala3"))
```
(and make `scala3`, which should be a `val`, above accessible from here)

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -334,21 +342,85 @@ final case class BuildOptions(
...
+        val res = finalCache.logger.use {
+          Versions()
+            .withModule(scalaNightly2Module)
+            .withRepositories(Seq(coursier.Repositories.scalaIntegration))
+            .result()
+            .unsafeRun()(ec)
+        }
+        res.versions.latest(Latest.Release)
+      }
+      val scalaVersion       = moduleVersion.get
```
alexarchambault, line 406: Same as above, no `Option.get`.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -334,21 +342,85 @@ final case class BuildOptions(
...
+  private def computeLatestScalaThreeNightlyVersions(): Either[BuildException, (String, String)] =
+    either {
+      import coursier.Versions
+      import coursier.core.Latest
+      import coursier._
+      import scala.concurrent.ExecutionContext.{global => ec}
+      val moduleVersion: Option[String] = {
+        def scala3 = mod"org.scala-lang:scala3-library_3"
+        val res = finalCache.logger.use {
+          Versions()
```
alexarchambault: The coursier cache coming from the options should be passed here
```suggestion
          Versions(finalCache)
```

(this might need to be fixed below too, and in the `computeScalaVersions` method that was there before this PR too)

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -334,21 +342,85 @@ final case class BuildOptions(
...
+            .withModule(scala3)
+            .result()
+            .unsafeRun()(ec)
```
alexarchambault: Let's use the `ExecutionContext` of `finalCache` (which should have threads ready to run things):
```suggestion
            .unsafeRun()(finalCache.ec)
```

(this too might need to be fixed below, and in the `computeScalaVersions` method that was there before this PR)

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
alexarchambault: Actually, we might need to create a new `BuildException` sub-class in `scala.build.errors`, and create it here
```suggestion
      val scalaVersion       = value(moduleVersion.toRight(new NoScalaVersionFoundError(scala3)))
```

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -28,6 +28,43 @@ class BuildOptionsTests extends munit.FunSuite {
     )
   }
 
+  test("-S 3.nightly option works") {
+    val options = BuildOptions(
+      scalaOptions = ScalaOptions(
+        scalaVersion = Some("3.nightly"),
+        scalaBinaryVersion = None,
+        supportedScalaVersionsUrl =
```
lwronski: It' seems, that you don't have to set `supportedScalaVersionsUrl`, because this test doesn't test this functionality.

```suggestion
        supportedScalaVersionsUrl = None
```

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -28,6 +28,43 @@ class BuildOptionsTests extends munit.FunSuite {
...
+      "-S 3.nightly argument does not lead to scala3 nightly build option"
+    )
+  }
+
+  test("-S 2.nightly option works") {
+    val options = BuildOptions(
+      scalaOptions = ScalaOptions(
+        scalaVersion = Some("2.nightly"),
+        scalaBinaryVersion = None,
+        supportedScalaVersionsUrl =
```
lwronski, line 54: Same as above

modules/build/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -129,12 +129,23 @@ object Artifacts {
...
 
+    val scala2NightlyRepo = Seq(coursier.Repositories.scalaIntegration.root)
+
     val scalaNativeCliDependency =
       scalaNativeCliVersion.map(version =>
         Seq(dep"org.scala-native:scala-native-cli_2.12:$version")
       )
 
-    val allExtraRepositories = maybeSnapshotRepo ++ extraRepositories
+    val scala2NightlyRegex = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
```
lwronski, line 139: I think, that this regex should start from `2`

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -28,6 +28,43 @@ class BuildOptionsTests extends munit.FunSuite {
...
+          Some(
+            Random.alphanumeric.take(10).mkString("")
+          ) // invalid url, it should use defaults from Deps.sc
+      )
+    )
+    val scalaParams        = options.scalaParams.orThrow
+    val scala2NightlyRegex = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
```
lwronski, line 61: I think that regex for scala 2 nightly should only be defined once in code. I see, that was defined 3 times.

modules/build/src/main/scala/scala/build/Artifacts.scala
ghost: What do you mean?

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
ghost: @lwronski What is the best place for defining it? I wanted to get it from coursier, but the api did not provide a means for that.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: I have solved this in my scala-version-error commit which is in draft stage.

ghost: I return `Either` from the `moduleVersion`, then use `value` for extracting it.

ghost: These functions are changed quite a bit in the scala-version-error PR

ghost: I am thinking about closing this PR and only keeping the scala-version-error, since the same functions are modified in there.

modules/build/src/main/scala/scala/build/Artifacts.scala
lwronski: I think, this regex should works for input `3.13.2-bin-c4434f`.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
lwronski: I think we should open PR per one issue. It's easier to make a code review.

ghost:
For now, things are working. All of these problems are either addressed in the next PR for scala-version-error, or actually belong there. This commit history has become too messy. I am closing this pull request for now and create an original identical one with a clean commit history. Then these matters need to be assessed on the scala-version-error.

modules/build/src/main/scala/scala/build/Artifacts.scala
ghost: Definitely. I am going to correct that. Thank you!

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
ghost: Yep

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: It makes a lot of sense.

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
zmerr: @lwronski

ghost: @alexarchambault do you think it is a good idea for now to create a pull request to add a function to coursier Versions, as I have DM’ed you for solving this?
Something like this:
```
def getBinaryType(scalaVersion: String): ScalaBinaryType =
    scalaVersion match {
      case Scala3EarlyVersion(milestone)         =>  ScalaBinaryType.Scala3EarlyVersion
      case Scala3Version(_, _)                   =>  ScalaBinaryType.Scala3
      case ReleaseVersion(major, minor, _)       => ScalaBinaryType.ReleaseVersion
      case MinorSnapshotVersion(major, minor, _) => ScalaBinaryType.MinorSnapshotVersion
      case NightlyVersion(major, minor, _)       => ScalaBinaryType.NightlyVersion
      case DottyVersion(minor, _)                => ScalaBinaryType.DottyVersion
      case TypelevelVersion(major, minor, _)     => ScalaBinaryType.TypeLevelVersion
      case _                                     => ScalaBinaryType.Default
    }
```

zmerr: Something like this:
```
def getBinaryType(scalaVersion: String): ScalaBinaryType =
    scalaVersion match {
      case Scala3EarlyVersion(milestone)         =>  ScalaBinaryType.Scala3EarlyVersion
      case Scala3Version(_, _)                   =>  ScalaBinaryType.Scala3
      case ReleaseVersion(major, minor, _)       => ScalaBinaryType.ReleaseVersion
      case MinorSnapshotVersion(major, minor, _) => ScalaBinaryType.MinorSnapshotVersion
      case NightlyVersion(major, minor, _)       => ScalaBinaryType.NightlyVersion
      case DottyVersion(minor, _)                => ScalaBinaryType.DottyVersion
      case TypelevelVersion(major, minor, _)     => ScalaBinaryType.TypeLevelVersion
      case _                                     => scalaVersion
    }
```

lwronski: I think, it should be added in `ScalaParse.scala`

ghost: I would be doing that. Just a moment.

ghost: I have put it in the BuildOptions companion object, since it is only related to BuildOptions.

PR #659, opened 2022-02-16
Compute common root for inputs elements

### LINKED ISSUE #646: Scala CLI does not compute root project properly
**Version(s)**

`0.1.0`

**Describe the bug**

With the following directory structure:

```
.
├── a
│   └── A.scala
└── b
    └── B.scala
```

When runnning `scala-cli a/A.scala b/B.scala` the `.scala-build` and `.bap` directories are generated in `a` instead of `cwd` (current directory).

I also think that Scala CLI should expose a command line option to manually provide a project root, mainly to generate `.bsp` directory in the right place.

lwronski:
#646

### DIFF
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -238,7 +238,8 @@ object Inputs {
   private def forValidatedElems(
     validElems: Seq[Element],
     baseProjectName: String,
-    directories: Directories
+    directories: Directories,
+    cwd: os.Path
   ): Inputs = {
 
     assert(validElems.nonEmpty)
@@ -249,7 +250,7 @@ object Inputs {
       }
       .getOrElse {
         validElems.head match {
-          case elem: SourceFile => (elem.path / os.up, true)
+          case _: SourceFile => (cwd, true)
           case _: Virtual =>
             val dir = homeWorkspace(validElems, directories)
             (dir, false)
@@ -368,7 +369,7 @@ object Inputs {
       }.flatten
       assert(validElems.nonEmpty)
 
-      Right(forValidatedElems(validElems, baseProjectName, directories))
+      Right(forValidatedElems(validElems, baseProjectName, directories, cwd))
     }
     else
       Left(invalid.mkString(System.lineSeparator()))
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -235,27 +235,44 @@ object Inputs {
     dir
   }
 
+  private def computeCommonRoot(validElems: Seq[Element]): os.Path = {
+    val sourceDirectories = validElems.collect {
+      case s: SourceFile => s.path / os.up
+      case d: Directory  => d.path
+    }.toList
+
+    assert(sourceDirectories.nonEmpty)
+
+    def findCommonRoot(
+      candidateRootSegments: Iterator[String],
+      directories: Seq[os.Path]
+    ): os.Path =
+      directories match {
+        case dirPath :: rest =>
+          val updatedCandidate = candidateRootSegments.zip(dirPath.segments)
+            .takeWhile(Function.tupled(_ == _))
+            .map(_._1)
+          findCommonRoot(updatedCandidate, rest)
+        case Nil => os.root / candidateRootSegments.toList
+      }
+
+    findCommonRoot(sourceDirectories.head.segments, sourceDirectories.tail)
+  }
+
   private def forValidatedElems(
     validElems: Seq[Element],
     baseProjectName: String,
-    directories: Directories,
-    cwd: os.Path
+    directories: Directories
   ): Inputs = {
 
     assert(validElems.nonEmpty)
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
I don't think we should use `os.pwd` here… That means the following commands will write `.scala-build` directories at two different places:
```text
$ scala-cli a/A.scala b/B.scala # writes under ./.scala-build
$ mkdir c && cd c
$ scala-cli ../a/A.scala ../b/B.scala # writes under ./c/.scala-build
```

When individual files are passed, and don't live in the same directory or sub-directory of others (like `a/A.scala a/b/B.scala` should use `a/.scala-build`), we should probably assume we have to write things under the dedicated directory in the home dir.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -235,6 +235,30 @@ object Inputs {
...
+      candidateRootSegments: Iterator[String],
+      directories: Seq[os.Path]
+    ): os.Path =
+      directories match {
+        case dirPath :: rest =>
+          val updatedCandidate = candidateRootSegments.zip(dirPath.segments)
+            .takeWhile(Function.tupled(_ == _))
+            .map(_._1)
+          findCommonRoot(updatedCandidate, rest)
+        case Nil => os.root / candidateRootSegments.toList
```
alexarchambault, line 256: Not sure `os.root` works fine on Windows, when several drives (`C:\`, `D:\`, …) are involved.

alexarchambault: When `candidateRootSegments.toList` is empty, we should just use a directory under home dir at the end.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -243,18 +267,12 @@ object Inputs {
 
     assert(validElems.nonEmpty)
 
-    val (workspace, needsHash) = validElems
-      .collectFirst {
-        case d: Directory => (d.path, true)
-      }
```
alexarchambault, line 249: I would have kept this section. It's not really the problem here.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -243,18 +267,12 @@ object Inputs {
...
-      .getOrElse {
-        validElems.head match {
-          case elem: SourceFile => (elem.path / os.up, true)
-          case _: Virtual =>
-            val dir = homeWorkspace(validElems, directories)
-            (dir, false)
-          case _: Directory => sys.error("Can't happen")
-        }
```
alexarchambault, line 257: `computeCommonRoot` should only apply from here IMO.

modules/build/src/main/scala/scala/build/Inputs.scala
lwronski: How about this case?
```
$ scala-cli a/ b/ # writes under ./a/.scala-build
$ scala-cli b/ a/  # writes under ./b/.scale-build
```

I think, it should generate only one target. Therefore, I compute common root also for directories.

modules/build/src/main/scala/scala/build/Inputs.scala
lwronski: What if someone runs scala-cli in the `/tmp` directory. I wouldn't expect that `scala-cli` generate the files under home dir.

lwronski: Screenshot 2022-02-21 at 12 20 34
It seems, that `os.root` working fine on Windows, if there are several drives

alexarchambault: What I mean is that `os.root` is usually `C:\` on Windows. So if we're passed files on `D:\` for example, that will be a problem. We just shouldn't rely on `os.root` on Windows.

modules/build/src/main/scala/scala/build/Inputs.scala
alexarchambault: Both look fine to me. The order matters.

I'd expect users to pass the "main" directory first, and we put `.scala-build` in that one.

alexarchambault:
I thought a bit about this, and I'm really wary… I still think favoring the first argument (if it's a directory, just use it as workspace, if it's a file, use its parent as workspace) works better.

When trying to guess a workspace like here, we might pick a directory not really related to the arguments, which would be surprising, like in these:
```text
$ scala-cli ~/projects/first/app ~/projects/second/Helper.scala # will pick ~/projects
$ scala-cli ~/projects/first/app ~/scripts/Helper.scala # will pick ~/
```
Here, we're picking some top directories that we just shouldn't pick IMO. When stumbling upon `.scala-build` directories in these, users will likely be annoyed by those and wonder where those come from…

prolativ:
For me personally, when using scala-cli, it was a big surprise that the order of arguments mattered. IMHO that would only make sense e.g. to guarantee that some files are compiled in a specific order in separate runs of the compiler (such thing is sometimes done internally in test of the compiler but I've probably never had a need to do that in a different context).
So maybe a warning or error should be raised in cases when there is no reasonable default for the output directory? Then a user would have to specify it explicitly.

PR #663, opened 2022-02-18
show update info in About command on linux

### LINKED ISSUE #543: diffucult to understand which version of scala-cli I have or if it is the latest or if I need to update or if I have duplicates on path
* scala-cli should accept --version and version and perhaps also empty -v:
* if I run `scala-cli about` I'm not getting the real version 
(I have 0.0.9 but below `about` is showing something else 0.0.0+274-... so user is confused if it is the latest version or not)
* can scala-cli let me know if there is a new version to install?
* can scala-cli update itself?
```
$ scala-cli --version
Unrecognized argument: --version

$ scala-cli -v
No inputs provided (expected files with .scala or .sc extensions, and / or directories).

$ scala-cli version
version: not found

$ scala-cli about
Scala CLI version 0.0.0+274-gb0d66b5-SNAPSHOT (latest-274-b0d66b)

$ curl -sSLf https://virtuslab.github.io/scala-cli-packages/scala-setup.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   667  100   667    0     0   2744      0 --:--:-- --:--:-- --:--:--  2744
100 36.7M  100 36.7M    0     0  23.2M      0  0:00:01  0:00:01 --:--:-- 29.2M
Scala-cli 0.0.9 is already installed and up-to-date.
```

ghost:
It resolves #543

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
@@ -12,14 +12,10 @@ class ScalaCliCommands(
   isSipScala: Boolean
 ) extends CommandsEntryPoint {
 
-  val appName =
-    if (isSipScala) "Scala command"
-    else "Scala CLI"
-
   lazy val actualDefaultCommand = new Default(help)
 
   private def allCommands = Seq[ScalaCommand[_]](
-    new About(appName),
+    new About(isSipScala = isSipScala),
     AddPath,
     BloopExit,
     BloopStart,
--- a/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
@@ -12,10 +12,14 @@ class ScalaCliCommands(
   isSipScala: Boolean
 ) extends CommandsEntryPoint {
 
+  val appName =
+    if (isSipScala) "Scala command"
+    else "Scala CLI"
+
   lazy val actualDefaultCommand = new Default(help)
 
   private def allCommands = Seq[ScalaCommand[_]](
-    new About(isSipScala = isSipScala),
+    new About(appName),
     AddPath,
     BloopExit,
     BloopStart,
--- a/modules/cli/src/main/scala/scala/cli/commands/About.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/About.scala
@@ -2,15 +2,19 @@ package scala.cli.commands
 
 import caseapp._
 
+import scala.build.internal.Constants
 import scala.cli.CurrentParams
 
-class About(appName: String) extends ScalaCommand[AboutOptions] {
+class About(isSipScala: Boolean) extends ScalaCommand[AboutOptions] {
   override def group = "Miscellaneous"
   def run(options: AboutOptions, args: RemainingArgs): Unit = {
     CurrentParams.verbosity = options.verbosity.verbosity
-    println(s"$appName version ${Version.version}" + Version.detailedVersionOpt.fold("")(
-      " detailed version: (" + _ + ")"
-    ))
+    val version            = Constants.version
+    val detailedVersionOpt = Constants.detailedVersion.filter(_ != version)
+    val appName =
+      if (isSipScala) "Scala command"
+      else "Scala CLI"
+    println(s"$appName version $version" + detailedVersionOpt.fold("")(" (" + _ + ")"))
     if (Version.isOutdated(None)) println(Update.updateInstructions)
   }
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/About.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/About.scala
@@ -2,18 +2,15 @@ package scala.cli.commands
 
 import caseapp._
 
-import scala.build.internal.Constants
 import scala.cli.CurrentParams
 
-class About(isSipScala: Boolean) extends ScalaCommand[AboutOptions] {
+class About(appName: String) extends ScalaCommand[AboutOptions] {
   override def group = "Miscellaneous"
   def run(options: AboutOptions, args: RemainingArgs): Unit = {
     CurrentParams.verbosity = options.verbosity.verbosity
-    val version            = Constants.version
-    val detailedVersionOpt = Constants.detailedVersion.filter(_ != version)
-    val appName =
-      if (isSipScala) "Scala command"
-      else "Scala CLI"
-    println(s"$appName version $version" + detailedVersionOpt.fold("")(" (" + _ + ")"))
+    println(s"$appName version ${Version.version}" + Version.detailedVersionOpt.fold("")(
+      " detailed version: (" + _ + ")"
+    ))
+    if (Version.isOutdated(None)) println(Update.updateInstructions)
   }
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
@@ -8,10 +8,11 @@ import scala.util.Try
 
 object CommandUtils {
 
-  /** we only want to check for update within 10% random calls; not everytime
-    */
-  lazy val shouldCheckUpdate: Boolean =
-    scala.util.Random.nextInt % 10 == 1
+  def isOutOfDateVersion(newVersion: String, oldVersion: String): Boolean = {
+    import coursier.core.Version
+
+    Version(newVersion) > Version(oldVersion)
+  }
 
   // Ensure the path to the CLI is absolute
   def getAbsolutePathToScalaCli(programName: String): String =
@@ -34,4 +35,5 @@ object CommandUtils {
         }
         .getOrElse(programName)
 
+  lazy val shouldCheckUpdate: Boolean = scala.util.Random.nextInt() % 10 == 1
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
@@ -1,12 +1,37 @@
 package scala.cli.commands
 
+import java.io.File
+import java.nio.file.Paths
+
+import scala.build.Os
+import scala.util.Try
+
 object CommandUtils {
 
-  def isOutOfDateVersion(newVersion: String, oldVersion: String): Boolean = {
-    import coursier.core.Version
+  /** we only want to check for update within 10% random calls; not everytime
+    */
+  lazy val shouldCheckUpdate: Boolean =
+    scala.util.Random.nextInt % 10 == 1
 
-    Version(newVersion) > Version(oldVersion)
-  }
+  // Ensure the path to the CLI is absolute
+  def getAbsolutePathToScalaCli(programName: String): String =
+    if (programName.contains(File.separator))
+      os.Path(programName, Os.pwd).toString
+    else
+      /*
+  In order to get absolute path we first try to get it from coursier.mainJar (this works for standalone launcher)
+  If this fails we fallback to getting it from this class and finally we may also use rawArg if there is nothing left
+       */
+      sys.props.get("coursier.mainJar")
+        .map(Paths.get(_).toAbsolutePath.toString)
+        .orElse {
+          Try(
+            // This is weird but on windows we get /D:\a\scala-cli...
+            Paths.get(getClass.getProtectionDomain.getCodeSource.getLocation.toURI)
+              .toAbsolutePath
+              .toString
+          ).toOption
+        }
+        .getOrElse(programName)
 
-  lazy val shouldCheckUpdate: Boolean = scala.util.Random.nextInt() % 10 == 1
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
@@ -82,6 +82,10 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     if (virtualInputs.nonEmpty)
       value(Left(new FoundVirtualInputsError(virtualInputs)))
 
+    val progName = argvOpt.flatMap(_.headOption).getOrElse {
+      sys.error("setup-ide called in a non-standard way :|")
+    }
+
     val logger = options.shared.logger
 
     if (options.buildOptions.classPathOptions.extraDependencies.toSeq.nonEmpty)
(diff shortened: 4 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
```diff
@@ -8,5 +8,11 @@ object CommandUtils {
     Version(newVersion) > Version(oldVersion)
   }
 
-  lazy val shouldCheckUpdate: Boolean = scala.util.Random.nextInt() % 10 == 1
+  /** there is no implementation for this val
```
romanowski: I think the comment should stay that we only want to check for update within 10% random calls not everytime.

modules/cli/src/main/scala/scala/cli/commands/Update.scala
```diff
@@ -50,48 +57,97 @@ object Update extends ScalaCommand[UpdateOptions] {
...
+    val base = s"""Your Scala CLI version is outdated. The newest version is $newestScalaCliVersion
+                  |It is recommended that you update Scala CLI through the same tool or method you used for its initial installation for avoiding the creation of outdated duplicates.""".stripMargin
+    if (pathInstances.isRight)
+      pathDuplicatesWarning.right.get
+    else base
+  }
+
+  /** Either an error or the warning about more than one instance of scala-cli on PATH
+    */
+  lazy val pathDuplicatesWarning: Either[String, String] = either {
```
romanowski: It is just an message so I think it should not return `Either` but just a `String`

modules/cli/src/main/scala/scala/cli/commands/Update.scala
```diff
@@ -50,48 +57,97 @@ object Update extends ScalaCommand[UpdateOptions] {
...
+    if (scalaCliPathInstances.exitCode == 0)
+      Right(scalaCliPathInstances.out.text().trim.split(sys.props("line.separator")).toList)
+    else Left(scalaCliPathInstances.err.text().trim())
+  }
+
+  def checkUpdate(options: UpdateOptions): Unit = {
+
+    val scalaCliBinPath = options.installDirPath / options.binaryName
+
+    lazy val absolutePathToScalaCli: String = {
```
romanowski: This looks like a copied logic from `SetupIde.scala`. Can we extract it to some common place?

romanowski: also, why this is a lazy val inside def?

modules/cli/src/main/scala/scala/cli/commands/Update.scala
romanowski: Also there is a hey problem with this feature namely on Ubunbu I've got scala-cli installed in 2 places:

```
⋊> ~ type -ap scala-cli                                               
/usr/bin/scala-cli
/bin/scala-cli
```

I was using provided installer and both are a scripts that calls the same scala-cli instance on `/usr/share/scala/scala-cli` - so I would get the warning everytime and it would be really annoying. I think we need to discuss how this should work on our standup.

romanowski (commented):
I am not sure if there is a good way to implement this feature. We need to discuss it on our team meeting.

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: So if these two results are both fed to `ls -l` it would show that they both refer to the same location. So this requires one extra step of feeding to `ls -l` and then parsing the returned result to make sure these are real duplicates.
The warning would only be shown on update, about and version commands.

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: yes. a good idea.

ghost: yes. it would get calculated any time the def is called anyway. So perhaps, after the extraction to another different location, it would make sense to keep it lazy.

modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
ghost: ok, I would remove the `TODO`  then

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: regarding the `Either`, for example, if the operating system is windows, it cannot calculate whether there is any duplicates on path or not. So, in those cases, it should not show a warning to the user.

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: I extracted to `CommandUtils`

modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
ghost: solved!

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: solved!

modules/cli/src/main/scala/scala/cli/commands/Update.scala
ghost: I removed the duplicate finding code!

modules/cli/src/main/scala/scala/cli/commands/Update.scala
```diff
@@ -50,48 +52,46 @@ object Update extends ScalaCommand[UpdateOptions] {
...
-      else
-        "0.0.0"
+  lazy val updateInstructions: String =
+    s"""Your Scala CLI version is outdated. The newest version is $newestScalaCliVersion
+       |It is recommended that you update Scala CLI through the same tool or method you used for its initial installation for avoiding the creation of outdated duplicates.""".stripMargin
+
+  def update(options: UpdateOptions, maybeScalaCliBinPath: Option[os.Path]): Unit = {
+    val maybeCurrentVersion = maybeScalaCliBinPath.map {
+      scalaCliBinPath =>
+        val res = os.proc(scalaCliBinPath, "version").call(cwd = os.pwd, check = false)
```
lwronski: It's seems, that you changed `Version` command and now prints 
```
s"$appName version ${Version.version}" + Version.detailedVersionOpt.fold("")(
       " detailed version: (" + _ + ")"
     )
```
so this logic won't be working, because `os.proc(scalaCliBinPath, "version")` return `printVersionInfo`. 

`maybeCurrentVersion` can contains only valid `Versions`, which can be parsed by `coursier.core.Version`. 

We use `maybeCurrentVersion` and `newestScalaCliVersion`, to check if current version of scala-cli is out to date - [CommandUtils. isOutOfDateVersion](https://github.com/VirtusLab/scala-cli/blob/main/modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala#L5)

ghost: So, how is this creating a problem? I just don’t get it...

ghost: after  `os.proc(scalaCliBinPath, "version”)` returns `printVersionInfo` perhaps I can add logic for finding the first string matching the version regex as you have in the code.

ghost: Plus, That code would only get executed in tests, because in normal running circumstance, the `scalaCliBinPath` is going to be `None`, so only `Constants.version` is going to be used.

ghost: solved it by refactoring and making the version command only print the version.

modules/cli/src/main/scala/scala/cli/commands/Version.scala
```diff
@@ -4,11 +4,47 @@ import caseapp._
 
 import scala.build.internal.Constants
 import scala.cli.CurrentParams
+import scala.cli.internal.ProcUtil
 
-object Version extends ScalaCommand[VersionOptions] {
+class Version(appName: String) extends ScalaCommand[VersionOptions] {
```
lwronski:
```suggestion
object Version extends ScalaCommand[VersionOptions] {
```
`appName` is unused in `Version` class.

modules/cli/src/main/scala/scala/cli/commands/Version.scala
```diff
@@ -4,11 +4,47 @@ import caseapp._
...
   def run(options: VersionOptions, args: RemainingArgs): Unit = {
     CurrentParams.verbosity = options.verbosity.verbosity
-    println(Constants.version)
+    println(Version.version)
   }
 }
+
+object Version {
+
+  val scalaCliVersionRegex = "tag/v(.*?)\"".r
```
lwronski: This is specific regex for GitHub releases, when we discover scala version. We can move this regex to block in `newestScalaCliVersion`. This regex should be used only when, some download https://github.com/VirtusLab/scala-cli/releases/latest and trying to resolve scala version.

modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
```diff
@@ -36,7 +40,7 @@ class ScalaCliCommands(
     Shebang,
     Test,
     Update,
-    Version
+    new Version(appName)
```
lwronski: This changes in [ScalaCliCommands.scala](https://github.com/VirtusLab/scala-cli/pull/663/files#diff-8ed11a35d409eda367a95fc06934cb06407b69465e219d1eb05b3027eef9fe7d) can be removed.  They don't seem to change anything.

modules/cli/src/main/scala/scala/cli/commands/About.scala
```diff
@@ -2,18 +2,19 @@ package scala.cli.commands
...
-      else "Scala CLI"
-    println(s"$appName version $version" + detailedVersionOpt.fold("")(" (" + _ + ")"))
+    printDetailedVersionInfo
+  }
+
+  def printDetailedVersionInfo = {
+    println(s"$appName version ${Version.version}" + Version.detailedVersionOpt.fold("")(
+      " detailed version: (" + _ + ")"
+    ))
+    if (Version.isOutdated(None)) println(Update.updateInstructions)
```
lwronski, line 18: I would not except, that `printDetailedVersionInfo` check, if scala-cli is out to date. I suggest, to move execution of this method to `def run(options: AboutOptions, args: RemainingArgs)`

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -82,7 +79,7 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     if (virtualInputs.nonEmpty)
       value(Left(new FoundVirtualInputsError(virtualInputs)))
 
-    val progName = argvOpt.flatMap(_.headOption).getOrElse {
+    argvOpt.flatMap(_.headOption).getOrElse {
```
lwronski: Why this variable was removed? I think, it should be used to resolve `absolutePathToScalaCli`.

modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
ghost: sending the appName to `Version` you mean?

modules/cli/src/main/scala/scala/cli/commands/About.scala
ghost: I didn’t get it! It’s getting called in the `def run` right now! So can you please explain a little bit more?

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
ghost: aha, `progName` was not used anywhere in the code! so scalafix was showing error for it! But the implementation remains.

modules/cli/src/main/scala/scala/cli/commands/Version.scala
ghost: Yep, I’m gonna fix it!

modules/cli/src/main/scala/scala/cli/commands/Version.scala
ghost: good idea! So I should move it inside `newestScalaCliVersion`?

modules/cli/src/main/scala/scala/cli/commands/About.scala
lwronski: yea, my mistake

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
ghost: I restored it inside the absolutePath calculation in `CommandUtils`
Can you please take a look?

ghost:
FYI, the conflict with the main branch is caused by the refactoring of the absolute path calculation inside `CommandUtils`

ghost:
@romanowski  this is ready to get merged!

ghost:
@romanowski  the functionality of PR #741  is overlapping this one

modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
```diff
@@ -1,12 +1,43 @@
...
+  /** we only want to check for update within 10% random calls; not everytime
+    */
+  lazy val shouldCheckUpdate: Boolean = {
+    scala.util.Random.nextInt % 10 == 1
+  }
+
+  // Ensure the path to the CLI is absolute
+  lazy val absolutePathToScalaCli: String = {
+
+    val progName = getProgName
```
lwronski: if I use `absolutePathToScalaCli`  in new command, I can get unexpected error `sys.error("setup-ide called in a non-standard way :|")`.

ghost: I guess I don’t know how to solve it! Because passing it as parameter also causes confusion with other `progName` values in the code, which can get passed by mistake! Any suggestion welcome!

ghost: I think it is unlikely that you would get the error though, since `setup-side`’s `run` method would not be called in here!

ghost: Currently, `absolutePathToScalaCli` is getting called in the `update` command! so if the integration tests pass, then this shows that the error would not cause a problem.

ghost: Ok, solved by passing `programName` to `getAbsolutePathToScalaCli`

PR #666, opened 2022-02-18
Allow to disable Bloop file content comparison

alexarchambault:
~Not sure we should merge that right now.~ While trying to find the origin of some slowdowns when preparing my Scala Love talk, I found that generating Bloop file JSONs and comparing them to their values on disk was quite costly (~200 ms, but I'd like to re-check those benchmarks).

On top of that, they're unnecessary… most of the time. As we write them under `project_-inputs-hash-` directories, we only compute them from the same set of inputs, so they're very unlikely to change.

Yet, they *might* change. They reference many artifacts, and these could change over time. That can be the case for snapshot artifacts, or if users change their coursier cache location. For that reason, I'm not sure we should ~merge that right now~ enable this by default. I'd like to keep refactoring things, so that we know if some artifacts are considered "changing" (snapshots), and we take into account the coursier cache location in the project hash. Maybe that should be enough to make it safe to enable that optimization by default.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -115,6 +115,8 @@ object Build {
     def diagnostics: None.type   = None
   }
 
+  def defaultStrictBloopJsonCheck = true
+
   def updateInputs(
     inputs: Inputs,
     options: BuildOptions,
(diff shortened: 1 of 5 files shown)


alexarchambault:
It's disabled by default now, so this could be merged.

It could be interesting to enable it in the Scala CLI build itself, to possibly save quite some time when watching sources in particular.

PR #667, opened 2022-02-18
More lightweight coursier logging

alexarchambault:
This is the logging that was enabled in my Scala Love talk earlier today. This basically:
- prints a message if and only if anything is being downloaded (so if all artifacts are in cache, nothing is printed)
- this doesn't keep on screen the progress bars usually printed by coursier, so that these don't flood the screen

To specify the message to be printed, see the `cache.withMessage` calls. The message is somehow passed to the coursier cache, that passes it to its logger (`withMessage` is an extension method, added in this PR too).


alexarchambault:
I might have seen some spurious output, that might need investigation (previous lines being cleared by the progress bars), but I could be wrong on this.

PR #672, opened 2022-02-21
Stop reading global bloop config

### LINKED ISSUE #640: Scala-cli's bloop instance isn't working properly and interferes with original one.
**Version(s)**

Scala CLI version `0.1.0` (v0.1.0)

**Describe the bug**

As a Metals user, I have Bloop instance running in the background almost all of the time. Yesterday was the first time I've tried to use scala-cli `0.1.0` (which uses its own bloop instance) and I was surprised that Metals cannot auto-connect to the scala-cli as a build server. The next thing I discovered is that scala-cli doesn't work because it can't start its own bloop instance, namely every command such as `run` or `compile` doesn't work and yields:
```shell
Bloop daemon status: not running
Starting Bloop 1.4.19 at /Users/kpodsiadlo/Library/Caches/ScalaCli/bloop/daemon using JVM /Users/kpodsiadlo/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17%252B35/OpenJDK17-jdk_x64_mac_hotspot_17_35.tar.gz/jdk-17+35/Contents/Home/bin/java
Error: java.lang.Exception: Server didn't start
For more details, please see '/Users/kpodsiadlo/software/scala-cli-sample/.scala-build/stacktraces/1644385522-808445731027933856.log'
``` 
The content of the `1644385522-808445731027933856.log` is:
```scala
java.lang.Exception: Server didn't start
  scala.build.blooprifle.internal.Operations$.$anonfun$startServer$2(Operations.scala:153)
  scala.build.blooprifle.BloopRifleLogger.$anonfun$runnable$1(BloopRifleLogger.scala:12)
  java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
  java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
  java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
  java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
  java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
  java.lang.Thread.run(Thread.java:833)
  com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:597)
  com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:194)
```

**_I managed to successfully run scala-cli only after I manually killed the "normal" bloop instance._**

However, after that I encountered another problem - Metals won't be able to start the original Bloop for their own purposes. So in order to get Bloop started I had to manually kill scala-cli's Bloop instance. 

Recently I've been observing many Metals users having trouble with connecting to the Bloop in Metals and I'm afraid that it's not a coincidence that it started after scala-cli `0.1.0` release. I hope not, but I experienced the same issues as users on my own, something isn't right :/

**To Reproduce**
1. Start with Bloop instance running in the background (may be one started by Metals)
2. Check if scala-cli is working
2a. Try to run scala-cli command such as `run` or `compile`
2b. Alternatively, open scala-cli project and wait for Metal aren't able to connect to the build server
3. Kill Bloop manually
4. Scala-cli starts working

**Expected behaviour**
scala-cli's bloop instance doesn't interfere with original one

lwronski:
https://github.com/VirtusLab/scala-cli/issues/640#issuecomment-1033499643

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/SharedCompilationServerOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/SharedCompilationServerOptions.scala
@@ -78,7 +78,7 @@ final case class SharedCompilationServerOptions(
   @Group("Compilation server")
   @HelpMessage("Bloop global options file")
   @Hidden
-    bloopGlobalOptionsFile: String = (os.home / ".bloop" / "bloop.json").toString,
+    bloopGlobalOptionsFile: Option[String] = None,
 
   @Group("Compilation server")
   @HelpMessage("JVM to use to start Bloop (e.g. 'system|11', 'temurin:17', …)")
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/scala/cli/commands/SharedCompilationServerOptions.scala
```diff
@@ -185,26 +185,28 @@ final case class SharedCompilationServerOptions(
...
-        )
-        val bloopJson = upickle.default.read(json)(BloopJson.jsonCodec)
-        bloopJson.javaOptions
-      }
-      catch {
-        case e: Throwable =>
-          System.err.println(s"Error parsing global bloop config in '$filePath':")
-          e.printStackTrace()
+  def bloopDefaultJvmOptions(): Option[List[String]] = {
+    val filePathOpt = bloopGlobalOptionsFile.map(os.Path(_, Os.pwd))
```
alexarchambault: NIT (allows to ignore `--bloop-global-options-file ""`)
```suggestion
    val filePathOpt = bloopGlobalOptionsFile.filter(_.trim.nonEmpty).map(os.Path(_, Os.pwd))
```

modules/cli/src/main/scala/scala/cli/commands/SharedCompilationServerOptions.scala
```diff
@@ -185,26 +185,28 @@ final case class SharedCompilationServerOptions(
         BloopVersion(Constants.bloopVersion)
       ))(v => BloopRifleConfig.Strict(BloopVersion(v)))
 
-  def bloopDefaultJvmOptions(logger: Logger): List[String] = {
```
alexarchambault, line 188: Why don't we use the `logger` to print stuff? (I saw there were some `println`s before already, but I'm not sure why they're here…)

lwronski: Right, a logger should be used

alexarchambault (approved):

PR #673, opened 2022-02-21
Tweaking

alexarchambault:
This:
- ~enables Scala.JS linking on Windows (used to be disabled because of weird GraalVM errors)~ (still disabled because of possible memory issues on CI, moved to https://github.com/VirtusLab/scala-cli/pull/676)
- ~enables more Scala Native tests (many were disabled for Scala 3 and now pass)~ (seems munit and all [aren't published](https://github.com/scalameta/munit/pull/477) for Scala 3 + Native yet, this will have to wait)
- passes a logger taking into account verbosity to the Scala.JS linker
- tweaks other minor stuff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,33 +13,6 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  bloop-memory-footprint:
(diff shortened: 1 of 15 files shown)


alexarchambault:
Seems we're a bit short on memory on Windows again, a bit worrying… 🤔

PR #674, opened 2022-02-21
Fix scala-cli version used in our build

romanowski:


alexarchambault:
Why not just disable the use of Scala CLI on Windows?

romanowski:
It is not only Windows it also fails on Linux (my machine) and macs (Ina's). It is caused by the problem with handling multiple `-O` options

alexarchambault:
A problem is that `--cli-version` makes us use the JVM version of Scala CLI, which is kind of slow.

romanowski:
problem will be fixed in next release, we do not want to have it merged

PR #655, opened 2022-02-15
Scala version error support

ghost:
it resolves #574 and #571

PR #677, opened 2022-02-21
Convinient way to support nightly versions of compiler + more meaningfull error messages about missconfigured Scala versions

romanowski:
Rebased version of #655 with some latest touches applied.

All the hard work by @zmerr, I am only a merger tool ;)


alexarchambault:
Shouldn't this have been squashed-and-merged? Commits such as "Apply suggestions from code review" don't really make sense on the main branch (they make it hard to bisect things for example), the numerous "Update website/docs/commands/compile.md" could probably have been squashed, and it seems some other commits could be been squashed too (the two "scala2NightlyRegex to ScalaParse.scala" for example).

romanowski:
It is my fault, commits should be squashed merged.

PR #688, opened 2022-02-22
Setup-ide fails if is call by non-owner

### LINKED ISSUE #656: Invalid `canWrite` check
https://github.com/VirtusLab/scala-cli/blob/6efbefb1d864c0ee36156f9ac8489d0e14ee54c4/modules/build/src/main/scala/scala/build/Inputs.scala#L129

`scala-cli` checks if the directory is writable, but it's doing it incorrectly. If file permissions are `000`, then even the owner can't write to it. On the other hand, if it is `777` then anyone (so non-owners) are allowed to write to it.

lwronski:
I'm not sure if the condition is sufficient.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
@@ -11,7 +11,7 @@ import java.nio.file.Paths
 
 import scala.build.EitherCps.{either, value}
 import scala.build.Inputs.WorkspaceOrigin
-import scala.build.errors.BuildException
+import scala.build.errors.{BuildException, WorkspaceError}
 import scala.build.internal.{Constants, CustomCodeWrapper}
 import scala.build.options.{BuildOptions, Scope}
 import scala.build.{Artifacts, CrossSources, Inputs, Logger, Os, Sources}
@@ -52,7 +52,7 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     val inputs = options.shared.inputsOrExit(args)
     CurrentParams.workspaceOpt = Some(inputs.workspace)
 
-    val bspPath = saveBspConfiguration(
+    val bspPath = writeBspConfiguration(
       options,
       inputs,
       previousCommandName = None
@@ -67,14 +67,13 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     logger: Logger,
     previousCommandName: Option[String]
   ): Unit =
-    saveBspConfiguration(SetupIdeOptions(shared = options), inputs, previousCommandName) match {
+    writeBspConfiguration(SetupIdeOptions(shared = options), inputs, previousCommandName) match {
       case Left(ex) =>
         logger.debug(s"Ignoring error during setup-ide: ${ex.message}")
-      case Right(bspPath) =>
-        bspPath.foreach(path => logger.debug(s"Wrote $path"))
+      case Right(_) =>
     }
 
-  private def saveBspConfiguration(
+  private def writeBspConfiguration(
     options: SetupIdeOptions,
     inputs: Inputs,
     previousCommandName: Option[String]
@@ -153,11 +152,11 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     val scalaCliOptionsForBspJson = write(options.shared)
 
     if (inputs.workspaceOrigin.contains(WorkspaceOrigin.HomeDir))
-      value(Left(new BuildException(
-        """scala-cli can not determine where write bsp configuration.
-          |Set an explicit bsp destination path via `--bsp-directory`.
+      value(Left(new WorkspaceError(
+        """scala-cli can not determine where to write its BSP configuration.
+          |Set an explicit BSP directory path via `--bsp-directory`.
           |""".stripMargin
-      ) {}))
+      )))
 
     if (previousCommandName.isEmpty || !bspJsonDestination.toIO.exists()) {
       os.write.over(bspJsonDestination, json.getBytes(charset), createFolders = true)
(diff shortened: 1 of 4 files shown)


modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +149,22 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     val json                      = gson.toJson(details)
     val scalaCliOptionsForBspJson = write(options.shared)
 
+    if (bspJsonDestination.toString.contains(".cache"))
```
alexarchambault: We shouldn't detect that `.bsp` would land in the home dir this way… At a minimum, that check should have been something like `bspJsonDestination.segments.contains(".cache")`, but that would only work on Linux, and wouldn't always mean that Scala CLI decided on its own to put things there (users might have asked for it, and we should just let them in this case).

We should probably keep track of how `workspace` is computed in `Inputs` (from source paths / in home dir / forced by users). If this says the workspace was put in home dir by Scala CLI, and that workspace is used by `SharedBspFileOptions.bspDetails`, then we should print an error message.

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +149,22 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+      value(Left(new FilePermissionsError(
```
alexarchambault: This is not about file permissions. It's just that Scala CLI doesn't know where to write `.bsp`, so that users should pass an explicit workspace (via `--workspace`, if it's accepted by `setup-ide`, or via `--bsp-directory`).

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +149,22 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+
     if (previousCommandName.isEmpty || !bspJsonDestination.toIO.exists()) {
       os.write.over(bspJsonDestination, json.getBytes(charset), createFolders = true)
       os.write.over(
         scalaCliBspJsonDestination,
         scalaCliOptionsForBspJson.getBytes(charset),
         createFolders = true
       )
-      logger.debug(s"Wrote $bspJsonDestination")
+      if (isRunSafe) logger.debug(s"Wrote $bspJsonDestination")
```
alexarchambault: We shouldn't need this variable (`runSafe`). If you really want to print a message from `setup-ide`, return `bspJsonDestination` from that method, and use it in the `run` method above.

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -64,17 +67,18 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+    saveBspConfiguration(SetupIdeOptions(shared = options), inputs, previousCommandName) match {
       case Left(ex) =>
         logger.debug(s"Ignoring error during setup-ide: ${ex.message}")
-      case Right(()) =>
+      case Right(bspPath) =>
+        bspPath.foreach(path => logger.debug(s"Wrote $path"))
     }
 
-  private def doRun(
+  private def saveBspConfiguration(
```
alexarchambault: ?
```suggestion
  private def writeBspConfiguration(
```

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +152,23 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     val json                      = gson.toJson(details)
     val scalaCliOptionsForBspJson = write(options.shared)
 
+    if (inputs.workspaceOrigin.contains(WorkspaceOrigin.HomeDir))
+      value(Left(new BuildException(
```
alexarchambault: Let's create a dedicated class under `scala.cli.errors`, rather than an anonymous one here.

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +152,23 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+        """scala-cli can not determine where write bsp configuration.
```
alexarchambault:
```suggestion
        """scala-cli can not determine where to write its BSP configuration.
```

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +152,23 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+          |Set an explicit bsp destination path via `--bsp-directory`.
```
alexarchambault:
```suggestion
          |Set an explicit BSP directory path via `--bsp-directory`.
```

modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -148,14 +152,23 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
...
+      ) {}))
+
     if (previousCommandName.isEmpty || !bspJsonDestination.toIO.exists()) {
       os.write.over(bspJsonDestination, json.getBytes(charset), createFolders = true)
       os.write.over(
         scalaCliBspJsonDestination,
         scalaCliOptionsForBspJson.getBytes(charset),
         createFolders = true
       )
-      logger.debug(s"Wrote $bspJsonDestination")
```
alexarchambault, line 158: We can keep this one, it's only there for debugging (it's fine if we print more-or-less the same thing twice when debug output is enabled…)

alexarchambault (approved):

PR #712, opened 2022-02-28
Disable using scala-cli to build scala-cli on unsupported architecture

lwronski:


project/settings.sc
```diff
@@ -745,10 +745,16 @@ trait ScalaCliCompile extends ScalaModule {
     }
     PathRef(file)
   }
+  // disable using scala-cli to build scala-cli on unsupported architectures
+  lazy val isUnsupportedArch: Boolean = {
+    val unsupportedArch = Seq("arm64", "aarch64")
```
alexarchambault: Let's do the opposite, maybe, that is checking only for supported architectures (should be `x86_64` or `amd64`). We know all the others shouldn't be supported for now.

alexarchambault (approved):

PR #714, opened 2022-02-28
Package - overwrite a file generated by scala-cli with --watch flag

### LINKED ISSUE #710: `package --watch` requires `--force`
**Version(s)**

0.1.2

**Describe the bug**

At least when used with Scala.js. Running `package --watch` without `--force` causes it to fail when it attempt to overwrite the file with the changes.

**To Reproduce**

Happy to follow-up with a reproducer if necessary.

**Expected behaviour**
`--force` mode should be automatically enabled.

lwronski:


alexarchambault (commented):
Rather than setting force to true by default, I think, as much as possible, we should only overwrite a file we previously wrote (in the same run of scala-cli). A way of doing that would consist in keeping track of the last modified time of the file we just wrote, and compare with the previous one right before we overwrite it.

There might still be concurrency issues here (if the file changes right before or right after we read the last modified time), but it would still be better than blindly overwriting things IMO.

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -58,8 +59,16 @@ object Package extends ScalaCommand[PackageOptions] {
...
+            val mtimeDestPath = doPackage(
+              inputs,
+              logger,
+              options.output.filter(_.nonEmpty),
+              options.force,
+              s,
+              expectedModifyEpochSecondOpt
+            )
               .orReport(logger)
+            expectedModifyEpochSecondOpt = mtimeDestPath.flatten
```
alexarchambault: Let's not make `expectedModifyEpochSecondOpt` empty if the last run just failed (because compilation failed, or anything else). Something like this should retain the value of the last successful run:
```suggestion
            for (valueOpt <- mtimeDestPath)
              expectedModifyEpochSecondOpt = valueOpt
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -152,14 +162,24 @@ object Package extends ScalaCommand[PackageOptions] {
       else destPath.toString
 
     def alreadyExistsCheck(): Unit =
-      if (!force && os.exists(destPath)) {
+      if (!force && os.exists(destPath) && expectedModifyEpochSecondOpt.isEmpty) {
         System.err.println(
           s"Error: $printableDest already exists. Pass -f or --force to force erasing it."
         )
         sys.exit(1)
       }
```
alexarchambault: Maybe we can merge both checks, like
```suggestion
      val alreadyExists = !force &&
        os.exists(destPath) &&
        expectedModifyEpochSecondOpt.forall(exp => os.mtime(destPath) != exp)
      if (alreadyExists) {
          val msg =
            if (expectedModifyEpochSecondOpt.isEmpty) s"$printableDest already exists"
            else s"$printableDest was overwritten by another process"
        System.err.println(s"Error: $msg. Pass -f or --force to force erasing it.")
        sys.exit(1)
      }
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -274,6 +294,9 @@ object Package extends ScalaCommand[PackageOptions] {
         else
           s"Wrote $dest"
       }
+
+    val mTimeDestPathOpt = if (!packageOptions.isDockerEnabled) Some(os.mtime(destPath)) else None
```
alexarchambault: Maybe we could check `packageType` instead of `packageOptions.isDockerEnabled` (as the former is actually used above to decide to build a docker image or not), and we could also change `PackageType.runnable` to an `Option[Boolean]`, that would be empty if the packaging type doesn't build a file (so `PackageType.Docker.runnable` would be `None`), and use that here as a check: `if (packageType.runnable.isEmpty) None else Some(…)`

armanbilge:
Thanks for working on this!

> There might still be concurrency issues here (if the file changes right before or right after we read the last modified time), but it would still be better than blindly overwriting things IMO.

I agree that blindly overwriting is not a good default in general. However, specifically the case with the `--watch` flag, blindly overwriting is exactly what I'd expect. If there's an occasional concurrency issue that causes it to crash because it couldn't overwrite, this would be annoying.

alexarchambault (approved):

lwronski:
> Thanks for working on this!
> 
> > There might still be concurrency issues here (if the file changes right before or right after we read the last modified time), but it would still be better than blindly overwriting things IMO.
> 
> I agree that blindly overwriting is not a good default in general. However, specifically the case with the `--watch` flag, blindly overwriting is exactly what I'd expect. If there's an occasional concurrency issue that causes it to crash because it couldn't overwrite, this would be annoying.

I don't think, we should  blindly overwriting. For now, Scala CLI overwrite a file that was previously written in the same run of scala-cli. But, if it would be annoying for you, please feel free to open discussion.  - https://github.com/VirtusLab/scala-cli/discussions

alexarchambault:
Blindly overwriting means that `scala-cli package -w …` might overwrite some prior file that just `scala-cli package …` wouldn't have overwritten, which is a bit surprising IMO.

armanbilge:
Hmm, it sounds like `-w` shouldn't blindly overwrite when you _first_ start it, but after it's started it should overwrite. I was primarily responding to this comment.
> There might still be concurrency issues here (if the file changes right before or right after we read the last modified time), but it would still be better than blindly overwriting things IMO.

If this concurrency issue is not a problem in practice, then it should be fine.

Edit: oh, I think I misunderstood the comment. "if the file changes right before or right after" == if some non-scala-cli process changes the file. Which I don't expect to happen. So I think this is good, sorry! :)

PR #736, opened 2022-03-01
added support for 2.12.nightly and 2.13.nightly and some non stable versions

### LINKED ISSUE #687: Regression: can't use Scala 3.1.2-RC1
this was working before I upgraded to 0.1.2

```
% scala-cli repl -S 3.1.2-RC1
[error]  Cannot find matching Scala version for '3.1.2-RC1'
You can only choose one of the 3.x, 2.13.x, and 2.12.x. versions.
The latest supported stable versions are 3.1.1, 2.13.8, 2.12.15.
In addition, you can request the latest Scala 2 and Scala 3 nightly versions by passing 2.nightly, and 3.nightly arguments respectively.
Specific Scala 2 or Scala 3 nightly versions are also accepted.
```

the same happens if I try to use 2.13.0-RC1

### LINKED ISSUE #702: Add support for `2.12.nightly`
it's awesome that `2.nightly` now gets you a 2.13 nightly, but I would suggest also supporting `2.12.nightly` — we still publish 2.12 nightlies, and probably will continue doing so for some time yet to come

for completeness, it would be good to also add `2.13.nightly` at the same time, to mean the same thing as `2.nightly`

cc @zmerr

ghost:
resolves #702 and resolves #687

### DIFF
--- a/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/build/src/main/scala/scala/build/options/BuildOptions.scala
@@ -49,9 +49,18 @@ final case class BuildOptions(
     Seq(s"Scala ${value(scalaParams).scalaVersion}", platform0)
   }
 
+  lazy val scalaVersionIsExotic: Boolean = scalaParams match {
+    case Left(_) => false
+    case Right(ScalaParameters(scalaVersion, _, _)) =>
+      if (scalaVersion.startsWith("2") && "[a-zA-Z]".r.findAllIn(scalaVersion).nonEmpty) true
+      else false
+  }
+
   def addRunnerDependency: Option[Boolean] =
     internalDependencies.addRunnerDependencyOpt
-      .orElse(if (platform.value == Platform.JVM) None else Some(false))
+      .orElse(if (scalaVersionIsExotic) Some(false)
+      else if (platform.value == Platform.JVM) None
+      else Some(false))
 
   private def scalaLibraryDependencies: Either[BuildException, Seq[AnyDependency]] = either {
     if (platform.value != Platform.Native && scalaOptions.addScalaLibrary.getOrElse(true)) {
@@ -244,8 +253,9 @@ final case class BuildOptions(
     val scalaCliVersion = version
     val launchersTask   = cache.logger.using(task)
 
-    //  If an error occurred while downloading stable versions,
-    //  it uses stable scala versions from Deps.sc
+    /** If an error occurred while downloading stable versions, it uses stable scala versions from
+      * `Deps.sc`
+      */
     val supportedScalaVersions =
       launchersTask.attempt.unsafeRun()(cache.ec) match {
         case Left(_) =>
@@ -292,9 +302,66 @@ final case class BuildOptions(
   def finalRepositories: Seq[String] =
     classPathOptions.extraRepositories ++ internal.localRepository.toSeq
 
-  private lazy val maxSupportedStableScalaVersions = latestSupportedStableScalaVersion()
+  private lazy val maxSupportedStableScalaVersions: Seq[Version] =
+    latestSupportedStableScalaVersion()
 
-  private lazy val latestSupportedStableVersions = maxSupportedStableScalaVersions.map(_.repr)
+  private lazy val latestSupportedStableVersions: Seq[String] =
+    maxSupportedStableScalaVersions.map(_.repr)
+
+  private def getAllMatchingStableVersions(scalaVersionArg: Option[String]): Seq[String] = {
+    val modules = {
+      import coursier.moduleString
+      def scala2 = mod"org.scala-lang:scala-library"
+      // No unstable, that *ought* not to be a problem down-the-line…?
+      def scala3 = mod"org.scala-lang:scala3-library_3"
+      if (scalaVersionArg.contains("2") || scalaVersionArg.exists(_.startsWith("2."))) Seq(scala2)
+      else if (scalaVersionArg.contains("3") || scalaVersionArg.exists(_.startsWith("3.")))
+        Seq(scala3)
+      else Seq(scala2, scala3)
+    }
+    def isStable(v: String): Boolean =
+      !v.endsWith("-NIGHTLY") && !v.contains("-RC")
+    def moduleVersions(mod: Module): Seq[String] = {
+      val res = finalCache.logger.use {
+        try Versions(finalCache)
+            .withModule(mod)
+            .result()
+            .unsafeRun()(finalCache.ec)
+        catch {
+          case NonFatal(e) => throw new Exception(e)
+        }
+      }
+      res.versions.available.filter(isStable)
+    }
+    modules.flatMap(moduleVersions).distinct
+  }
+
+  private def getAllMatchingVersions(scalaVersionArg: Option[String]): Seq[String] = {
+    val modules = {
+      import coursier.moduleString
+      def scala2 = mod"org.scala-lang:scala-library"
+      // No unstable, that *ought* not to be a problem down-the-line…?
+      def scala3 = mod"org.scala-lang:scala3-library_3"
+      if (scalaVersionArg.contains("2") || scalaVersionArg.exists(_.startsWith("2."))) Seq(scala2)
+      else if (scalaVersionArg.contains("3") || scalaVersionArg.exists(_.startsWith("3.")))
+        Seq(scala3)
+      else Seq(scala2, scala3)
+    }
+
+    def moduleVersions(mod: Module): Seq[String] = {
+      val res = finalCache.logger.use {
+        try Versions(finalCache)
+            .withModule(mod)
+            .result()
+            .unsafeRun()(finalCache.ec)
+        catch {
+          case NonFatal(e) => throw new Exception(e)
+        }
+      }
+      res.versions.available
+    }
+    modules.flatMap(moduleVersions).distinct
+  }
 
   /** @param scalaVersionArg
     *   the command line, using directive, or default argument passed as scala version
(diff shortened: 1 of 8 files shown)


zmerr:
It yet requires the repository to be specified for Scala 2 nightlies with `repl` .

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -49,9 +49,18 @@ final case class BuildOptions(
     Seq(s"Scala ${value(scalaParams).scalaVersion}", platform0)
   }
 
+  lazy val scalaVersionIsExotic: Boolean = scalaParams match {
+    case Left(_) => false
+    case Right(ScalaParameters(scalaVersion, _, _)) =>
+      if (scalaVersion.startsWith("2") && "[a-zA-Z]".r.findAllIn(scalaVersion).nonEmpty) true
+      else false
+  }
```
alexarchambault:
```suggestion
  lazy val scalaVersionIsExotic = scalaParams.exists { params =>
    val sv = params.scalaVersion
    sv.startsWith("2") && sv.exists(_.isLetter)
  }
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -49,9 +49,18 @@ final case class BuildOptions(
...
+
   def addRunnerDependency: Option[Boolean] =
     internalDependencies.addRunnerDependencyOpt
-      .orElse(if (platform.value == Platform.JVM) None else Some(false))
+      .orElse(if (scalaVersionIsExotic) Some(false)
+      else if (platform.value == Platform.JVM) None
+      else Some(false))
```
alexarchambault: NIT
```suggestion
      .orElse {
        if (platform.value == Platform.JVM && !scalaVersionIsExotic) None
        else Some(false)
      }
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -244,8 +253,9 @@ final case class BuildOptions(
     val scalaCliVersion = version
     val launchersTask   = cache.logger.using(task)
 
-    //  If an error occurred while downloading stable versions,
-    //  it uses stable scala versions from Deps.sc
+    /** If an error occurred while downloading stable versions, it uses stable scala versions from
```
alexarchambault: Not sure this comment was meant to be a scaladoc.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -292,9 +302,66 @@ final case class BuildOptions(
   def finalRepositories: Seq[String] =
     classPathOptions.extraRepositories ++ internal.localRepository.toSeq
 
-  private lazy val maxSupportedStableScalaVersions = latestSupportedStableScalaVersion()
+  private lazy val maxSupportedStableScalaVersions: Seq[Version] =
+    latestSupportedStableScalaVersion()
+
+  private lazy val latestSupportedStableVersions: Seq[String] =
+    maxSupportedStableScalaVersions.map(_.repr)
```
alexarchambault: Would it be possible _not_ to add types to existing code this way? This makes the diff harder to read, and doesn't really bring much in itself…

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -500,13 +612,20 @@ final case class BuildOptions(
...
+          case Some("3.nightly")    => computeLatestScalaThreeNightlyVersions()
+          case Some("2.nightly")    => computeLatestScalaTwoNightlyVersions()
+          case Some("2.13.nightly") => computeLatestScalaTwoNightlyVersions()
+          case Some("2.12.nightly") => computeLatestScalaTwoTwelveNightlyVersions()
           case Some(versionString) if isScala3Nightly(versionString) =>
             turnScala3NightlyVersionArgIntoVersion(versionString)
           case Some(versionString) if isScala2Nightly(versionString) =>
             turnScala2NightlyVersionArgToVersions(versionString)
-          case _ => turnScalaVersionArgToScalaVersions(
+          case Some(versionString) if "[a-zA-Z]".r.findAllIn(versionString).nonEmpty =>
```
alexarchambault:
```suggestion
          case Some(versionString) if versionString.exists(_.isLetter) =>
```

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -124,6 +124,22 @@ class BuildOptionsTests extends munit.FunSuite {
     )
   }
 
+  test("Scala 3.1.2-RC1 works") {
+
+    val options = BuildOptions(
+      scalaOptions = ScalaOptions(
+        scalaVersion = Some("3.1.2-RC1"),
+        scalaBinaryVersion = None,
+        supportedScalaVersionsUrl = None
```
alexarchambault: No need of the explicit `None`s
```suggestion
        scalaVersion = Some("3.1.2-RC1")
```

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -171,6 +187,36 @@ class BuildOptionsTests extends munit.FunSuite {
     )
   }
 
+  test("-S 2.13.nightly option works") {
+    val options = BuildOptions(
+      scalaOptions = ScalaOptions(
+        scalaVersion = Some("2.13.nightly"),
+        scalaBinaryVersion = None,
+        supportedScalaVersionsUrl = None
```
alexarchambault:
```suggestion
        scalaVersion = Some("2.13.nightly")
```

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
```diff
@@ -171,6 +187,36 @@ class BuildOptionsTests extends munit.FunSuite {
...
+      "-S 2.13.nightly argument does not lead to scala2 nightly build option"
+    )
+  }
+
+  test("-S 2.12.nightly option works") {
+    val options = BuildOptions(
+      scalaOptions = ScalaOptions(
+        scalaVersion = Some("2.12.nightly"),
+        scalaBinaryVersion = None,
+        supportedScalaVersionsUrl = None
```
alexarchambault:
```suggestion
        scalaVersion = Some("2.12.nightly")
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -292,9 +302,66 @@ final case class BuildOptions(
...
+    val modules = {
+      import coursier.moduleString
+      def scala2 = mod"org.scala-lang:scala-library"
+      // No unstable, that *ought* not to be a problem down-the-line…?
+      def scala3 = mod"org.scala-lang:scala3-library_3"
+      if (scalaVersionArg.contains("2") || scalaVersionArg.exists(_.startsWith("2."))) Seq(scala2)
+      else if (scalaVersionArg.contains("3") || scalaVersionArg.exists(_.startsWith("3.")))
+        Seq(scala3)
+      else Seq(scala2, scala3)
+    }
```
alexarchambault: Maybe this should be factored to a private method, there's the exact same code just below.

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -303,87 +370,108 @@ final case class BuildOptions(
...
+          Right(validVersions.max.repr)
+    }
+
+  private def matchNewestNonStableScalaVersion(
+    maybeScalaVersionStringArg: Option[String],
+    versionPool: Seq[String]
+  ): Either[ScalaVersionError, String] =
+    maybeScalaVersionStringArg match {
+      case Some(scalaVersionStringArg) =>
+        val matchingVersions =
```
alexarchambault: I think the `case Some(scalaVersionStringArg) =>` case should look like this (this should compile, and is simpler):
```scala
        val matches = versionPool.contains(scalaVersionStringArg)
        if (matches) {
          val isSupportedVersion0 = isSupportedVersion(scalaVersionStringArg)
          if (isSupportedVersion0)
            Right(scalaVersionStringArg)
          else
            Left(new UnsupportedScalaVersionError(
              scalaVersionStringArg,
              latestSupportedStableVersions
            ))
        }
        else
          Left(new InvalidBinaryScalaVersionError(
            scalaVersionStringArg,
            latestSupportedStableVersions
          ))
```

alexarchambault: Or even:

```scala
        val matches = versionPool.contains(scalaVersionStringArg)
        lazy val isSupportedVersion0 = isSupportedVersion(scalaVersionStringArg)
        if (matches && isSupportedVersion0)
          Right(scalaVersionStringArg)
        else if (matches)
          Left(new UnsupportedScalaVersionError(
            scalaVersionStringArg,
            latestSupportedStableVersions
          ))
        else
          Left(new InvalidBinaryScalaVersionError(
            scalaVersionStringArg,
            latestSupportedStableVersions
          ))
```

website/docs/commands/compile.md
```diff
@@ -68,7 +68,7 @@ scala-cli compile --scala 3 Hello.scala
 
 The nightly builds of Scala compiler are unstable ones which are published on a nightly basis.
 
-For using the latest Scala 2 and Scala 3 nightly builds, you should pass `2.nightly` and `3.nightly`, respectively.
+For using the latest Scala 2 and Scala 3 nightly builds, you should pass `2.nightly` and `3.nightly`, respectively. You can also request the last `2.12.nightly` and `2.13.nightly` versions. Currently, the`2.13.nightly` version is the same as `2.nightly`
```
alexarchambault:
```suggestion
To use the latest Scala 2 and Scala 3 nightly builds, pass `2.nightly` and `3.nightly`, respectively. You can also request the last `2.12.nightly` and `2.13.nightly` versions. `2.13.nightly` is the same as `2.nightly`.
```

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Thank you very much! I would apply it!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Good style!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: It helps in IDEs to leave the comments in this way, so that you can get this information when hovering over the name without looking at the location of the implementation!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Well, frankly, for myself, reading the code without marking with types is very difficult, because I then have to hover over something or look at the source of its implementation to know what the type is!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Thanks for this style tip! I would apply it, then!

modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
ghost: Good idea!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Good idea!

ghost: done!

modules/build/src/main/scala/scala/build/options/BuildOptions.scala
ghost: Yes! Thank you very much!

SethTisue:
Awesome! This will have at least one enthusiastic user. cc @dwijnand

PR #740, opened 2022-03-01
Allow use of main.sc scripts with scala-cli

### LINKED ISSUE #314: `main.sc` script file execution fails at compile time
We don't support this specific filename yet, so we have to print more descriptive message

ghostbuster91:
This closes https://github.com/VirtusLab/scala-cli/issues/314

Joint work with @mtk, @lwronski in the framework of the issue spree.


alexarchambault (approved):

alexarchambault:
Nice!

PR #742, opened 2022-03-01
add output option to compile command

### LINKED ISSUE #696: Scala CLI should be able to use a fixed output for the compilation artifacts

In some cases it is important to access compiled class using some other tool (e.g. to process .class or .tasy files) or place the classes in the right spot so e.g. IDE can pick them up. 

This can be achieved with Scala CLI by taking the first of the path reported when `--classpath` is provided or using `package` command to produce a `jar` and then extract it. Neither of the above is convenient. Otherwise, Scala CLI keeps .classfiles and other artefacts within `.scala-build` directory, however, the path contains a hash of the inputs so may change and is really to figure out which one is being currently used.

I propose to add `--output` or `--output-directory` option. When provided, Scala CLI, after each compilation will update the provided directories with current set of class files and resources. Initially, it does not need to be really efficient since it will be turn off by default.

wiacekm:
Closes #696

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
@@ -52,7 +52,7 @@ object Compile extends ScalaCommand[CompileOptions] {
           sys.exit(1)
       }
       else {
-        if (options.classPath) {
+        if (options.classPath)
           for {
             build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
             s     <- build.successfulOpt
@@ -60,16 +60,14 @@ object Compile extends ScalaCommand[CompileOptions] {
             val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
             println(cp)
           }
-        }
         options.outputPath()
-          .foreach{
+          .foreach {
             output =>
               for {
                 build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
                 s     <- build.successfulOpt
-              } {
-                os.copy.over(s.output, output)
               }
+                os.copy.over(s.output, output)
           }
       }
     }
--- a/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
@@ -51,14 +51,27 @@ object Compile extends ScalaCommand[CompileOptions] {
         if (allowExit)
           sys.exit(1)
       }
-      else if (options.classPath)
-        for {
-          build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
-          s     <- build.successfulOpt
-        } {
-          val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
-          println(cp)
+      else {
+        if (options.classPath) {
+          for {
+            build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
+            s     <- build.successfulOpt
+          } {
+            val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
+            println(cp)
+          }
         }
+        options.outputPath()
+          .foreach{
+            output =>
+              for {
+                build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
+                s     <- build.successfulOpt
+              } {
+                os.copy.over(s.output, output)
+              }
+          }
+      }
     }
 
     val buildOptions     = options.buildOptions
--- a/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Compile.scala
@@ -52,23 +52,18 @@ object Compile extends ScalaCommand[CompileOptions] {
           sys.exit(1)
       }
       else {
-        if (options.classPath)
+        val successulBuildOpt =
           for {
             build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
             s     <- build.successfulOpt
-          } {
+          } yield s
+        if (options.classPath)
+          for (s <- successulBuildOpt) {
             val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
             println(cp)
           }
-        options.outputPath()
-          .foreach {
-            output =>
-              for {
-                build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
-                s     <- build.successfulOpt
-              }
-                os.copy.over(s.output, output)
-          }
+        for (output <- options.outputPath; s <- successulBuildOpt)
+          os.copy.over(s.output, output)
       }
     }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
@@ -3,6 +3,7 @@ package scala.cli.commands
 import caseapp._
 import caseapp.core.help.Help
 
+import java.nio.file.Paths
 import scala.build.options.BuildOptions
 
 // format: off
@@ -19,6 +20,12 @@ final case class CompileOptions(
   @Name("classpath")
   @HelpMessage("Print the resulting class path")
     classPath: Boolean = false,
+
+  @Name("output-directory")
+  @HelpMessage("Copy compilation results to output directory using either relative or absolute path")
+  @ValueDescription("/example/path")
+    output: Option[String] = None,
+
   @HelpMessage("Compile test scope")
     test: Boolean = false
 ) {
@@ -26,6 +33,13 @@ final case class CompileOptions(
 
   def buildOptions: BuildOptions =
     shared.buildOptions(enableJmh = false, jmhVersion = None)
+
+  def outputPath() =
+    output
+      .filter(_.nonEmpty)
+      .map{p =>
+        os.Path(Paths.get(p), os.pwd)
+      }
 
 }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
@@ -3,8 +3,7 @@ package scala.cli.commands
 import caseapp._
 import caseapp.core.help.Help
 
-import java.nio.file.Paths
-
+import scala.build.Os
 import scala.build.options.BuildOptions
 
 // format: off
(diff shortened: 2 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -50,11 +50,21 @@ object Compile extends ScalaCommand[CompileOptions] {
...
+        if (options.classPath) {
+          for (s <- builds.main.successfulOpt) {
+            val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
+            println(cp)
+          }
         }
+        options.output
+        .foreach{
+          output =>
+            for (s <- builds.main.successfulOpt) {
```
lwronski: I think, we should also copy `output` from scope test - `builds.get(Scope.Test)`

Scala CLI have two different output for scopes, for example:
```
Some(/Users/lwronski/projects/scala-cli/.scala-build/project_8fa0e83754_463258e567-911def99dc/classes/test)
Some(/Users/lwronski/projects/scala-cli/.scala-build/project_8fa0e83754_463258e567-911def99dc/classes/main)
```

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
```diff
@@ -29,4 +37,16 @@ final case class CompileOptions(
 object CompileOptions {
   implicit lazy val parser: Parser[CompileOptions] = Parser.derive
   implicit lazy val help: Help[CompileOptions]     = Help.derive
+  implicit val customArgParser: ArgParser[Path] =
+    SimpleArgParser.from("output") { pathString =>
+      try Right{
+        os.Path(Paths.get(pathString).toAbsolutePath())
```
lwronski:
```suggestion
        os.Path(pathString, Os.pwd)
```

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
```diff
@@ -29,4 +37,16 @@ final case class CompileOptions(
 object CompileOptions {
   implicit lazy val parser: Parser[CompileOptions] = Parser.derive
   implicit lazy val help: Help[CompileOptions]     = Help.derive
+  implicit val customArgParser: ArgParser[Path] =
```
alexarchambault: We've not used a custom `ArgParser` for `os.Path` until now, we just accept an `Option[String]`, then convert its value to an `os.Path` with `os.Path(str, Os.pwd)`. That allows to filter out empty strings before the conversion in particular.

Maybe we could use an `ArgParser`, but it would need to be an `ArgParser[Option[os.Path]]`, and have empty strings (or empty after trimming) result in `None`. That can be done later.

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
```diff
@@ -2,8 +2,11 @@ package scala.cli.commands
 
 import caseapp._
 import caseapp.core.help.Help
+import caseapp.core.argparser.{ArgParser, SimpleArgParser}
 
 import scala.build.options.BuildOptions
+import os.Path
```
alexarchambault: Let's not import that. We just use `os.Path` when we need a path.

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
alexarchambault: @lwronski We don't do that for `--classpath` though (see the `if (options.classPath)` just above). I guess it's fine to only take main into account for now… (and we can to fix both later).

wiacekm: Done for both main and test scope the same way it is done for classpath

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
wiacekm: `os.Path` import removed

modules/cli/src/main/scala/scala/cli/commands/CompileOptions.scala
```diff
@@ -27,6 +35,13 @@ final case class CompileOptions(
   def buildOptions: BuildOptions =
     shared.buildOptions(enableJmh = false, jmhVersion = None)
 
+  def outputPath() =
+    output
+      .filter(_.nonEmpty)
+      .map { p =>
+        os.Path(Paths.get(p), os.pwd)
+      }
```
alexarchambault: This should do:
```suggestion
  def outputPath = output.filter(_.nonEmpty).map(p => os.Path(p, Os.pwd))
```
(`Os` being `scala.build.Os`)

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -57,40 +87,43 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
-            |      assert(Main.message.startsWith("Hello"))
-            |    }
-            |  }
-            |}
-            |""".stripMargin
+        mainInput,
+        testInput
+      )
+    )
+    val tempOutput = os.temp.dir()
```
alexarchambault: Let's avoid such temporary directories in tests. `root` can be assumed to be a temporary directory (where we write the inputs). A sub-directory of it would be fine for `--output`.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -57,40 +87,43 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
+    inputs.fromRoot { root =>
+      val outputOptions = Seq("--output", tempOutput.toString)
+      val cmd           = os.proc(TestUtil.cli, "compile", outputOptions, extraOptions, ".")
+      cmd.call(cwd = root, check = false, stderr = os.Pipe, mergeErrIntoOut = true)
+    }
+    val compileOutputCoppied = os.list(tempOutput).forall(_.baseName.startsWith("Main"))
```
alexarchambault: These two lines should be moved  inside the `inputs.fromRoot { … }`, so that they can use the output dir under `root`.

alexarchambault: We could also check for the actual file names, like
`os.list(tempOutput).map(_.relativeTo(tempOutput)) == Seq(os.rel / "Main.class", …)`

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -57,40 +87,43 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
+  }
+
+  test("test scope with copy output") {
+    val inputs = TestInputs(
+      Seq(
+        mainInput,
+        testInput
       )
     )
+    val tempOutput    = os.temp.dir()
```
alexarchambault: Same as above (no temp dir).

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -51,14 +51,25 @@ object Compile extends ScalaCommand[CompileOptions] {
...
-        }
+      else {
+        if (options.classPath)
+          for {
+            build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
+            s     <- build.successfulOpt
+          } {
+            val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
+            println(cp)
+          }
```
alexarchambault: Let's factor things a bit
```suggestion
        val successulBuildOpt =
          for {
            build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
            s     <- build.successfulOpt
          } yield s
        if (options.classPath)
          for (s <- successulBuildOpt) {
            val cp = s.fullClassPath.map(_.toAbsolutePath.toString).mkString(File.pathSeparator)
            println(cp)
          }
```

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -51,14 +51,25 @@ object Compile extends ScalaCommand[CompileOptions] {
...
+        options.outputPath()
+          .foreach {
+            output =>
+              for {
+                build <- builds.get(Scope.Test).orElse(builds.get(Scope.Main))
+                s     <- build.successfulOpt
+              }
+                os.copy.over(s.output, output)
```
alexarchambault: NIT
```suggestion
        for (output <- options.outputPath; s <- successulBuildOpt)
          os.copy.over(s.output, output)
```

alexarchambault (commented):
Thanks @wiacekm, I added a few more suggestions.

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
wiacekm: done as suggested

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
wiacekm: done as suggested

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
wiacekm: done

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
wiacekm: let me know if it is ok now.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -57,40 +88,47 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
-            |""".stripMargin
+        mainInput,
+        testInput
+      )
+    )
+    inputs.fromRoot { root =>
+      val tempOutput    = root / "output"
+      val outputOptions = Seq("--output", tempOutput.toString)
+      val cmd           = os.proc(TestUtil.cli, "compile", outputOptions, extraOptions, ".")
+      cmd.call(cwd = root, check = false, stderr = os.Pipe, mergeErrIntoOut = true)
```
alexarchambault: No need to pass `check = false` and all, it seems:
```suggestion
      os.proc(TestUtil.cli, "compile", outputOptions, extraOptions, ".").call(cwd = root)
```

PR #747, opened 2022-03-02
Run process in background

### LINKED ISSUE #711: Can `run --watch` send `ctrl+c` on source changes?
If I'm working on a server, then `main` starts the server and sleeps indefinitely e.g. until `ctrl+c` or `enter`. If I make changes I'd like the server to restart. `run --watch` would almost work perfectly except it only works if the process is already terminated i.e. it will not kill/restart a currently running process.

If `run --watch` detects changes, it kills the currently running process and restarts it.

For now, I manually terminate my server with `enter` and then scala-cli starts it again with the changes.

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -72,10 +76,12 @@ object Run extends ScalaCommand[RunOptions] {
         postAction = () => WatchUtil.printWatchMessage(),
         buildTests = false
       ) { res =>
+        processOpt.map(_.destroyForcibly())
```
alexarchambault: Don't know if we want to destroy the app forcibly here… (Not sure, but I guess this send a SIGKILL to the app.) @armanbilge suggested a Ctrl+C, so a SIGINT, which would be a good default IMO.

@lwronski To send a SIGINT to a process, [this method](https://github.com/VirtusLab/scala-cli/blob/1471648fafe6365d0f138a25019deefa67d8932f/modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala#L116-L120) can be adapted (we need to pass a `-2` to `kill` for Linux / macOS, don't know for Windows). On Linux / macOS, we can even define a GraalVM substitution like done [here](https://github.com/scala-cli/libdaemon-jvm/blob/bd3d8c7d89d4b559505f3b77e579e3135e758cb9/library/src/libdaemonjvm/internal/IsRunningSubst.java), so that from native launchers, we call the `kill` sys call (see `man 2 kill`), rather than calling the external `kill` command (which is `man 1 kill`), which should be faster and more reliable. (You can get the PID of a `java.lang.Process` by calling `pid` on it.)

alexarchambault (commented):
Don't know if we want to enable that by default, it might be a bit surprising… Better only enable it when a new option, such as `--revolver` (as a reference to [sbt-revolver](https://github.com/spray/sbt-revolver)) is passed (and `--revolver` would automatically enable watch mode too).

modules/cli/src/main/scala/scala/cli/commands/SharedWatchOptions.scala
```diff
@@ -7,10 +7,13 @@ final case class SharedWatchOptions(
...
   @Name("w")
-    watch: Boolean = false
+    watch: Boolean = false,
+  @HelpMessage("Run your application in background and automatically restart if sources have been changed") 
+    revolver: Boolean = false
+) { // format: on
 
-)
-// format: on
+  lazy val isWatchMode = watch || revolver
```
alexarchambault: NIT Rename that to just `watchMode`?

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -43,4 +45,15 @@ object ProcUtil {
...
 
+  def interruptProcess(process: Process): Unit = {
+    def interrupt(pid: Long) =
+      if (Properties.isWin)
+        os.proc("taskkill", "/PID", pid).call()
+      else
+        os.proc("kill", "-2", pid).call()
+
+    val pid = process.pid()
+    interrupt(pid)
```
alexarchambault: NIT No need of the local method:
```suggestion
    val pid = process.pid()
    if (Properties.isWin)
      os.proc("taskkill", "/PID", pid).call()
    else
      os.proc("kill", "-2", pid).call()
```

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -196,17 +208,20 @@ object Run extends ScalaCommand[RunOptions] {
...
+      val retCode = p1.exitValue()
+      if (retCode != 0)
+        if (exitOnError)
+          sys.exit(retCode)
+        else {
+          val red      = Console.RED
+          val lightRed = "\u001b[91m"
+          val reset    = Console.RESET
+          System.err.println(s"${red}Program exited with return code $lightRed$retCode$red.$reset")
+        }
```
alexarchambault: I'm not sure we want to do that when we kill the process ourselves (this is going to print a weird `Program exited with return code 130`). We could keep the future that `thenApply` returns, and maybe call `cancel(false)` on it right before we try to kill it.

For that to be more convenient, maybe the call to `thenApply` should be done elsewhere, like when we call `maybeRun` (and only the call in watch mode should need fancy things).

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -73,10 +78,16 @@ object Run extends ScalaCommand[RunOptions] {
         partial = None,
         postAction = () => WatchUtil.printWatchMessage()
       ) { res =>
+        for (process <- processOpt)
+          ProcUtil.interruptProcess(process)
```
alexarchambault: Maybe we should check if the process if running right before calling that.

When verbosity is increased, we should also log that we are sending a SIGINT to the process (we could even print the command we run in debug mode, just in case).

If the command to kill the process fails, we should maybe ignore the failure if the process isn't running anymore (might mean it exited between the first check and the call of the command to kill it).

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -75,10 +99,18 @@ object Run extends ScalaCommand[RunOptions] {
...
+        }
         res.orReport(logger).map(_.main).foreach {
           case s: Build.Successful =>
-            maybeRun(s, allowTerminate = false)
+            val maybeProcess = maybeRun(s, allowTerminate = false)
               .orReport(logger)
+            if (options.watch.revolver)
+              processOpt = maybeProcess
+            else
+              maybeProcess.map(_._1.waitFor())
```
alexarchambault: NIT Maybe we should also "join" the future here, so that it doesn't run in parallel to the next watch loop iteration. Something along those lines should do
```suggestion
              for ((proc, onExit) <- maybeProcess) {
                proc.waitFor()
                try onExit.join()
                catch {
                  case _: CancellationException | CompletionException => // ignored
                }
              }
```

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -100,8 +132,9 @@ object Run extends ScalaCommand[RunOptions] {
           .orExit(logger)
       builds.main match {
         case s: Build.Successful =>
-          maybeRun(s, allowTerminate = true)
+          val (process, _) = maybeRun(s, allowTerminate = true)
             .orExit(logger)
+          process.waitFor()
```
alexarchambault: Just like above, maybe we should wait for the on-exit future to finish (it's not clear whether it runs from a daemon thread or not… if that's a daemon thread, we have to wait for it, or it might not run if the main method exits before it runs).

modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -145,14 +145,17 @@ object Test extends ScalaCommand[TestOptions] {
...
-            Runner.testJs(
+            val res = Runner.testJs(
               build.fullClassPath,
               js.toIO,
               requireTests,
               args,
               testFrameworkOpt,
               logger
             )
+            // clean after linking js TODO it should be move to Run.withLinkedJs
```
alexarchambault: We shouldn't call `withLinkedJs` here. `withLinkedJs`'s only purpose is cleaning-up that file. We can just call `Package.linkJs` ourselves here.

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -43,4 +47,24 @@ object ProcUtil {
     new String(data, StandardCharsets.UTF_8)
   }
 
+  def interruptProcess(process: Process, logger: Logger): Unit = {
+    val pid = process.pid()
+    try
+      if (process.isAlive) {
+        logger.debug(s"Sending a SIGINT signal to process PID:$pid")
```
alexarchambault: Shouldn't this be a standard message?
```suggestion
        logger.debug("Interrupting running process")
```

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -75,10 +99,18 @@ object Run extends ScalaCommand[RunOptions] {
         partial = None,
         postAction = () => WatchUtil.printWatchMessage()
       ) { res =>
+        for ((process, onExitProcess) <- processOpt) {
+          onExitProcess.cancel(true)
+          ProcUtil.interruptProcess(process, logger)
```
alexarchambault, line 120: It might not be enough to send a SIGINT… If the process doesn't exit, we shouldn't proceed with running a new one straightaway. Maybe just before the call to `maybeRun` below, we could check again if the process is still alive, and send it a SIGKILL with `Process.destroyForcibly` (and only then, warn users if ever it's still alive and start a new one anyway).

modules/cli/src/main/scala/scala/cli/commands/Test.scala
lwronski: If I use `Package.linkJs`, it should be the same effect because `Package.linkJs` required `dest: os.Path` argument. 

Do you mean to generate the output directory here, call `Package.linkJs` and then delete it?

modules/cli/src/main/scala/scala/cli/commands/Test.scala
```diff
@@ -144,15 +144,19 @@ object Test extends ScalaCommand[TestOptions] {
...
-            Runner.testJs(
-              build.fullClassPath,
-              js.toIO,
-              requireTests,
-              args,
-              testFrameworkOpt,
-              logger
-            )
+          val jsDest = os.temp(prefix = "main", suffix = ".js")
+          Package.linkJs(build, jsDest, None, addTestInitializer = true, linkerConfig, logger).map {
```
alexarchambault: Why not keep `withLinkedJs` here? We don't need the linked file outside the wrapped call.

modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala
```diff
@@ -43,4 +48,39 @@ object ProcUtil {
     new String(data, StandardCharsets.UTF_8)
   }
 
+  def forceKillProcess(process: Process, logger: Logger): Unit = {
+    if (process.isAlive) {
+      process.destroyForcibly()
+      logger.debug("Force killing running process")
```
alexarchambault: NIT Maybe specify the PID in the debug message:
```suggestion
      logger.debug(s"Killing user process ${process.pid()}")
```

modules/cli/src/main/scala/scala/cli/commands/Test.scala
lwronski: Good point, yes, we can keep `withLinkedJs` here. Thanks

alexarchambault (approved):

kubukoz:
I feel like `--revolver` is too obscure of a name, only people already familiar with `sbt-revolver` will immediately know what it does. Why not name it something closer to what it does like `--watch` or `--restart` instead?

PR #748, opened 2022-03-02
Add support for signing in publish command

alexarchambault:


alexarchambault:
I'd like to add some tests for `--compute-version` prior to merging this.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -221,44 +230,70 @@ object Build {
         }
 
       val mainBuild = value(doBuildScope(mainOptions, mainSources, Scope.Main))
+      val mainDocBuildOpt = docCompilerOpt match {
```
romanowski, line 233: Nit: can't it be replaced by `map`?

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -221,44 +230,70 @@ object Build {
...
-                  Scope.Test,
-                  "Parent build failed or cancelled"
-                ))
-            }
+          val actualCompilerOpt =
+            if (doc) docCompilerOpt
+            else Some(compiler)
+          actualCompilerOpt match {
+            case None => None
+            case Some(actualCompiler) =>
```
romanowski, line 251: Same here, can be a `map`

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -221,44 +230,70 @@ object Build {
...
+                }
+              }
+              Some(testBuild)
           }
-          Some(testBuild)
         }
         else None
+      }
 
+      val testBuildOpt0 = value(testBuildOpt())
```
romanowski, line 282: Can we rename `testBuildOpt0` to something more meaningful?

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -221,44 +230,70 @@ object Build {
...
       doPostProcess(mainBuild, inputs0, Scope.Main)
-      for (testBuild <- testBuildOpt)
+      for (testBuild <- testBuildOpt0)
         doPostProcess(testBuild, inputs0, Scope.Test)
 
-      (mainBuild, testBuildOpt)
+      val docTestBuildOpt0 = value(testBuildOpt(doc = true))
```
romanowski, line 287: Same as above

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -267,21 +302,31 @@ object Build {
...
+              extraBuilds.flatMap(_.testDocOpt)
+            )
           }
           else
-            (Nil, Nil)
+            (Nil, Nil, Nil, Nil)
 
-        (mainBuild, extraMainBuilds, testBuild, extraTestBuilds)
+        Builds(
+          Seq(nonCrossBuilds.main) ++ nonCrossBuilds.testOpt.toSeq,
```
romanowski, line 316: Can we have both crossed and not-crossed build? If not this code can be really simlified

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -416,16 +462,39 @@ object Build {
...
-        buildClient = buildClient,
-        compiler = compiler,
-        crossBuilds = crossBuilds,
-        buildTests = buildTests,
-        partial = partial
-      )
+      docCompilerMakerOpt match {
+        case None =>
+          build(
+            inputs = inputs,
```
romanowski, line 468: I wonder if turning all this parameter into a cases class wouldn't make this code mush shorter and easier to read

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -416,16 +462,39 @@ object Build {
...
+            compiler = compiler,
+            docCompilerOpt = None,
+            crossBuilds = crossBuilds,
+            buildTests = buildTests,
+            partial = partial
+          )
+        case Some(docCompilerMaker) =>
+          docCompilerMaker.withCompiler(
+            inputs.workspace / Constants.workspaceDirName,
+            classesDir0, // ???
```
romanowski, line 481: ?

modules/build/src/main/scala/scala/build/compiler/SimpleScalaCompiler.scala
```diff
@@ -31,56 +34,60 @@ final case class SimpleScalaCompiler(
...
+          project.scalaCompiler.scalacOptions ++
+            Seq(
+              "-d",
+              outputDir.toString,
+              "-cp",
+              project.classPath.map(_.toString).mkString(File.pathSeparator)
+            ) ++
+            project.sources.map(_.toString)
+
+        val javaCommand = project.javaHomeOpt match {
```
romanowski, line 66: We have similar implementation in `OsLibc- probably we can reuse it

modules/cli/src/main/scala/scala/cli/commands/Publish.scala
```diff
@@ -270,43 +315,129 @@ object Publish extends ScalaCommand[PublishOptions] {
...
   private def doPublish(
     builds: Seq[Build.Successful],
+    docBuilds: Seq[Build.Successful],
     workingDir: os.Path,
     logger: Logger
   ): Either[BuildException, Unit] = either {
 
+    assert(docBuilds.isEmpty || docBuilds.length == builds.length)
+
+    val it = builds.iterator.zip {
```
romanowski, line 337: what is `it` ? Maybe `buildsWithDocs`?

modules/cli/src/main/scala/scala/cli/commands/Publish.scala
```diff
@@ -270,43 +315,129 @@ object Publish extends ScalaCommand[PublishOptions] {
...
+      new InteractiveSignerLogger(new OutputStreamWriter(System.err), verbosity = 1)
+    val signRes = signer.signatures(
+      fileSet0,
+      now,
+      ChecksumType.all.map(_.extension).toSet,
+      Set("maven-metadata.xml"),
+      signerLogger
+    )
+
+    val fileSet1 = value {
```
romanowski, line 415: can we call it more meaningfull then `fileSet1`?

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCreate.scala
```diff
@@ -0,0 +1,16 @@
+package scala.cli.commands.pgp
+
+import caseapp.core.RemainingArgs
+
+import scala.cli.commands.ScalaCommand
+import scala.cli.signing.commands.{PgpCreate => OriginalPgpCreate, PgpCreateOptions}
+
+object PgpCreate extends ScalaCommand[PgpCreateOptions] {
```
romanowski, line 8: Does `ScalaCommand` refers to the fact that it is written in Scala?

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpExternalCommand.scala
```diff
@@ -0,0 +1,69 @@
...
+
+    val platformSuffix = FetchExternalBinary.platformSuffix()
+    val version = versionOpt
+      .getOrElse(Constants.scalaCliSigningVersion)
+    val (tag, changing) =
+      if (version == "latest") ("launchers", true)
+      else ("v" + version, false)
+    val ext = if (Properties.isWin) ".zip" else ".gz"
+    val url =
+      s"https://github.com/scala-cli/scala-cli-signing/releases/download/$tag/scala-cli-signing-$platformSuffix$ext"
```
romanowski, line 65: It seems that `scala-cli-signing` containts both bouncy castle and pgp things, should it be move to some util like `ExternalCommand`?

romanowski (commented):
Some general questions:

1. can't we include in our repo it should simply development in the future and it seems there are no limits per files in release or release count ([see here](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases))

2. I am not sure if the model of keeping docs and sources per scope / scala version based on order (zipping lists etc.) is maintable in long run. I think we should at least build some abstractions on top of it (even if sync it based on order under the hood).

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: I don't think, the call to `value` (either CPS thing) in one of the branches prevents that.

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: Same as above, there's a `value` call in one of the branches.

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: It's already a meaningful name, isn't it? (it's not named… `b` 😅) It's optionally the test build.

It should have been called just `testBuildOpt`, but that name is already used by the method we call on the same line, so we add a `0` prefix (which we use as a substitute to `'` - prime, that we can't use in scala variable names).

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: Admittedly, this one could be called `docTestBuildOpt` (it's optionally the scaladoc build of the test scope). We don't need the `0` suffix here.

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: I'm not sure I understand your point. `Builds.{builds, docBuilds}` are the "main" builds, and `Builds.{crossBuilds, docCrossBuilds}` are the crossed ones. If cross builds are enabled, we get both the main and the crossed builds.

modules/build/src/main/scala/scala/build/Build.scala
alexarchambault: I don't know, it might. I think I didn't introduce one yet because of the 3 overloads, with subtly different input params.

modules/build/src/main/scala/scala/build/compiler/SimpleScalaCompiler.scala
alexarchambault: You mean, for this computation?
```scala
val ext  = if (Properties.isWin) ".exe" else ""
val path = javaHome / "bin" / s"java$ext"
```

It's only two lines…

modules/cli/src/main/scala/scala/cli/commands/Publish.scala
alexarchambault: `it` stands for a (short lived) iterator. `buildsWithDocsIt` could have worked too.

modules/cli/src/main/scala/scala/cli/commands/Publish.scala
alexarchambault: `fileSet0`, `fileSet1`, `fileSet2`, and `finalFileSet` each add some more data, and are there to avoid using mutable variables. But maybe we should just use a mutable `fileSet` variable here…

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCreate.scala
alexarchambault: It stands for Scala CLI command. It adds settings to `caseapp.core.app.Command` shared by all Scala CLI commands.

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpExternalCommand.scala
alexarchambault: I'm not sure I see what you mean. scala-cli-signing only deals with bouncycastle. These `pgp {create,sign,verify}` commands are low level commands somehow, only relying on bouncycastle, and that can be helpful for testing.

gpg signing is handled by `GpgSigner`, which lives [here](https://github.com/coursier/publish/blob/97a879504db4f13e94f18b1f2bfc31ebdd890512/publish/src/coursier/publish/signing/GpgSigner.scala).

alexarchambault:
> can't we include in our repo it should simply development in the future and it seems there are no limits per files in release or release count

I'd expect scala-cli-signing to move at a much lower pace than Scala CLI (it's only low level stuff). I think keeping it here would unnecessarily lengthen our CI build times.

alexarchambault:
> I am not sure if the model of keeping docs and sources per scope / scala version based on order (zipping lists etc.) is maintable in long run. I think we should at least build some abstractions on top of it (even if sync it based on order under the hood).

I kind of agree. That should require changes in `BuildOptions` too, as cross options [are returned as a sequence for now](https://github.com/VirtusLab/scala-cli/blob/d469bc1aacfec4dff00925737923c72e49611ed1/modules/options/src/main/scala/scala/build/options/BuildOptions.scala#L699).

PR #749, opened 2022-03-03
Add missing Scala Native using directives

jchyb:
While working on a Scala Native related cookbook I noticed that not all SN CLI options had using directives counterparts. This PR add all 3 missing: native-mode (important in my opinion), clang and clang++ paths (not that important but added for the sake of consistency). 

Since the tests for SN using directives were starting to take a fourth of all BuildTests I moved them to a separate suite (but I can revert if necessary).

PR #750, opened 2022-03-03
Add a cookbook with Scala Native scripting

jchyb:
Inspired by @alexarchambault's Scala Love talk, this PR is introduces a cookbook called "Scala Scripting with instant startup times". It's meant to be an introduction to Scala Native in scala-cli, showing some good practices and some caveats as well.

Requires #749 (since native-mode using directive is used at one point).


alexarchambault (approved):
Not sure what @romanowski thinks, but AFAIC, LGTM!

romanowski (approved):

PR #760, opened 2022-03-07
Pass programArgs to scala repl

### LINKED ISSUE #738: REPL arguments not being recognized
**Version(s)**

0.1.2
**Describe the bug**

I'd like to pass flags into the REPL, however this seems to have no effect.

**To Reproduce**

For example,  try `scala-cli repl --scala 3.nightly -Xshow-phases`. This starts up the REPL when it should print out the phases. I thought maybe I could do `scala-cli repl --scala 3.nightly -- -Xshow-phases`, but this also seems to behave the same.

**Expected behaviour**
I'd expect this to print out the phases.

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -183,14 +186,15 @@ object Repl extends ScalaCommand[ReplOptions] {
...
-          options.notForBloopOptions.replOptions.ammoniteArgs,
+        replArtifacts.replMainClass, {
+          if (Properties.isWin)
+            options.notForBloopOptions.replOptions.ammoniteArgs.map { a =>
+              if (a.contains(" ")) "\"" + a.replace("\"", "\\\"") + "\""
+              else a
+            }
+          else
+            options.notForBloopOptions.replOptions.ammoniteArgs
+        } ++ programArgs,
```
alexarchambault: We probably need to do the `if (a.contains(" ")) "\"" + a.replace("\"", "\\\"") + "\""` stuff to what's in `programArgs` too. Better factor `options.notForBloopOptions.replOptions.ammoniteArgs ++ programArgs` a bit before, then do the `if (Properties.isWin) …` stuff on the result.

lwronski: It's make sense, good point

alexarchambault (approved):

PR #770, opened 2022-03-09
Add option to debug BSP connection.

romanowski:


modules/cli/src/main/scala/scala/cli/commands/SharedJvmOptions.scala
```diff
@@ -47,7 +47,13 @@ final case class SharedJvmOptions(
   @HelpMessage("Javac options")
   @Name("javacOpt")
   @Hidden
-    javacOption: List[String] = Nil
+    javacOption: List[String] = Nil,
+
+  @Group("Java")
+  @HelpMessage("Port for BSP debugging")
+  @Name("bspDebugPort")
```
alexarchambault: Seems this makes the duplicated option check fail (this adds the same name as the variable).
```suggestion
```

alexarchambault (commented):
Apart from the comment, and the reference doc that needs to be updated (addressing both should make the CI pass), LGTM!

modules/cli/src/main/scala/scala/cli/commands/SharedJvmOptions.scala
romanowski: Good point, I was copy-pasting :)

alexarchambault (approved):

alexarchambault:
@romanowski Seems wrong stuff (bad rebase?) were pushed.

Gedochao (approved):

PR #771, opened 2022-03-10
Add server less mode (enable with --server=false)

alexarchambault:


alexarchambault:
These are bits of https://github.com/VirtusLab/scala-cli/pull/748. We'll hopefully have less memory issues on the CI here.

SethTisue:
awesome. 👏

PR #773, opened 2022-03-10
Add native image packaging (scala-cli package --native-image)

alexarchambault:


romanowski (changes requested):
We miss user facing documentation, mainly within a package command but I think we should also a have a cookbook

PR #777, opened 2022-03-14
Use jsoniter-scala rather than upickle

alexarchambault:
jsoniter-scala is slimmer than upickle, which should hopefully help during native image generation (less dependencies -> lower memory requirements?), and could also help generate smaller executables.


romanowski:
It seems that ubuntu worker is faster (14:35 vs 11:43 to generate native image) but windows was much slower (15:26 vs 29:22) but this may be a problem with given runner.

romanowski (approved):

alexarchambault:
> It seems that ubuntu worker is faster (14:35 vs 11:43 to generate native image) but windows was much slower (15:26 vs 29:22) but this may be a problem with given runner.

There's also quite a bit of variance on those times though, don't know if they mean much.

PR #780, opened 2022-03-15
Adjust BSP project base directory to scala-cli content root to fix IntelliJ import

Gedochao:
This fixes the BSP project import for scenarios where there are Scala sources directly within the `scala-cli` content root directory (no subdirectories in-between).
The root cause of the issue comes indirectly from this IntelliJ limitation: [IDEABKL-6745](https://youtrack.jetbrains.com/issue/IDEABKL-6745) (and all of its spin-off bugs created since).

Long story short, when running `scala-cli setup-ide .`, we create a `BSP` project to be imported.
The base directory for the `BSP` project is `./.scala-build`.
When that is imported into IntelliJ, a synthetic `folder-name-root` module is created by the `IntelliJ Scala Plugin`, because there is no importable project with the actual content root (as in, `.`) set as its base directory [(relevant code in `intellij-scala`)](https://github.com/JetBrains/intellij-scala/blob/f4e6caf1143f4a6ba3fa629d45ade773510f50b0/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala#L536)

image

As a result, we have 2 modules: a synthetic `folder-name-root` coming from `Intellij Scala Plugin` and our `BSP` module, both of which at least partially sharing their sources root (if any sources are directly in the root folder)
Now, because IntelliJ doesn't support multiple modules to have the same root (as described in the JetBrains ticket linked above), one of them is being filtered out, effectively breaking it.

image

What seems to be sufficient to fix it, is to just translate the `BSP` project's base directory to its parent when communicating with IntelliJ.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
@@ -3,10 +3,11 @@ package scala.build.bsp
 import ch.epfl.scala.bsp4j.{BuildClient, LogMessageParams, MessageType}
 import ch.epfl.scala.{bsp4j => b}
 
-import java.io.{PrintWriter, StringWriter}
+import java.io.{File, PrintWriter, StringWriter}
+import java.net.URI
 import java.util.concurrent.{CompletableFuture, TimeUnit}
+import java.util.concurrent.atomic.AtomicBoolean
 import java.{util => ju}
-
 import scala.build.Logger
 import scala.build.bloop.{ScalaDebugServer, ScalaDebugServerForwardStubs}
 import scala.build.internal.Constants
@@ -25,6 +26,7 @@ class BspServer(
     with ScalaBuildServerForwardStubs with JavaBuildServerForwardStubs with HasGeneratedSources {
 
   private var client: Option[BuildClient] = None
+  private val isIntelliJ: AtomicBoolean   = new AtomicBoolean(false)
 
   override def onConnectWithClient(client: BuildClient): Unit = this.client = Some(client)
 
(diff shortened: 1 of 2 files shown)


modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -145,6 +147,10 @@ class BspServer(
       scala.build.blooprifle.internal.Constants.bspVersion,
       capabilities
     )
+    val intellijBSPName        = "IntelliJ-BSP"
+    val buildComesFromIntelliJ = params.getDisplayName == "IntelliJ-BSP"
```
romanowski: IMHO we should check if `Intellij` is present in the name (case insensitive) - it should be a bit more roboust

website/docs/guides/ide.md
```diff
@@ -53,6 +53,5 @@ Once Metals picks up the project structure that’s created by Scala CLI, basic
 
 Here are a few notes related to IntelliJ support:
 
-- The most significant problem with IntelliJ is that it requires sources to be placed within a directory. Therefore, if you use IntelliJ, we strongly suggest that you place your sources in a directory (like `src`).
```
romanowski, line 56: Imho we should warn that Intellij will not precisely recognize what is source and what is not. E.g. in case of:

```
A.scala
B.scala
```

that both contains `Main` object running `scala-cli A.scala` and then open the project will (as far as I understand) add both A.scala and B.scala to Intellij project and highlight that Main is duplicated even though project compiles fine in scala-cli.

romanowski (changes requested):
Looks good however we need to improve the documentation. 
Also, I think we should experiment with a warning in Intellij that the importanted sources may not match what Scala CLI is running (but this can be done as a follow up PR).

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
Gedochao: fixed

website/docs/guides/ide.md
Gedochao: image
Ah yes, you're right. 
No matter what we pass to `scala-cli`, IntelliJ always operates on whole directories they are in.
I don't think we'll ever be able to fix that because of how IntelliJ itself works.
Alright, lemme explain it in the docs.

Gedochao: fixed

romanowski (approved):

PR #784, opened 2022-03-16
Emit source maps for package command

### LINKED ISSUE #779: jsSourceMaps are not emitted.
**Version(s)**

0.1.2

**Describe the bug**

When told to emit js source maps, scala cli does not

**To Reproduce**

```scala
//> using jsEmitSourceMaps "true"

import org.scalajs.dom
import dom.html


@main def main() = 
  val container = dom.document.querySelector("#container")
  val content = dom.document.createElement("div")
  content.classList.add("content")
  content.textContent = "This is the glorious text-content!"

  container.appendChild(content)
```

`scala-cli package scalajs-file.scala --js --js-emit-source-maps --dependency org.scala-js::scalajs-dom::2.1.0`

No js source maps emitted.

**Expected behaviour**
Source maps should be produced alongside the js file.

lwronski:


romanowski (changes requested):
I think we should also add information to `--js-emit-source-map` where  source map would be created and how to change it. I think it should also have an entry on JS guide.

Beside that looks ok.

modules/cli/src/main/scala/scala/cli/commands/ScalaJsOptions.scala
```diff
@@ -60,6 +63,7 @@ final case class ScalaJsOptions(
       moduleKindStr = jsModuleKind,
       checkIr = jsCheckIr,
       emitSourceMaps = jsEmitSourceMaps,
+      sourceMapsDest = jsSourceMapsPath.map(os.Path(_, Os.pwd)),
```
alexarchambault: NIT
```suggestion
      sourceMapsDest = jsSourceMapsPath
        .filter(_.trim.nonEmpty)
        .map(os.Path(_, Os.pwd)),
```

alexarchambault (commented):
Don't know for @romanowski's points, but apart from the minor NIT, it's fine by me.

alexarchambault (approved):

PR #788, opened 2022-03-17
3.x.nightly support

ghost:


alexarchambault (approved):

lwronski (changes requested):
I think there should also be updated entry on [Scala nightlies](https://scala-cli.virtuslab.org/docs/commands/compile/#scala-nightlies) section.

Beside that looks ok.

lwronski (approved):

PR #796, opened 2022-03-21
JsDom - run tests with JSDOM JSEnv

### LINKED ISSUE #709: `--js-dom` / `using jsDom` doesn't seem to work?
**Version(s)**

0.1.2

**Describe the bug**

I added this to a project but the tests fail with this error, suggesting it's not actually working.
```
JavaScriptException: ReferenceError: document is not defined
```

I looked at the scala-cli sources and couldn't tell if this option has actually been implemented?

**To Reproduce**

If this _has_ been implemented then I'll follow-up with a reproducer.

**Expected behaviour**
Tests should run with the JSDOM JSEnv :) or, this option should not exist.

lwronski:


project/deps.sc
```diff
@@ -50,11 +50,12 @@ object InternalDeps {
...
   object Versions {
     // jni-utils version may need to be sync-ed when bumping the coursier version
-    def coursier      = "2.1.0-M5-18-gfebf9838c"
-    def jsoniterScala = "2.13.8"
-    def scalaMeta     = "4.5.1"
-    def scalaNative   = "0.4.4"
-    def scalaPackager = "0.1.26"
+    def coursier              = "2.1.0-M5-18-gfebf9838c"
+    def jsoniterScala         = "2.13.8"
+    def scalaJsEnvJsdomNodejs = "1.1.0"
```
alexarchambault: NIT Do we really need to put this version here? It's only used once, so it can just be inlined where it's used.

website/docs/guides/scala-js.md
```diff
@@ -39,6 +39,39 @@ scala-cli package Hello.scala --js --js-emit-source-maps
 #  node ./Hello.js
 ```
 
+## Scala.Js DOM support
```
alexarchambault: NIT
```suggestion
## Scala.js DOM support
```
(just realized I've been making a typo too, `Scala.JS` instead of `Scala.js` - but it's not `Scala.Js`. We should unify those at some point…)

website/docs/guides/scala-js.md
```diff
@@ -39,6 +39,39 @@ scala-cli package Hello.scala --js --js-emit-source-maps
...
+
+Passing `--js-dom` to the sub-commands to simulate a DOM in Nodej.js
```
alexarchambault: Phrasing:
```suggestion
Pass `--js-dom` to `compile`, `run`, or `test` to simulate a DOM in Node.js
```

website/docs/guides/scala-js.md
```diff
@@ -39,6 +39,39 @@ scala-cli package Hello.scala --js --js-emit-source-maps
...
+
+:::note
+If you see the following error, it means that you don't have installed `jsdom` library to simulate DOM.
```
alexarchambault:
```suggestion
If you see the following error, it means that you don't have the `jsdom` library installed to simulate the DOM.
```

alexarchambault (approved):

alexarchambault (approved):

PR #798, opened 2022-03-21
Cross compile to Scala 3

romanowski:
Cross compile build and its dependencies to Scala 3


alexarchambault:
Merged via https://github.com/VirtusLab/scala-cli/pull/807.

PR #812, opened 2022-03-25
Add resources support to Scala Native and fixes to overall resources

### LINKED ISSUE #808: Deleted resources are not being removed from `.scala-build` and the classpath
**Version(s)**
0.1.2, also on e968a10797b2a14b835c6555ffc6c3d7aad14b10

**Describe the bug**
When `--resource-dirs` option or using directives is used resources are copied to the project directory in `.scala-build`, which is then put on a classpath. If the resource will then be deleted (or have it's name changed) and the project recompiled, the previous resource will not be removed from the `.scala-build` and can still be used in the program.

**To Reproduce**

reproduction/main.scala
```scala
//> using resourceDir "resources"

import java.nio.charset.StandardCharsets
import java.io.{BufferedReader, InputStreamReader}
import java.io.FileInputStream

@main def main(): Unit =
  val inputStream = getClass().getResourceAsStream("/resourcefile.txt")
  val text = new BufferedReader(
    new InputStreamReader(inputStream, StandardCharsets.UTF_8)
  ).readLine()
  println(text)
```
reproduction/resources/resource1
```
resource1
``` 

1. Run `scala-cli project`
2. Remove `project/resources/resource1` file
3. Run `scala-cli project`

**Expected behaviour**
Deleted resource files should be removed from `.scala-build`, and resource files with changed filenames should not be duplicated there - instead the previous version should be removed.

jchyb:
This PR allows to embed resources (by default) in a SN binary, as well as fixes some bugs that ended up being related. Those can be then accessed via getResourceAsStream Java API. An integration test was also added.

While working on this I ended up fixing a couple bugs that did not have a chance to show until now, connected with scala-cli SN support:
* SN caching: previously only resources passed with a cli-option would be cached, using directive ones would be ignored.
* classpath passed to SN: previously I think I unnecessarily doubled it for non-script projects, which ended up messing with the resources

I later got some feedback and decided to also allow using c files as regular inputs, instead of using them via resources (which is basically how the underlying Scala Native handles them). This required me to make a very simple fix of #808 (and #1209), which I did for both the newly added C files, and regular java-style resources.
We now store in .scala-build an additional file which stores mapping between the path of the original c file/resource and the one copied internally for building. This way we can delete the internal files in case it disappears in the original path, which eliminates files doubling when renamed etc.

In summary:
* embedding resources was added by default for SN resources
* a new option "-no-embed", was added for SN which will not embed any resources (useful for keeping overall binary size low in the more complex builds)
* various improvements/ fixes were done with SN builds
* passing c files as regular inputs was added, which are then handled by SN allowing for easier interop in scala-cli
* Both c files and regular resources have internal .scala-build states consistent with the project by the use of ResourceMapper, which maps source file (in project) to destination (in .scala-build), before copying
* fixes #808 and #1209


jchyb:
I have heavily updated the PR message to reflect the new scope of the PR (now with fixes for the java resources outside of SN)

modules/build/src/main/scala/scala/build/ResourceMapper.scala
```diff
@@ -0,0 +1,122 @@
+package scala.build
+
+import scala.collection.Map
+import scala.jdk.CollectionConverters._
+import scala.build.internal.Constants
+
+trait BaseResourceMapper {
```
alexarchambault: Can this trait be moved to its own file?

modules/build/src/main/scala/scala/build/ResourceMapper.scala
```diff
@@ -0,0 +1,122 @@
...
+  def copyResourceToClassesDir(build: Build): Unit = build match {
+    case b: Build.Successful =>
+      val fullFilePath = resolveResourceMappingPath(b.output)
+      copyResourcesToDirWithMapping(b, fullFilePath, resourceMapping)
+
+    case _ =>
+  }
+}
+
+object NativeResourceMapper extends BaseResourceMapper {
```
alexarchambault: Same as above, this should live in its own `NativeResourceMapper.scala` file.

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
```diff
@@ -43,9 +43,14 @@ final case class ScalaNativeOptions(
...
-    nativeCompileDefaults: Option[Boolean] = None //TODO does it even work when we default it to true while handling?
+    nativeCompileDefaults: Option[Boolean] = None, //TODO does it even work when we default it to true while handling?
+
+  @Group("Scala Native")
+  @HelpMessage("Do not embed resources into the Scala Native binary (java style Resources will not be able to be used)")
+    noEmbed: Option[Boolean] = None
+
 )
-// format: on
+  // format: on
```
alexarchambault: NIT Why change that?
```suggestion
// format: on
```

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
```diff
@@ -43,9 +43,14 @@ final case class ScalaNativeOptions(
   @Group("Scala Native")
   @Hidden
   @HelpMessage("Use default compile options")
-    nativeCompileDefaults: Option[Boolean] = None //TODO does it even work when we default it to true while handling?
+    nativeCompileDefaults: Option[Boolean] = None, //TODO does it even work when we default it to true while handling?
+
+  @Group("Scala Native")
+  @HelpMessage("Do not embed resources into the Scala Native binary (java style Resources will not be able to be used)")
+    noEmbed: Option[Boolean] = None
```
alexarchambault: No need to use a negation in the name here IMO.
```suggestion
    embedResources: Option[Boolean] = None
```
Users can pass `--embed-resources=false` to negate the flag.

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
```diff
@@ -43,9 +43,14 @@ final case class ScalaNativeOptions(
   @Group("Scala Native")
   @Hidden
   @HelpMessage("Use default compile options")
-    nativeCompileDefaults: Option[Boolean] = None //TODO does it even work when we default it to true while handling?
+    nativeCompileDefaults: Option[Boolean] = None, //TODO does it even work when we default it to true while handling?
+
+  @Group("Scala Native")
+  @HelpMessage("Do not embed resources into the Scala Native binary (java style Resources will not be able to be used)")
```
alexarchambault: ?
```suggestion
  @HelpMessage("Embed resources into the Scala Native binary (Java style resources will not be used)")
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -923,10 +924,11 @@ object Package extends ScalaCommand[PackageOptions] {
         nativeWorkDir
       )
 
-    if (cacheData.changed)
+    if (cacheData.changed) {
+      NativeResourceMapper.copyCFilesToScalaNativeDir(build, nativeWorkDir)
       Library.withLibraryJar(build, dest.last.stripSuffix(".jar")) { mainJar =>
 
-        val classpath = build.fullClassPath.map(_.toString) :+ mainJar.toString
+        val classpath = build.artifacts.classPath.map(_.toString) :+ mainJar.toString
```
alexarchambault: Don't know how Scala Native handles it down the line, but I'd put `mainJar` first just in case, so that its classes are used in priority over those of the dependencies:
```suggestion
        val classpath = mainJar.toString +: build.artifacts.classPath.map(_.toString)
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -122,8 +122,15 @@ final case class Inputs(
       case v: Inputs.Virtual =>
         Iterator(v.content, bytes("\n"))
     }
+    val resourceDirsIt = (resourceDirs.flatMap { dir =>
+      os.walk(dir)
+        .filter(os.isFile(_))
+        .map(filePath => s"$filePath:" + os.read(filePath))
```
alexarchambault: Some resource files might not be readable as strings. Better read things as binary data, compute a SHA-1 of that data say, and use that here.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -103,7 +103,7 @@ final case class Inputs(
     if (canWrite) this
     else inHomeDir(directories)
   }
-  def sourceHash(): String = {
+  def sourceHash(resourceDirs: Seq[os.Path] = Seq.empty): String = {
```
alexarchambault: We shouldn't read resource directories here IMO. `sourceHash` is meant to be used in directory names under `.scala-build`, so that `scala-cli` concurrent runs with different options don't use the same folders. If only some resource files change, there's no reason to change that hash. Better manually add the resource directories hash right where it's needed, that is in `CachedBinary`.

modules/build/src/main/scala/scala/build/ResourceMapper.scala
```diff
@@ -0,0 +1,122 @@
...
+import scala.jdk.CollectionConverters._
+import scala.build.internal.Constants
+
+trait BaseResourceMapper {
+  private def readMappingIfExists(mappingFile: os.Path): Option[Map[os.Path, os.Path]] =
+    if (os.exists(mappingFile)) {
+      val mappingContent = os.read(mappingFile)
+      if (mappingContent.isEmpty()) Some(Map.empty)
+      else Some(
+        mappingContent.split('\n').map { pair =>
```
alexarchambault: ?
```suggestion
        mappingContent.linesIterator.filter(_.nonEmpty).map { pair =>
```

alexarchambault (commented):
I see you just pushed things. Here's my in progress review comments.

modules/directives/src/main/scala/scala/build/preprocessing/directives/UsingScalaNativeOptionsDirectiveHandler.scala
```diff
@@ -40,13 +42,15 @@ case object UsingScalaNativeOptionsDirectiveHandler extends UsingDirectiveHandle
       "native-linking",
       "native-clang",
       "native-clang-pp",
+      "native-no-embed",
```
alexarchambault, line 47: Same as the CLI option, better use a positive rather than negative form here.

jchyb:
Sorry about sudden pushing, I did it to (hopefully) fix the issues with the failing gif tests in the CI, the committed code was not changed, only rebased

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -256,6 +256,91 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+          .out.text().trim
+      expect(output == interopMsg)
+
+      os.move(root / projectDir / interopFileName, root / projectDir / "bindings2.c")
+      val output2 =
+        os.proc(TestUtil.cli, extraOptions, projectDir, "-q")
+          .call(cwd = root)
+          .out.text().trim
+
+      // LLVM throws linking errors if scalanative_print is internally repeated.
```
alexarchambault, line 338: Why does `copyResourcesToDirWithMapping` get "internally repeated" when renaming `bindings.c`? The exact name of the C file (apart from the `.c` extension) shouldn't matter, does it?

modules/build/src/main/scala/scala/build/ResourceMapper.scala
```diff
@@ -0,0 +1,122 @@
...
+      s"$inputPath,$outputPath"
+    }.mkString("\n")
+    os.write.over(mappingPath, mappingContent)
+  }
+
+  protected def copyResourcesToDirWithMapping(
+    build: Build.Successful,
+    mappingFilePath: os.Path,
+    mappingFunction: Build.Successful => Map[os.Path, os.Path]
+  ): Unit = {
```
alexarchambault: Wouldn't the signature be clearer with these arguments? (`mappingFunction` can be applied by the caller, and then there's no need of the whole build here, only the output dir)
```suggestion
  protected def copyResourcesToDirWithMapping(
    output: os.Path,
    mappingFilePath: os.Path,
    newMapping: Map[os.Path, os.Path]
  ): Unit = {
```

modules/build/src/main/scala/scala/build/ResourceMapper.scala
```diff
@@ -0,0 +1,122 @@
...
+
+    def isInDirectory(output: os.Path, filePath: os.Path) = {
+      val outputFullPath = output.toNIO.iterator().asScala.toSeq
+      val fileFullPath   = output.toNIO.iterator().asScala.toSeq
+      fileFullPath.startsWith(outputFullPath)
+    }
```
alexarchambault: I believe this should work (and this can probably be inlined then):
```suggestion
    def isInDirectory(output: os.Path, filePath: os.Path) =
      filePath.startsWith(output)
```

alexarchambault (commented):
About the resource mapping stuff: do we really need to write an actual mapping file on disk like this? Couldn't we instead write a list of relative paths of the resources written in the output dir, and right before copying / updating resources, look at those that will not be overwritten, and remove them?

The resource safety thing (ensuring we remove stale resources) is really great, but maybe it could be done in a separate PR?

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
jchyb: It would previously, since the binding.c and the renamed binding.c files would both appear in .scala-build (the original file would not be deleted without the mapping feature), leading to repeated symbols which in turn would give us linking errors. This test checks whether that still happens or not. I'll adjust the phrasing of the comment to make it more clear

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
jchyb: Embedding is meant to be enabled by default and I feel that having "--embed-resources" do nothing and asking users to use "--embed-resources=false" to have any kind of effect would be unintuitive. Is that a problem? wdyt

jchyb:
I have updated the mapping file contents to store only the relative output paths, as was suggested. I am not sure about the gains splitting the PR would give us (outside of easier review), since both the SN improvements and fixes to overall fixes to resources depend on the mapping mechanizm - I imagine if something goes wrong we would still have to revert both.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -237,6 +236,8 @@ object Inputs {
           Inputs.ScalaFile(d.path, p.subRelativeTo(d.path))
         case p if p.last.endsWith(".sc") =>
           Inputs.Script(d.path, p.subRelativeTo(d.path))
+        case p if p.last.endsWith(".c") || p.last.endsWith(".h") =>
```
Gedochao, line 247: this will include `.c`/`.h` files as `Inputs` regardless if we're in Scala native mode, no?
we should have an extra check here to only include those for Scala native builds

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -423,6 +425,7 @@ object Inputs {
       else if (arg.endsWith(".sc")) Right(Seq(Script(dir, subPath)))
       else if (arg.endsWith(".scala")) Right(Seq(ScalaFile(dir, subPath)))
       else if (arg.endsWith(".java")) Right(Seq(JavaFile(dir, subPath)))
+      else if (arg.endsWith(".c") || arg.endsWith(".h")) Right(Seq(CFile(dir, subPath)))
```
Gedochao, line 465: Similarly, this will include `.c` and `.h` sources passed explicitly as inputs, even for non-scala-native builds.
I think we should at least print a warning if someone passes such a source explicitly when not running Scala native (?)

jchyb: Good point, I kept thinking about it in the context of passing a directory (where unrecognized files are ignored, which lined up with what we are doing here), but I see there should be an error when unrecognized/irrelevant files are passed directly

alexarchambault:
I reworked a bit the commit history, to make it easier to review this. I'll have a closer look a bit later.

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
alexarchambault: I hold by `--embed-resources=false`.  IIRC there are no negative CLI options such as this in Scala CLI. For example, users can pass `--server=false` to disable using Bloop. Better stick to that convention for now.

To give you some more context, I added a `--no-default` option in coursier [quite some time ago](https://github.com/coursier/coursier/commit/3b4b773c6433de070d19239b1ad182037e5b413c), and I still regret it up to this day…

alexarchambault:
I'm in the midst of reviewing this. I might push a slightly reworked version of it, either here or in another PR.

modules/build/src/main/scala/scala/build/Inputs.scala
alexarchambault: That's a good point, but OTOH, users might be enabling / disabling Scala Native on-the-fly in watch mode by using a `//> using platform "native"`. So we should accept them upfront anyway IMO, and they'll get picked or not later on.

alexarchambault:
So I rebased on the latest `main`, and refactored things a bit further. AFAIC, I'm happy with the result (hoping the CI will be too).

alexarchambault:
Rebased again, hoping it will fix CI.

modules/cli-options/src/main/scala/scala/cli/commands/ScalaNativeOptions.scala
jchyb: Makes sense, I honestly forgot about `--server=false` when I wrote the original comment

alexarchambault (approved):

modules/build/src/main/scala/scala/build/Inputs.scala
Gedochao: yeah, that makes sense.
however, when Scala native is disabled and any `.c`/`.h` input are passed, we should at least show a warning that they're being ignored, IMO

modules/build/src/main/scala/scala/build/Inputs.scala
lwronski: https://github.com/VirtusLab/scala-cli/issues/1346

Gedochao (approved):

romanowski (approved):

PR #815, opened 2022-03-28
Add initial implementation for BSP workspace reload

Gedochao:
Just to make sure things are clear, here's what this PR actually does:

- makes sure IDEs pick up any sources included in a Scala CLI build after subsequent runs of `scala-cli setup-ide` without the need for a BSP server restart

and here's what this PR does not yet do (but should, and I hope to make it happen in future PRs)

- make sure IDEs pick up any sources changes when running and building the project without the need for a BSP server restart
- make sure IDEs pick up any changes to library dependencies, java options or scala options without a BSP server restart
- return a valid json-rpc error code & message if any failures occur

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
@@ -261,8 +261,9 @@ class BspServer(
   override def onBuildExit(): Unit = ()
 
   override def workspaceReload(): CompletableFuture[Object] =
-    super.workspaceReload().thenApply { res =>
-      for {
+    super.workspaceReload().thenApply[Object] { res =>
+      for { // this is just a workaround hack to emulate workspace reloading until Bloop starts supporting it
+        // TODO: reload dependencies
         case (targetUri, targetName) <- buildTargetNamesByUri.toSeq
         scalaBuildDirPath = os.Path(new URI(targetUri).getRawPath)
         if os.isDir(scalaBuildDirPath)
@@ -282,7 +283,7 @@ class BspServer(
         val updatedBloopFileJson: Array[Byte] = core.writeToArray(updatedBloopFile)(BloopCodecs.codecFile)
         os.write.over(bloopFileJsonPath, updatedBloopFileJson, createFolders = true)
       }
-      res
+      res // TODO: return a valid message in case of an error
     }
 
   def initiateShutdown: Future[Unit] =
--- a/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspServer.scala
@@ -1,7 +1,7 @@
 package scala.build.bsp
 
 import _root_.bloop.config.{Config, ConfigCodecs as BloopCodecs}
-import ch.epfl.scala.bsp4j.{BuildClient, LogMessageParams, MessageType}
+import ch.epfl.scala.bsp4j.{BuildClient, BuildTargetIdentifier, CompileParams, LogMessageParams, MessageType}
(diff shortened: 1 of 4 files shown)


modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -1,7 +1,9 @@
 package scala.build.bsp
 
-import ch.epfl.scala.bsp4j.{BuildClient, LogMessageParams, MessageType}
+import _root_.bloop.config.{Config, ConfigCodecs => BloopCodecs}
```
alexarchambault, line 3: NIT Is the `_root_` really necessary?

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -254,6 +259,55 @@ class BspServer(
...
   override def onBuildExit(): Unit = ()
 
+  // TODO: reload dependencies
+  // TODO: reload scalac & javac options
+  // TODO: fix IntelliJ run tasks not picking up new sources
+  override def workspaceReload(): CompletableFuture[Object] =
+    super.workspaceReload().thenApply[Object] { res =>
+      val buildTargetsToRecompile: List[b.BuildTargetIdentifier] =
+        for { // this is just a workaround hack to emulate workspace reloading until Bloop itself supports it
+          case (targetUri, targetName) <- buildTargetNamesByUri.toList
```
alexarchambault, line 269: NIT no need of `case` here
```suggestion
          (targetUri, targetName) <- buildTargetNamesByUri.toList
```

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -254,6 +259,55 @@ class BspServer(
...
+          scalaBuildDirPath = os.Path(new URI(targetUri).getRawPath)
```
alexarchambault, line 270: I don't think this way of converting URIs to paths will work in all cases (on Windows in particular). This should work better:
```suggestion
          scalaBuildDirPath = os.Path(Paths.get(new URI(targetUri)))
```
(using `Paths` from NIO)

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -254,6 +259,55 @@ class BspServer(
...
+          if os.isDir(scalaBuildDirPath)
```
alexarchambault, line 271: Isn't this unnecessary thanks to the `if os.isFile(ideInputsJsonPath)` right after?
```suggestion
```

modules/build/src/main/scala/scala/build/bsp/BspServer.scala
```diff
@@ -254,6 +259,55 @@ class BspServer(
...
+          ideInputsJsonPath = scalaBuildDirPath / "ide-inputs.json"
+          if os.isFile(ideInputsJsonPath)
+          bloopDirPath = scalaBuildDirPath / ".bloop"
+          if os.isDir(bloopDirPath)
```
alexarchambault, line 275: Same as above, this should be unnecessary:
```suggestion
```

alexarchambault (commented):
I'm not sure we want to override Bloop files like this… Bloop files in a given `project_…` directory are supposed to be the ones corresponding to the `Inputs` instance used when computing the hash in the directory name. Overwriting them for different set of inputs breaks what they are trying to address: isolating Scala CLI invocations having different sets of inputs.

Gedochao:
This PR will probably be succeeded by the implementation from #852, closing this.

PR #816, opened 2022-03-29
Package command - add support for multi modules

### LINKED ISSUE #530: Scala.js - package command - support for multi modules
It would be nice to support js modules in `package` command 

```
  @JSExportTopLevel(name = "onRequest", moduleID = "request_type")
  def func(...)

  @JSExportTopLevel(name = "onRequest", moduleID = "request_headers")
  def func(...)
```

for now,  `package ... --js ` throws error, that main class is missing.

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -698,16 +707,25 @@ object Package extends ScalaCommand[PackageOptions] {
...
-        if (os.exists(mainJs)) {
-          os.copy(mainJs, dest, replaceExisting = true)
-          if (build.options.scalaJsOptions.emitSourceMaps && os.exists(sourceMapJs)) {
-            val sourceMapDest =
-              build.options.scalaJsOptions.sourceMapsDest.getOrElse(os.Path(s"$dest.map"))
-            os.copy(sourceMapJs, sourceMapDest, replaceExisting = true)
-            logger.message(s"Emitted js source maps to: $sourceMapDest")
+
+        if (os.exists(mainJs))
+          if (os.walk(linkingDir).filterNot(_ == mainJs).filterNot(_ == sourceMapJs).nonEmpty) {
```
alexarchambault: Let's use `os.walk.stream` here, that should allow not to list all files if ever there are a lot of them, or the directory contains garbage, etc.

```suggestion
          if (os.walk.stream(linkingDir).filterNot(_ == mainJs).filterNot(_ == sourceMapJs).nonEmpty) {
```
(if `geny.Generator` doesn't have a `nonEmpty` method, I guess `headOption.nonEmpty` should do…)

alexarchambault (commented):
Apart from the minor comment, LGTM!

alexarchambault (approved):

romanowski (changes requested):
Shouldn't we update js documentation?

armanbilge:
This feature is already mentioned in the documentation, it just wasn't working :)

https://scala-cli.virtuslab.org/docs/reference/cli-options/#--js-module-split-style

lwronski:
> Shouldn't we update js documentation?

I think, that I confusingly named PR, it is more bug fixes. Because if someone run `package --js` command with the following scala file:
```
  @JSExportTopLevel(name = "onRequest", moduleID = "request_type")
  def func(...)

  @JSExportTopLevel(name = "onRequest", moduleID = "request_headers")
  def func(...)
```
he gets an error that `main class is missing`, when run generated js file. This PR fixed it.

romanowski:
That is valid, however current Scala JS guide does not have a section about publishing so I think we should describe the publishing for JS. We have a [small section ](https://scala-cli.virtuslab.org/docs/guides/scala-js)at `publish` command but I think it only covers a basic scenario.

So I am merging this PR for now, @lwronski could you review the documentation for packaging in JS and create a follow up PR if needed?

armanbilge:
Note that publishing (i.e., uploading SJSIR files to Maven) is different from packaging (linking SJSIR files into JS files).

PR #819, opened 2022-03-29
Fix sclicheck

romanowski:
- fix script to run tests on CI
- create status file to make sure that tests are actually run
- add support for compiling snippets

### DIFF
--- a/.github/scripts/check_docs.sh
+++ b/.github/scripts/check_docs.sh
@@ -3,15 +3,13 @@ set -e
 
 dest="$(pwd)/out/sclicheck/bin"
 
-allArgs='"$@"'
-
 ./mill copyTo cli.launcher "$dest/scala-cli.sh"
 
 # work around sh issues (sh ignores PATH entries with '+' or '%', and the cs-provided Java 17 entry has one)
 # so we try to run the scala-cli launcher with bash instead
 cat > "$dest/scala-cli" << EOF
 #!/usr/bin/env bash
-exec bash "$dest/scala-cli.sh" $allArgs
+exec bash "$dest/scala-cli.sh" "\$@"
 EOF
 chmod +x "$dest/scala-cli"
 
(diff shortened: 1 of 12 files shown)


.github/scripts/check_docs.sh
```diff
@@ -3,26 +3,35 @@ set -e
...
+allArgs='"$@"'
+
 ./mill copyTo cli.launcher "$dest/scala-cli.sh"
 
 # work around sh issues (sh ignores PATH entries with '+' or '%', and the cs-provided Java 17 entry has one)
 # so we try to run the scala-cli launcher with bash instead
 cat > "$dest/scala-cli" << EOF
 #!/usr/bin/env bash
-exec bash "$dest/scala-cli.sh"
+exec bash "$dest/scala-cli.sh" $allArgs
```
alexarchambault: Wouldn't this work?
```suggestion
exec bash "$dest/scala-cli.sh" "\$@"
```

.github/scripts/check_docs.sh
```diff
@@ -3,26 +3,35 @@ set -e
...
+else
     toCheck=("$@")
 fi
 
-# adding --resource-dirs is a hack to get file watching for free on .md files
-scala-cli sclicheck/sclicheck.scala docs -- "${toCheck[@]}"
+statusFile="$(pwd)/out/sclicheck/.status"
+scala-cli sclicheck/sclicheck.scala -- --status-file "$statusFile" "${toCheck[@]}" || (
+  echo "Checking documentation failed. To run tests locally run `.github/scripts/check_docs.sh <failing_file>`" &&
+  echo "You can find more about automatic documentaiton testing in sclicheck/Readme.md file." &&
```
alexarchambault: I don't think the `&&` are necessary here.

.github/scripts/check_docs.sh
```diff
@@ -3,26 +3,35 @@ set -e
...
+  exit 1
+)
+
+test -f "$statusFile" || 
+  ( echo "Fatal error. Status file: $statusFile does not exists what signal problem with running tests." &&
+  exit 1)
```
alexarchambault: Just like above, no need of the `&&` I think:
```suggestion
test -f "$statusFile" || (
  echo "Fatal error. Status file: $statusFile does not exists what signal problem with running tests."
  exit 1
)
```

.gitignore
```diff
@@ -5,3 +5,5 @@ out/
 .idea/
 
 *.semanticdb
+.bsp
+.scala-build
```
alexarchambault, line 9: The `.scala-build` shouldn't be needed here. Ideally, no `.scala-build` directory should be created at the root (this is checked in the tests, to ensure `.scala-build` is created where expected, rather than naively in `os.pwd`). Better put that line in a `.gitignore` in a sub-directory.

sclicheck/sclicheck.scala
```diff
@@ -6,8 +6,11 @@ import scala.util.matching.Regex
 import scala.io.StdIn.readLine
 import fansi.Color.{Red, Blue, Green}
 import java.security.SecureRandom
+import scala.util.Random
+import os.stat
```
alexarchambault: Seems unused?
```suggestion
```

.gitignore
romanowski: TBH I often run some quick tests using `scala-cli-src` and it sometimes creates top-level `.scala-build`. But I can move that to my personal .gitignore

alexarchambault (approved):

PR #821, opened 2022-03-30
Correction in Scala.js docs

armanbilge:
Scala CLI doesn't have support for _bundling_ JS files.

However, Scala CLI does in fact support _linking_ of SJSIR files, since this is the only way to get a runnable JS file as output :)


lwronski (approved):
Thanks @armanbilge!

PR #827, opened 2022-03-31
Resolve latest version for --cli-version

lwronski:


modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala
```diff
@@ -15,23 +19,22 @@ object LauncherCli {
...
+    val scalaVersion  = options.cliScalaVersion.getOrElse(Properties.versionNumberString)
+    val snapshotsRepo = Seq(Repositories.central, Repositories.sonatype("snapshots"))
 
-    val scalaCliDependency = Seq(dep"org.virtuslab.scala-cli::cli:$version")
-    val snapshotsRepo = Seq(
-      Repositories.central.root,
-      Repositories.sonatype("snapshots").root
-    )
+    val cliVersion: String =
+      if (version == "latest") resolveLatestScalaCliVersion(cache, snapshotsRepo) else version
```
romanowski: I think `latest` should point to latest stable version and `nightly` should point to latest snapshot. 

TBH I do not know if we need  `latest` so we may just rename it to `nightly`

romanowski (changes requested):
I know it is a draft but it looks good already!

I only miss tests and changes in docs

modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala
lwronski: Right, `nightly` sounds better, so I just renamed it.

lwronski:
I added `CoursierUtil` because I don't have access to `WebPage.listDirectories` which is declared as a private package:
```
private[coursier] object WebPage extends WebPageCompatibility {
```

modules/cli/src/main/scala/coursier/CoursierUtil.scala
```diff
@@ -0,0 +1,9 @@
+package coursier
+
+import coursier.util.WebPage
+
+case object CoursierUtil {
```
alexarchambault: NIT (no need to make this a `case object`)
```suggestion
object CoursierUtil {
```

modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala
```diff
@@ -1,37 +1,41 @@
 package scala.cli.launcher
-
```
alexarchambault, line 2: Why remove this empty line?

modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala
```diff
@@ -65,4 +69,34 @@ object LauncherCli {
...
+      case Left(_) =>
+        System.err.println("Unable to find nightly Scala CLI version")
+        sys.exit(1)
+      case Right(f) =>
+        val snapshotRepoPage = os.read(os.Path(f, Os.pwd))
+        val rawVersions      = coursier.CoursierUtil.rawVersions(snapshotRepoUrl, snapshotRepoPage)
+        val versions         = rawVersions.map(Version(_))
+        val nightlyVersion   = versions.max
+
+        nightlyVersion.repr
```
alexarchambault: Maybe let's check that `versions` is non-empty, just in case, and print a better error message if it is:
```suggestion
        if (versions.isEmpty)
          sys.error(s"No versions found in $snapshotRepoUrl (locally at $f)")
        else
          versions.max.repr
```

alexarchambault (commented):
Apart from the minor comments, LGTM!

alexarchambault (approved):

PR #830, opened 2022-04-01
Graal scala3

romanowski:


build.sc
```diff
@@ -416,6 +422,38 @@ trait ScalaParse extends SbtModule with ScalaCliPublishModule with settings.Scal
   def scalaVersion = Scala.defaultInternal
 }
 
+trait Scala3Runtime extends SbtModule with ScalaCliPublishModule with settings.ScalaCliCompile {
```
alexarchambault: NIT Why do we need the `setting.` prefix here? We could just import `ScalaCliCompile` above.

build.sc
```diff
@@ -416,6 +422,38 @@ trait ScalaParse extends SbtModule with ScalaCliPublishModule with settings.Scal
...
+  def ivyDeps      = super.ivyDeps() ++ Agg(Deps.scalaparse)
```
alexarchambault: Seems scalaparse is unused.
```suggestion
```

build.sc
```diff
@@ -416,6 +422,38 @@ trait ScalaParse extends SbtModule with ScalaCliPublishModule with settings.Scal
...
+  def scalaVersion = Scala.scala3
+  def jar = T {
+    val original = super.jar().path
+    // scala3RuntimeFixes.jar is also used within
+    // resource-config.json and BytecodeProcessor.scala
+    val dest = original / os.up / "scala3RuntimeFixes.jar"
+    os.copy(original, dest)
+    PathRef(dest)
+  }
```
alexarchambault: I'm not sure that's the right place to do this. If this file needs to be added under a specific name as resource, this can be handled in the module that adds it as a resource.
```suggestion
```

build.sc
```diff
@@ -416,6 +422,38 @@ trait ScalaParse extends SbtModule with ScalaCliPublishModule with settings.Scal
...
+}
+
+class Scala3Graal(val crossScalaVersion: String) extends BuildLikeModule {
+  def ivyDeps = super.ivyDeps() ++ Agg(
+    Deps.asm,
+    Deps.osLib
+  )
+
+  def resources = T.sources {
+    super.resources() ++ Seq(`scala3-runtime`.jar())
```
alexarchambault: Something like this ought to do:
```suggestion
    val extraResourceDir = T.dest / "extra"
    os.copy(`scala3-runtime`.jar().path, extraResourceDir / "scala3RuntimeFixes.jar", createFolders = true)
    super.resources() ++ Seq(PathRef(extraResourceDir))
```

build.sc
```diff
@@ -416,6 +422,38 @@ trait ScalaParse extends SbtModule with ScalaCliPublishModule with settings.Scal
...
+  }
+}
+
+trait Scala3GraalProcessor extends ScalaModule {
+  def moduleDeps   = Seq(`scala3-graal`(Scala.scala3))
+  def scalaVersion = Scala.scala3
+  // Make sure that we put jar on run classpath
+  def runClasspath   = T(`scala3-graal`(Scala.scala3).jar() +: super.runClasspath())
```
alexarchambault: Why do we need this? Why isn't the `moduleDeps` enough?

build.sc
```diff
@@ -509,9 +547,13 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
   def moduleDeps = Seq(
     build(myScalaVersion),
     `cli-options`,
-    `test-runner`(myScalaVersion)
+    `test-runner`(myScalaVersion),
+    `scala3-graal`(myScalaVersion)
   )
 
+  // We are adding graal as compile deps and adding jat on classpath since we want to build a native image from jar not from directories
+  def runClasspath = T(Seq(`scala3-graal`(myScalaVersion).jar()) ++ super.runClasspath())
```
alexarchambault: Same as above. `native-image` should be fine with directories.

build.sc
```diff
@@ -539,6 +581,27 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
   }
 }
 
+trait Cli3 extends Cli {
+  override def myScalaVersion = Scala.scala3
+
+  val argsFile = os.pwd / os.RelPath("out") / ".graal-cp-args"
+
```
alexarchambault: Unused?
```suggestion
```

build.sc
```diff
@@ -539,6 +581,27 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
...
+  override def nativeImageClassPath = T {
+    // TODO - results are cached - need to add inputs or caching method
+    val classpath     = super.nativeImageClassPath().map(_.path).mkString(File.pathSeparator)
+    val coursierCache = coursier.cache.FileCache[coursier.util.Task]().location.toString
```
alexarchambault: You shouldn't use the coursier cache path here… It's not meant to be used this way. `coursier.cache.shaded.dirs.ProjectDirectories` could be used instead (see `Directories.scala` for an example of use).

modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala
```diff
@@ -231,46 +234,77 @@ object NativeImage {
...
-            "-cp",
-            classPath.map(_.toString).mkString(File.pathSeparator),
-            mainClass
-          )
+        ) { processedClasspath =>
+          val (classPath, toClean, scala3extraOptions) =
+            if (!build.scalaParams.scalaBinaryVersion.startsWith("3"))
+              (processedClasspath, Seq[os.Path](), Seq[String]())
+            else {
+              val coursierCacheLocation = os.Path(FileCache().location.toPath())
```
alexarchambault: Same as above, the coursier cache isn't really meant for that… The graal processor should just compute its own default cache path with the directories-jvm library.

modules/scala3-graal/src/main/scala/scala/cli/graal/BytecodeProcessor.scala
```diff
@@ -0,0 +1,253 @@
...
+
+import org.objectweb.asm._
+
+import java.io.{File, InputStream}
+import java.nio.file.{Files, Path, StandardOpenOption}
+import java.util.jar.{Attributes, JarEntry, JarFile, JarOutputStream, Manifest}
+
+import scala.jdk.CollectionConverters._
+
+trait JarCache {
```
alexarchambault: Put those extra classes (`JarCache`, `ClassPathEntry`, `TempCache`) in their own file?

modules/scala3-graal/src/main/scala/scala/cli/graal/BytecodeProcessor.scala
```diff
@@ -0,0 +1,253 @@
...
+    }
+    finally jarFile.close()
+  }
+
+  def processClasspath(classpath: String, cache: JarCache = TempCache): Seq[ClassPathEntry] = {
+    val cp = classpath.split(File.pathSeparator)
+    if (cp.size == 1 && cp.head.endsWith(".jar"))
+      processPathingJar(cp.head, cache)
+    else
+      processClasspathEntries(cp.toSeq, cache)
```
alexarchambault: NIT Match so that we don't call `head`?
```suggestion
    cp match {
      case Array(jar) if jar.endsWith(".jar") => processPathingJar(jar, cache)
      case _ => processClasspathEntries(cp.toSeq, cache)
    }
```

modules/scala3-graal/src/main/scala/scala/cli/graal/BytecodeProcessor.scala
```diff
@@ -0,0 +1,253 @@
...
+      sig: String,
+      exceptions: Array[String]
+    ): MethodVisitor =
+      if (name == "<clinit>")
+        new StaticInitVistor(super.visitMethod(access, name, desc, sig, exceptions))
+      else
+        super.visitMethod(access, name, desc, sig, exceptions)
+  }
+
+  def toPath(str: String) = os.FilePath(str) match {
```
alexarchambault: Do we really need that method? It just does `os.Path(str, os.pwd)`.

modules/cli/src/main/resources/META-INF/native-image/org.virtuslab/scala-cli-core/resource-config.json
```diff
@@ -21,6 +21,9 @@
       },
       {
         "pattern": "\\Qcom/google/javascript/jscomp/js/polyfills.txt\\E"
+      },
+      {
+        "pattern": ".*scala3RuntimeFixes.jar$"
```
alexarchambault, line 23: It seems this is redundant with the `-H:IncludeResources=` added in `project/settings.sc`. Better keep things there, as this should allow to keep the resource name in a single place.

alexarchambault (commented):
I left some comments, but don't we want to move the graal processing stuff to its own project? It does pretty general things, that may also be used by projects others than Scala CLI.

Also, I'm not sure the global caching logic is fine with concurrency (file locks or atomic moves should be used to ensure it's fine). But are the computations it caches that slow? Couldn't a per-project cache work?

build.sc
romanowski: I think it was autocompetion in Metals I've missed

build.sc
romanowski: there was a porblem with hadling resources, namely jar was not visible. I guess this may be caused by my setup of resources, so I will roll it back and see if it works

romanowski:
> I left some comments, but don't we want to move the graal processing stuff to its own project? It does pretty general things, that may also be used by projects others than Scala CLI.

Wouldn't it be published with Scala CLI anyway? I can also do that once its stabilize and it can be another lib that we can test  publishing from Scala CLI on.scala-cli

> Also, I'm not sure the global caching logic is fine with concurrency (file locks or atomic moves should be used to ensure it's fine). But are the computations it caches that slow? Couldn't a per-project cache work?

I think I can change it back to process everything every time (`TempCache`) and we can investigate a proper caching in follow up PR (actually, this can make a nice task for someone outside main Scala CLI team)

alexarchambault (approved):

PR #852, opened 2022-04-05
Add implementation for  `BSP` `workspace/reload`

Gedochao:
bsp_reload_2

The reload generally works now, but there are some limitations on IntelliJ:
- if there are sources directly in the primary workspace directory (first input arg passed to the `scala-cli setup-ide` command), they will not be included in the new build after the reload if the project hash has changed.
- The reason for this is the old problem with `intellij-scala` creating an empty `*-root` module. Even though we work around it on BSP initialization (as per changes in #780), it becomes harder when handling a `workspace/reload` request with changed sources. If the source directories change on reload, the build target ids also change (the bsp project has a new hash). As a result, some targets need to be deleted, while others are created and `intellij-scala` creates the `*-root` module in between, as it's their hack for there always being a module with the primary workspace source directory. 
 
image

- Now, this is tolerable if the sources are hidden in subdirectories (screenshot above), as then this only causes an empty, harmless module to be added to the IntelliJ project structure. It gets worse if the workspace root directory also contains sources, as then we have a duplicate, and IntelliJ immediately gets rid of it just as we reload. This is because IntelliJ never allows for 2 modules to have overlapping source directories (and in all likelihood, the pre-reload build & the post-reload build share the root workspace directory).  Thus, even though we reload correctly, we miss any sources in the root directory. (screenshot below)
image

- There may be some workaround for this, but I think it should be handled outside of the scope of this PR. Currently if a user runs into this, they have to either fix the project structure manually (by adding the correct sources directory to the BSP module and getting rid of the `*-root` module) or get rid of the `.idea` directory and reimport the project. Quite annoying 😞 
- This issue is absent from Metals, of course.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspServerProxy.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspServerProxy.scala
@@ -16,6 +16,29 @@ import scala.concurrent.duration.DurationInt
 import scala.jdk.CollectionConverters.*
 import scala.util.Try
 
+/** A wrapper layer for the actual [[BspServer]] & [[BloopCompiler]] instances. Added to enable live
+  * hot-swapping of both instances on a BSP:workspace/reload request. All RPC traffic is forwarded
+  * to the actual [[BspServer]] instance.
+  *
+  * @param bloopRifleConfig
+  *   Bloop rifle config
+  * @param threads
+  *   BSP threads
+  * @param localClient
+  *   BSP build client
+  * @param buildOptions
+  *   CLI build options
+  * @param compile
+  *   a function allowing to compile the bloop build targets
+  * @param logger
+  *   the logger
+  * @param initialInputs
+  *   the inputs to initialize BSP with (only used when the constructor is called)
+  * @param argsToInputs
+  *   a function allowing to convert raw input args to an Inputs instance
+  * @param prepareBuild
+  *   a function allowing to re-build bloop projects
+  */
 class BspServerProxy(
   bloopRifleConfig: BloopRifleConfig,
   threads: BspThreads,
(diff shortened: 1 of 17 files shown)


modules/cli/src/main/scala/scala/cli/commands/SetupIde.scala
```diff
@@ -99,9 +108,18 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     val (bspName, bspJsonDestination) = bspDetails(inputs.workspace, options.bspFile)
     val scalaCliBspJsonDestination =
       inputs.workspace / Constants.workspaceDirName / "ide-options-v2.json"
+    val scalaCliBspInputsJsonDestination =
+      inputs.workspace / Constants.workspaceDirName / "ide-inputs.json"
 
     val inputArgs = inputs.elements.collect { case d: Inputs.OnDisk => d.path.toString }
 
+    val ideInputs = IdeInputs(args =
+      options.shared.validateInputArgs(args)
```
Gedochao, line 117: It isn't a lot of overhead, but the `args` are effectively validated twice here (once for the `IdeInputs` instance, and once for the `Inputs` instance). I intend to try and refactor this in a future PR.

modules/build/src/main/scala/scala/build/bsp/BspServerProxy.scala
```diff
@@ -0,0 +1,126 @@
...
+    val event = new b.BuildTargetEvent(targetId)
+    event.setKind(eventKind)
+    event
+  }
+
+  private def responseError(
+    message: String,
+    errorCode: Int = JsonRpcErrorCodes.InternalError
+  ): ResponseError =
+    new ResponseError(errorCode, message, new Object())
```
Gedochao, line 146: even though we return a valid error response to the `workspace/reload` request, it doesn't seem like `IntelliJ` is handling it properly, as no error message is shown in the UI. We may consider sending a BSP `logMessage` request, so the `IntelliJ` user has at least a slight chance of noticing that something went wrong (it'd show in the BSP build log, not just the BSP logs in the IntelliJ directory). `showMessage` isn't an option either, as `IntelliJ` ignores that as well. TBD, I'm open to suggestions (although I'd rather do it in a separate PR as well)

Gedochao:
We're going ahead with a reworked version of this in #858, closing this.

PR #853, opened 2022-04-05
Send the Bloop server output to a file when possible

alexarchambault:
Needs https://github.com/scala-cli/bloop-core/pull/55, included in the upcoming (Scala CLI's) bloop-core `1.4.20`.

This sends the output of the Bloop server to a file under the user home directory (alongside the socket file, `~/Library/Caches/ScalaCli/bloop/daemon/output` on macOS for example), and adds a `scala-cli bloop output` command printing that file in the console, so that users can easily have a look at that file.

In order for that output file not to grow too big, we pass an option to the Bloop server so that it truncates that file periodically (as soon as it reaches 1 MB, https://github.com/scala-cli/bloop-core/pull/55).

### DIFF
--- a/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
@@ -47,7 +47,25 @@ object BloopRifle {
   ): Future[Unit] =
     config.classPath(version) match {
       case Left(ex) => Future.failed(new Exception("Error getting Bloop class path", ex))
-      case Right(cp) =>
+      case Right((cp, isScalaCliBloop)) =>
+        object IntValue {
+          def unapply(s: String): Option[Int] =
+            // no String.toIntOption in Scala 2.12.x
+            try Some(s.toInt)
+            catch {
+              case _: NumberFormatException => None
+            }
+        }
+        val bloopServerSupportsFileTruncating =
+          isScalaCliBloop && {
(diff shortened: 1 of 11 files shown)


modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
```diff
@@ -47,7 +47,25 @@ object BloopRifle {
...
+            catch {
+              case _: NumberFormatException => None
+            }
+        }
+        val bloopServerSupportsFileTruncating =
+          isScalaCliBloop && {
+            version.takeWhile(c => c.isDigit || c == '.').split('.') match {
+              case Array(IntValue(maj), IntValue(min), IntValue(patch)) =>
+                import scala.math.Ordering.Implicits._
+                Seq(maj, min, patch) >= Seq(1, 14, 20)
```
romanowski, line 64: I wonder why we do need to check the bloop version if we control the version of bloop ourselves? I think if we see too old version running (e.g. after update of scala-cli) we should restart it with newer version

romanowski (approved):

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifle.scala
alexarchambault: That's for if users manually force an earlier version to something else. Earlier Bloop versions can't truncate their output file, so redirecting the output this way would create a file that might grow too big.

PR #854, opened 2022-04-05
 docs: fix examples in scala-js.md and native-images.md

brianloveswords:
This change fixes the example command on https://scala-cli.virtuslab.org/docs/guides/scala-js/ to use `scala-cli` instead of `scala`.

### DIFF
--- a/website/docs/cookbooks/native-images.md
+++ b/website/docs/cookbooks/native-images.md
@@ -17,7 +17,7 @@ object Echo {
 
 The following command packages this application as a native executable:
 ```bash
-scala-cli package Echo.scala -o echo
+scala-cli package --native-image Echo.scala -o echo
 ```
 
 <!-- Expected-regex:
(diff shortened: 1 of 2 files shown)


brianloveswords:
After I made this I noticed the example calls on https://scala-cli.virtuslab.org/docs/cookbooks/native-images are missing the `--native-image` flag. I can amend this PR to include those as well, let me know if that's alright!

lwronski:
> After I made this I noticed the example calls on https://scala-cli.virtuslab.org/docs/cookbooks/native-images are missing the `--native-image` flag. I can amend this PR to include those as well, let me know if that's alright!

@brianloveswords  Thanks for contributing,

Yes, you have right. There should be added `--native-image` flag.

brianloveswords:
@lwronski alright, I pushed up a new commit adding those `--native-image` flags, so this should be good to go now. Thanks!

romanowski (approved):

PR #858, opened 2022-04-06
Add support for BSP workspace/reload (reworked version of #852)

alexarchambault:


modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -282,70 +282,87 @@ final class BspImpl(
...
-  val localClient: b.BuildClient with BloopBuildClient =
+  private val localClient: b.BuildClient with BloopBuildClient =
     if (verbosity >= 3)
       new BspImpl.LoggingBspClient(actualLocalClient)
     else
       actualLocalClient
 
-  var remoteServer: BloopCompiler  = null
-  var actualLocalServer: BspServer = null
+  private def newBloopSession(inputs: Inputs): BloopSession = {
```
Gedochao, line 278: Since this is essentially initializing a new session, I think it'd be better to make it a constructor/`apply` method for `BloopSession`. `BspImpl` is too big a source file already.

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -389,34 +402,114 @@ final class BspImpl(
...
+    val previousTargetIds = currentBloopSession.currentBspServer.targetIds
+    val newBloopSession0  = newBloopSession(newInputs)
+    bloopSession.update(currentBloopSession, newBloopSession0, "Concurrent reload of workspace")
+    currentBloopSession.dispose()
+
+    newBloopSession0.resetDiagnostics(actualLocalClient)
+    prepareBuild(newBloopSession0) match {
+      case Left((buildException, scope)) =>
+        CompletableFuture.completedFuture(
+          new ResponseError(
```
Gedochao: can use `BspImpl.responseError` here as well

alexarchambault:
(Latest push doesn't take into account the review comments yet.)

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
alexarchambault: I tried that [here](https://github.com/alexarchambault/scala-cli/commit/301d47b19b200e6288d4aeed82d2a3a765819b67), but I'm not too happy with the result: that method references a few others of `BspImpl`'s private methods, so we have to pass them as closures to `BloopSession.create` say, which is a bit ugly…

Gedochao: eh, neither is ideal, I guess. ok then.

Gedochao (approved):

PR #868, opened 2022-04-09
Start repl with scalac options

romanowski:
Add test to make sure that we pass options to repl correctly.

This fixes regression in #738

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Repl.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Repl.scala
@@ -211,7 +211,13 @@ object Repl extends ScalaCommand[ReplOptions] {
           " These will not be accessible from the REPL."
       )
 
-    val replArgs = options.notForBloopOptions.replOptions.ammoniteArgs ++ programArgs
+    val additionalArgs =
+      if (options.notForBloopOptions.replOptions.useAmmonite)
+        options.notForBloopOptions.replOptions.ammoniteArgs
+      else
+        options.scalaOptions.scalacOptions.toSeq.map(_.value.value)
+
+    val replArgs = additionalArgs ++ programArgs
 
     if (dryRun)
       logger.message("Dry run, not running REPL.")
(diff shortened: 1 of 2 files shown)


modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -83,6 +83,8 @@ object Runner {
         ) ++
         args
 
+    logger.debug(s"Running java command: ${command.mkString(" ")}")
+
```
alexarchambault: `run` should already log that stuff in debug mode.
```suggestion
```

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -82,7 +82,7 @@ object Repl extends ScalaCommand[ReplOptions] {
     ): Unit = {
       val res = runRepl(
         buildOptions,
-        programArgs,
+        programArgs ++ options.shared.scalac.scalacOption,
```
alexarchambault: - Better use `buildOptions.scalaOptions.scalacOptions.toSeq.map(_.value.value)` rather than `options.shared.scalac.scalacOption`. The latter corresponds to CLI options, while the former also takes into account using directives.
- I don't think this is going to work for Ammonite. Better add those options in `runRepl` itself, when when we specifically run the default REPL.

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -211,7 +211,10 @@ object Repl extends ScalaCommand[ReplOptions] {
           " These will not be accessible from the REPL."
       )
 
-    val replArgs = options.notForBloopOptions.replOptions.ammoniteArgs ++ programArgs
+    val replArgs =
+      options.notForBloopOptions.replOptions.ammoniteArgs ++ programArgs ++ options.scalaOptions.scalacOptions.toSeq.map(
+        _.value.value
```
alexarchambault: I think we shouldn't add those options if `options.notForBloopOptions.replOptions.useAmmonite` is true.

alexarchambault (approved):

alexarchambault:
(Looking at the current `main` code, it seems `options.notForBloopOptions.replOptions.ammoniteArgs` shouldn't be added either when `options.notForBloopOptions.replOptions.useAmmonite` is false…)

PR #886, opened 2022-04-12
Reload options on a BSP workspace/reload

Gedochao:
This allows for BSP workspace/reload to pick up options passed to the CLI, i.e. `--dependency` args.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -12,7 +12,6 @@ import java.util.concurrent.{CompletableFuture, Executor}
 import scala.build.EitherCps.{either, value}
 import scala.build._
 import scala.build.bloop.{BloopServer, ScalaDebugServer}
-import scala.build.blooprifle.BloopRifleConfig
 import scala.build.compiler.BloopCompiler
 import scala.build.errors.{BuildException, Diagnostic}
 import scala.build.internal.{Constants, CustomCodeWrapper}
@@ -24,15 +23,16 @@ import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success}
 
 final class BspImpl(
-  logger: Logger,
-  bloopRifleConfig: BloopRifleConfig,
   argsToInputs: Seq[String] => Either[String, Inputs],
-  getBuildOptions: () => BuildOptions,
-  verbosity: Int,
+  bspReloadableOptionsReference: BspReloadableOptions.Reference,
   threads: BspThreads,
   in: InputStream,
   out: OutputStream
 ) extends Bsp {
+  private def reloadableOptions: BspReloadableOptions = bspReloadableOptionsReference.get
+  private def logger: Logger                          = reloadableOptions.logger
+  private def buildOptions: BuildOptions              = reloadableOptions.buildOptions
+  private def verbosity: Int                          = reloadableOptions.verbosity
 
   import BspImpl.{PreBuildData, PreBuildProject, buildTargetIdToEvent, responseError}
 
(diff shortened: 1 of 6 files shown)


modules/build/src/main/scala/scala/build/bsp/BspReloadableOptions.scala
```diff
@@ -0,0 +1,26 @@
...
+
+case class BspReloadableOptions(
+  buildOptions: BuildOptions,
+  bloopRifleConfig: BloopRifleConfig,
+  logger: Logger,
+  verbosity: Int
+)
+
+object BspReloadableOptions {
+  case class Reference(getReloaded: () => BspReloadableOptions) {
```
alexarchambault: This shouldn't be a case class, as because of the `ref` field, `reference : Reference` and `reference.copy()` are not equivalent while case class makes them equal (*edited*)

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -24,15 +23,16 @@ import scala.jdk.CollectionConverters._
...
-  verbosity: Int,
+  bspReloadableOptionsReference: BspReloadableOptions.Reference,
   threads: BspThreads,
   in: InputStream,
   out: OutputStream
 ) extends Bsp {
+  private def reloadableOptions: BspReloadableOptions = bspReloadableOptionsReference.get
+  private def logger: Logger                          = reloadableOptions.logger
+  private def buildOptions: BuildOptions              = reloadableOptions.buildOptions
+  private def verbosity: Int                          = reloadableOptions.verbosity
```
alexarchambault: Getting `logger` / `buildOptions` / `verbosity` this way trumps the point of using an `AtomicReference`. As because of these methods, some methods below might get a `logger` from some options, and then get `buildOptions` from another options value if the options get updated in the mean time.

Instead of these methods, as much as possible, the `BspReloadableOptions` instance must be gotten once, and passed around to methods that need to see a consistent options value (and the `logger`, `buildOptions`, and `verbosity`, must be obtained through this passed around instance).

modules/build/src/main/scala/scala/build/bsp/BspReloadableOptions.scala
```diff
@@ -0,0 +1,26 @@
...
+    private val ref: AtomicReference[BspReloadableOptions] = new AtomicReference(getReloaded())
```
alexarchambault: I'm not sure we need to use an `AtomicReference` here, as we don't rely on its `compareAndSet` method. A simple private var would do (reading / writing a single reference var is thread safe).

alexarchambault: (This also applies to `atomicIntelliJ: AtomicBoolean` introduced in an earlier PR. A simple private var would have worked.)

alexarchambault: These private vars can be annotated with `volatile`, just-in-case.

modules/build/src/main/scala/scala/build/bsp/BspReloadableOptions.scala
Gedochao: fixed

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
Gedochao: Got it, fixed

modules/build/src/main/scala/scala/build/bsp/BspReloadableOptions.scala
Gedochao: Huh, [TIL](https://stackoverflow.com/questions/281132/java-volatile-reference-vs-atomicreference), indeed.
You're right, fixed (including `BspServer.atomicIntelliJ` from a previous PR you mentioned).

alexarchambault (approved):

PR #888, opened 2022-04-12
Make sure that Either CPS on Scala 3 handle Lefts correctly

romanowski:


modules/build-macros/src/test/scala/scala/build/CPSTest.scala
```diff
@@ -50,6 +53,14 @@ class CPSTest extends munit.FunSuite {
     new V
   }
 
+  // When changing something in EitherCPS.scala uncomment this and make sure it does not compile
```
lwronski: Maybe, we can use scala-cli to test such cases.

lwronski (approved):

modules/build-macros/src/test/scala/scala/build/CPSTest.scala
```diff
@@ -50,6 +53,14 @@ class CPSTest extends munit.FunSuite {
...
+  // def ee3: Either[E2, V] = either {
```
alexarchambault: In Scala 2, we could use [`shapeless.test.illTyped`](https://github.com/milessabin/shapeless/blob/19a9c3a5951e4197084285bd568c2fef6717ef81/core/src/main/scala/shapeless/test/typechecking.scala#L33) to test that this doesn't compile.

modules/build-macros/src/main/scala-2.13/scala/build/EitherCps.scala
```diff
@@ -10,7 +10,7 @@ object EitherCps {
   final case class Helper[E]() {
```
alexarchambault, line 10: Maybe we could remove the `case` here? (seems it was superfluous, and that would justify the use of `new` below)

alexarchambault:
To be safer, we should check that the output in negative tests prints an error we expect, IMO.

modules/build-macros/src/test/scala/scala/build/CPSTest.scala
romanowski: The problem is that it compiles on Scala 2

alexarchambault: Oh, ok.

alexarchambault (approved):
Approved, but squash-and-merge, please :)

PR #889, opened 2022-04-12
Add "gh secret {list,create}" commands

alexarchambault:


alexarchambault:
Included in https://github.com/VirtusLab/scala-cli/pull/926.

PR #913, opened 2022-04-20
Switch to Scala 3 by default

romanowski:


build.sc
```diff
@@ -653,7 +655,7 @@ trait Cli3 extends Cli {
       mainArgs = Seq(cache.toNIO.toString, classpath),
       workingDir = os.pwd
     )
-    val cp = res.out.text
+    val cp = res.out.lines.head
```
lwronski: Are you sure lines will contain at least one element? maybe it would be better to throw a descriptive error

lwronski (commented):
I think we should also update logic in method `scalaCliScalaVersion` which resolves the scala version used to build scala-cli. 

IMO we can just add new `else if` condition, like this:
```
if (cliVersion == "nightly") Properties.versionNumberString
    else if (Version(cliVersion) <= Version("0.1.2")) Constants.defaultScala212Version
    else if (Version(cliVersion) <= Version("0.1.4")) Constants.defaultScala213Version
    else Properties.versionNumberString
```

build.sc
romanowski: I've added a comment to address your point

build.sc
```diff
@@ -134,7 +136,7 @@ object packager extends ScalaModule with Bloop.Module {
 }
 
 object `generate-reference-doc` extends SbtModule with ScalaCliScalafixModule {
-  def scalaVersion = Scala.defaultInternal
+  def scalaVersion = Scala.scala3
```
alexarchambault: Why not change the value of `defaultInternal` to `scala3`? (and change `Scala.defaultInternal` to `Scala.scala2` when needed)

build.sc
alexarchambault: Wouldn't calling `.trim` on `res.out.text` just work?

build.sc
romanowski: I wanted to do it as next step - I want to first migrate to Scala 3 and then when all is working clean up the build

alexarchambault (approved):

PR #918, opened 2022-04-20
Update directective instruction for tests

amaalali:
The test documentation has the old directive instruction. This updates it to the newer comment directive format.


romanowski:
Thank you @amaalali for this change, indeed this piece of doc use outdated syntax. I've change a target to `stable` so it gets deployed once PR is merged.

PR #926, opened 2022-04-21
Add "publish setup" command

### LINKED ISSUE #12: Add publishing to Maven repositories capabilities
coursier has nascent publishing capabilities. These could be used for that, but these need some clean-up / updates before. In particular, it doesn't support "bundle uploads" to Sonatype repositories, that makes publishing there more reliable and that sbt-sonatype uses by default.

alexarchambault:
Fixes https://github.com/VirtusLab/scala-cli/issues/12.


website/docs/commands/publishing/publish-setup.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Configuration
+
+Configuring Scala CLI for publishing makes `publish setup` works more smoothly later on.
+
+In particular, one can configure:
+- details about oneself (name, email, website - written in the developers section of POM files)
+- a PGP key pair (to sign artifacts)
+- Sonatype credentials (to upload artifacts to Maven Central)
+- a GitHub token (to upload repository secrets to GitHub, and artifacts to Maven repositories of GitHub Packages)
```
romanowski, line 30: Shouldn't we also set up github actions with release flow? (snapshot, releases) ?

alexarchambault: You mean, have `scala-cli publish setup` write a workflow file? It does that too, yeah, via the file pulled for https://github.com/VirtusLab/scala-cli/pull/1027.

modules/cli/src/main/scala/scala/cli/commands/publish/PublishSetup.scala
```diff
@@ -0,0 +1,316 @@
...
+
+      val missingSetSecrets = missingFieldsWithDefaults
+        .flatMap {
+          case (_, default) => default.ghSecrets
+        }
+        .filter(s => s.force || !secretNames.contains(s.name))
+
+      if (missingSetSecrets.nonEmpty) {
+
+        logger.message("")
```
Gedochao: ?

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -23,6 +24,68 @@ import scala.util.Properties
...
+        catch {
+          case NonFatal(e) => throw new Exception(e)
+        }
+      }
+      res
+        .left.map(_.describe)
+        .map(f => os.read.bytes(os.Path(f, Os.pwd)))
+  }
+
+  def inputs(
```
Gedochao, line 44: `SharedOptionsOps` already contains an `inputs` method, as well as a few `inputsOrExit`.
Doesn't have to be in the scope of this PR, but I think some scaladocs and/or a refactor here could prevent the future confusion on which way of constructing an `Inputs` instance is the proper one in what circumstances.

modules/integration/src/test/scala/scala/cli/integration/PublishSetupTests.scala
```diff
@@ -0,0 +1,235 @@
...
+    os.proc(TestUtil.cli, "config", dirOptions, "--create-pgp-key")
+      .call(cwd = root, stdout = os.Inherit)
+  }
+
+  private val projDir = os.rel / projName
+  private val testInputs = TestInputs(
+    Seq(
+      os.rel / projDir / "Foo.scala" ->
+        """object Foo
+          |""".stripMargin
```
Gedochao, line 42: the contents of `Foo.scala` could just be `"object Foo"`, no triple-quote and stripping margins, no?

Gedochao (approved):
Added some NIT comments, none of which should block this PR;
LGTM, although I'd double check if everything works okay after the necessary rebase

modules/cli/src/main/scala/scala/cli/commands/publish/PublishSetup.scala
alexarchambault: That's to print an empty line, for readability. I just pushed some comments for that.

modules/integration/src/test/scala/scala/cli/integration/PublishSetupTests.scala
alexarchambault: I used that to get an empty line at the end of the file, just in case.

Gedochao: ah, okay then.
in that case I'd prefer just:
```scala
"object Foo\n"
```
The alternative is slightly verbose, but I guess it comes down to preference.

alexarchambault: Yeah, but on Windows that should be `"object Foo\r\n"`, so at the end, we'd get
```scala
"object Foo" + System.lineSeparator()
```
IMO, multiline + strip margin looks slightly better.

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
alexarchambault: Good point. Just pushed a few commits cleaning that up, and adding a scaladoc to the new `inputs` method.

romanowski (approved):

PR #933, opened 2022-04-25
Ensure a main class element is findable in Scala code passed by `stdin`

### LINKED ISSUE #907: Scala CLI does not see main classes from stdin passed as `_.scala`
**Version(s)**
`0.1.3`

**Describe the bug**

When provide a valid scala main class on stdin and run scala-cli it does not see any main classes.

**To Reproduce**

```
echo "@main def a = println(1)" | scala-cli run  _.scala
```

**Expected behaviour**
The main class should be found and the code should run

Gedochao:
Fixes #907

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
@@ -41,7 +41,7 @@ case object ScalaPreprocessor extends Preprocessor {
     updatedContent: Option[String]
   )
 
-  val usingDirectiveHandlers = Seq(
+  val usingDirectiveHandlers: Seq[UsingDirectiveHandler] = Seq(
     UsingDependencyDirectiveHandler,
     UsingScalaVersionDirectiveHandler,
     UsingRepositoryDirectiveHandler,
@@ -60,7 +60,7 @@ case object ScalaPreprocessor extends Preprocessor {
     UsingPublishDirectiveHandler
   )
 
-  val requireDirectiveHandlers = Seq[RequireDirectiveHandler](
+  val requireDirectiveHandlers: Seq[RequireDirectiveHandler] = Seq(
     RequireScalaVersionDirectiveHandler,
     RequirePlatformsDirectiveHandler,
     RequireScopeDirectiveHandler
@@ -240,7 +240,7 @@ case object ScalaPreprocessor extends Preprocessor {
       // for standard imports.
       val buf = content.toCharArray
       for (t <- dependencyTrees) {
-        val substitute = (t.prefix(0) + ".A").padTo(t.end - t.start, ' ')
+        val substitute = (t.prefix.head + ".A").padTo(t.end - t.start, ' ')
         assert(substitute.length == (t.end - t.start))
         System.arraycopy(substitute.toArray, 0, buf, t.start, substitute.length)
       }
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
@@ -114,6 +114,11 @@ case object ScalaPreprocessor extends Preprocessor {
 
       case v: Inputs.VirtualScalaFile =>
         val res = either {
+          val relPath = os.sub / "stdin.scala"
+          val className = {
+            val (pkg, wrapper) = AmmUtil.pathToPackageWrapper(relPath)
+            (pkg :+ wrapper).map(_.raw).mkString(".")
+          }
           val content = new String(v.content, StandardCharsets.UTF_8)
           val (requirements, scopedRequirements, options, updatedContentOpt) =
             value(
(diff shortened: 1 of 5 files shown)


lwronski (commented):
I think, we should fix following use case and update scala-cli integration test
```
lwronski@VL-D-0317 scala-cli % scala-cli version                         
0.1.4
lwronski@VL-D-0317 scala-cli % scala-cli https://gist.github.com/lwronski/ab522c25905a50d07d0e4cb3e1fcf5d5
Hello
lwronski@VL-D-0317 scala-cli % git branch
* fix-main-class-for-stdin-scala-code
  main
lwronski@VL-D-0317 scala-cli % ./mill scala https://gist.github.com/lwronski/ab522c25905a50d07d0e4cb3e1fcf5d5
Compiling project (Scala 3.1.1, JVM)
Warning: Flag -bootclasspath set repeatedly
Warning: Flag -classpath set repeatedly
[error] ab522c25905a50d07d0e4cb3e1fcf5d5-main/Hello.scala:3:11: Not found: Messages
Error compiling project (Scala 3.1.1, JVM)
Compilation failed
1 targets failed
cli.run subprocess failed
```

Gedochao:
> I think, we should fix following use case and update scala-cli integration test
> 
> ```
> lwronski@VL-D-0317 scala-cli % scala-cli version                         
> 0.1.4
> lwronski@VL-D-0317 scala-cli % scala-cli https://gist.github.com/lwronski/ab522c25905a50d07d0e4cb3e1fcf5d5
> Hello
> lwronski@VL-D-0317 scala-cli % git branch
> * fix-main-class-for-stdin-scala-code
>   main
> lwronski@VL-D-0317 scala-cli % ./mill scala https://gist.github.com/lwronski/ab522c25905a50d07d0e4cb3e1fcf5d5
> Compiling project (Scala 3.1.1, JVM)
> Warning: Flag -bootclasspath set repeatedly
> Warning: Flag -classpath set repeatedly
> [error] ab522c25905a50d07d0e4cb3e1fcf5d5-main/Hello.scala:3:11: Not found: Messages
> Error compiling project (Scala 3.1.1, JVM)
> Compilation failed
> 1 targets failed
> cli.run subprocess failed
> ```

Should be fixed now. The issue involved wrongly passing the `relPath` appropriate for a `stdin` input to multiple files inside of a zip archive.
I added a relevant integration test.

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -60,7 +60,7 @@ case object ScalaPreprocessor extends Preprocessor {
     UsingPublishDirectiveHandler
   )
 
-  val requireDirectiveHandlers = Seq[RequireDirectiveHandler](
+  val requireDirectiveHandlers: Seq[RequireDirectiveHandler] = Seq[RequireDirectiveHandler](
```
alexarchambault: Do we really need this? There's already a `Seq[RequireDirectiveHandler]` on the same line…

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -123,15 +128,15 @@ case object ScalaPreprocessor extends Preprocessor {
...
-            0,
-            Some(options),
-            Some(requirements),
+            ignoreLen = 0,
+            options = Some(options),
+            requirements = Some(requirements),
             scopedRequirements,
-            None,
-            v.scopePath
+            mainClassOpt = Some(className),
```
alexarchambault, line 138: Why do we need this? As it's a Scala file (not a script), we shouldn't have to come up with a main class name here. If users defined one, it should be found after compilation, doesn't it? (at least, that's how it should work in theory)

Gedochao: The short answer is, if we don't provide a fallback main class name here, a `NoMainClassFoundError` gets thrown in[ `scala.build.Build`
](https://github.com/VirtusLab/scala-cli/blob/main/modules/build/src/main/scala/scala/build/Build.scala#L63)
We create [an inferred class name for Inputs.ScalaFile](https://github.com/VirtusLab/scala-cli/blob/main/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala#L75) in a similar fashion, which my intension was to stay consistent with.
In turn, the `mainClassOpt`  seems to be used as fallback for the found main class in [`scala.build.Build`](https://github.com/VirtusLab/scala-cli/blob/main/modules/build/src/main/scala/scala/build/Build.scala#L60)

It seems to me that if we don't need to set the `mainClassOpt` here (and rely on the compilation instead), then we probably shouldn't need the inferred class name for `Inputs.ScalaFile`, either?
Is passing the class name here meant solely for the Scala script wrapper?

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
Gedochao: right, no need to pass the type param twice, in the same line, my bad.
should be better now.

alexarchambault (approved):

lwronski (approved):

PR #940, opened 2022-04-25
Add pure Java projects support

alexarchambault:


alexarchambault:
(If that can help, I can maybe split the last commit, which is a bit large.)

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -897,4 +904,18 @@ abstract class BuildTests(server: Boolean) extends munit.FunSuite {
       expect(scalaCompilerJarNameOpt.contains("scala3-compiler_3-3.1.0.jar"))
     }
   }
+
+  test("Pure Java") {
+    val inputs = TestInputs(
+      os.rel / "Foo.java" ->
+        """
+          |""".stripMargin
```
Gedochao, line 918: Do we want more tests for pure Java here?
Maybe smth with actual java code would be worth it?

modules/build/src/main/scala/scala/build/compiler/SimpleScalaCompiler.scala
```diff
@@ -25,70 +25,108 @@ final case class SimpleScalaCompiler(
...
+          mainClassOpt.forall { mainClass =>
+
+            val outputDir =
+              if (isScala2 && scaladoc) project.scaladocDir
+              else project.classesDir
+
+            runScalacLike(project, mainClass, outputDir, logger)
+          }
+
+        case None =>
```
Gedochao, line 95: Thinking if it wouldn't be less confusing if we extracted the pure Java scenario code to a separate class (`SimpleJavaCompiler`?), which could be called from here

Gedochao (approved):
The other comments are just nitpicks, LGTM

modules/build/src/test/scala/scala/build/tests/BuildTests.scala
alexarchambault: Oh yep, that's an oversight.

modules/build/src/main/scala/scala/build/compiler/SimpleScalaCompiler.scala
alexarchambault: Good idea, just did that.

PR #941, opened 2022-04-26
Define valid package type for Scala.js and Scala Native

### LINKED ISSUE #915: `package --js --library` (incorrectly) requires a main class to be present.
**Version(s)**

0.1.3, 0.1.4

**Describe the bug**

`scala-cli package --js --library test.scala` produces `[error]  No main class found` if there is no main class in test.scala.
Thus, no output is generated at all.

**To Reproduce**

sample `test.scala` for the above command.
```
object Main { val x = 4 }
```

**Expected behaviour**

This works with `scala-cli package --library test.scala` (i.e., no `--js`) and js packaging also works if one provides an empty main class, thus I would expect the js library to produce working output even if there is no main class.

lwronski:
Accept valid package type for Scala.js and Scala.Native, in other case throws error. 

For example following command should generate `test.jar` file:
```
scala-cli package --js --library test.scala
```

Closed  #915

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Package.scala
@@ -19,7 +19,7 @@ import java.util.zip.{ZipEntry, ZipOutputStream}
 
 import scala.build.EitherCps.{either, value}
 import scala.build._
-import scala.build.errors.{BuildException, ScalaNativeBuildError}
+import scala.build.errors.{BuildException, MalformedCliInputError, ScalaNativeBuildError}
 import scala.build.internal.{NativeBuilderHelper, Runner, ScalaJsLinkerConfig}
 import scala.build.options.{PackageType, Platform}
 import scala.cli.CurrentParams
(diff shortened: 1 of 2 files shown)


romanowski (changes requested):
Shouldn't we add info in the documentation for each of those options that they will not work with for `js` or `native`?

Maybe we can also have more generic method of checking if given option does not conflict with other?

lwronski:
> Shouldn't we add info in the documentation for each of those options that they will not work with for `js` or `native`?
> 
> Maybe we can also have more generic method of checking if given option does not conflict with other?

I created new issue to implement some global mechanism for option checking - https://github.com/VirtusLab/scala-cli/pull/941#pullrequestreview-953232236

romanowski (approved):

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -127,23 +127,47 @@ object Package extends ScalaCommand[PackageOptions] {
...
+            throw new MalformedCliInputError(
+              s"Unsuported package type: $basePackageType for Docker."
+            )
+          PackageType.Docker
+        case (_, Platform.JS) =>
+          val validatedPackageType =
+            for (basePackageType <- basePackageTypeOpt)
+              yield
+                if (validPackageScalaJS.contains(basePackageType)) basePackageType
+                else throw new MalformedCliInputError(
```
alexarchambault, line 156: As much as possible, we shouldn't throw here, but rather try to make the exception be returned via the left side of the either. Thrown exceptions aren't reported as nicely as the ones we propagate via eithers.

Issue #591, opened 2022-01-30
Fails to compile files with non-ASCII characters in the filename.


heksesang:
**Version(s)**

0.1.0

**Describe the bug**

With a file named `Løp.scala`, `scala-cli compile` produces the following error:
`Error: java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters`

**Expected behaviour**
The file compiles just fine.

romanowski:
Thank you @heksesang for reporting, I am able to reproduce the problem. Full stack trace is below:

```
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: L��p.scala
  sun.nio.fs.UnixPath.encode(UnixPath.java:121)
  sun.nio.fs.UnixPath.<init>(UnixPath.java:68)
  sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:279)
  java.nio.file.Path.resolve(Path.java:515)
  os.Path.$div(Path.scala:456)
  scala.build.Inputs$ScalaFile.path$lzycompute(Inputs.scala:191)
  scala.build.Inputs$ScalaFile.path(Inputs.scala:191)
  scala.build.preprocessing.ScalaPreprocessor$stateMachine$async$1.apply(ScalaPreprocessor.scala:87)
  scala.build.EitherStateMachine.start(EitherCps.scala:58)
  scala.build.preprocessing.ScalaPreprocessor$.preprocess(ScalaPreprocessor.scala:86)
  scala.build.CrossSources$stateMachine$async$2.$anonfun$forInputs$2(CrossSources.scala:104)
  scala.collection.Iterator$$anon$11.nextCur(Iterator.scala:486)
  scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:492)
  scala.collection.Iterator$SliceIterator.hasNext(Iterator.scala:268)
  scala.collection.Iterator.foreach(Iterator.scala:943)
  scala.collection.Iterator.foreach$(Iterator.scala:943)
  scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
  scala.collection.generic.Growable.$plus$plus$eq(Growable.scala:62)
  scala.collection.generic.Growable.$plus$plus$eq$(Growable.scala:53)
  scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:184)
  scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:47)
[...]

PR #966, opened 2022-05-04
Add UTF-8 chars in file name test

alexarchambault:


lwronski (approved):
We can also test the case reported in #591. There was a mention of `ø` character.

alexarchambault:
> We can also test the case reported in https://github.com/VirtusLab/scala-cli/issues/591. There was a mention of `ø` character.

It can be added, but we already use the same kind of characters in the added tests already…

PR #975, opened 2022-05-05
In Doctor, account for absolute and relative paths

anatoliykmetyuk:
When checking PATH, some of the entries there are relative.
According to os-lib documentation, Path will throw
exception if you try to use it to construct a relative
path. RelPath needs to be used in this case. This happens,
e.g. when the path starts with `~` which is not handled
by the previous filter in the code.

Co-authored with @mtk during 15th Issue Spree of 3rd May 2022.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Doctor.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Doctor.scala
@@ -79,7 +79,10 @@ object Doctor extends ScalaCommand[DoctorOptions] {
       .split(pathSeparator)
       .map(d => if (d == ".") pwd else d) // on unix a bare "." counts as the current dir
       .map(_ + "/scala-cli")
-      .filter(f => os.isFile(os.Path(f)))
+      .filter { f =>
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/scala/cli/commands/Doctor.scala
```diff
@@ -79,7 +79,10 @@ object Doctor extends ScalaCommand[DoctorOptions] {
       .split(pathSeparator)
       .map(d => if (d == ".") pwd else d) // on unix a bare "." counts as the current dir
       .map(_ + "/scala-cli")
-      .filter(f => os.isFile(os.Path(f)))
+      .filter { f =>
+        val p: os.Path = if f.startsWith("/") then os.Path(f) else os.RelPath(f).resolveFrom(os.pwd)
+        os.isFile(p)
+      }
```
alexarchambault: Just passing `os.pwd` to `os.Path` should fix this:
```suggestion
      .filter(f => os.isFile(os.Path(f, os.pwd)))
```

alexarchambault (commented):
Thanks @anatoliykmetyuk. I think there's a more straightforward way of fixing this, see my comment.

lwronski (approved):

Issue #957, opened 2022-04-29
scala-cli export: Long execution time for simple export


bjornregnell:
**Version(s)**
```
$ scala-cli version
0.1.4
```

**Describe the bug**

When running `scala-cli export --sbt main.scala -o sbt-proj``it takes a looong time to get any output so you might even think that it is hanging...

**To Reproduce**

```
$ cat main.scala 
@main def run = println("hello")

$ time scala-cli export --sbt main.scala -o sbt-proj

real    0m8,961s
user    0m6,808s
sys    0m2,177s
```

**Expected behaviour**
It would be good if 1) it didn't take so long to make a simple export and 2) if there was some initial output / feedback of what is coming and going on and if it takes a long time then there should be some kind of progress information.

lwronski:
@bjornregnell Thank you for reporting, and I apologize for the long response. 

This issue could be related to #1765, as the `scala-cli` searches for the `testFramework` name in the classpath for a long time.

PR #978, opened 2022-05-05
Improve the `export` command messages

### LINKED ISSUE #958: scala-cli export: Misleading error message when no dest
**Version(s)**

```
$ scala-cli version
0.1.4
```

**Describe the bug**

When `-o dest` is omitted in `scala-cli export` the error message that comes after a looong time (see #957) is misleading, as it says dest already exists, perhaps because when omitting `-o` then scala-cli assumes current dir and that already exists...

**To Reproduce**
```
$ cat main.scala 
@main def run = println("hello")

$ time scala-cli export --sbt main.scala
Error: dest already exists.
```

**Expected behaviour**

Either 1) An error message saying that `-o` is missing or else perhaps 2) if no dest then assume a default dest like `sbt-proj` or something and report the name of the created dest.

Gedochao:
Fixes #958 
Partially addresses #957 by adding some extra messages when the `export` command is executed

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Export.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Export.scala
@@ -72,6 +72,29 @@ object Export extends ScalaCommand[ExportOptions] {
   def run(options: ExportOptions, args: RemainingArgs): Unit = {
     CurrentParams.verbosity = options.shared.logging.verbosity
     val logger = options.shared.logger
+
+    val output = options.output.getOrElse("dest")
+    val dest   = os.Path(output, os.pwd)
+    if (os.exists(dest)) {
+      System.err.println(
+        s"""Error: $dest already exists.
+           |To change the destination output directory pass --output path or remove the destination directory first.""".stripMargin
+      )
+      sys.exit(1)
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/scala/cli/commands/Export.scala
```diff
@@ -72,12 +72,34 @@ object Export extends ScalaCommand[ExportOptions] {
...
+           |To change the destination output directory pass --output path or remove the destination directory first.""".stripMargin
+      )
+      sys.exit(1)
+    }
+
+    val shouldExportToMill = options.mill.getOrElse(false)
+    val shouldExportToSbt  = options.sbt.getOrElse(false)
+    if (shouldExportToMill && shouldExportToSbt) {
+      System.err.println(
+        s"Error: Cannot export to both mill and sbt. please pick one build tool to export."
```
lwronski:
```suggestion
        s"Error: Cannot export to both mill and sbt. Please pick one build tool to export."
```

lwronski (approved):

alexarchambault (approved):

PR #982, opened 2022-05-05
Don't download test bridges and supported Scala version listing for pure Java projects

alexarchambault:


modules/build/src/test/scala/scala/build/tests/BuildTests.scala
```diff
@@ -917,11 +918,32 @@ abstract class BuildTests(server: Boolean) extends munit.FunSuite {
...
+        ignoreSupportedScalaVersionsErrors = Some(false)
+      )
+    )
+    val gotExpectedError =
+      try {
+        options.scalaParams
+        false
+      }
+      catch {
+        case e if e.getCause.isInstanceOf[ArtifactError.DownloadError] =>
```
lwronski, line 939: If we want to avoid downloading supported Scala Version, why not change type `ScalaOptions.scalaVersionsUrl` to `Option[String]`. And download supportedScalaVersion only when variable `supportedScalaVersionsUrl ` have some value.

Instead of catching a download error, we just might not download the supported Scala Versions.

lwronski, line 934: If we want to avoid downloading supported Scala Version, why not change type `ScalaOptions.scalaVersionsUrl` to `Option[String]`. And download supportedScalaVersion only when variable `supportedScalaVersionsUrl ` have some value.

Instead of catching a download error, we just might not download the supported Scala Versions.

alexarchambault: I don't think that would work for this PR. Here, I want the supported Scala version check to fail loudly, so that the test just below those lines only passes if it doesn't rely at all on supported Scala version stuff.

We can't make `ScalaOptions.scalaVersionsUrl` have a default value like `Some("https://…")`, as that would make some `BuildOptions` calculations inconsistent (like `myOptions orElse BuildOptions()` would erase a `ScalaOptions.scalaVersionsUrl` set to `None` in `myOptions`, while we want that to be equal to just `myOptions` - this "breaks the monoid laws"). So we can't rely on `ScalaOptions.scalaVersionsUrl` to be `None` here.

We could set it to `Some("")` to disable downloading a listing, but then the most sensible thing to do would be to rely on the [hardcoded version list](https://github.com/VirtusLab/scala-cli/blob/673239d8c29d7152fca675061190bd05b010a991/modules/options/src/main/scala/scala/build/options/BuildOptions.scala#L263) like we do already when downloading the listing fails.

So a separate field for this is better I think.

Gedochao (approved):

alexarchambault:
(last force push: rebase on latest main + add comment in test in last commit)

lwronski (approved):

PR #994, opened 2022-05-09
Add "gh secret {list,create}" commands

alexarchambault:
(Part of https://github.com/VirtusLab/scala-cli/pull/926, pushed here for easier review.)


lwronski (approved):
I think, that documentation must be also created, but it can be included to https://github.com/VirtusLab/scala-cli/pull/926

PR #999, opened 2022-05-10
Allow forks to build their own launchers

alexarchambault:
This allows forks (like [mine](https://github.com/alexarchambault/scala-cli)) to build launchers and push them to a GitHub release named `nightly` of their own (like [here](https://github.com/alexarchambault/scala-cli/releases/tag/nightly)). This works automatically by having the fork just enable GitHub actions for the `main` branch, if they create a release with a tag named `nightly` (the actual commit of this release is unused and doesn't matter).

I'm planning to use this to automatically build launchers of https://github.com/VirtusLab/scala-cli/pull/926 until it's merged, but also for ongoing / upcoming developments on top of this. Having these launchers there allows to use them straightaway on other projects' CI, to test newer `publish` command features before they're merged here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,7 @@ jobs:
       run: ./mill -i nativeIntegrationTests
       env:
         COURSIER_JNI: force
+        UPDATE_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   native-mostly-static-tests:
     timeout-minutes: 120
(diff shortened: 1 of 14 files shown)


lwronski (commented):
IMO we should mention it feature in [internal section](https://scala-cli.virtuslab.org/docs/guides/internals) in documentation.

lwronski:
In the future, we could extend the `LauncherCLI` functionality so that it can run the `scala-cli` launcher from forks.

alexarchambault:
> IMO we should mention it feature in [internal section](https://scala-cli.virtuslab.org/docs/guides/internals) in documentation.

Not sure it should go there, that section is about how Scala CLI itself works, not about CI-related stuff.

It could be mentioned somewhere in the documentation, but it's quite an arcane feature… Not sure many people care about it.

lwronski (approved):

PR #1002, opened 2022-05-10
Fix bloop output command, and suggest it when it seems the Bloop server didn't start

alexarchambault:


lwronski (approved):
For me, there is missing some mention in documentation or `Readme.md`, that we have `bloop output` command to help debug bloop staff.

PR #1003, opened 2022-05-10
Allow to publish to Maven Central in publish command

alexarchambault:
Another batch of commits from https://github.com/VirtusLab/scala-cli/pull/926.


lwronski (approved):
IMO we need some unit tests to avoid regression in the future.

alexarchambault:
> IMO we need some unit tests to avoid regression in the future.

I wish too (unit or integration tests). But it seems we'd have to "mock" a Sonatype server instance, or use a docker-based one (which needs to be configured and all), and try make it look like the official Sonatype instances.

PR #865, opened 2022-04-07
Add workaround for ArchLinux + native-image CRC32 miscalculation issue

alexarchambault:
Fixes https://github.com/VirtusLab/scala-cli/issues/828.

~This vendors some GPL code, so I don't know if we're fine license-wise, here…~

Users need to set `SCALA_CLI_VENDORED_ZIS=true` in the environment to disable CRC32 calculations.

PR #1007, opened 2022-05-11
Suggest zip issue workaround when apparently running into it

alexarchambault:
Follow-up of https://github.com/VirtusLab/scala-cli/pull/865. This suggests users to set the right environment variable to disable checking zip CRC32 checksums.


lwronski (commented):
I think, in this kind of unobvious check we should add comment why there is exists case or at least add link to GitHub issue to see context of this exception.

lwronski (approved):

PR #1016, opened 2022-05-17
Add interactive mode to choose main method from multiple options

lwronski:


modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.build.Logger
+import scala.build.errors.BuildException
+import scala.build.options.BuildOptions
+import scala.io.StdIn.readLine
+
+case class Interactive(options: BuildOptions, logger: Logger) {
+
+  private val isInteractive = options.internal.interactive.getOrElse(false)
+
+  def chooseOneOptionsFromList[E <: BuildException](
```
alexarchambault: Typo
```suggestion
  def chooseOneOptionFromList[E <: BuildException](
```
but this could even be
```suggestion
  def chooseOne[E <: BuildException](
```

modules/cli-options/src/main/scala/scala/cli/commands/SharedOptions.scala
```diff
@@ -96,7 +96,10 @@ final case class SharedOptions(
   helpGroups: HelpGroupOptions = HelpGroupOptions(),
 
   @Hidden
-    strictBloopJsonCheck: Option[Boolean] = None
+    strictBloopJsonCheck: Option[Boolean] = None,
+  @HelpMessage("Interactive mode")
+  @Name("-i")
```
alexarchambault: No need of `-` here
```suggestion
  @Name("i")
```

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,48 @@
...
+      logger.message(msg)
+      options.zipWithIndex.foreach {
+        case (option, index) => logger.message(s"[$index] $option")
+      }
+      val response      = readLine()
+      val inputIndexOpt = parseIndexInput(response, options.length)
+      val chosedOption = for (inputIndex <- inputIndexOpt)
+        yield Right(options(inputIndex))
+
+      chosedOption.getOrElse(Left(fallbackError))
```
alexarchambault: NIT I think this could be
```suggestion
      inputIndexOpt.map(options(_)).toRight(fallbackError)
```

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,48 @@
...
+  def chooseOne[E <: BuildException](
+    msg: String,
+    options: List[String],
+    fallbackError: E
+  ): Either[E, String] =
+    if (isInteractive && coursier.paths.Util.useAnsiOutput()) {
+      logger.message(msg)
+      options.zipWithIndex.foreach {
+        case (option, index) => logger.message(s"[$index] $option")
+      }
```
alexarchambault: We should probably print directly to `System.err` here… Because we read directly from stdin anyway (via `readLine`), and not printing anything before waiting for input would be a problem.

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,48 @@
...
+        val isInRange = index <= range && index >= 0
+        if (isInRange) Some(index)
+        else {
+          logger.message(
+            s"The input index number is invalid, integer value from 0 to $range is expected."
+          )
+          None
+        }
+      case _ =>
+        logger.message(s"Unable to parse input: integer value from 0 to $range is expected.")
```
alexarchambault: Same here, we should probably print directly to `System.err`

alexarchambault (approved):

PR #1027, opened 2022-05-20
Add "default" command

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
```diff
@@ -0,0 +1,104 @@
...
+    case (_, d) =>
+      // d.path.toString and d.printablePath differ on Windows (one uses '/', the other '\')
+      Seq(
+        d.path.toString -> d,
+        d.printablePath -> d
+      )
+  }
+
+  private def unrecognizedFile(name: String): Nothing = {
+    System.err.println(
```
Gedochao: hmm... should we prefer using a logger rather than just stderr/stdout?
(it's been discussed, but I don't think we've agreed on a convention, really, just re-raising it here)

website/docs/commands/misc/default.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+Get the content of a default file with
+```text
+$ scala-cli default .gitignore
+/.bsp/
+/.scala-build/
+```
+
+Optionally, write the content of one or more default files by passing `--write`:
```
Gedochao: I'm thinking if it wouldn't be worth it to also enable writing all of the available default files when `--write` is used without inputs.
I mean, like:
```bash
scala-cli default --write
```
could possibly write both `.gitignore` and `.github/workflows/ci.yml` (as well as anything else we might decide to add in the future)

build.sc
```diff
@@ -621,6 +623,25 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
   }
   def generatedSources = super.generatedSources() ++ Seq(constantsFile())
 
+  def defaultFilesResources = T.persistent {
+    val dir = T.dest / "resources"
+    val resources = Seq(
+      "https://raw.githubusercontent.com/scala-cli/default-workflow/main/.github/workflows/ci.yml" -> (os.sub / "workflows" / "default.yml"),
```
lwronski, line 660: I see that definition of workflows contains:
```
- uses: coursier/setup-action@v1.2.0-M3
      with:
        jvm: temurin:17
```
which install scala-cli, so do we still need the `scala-cli-setup`? maybe Scala CLI should only promote `coursier/setup-action@v1.2.0-M3` to install `scala-cli`.

alexarchambault: I just changed that. I used `coursier/setup-action@v1.2.0-M3` out of habit, but scala-cli-setup is better here.

modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
alexarchambault: I was unsure about using a logger here, as this wasn't meant to print much… I changed that, but I added a new `Logger.error` method, for some messages here (that should be printed even when `-q` is passed).

website/docs/commands/misc/default.md
alexarchambault: Maybe we could have a `--all` option for that. An also optionally a `--missing` one, to write only missing files. But these could be added later I think (maybe when we'll have more default files, or if we need different shades of workflow files - for Scala.js, Scala Native, …).

Gedochao: Yeah, can be done separately later as well. 
Might still want to add a ticket for it, so we don't forget.

Gedochao (approved):

PR #1032, opened 2022-05-23
Use closest previous version when looking for stable Scala versions

alexarchambault:
This should allow not to have to update https://github.com/VirtusLab/scala-cli-scala-versions/blob/main/scala-versions-v1.json upon each release. This file only needs to be updated when new Scala versions are released (and known to be supported by Scala CLI).

### DIFF
--- a/.github/release/release-procedure.md
+++ b/.github/release/release-procedure.md
@@ -8,7 +8,6 @@
   ```bash
   git shortlog -sn --no-merges {old-release-tag}...{new-release-tag}
   ```
-- [ ] Create a pull request with supported scala-versions for the new release to this repository: [Virtuslab/scala-cli-scala-versions ](https://github.com/Virtuslab/scala-cli-scala-versions)
 - [ ] ScalaCLI Setup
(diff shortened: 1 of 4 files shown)


lwronski (commented):
For me, only the unit test is missing, the rest LGTM

a unit test might look like this:
```
val confFile = s"""[
                    | {
                    |  "scalaCliVersion": "0.1.0",
                    |  "supportedScalaVersions": ["3.0.1", "2.13.4", "2.12.13"]
                    | }
                    |]""".stripMargin

 test(s"Use closest previous version when looking for stable Scala versions") {
    TestInputs.withTmpDir("conf-scala-versions") { dirPath =>

      val confFilePath = dirPath / "conf-file.json"
      os.write(confFilePath, confFile)

      val options = BuildOptions(
        scalaOptions = ScalaOptions(
          scalaVersion = Some(MaybeScalaVersion("3")),
          scalaBinaryVersion = None,
          supportedScalaVersionsUrl = Some(s"file://${confFile2.toString()}")
        )
      )

      val scalaParams          = options.scalaParams.orThrow.getOrElse(???)
      val expectedScalaVersion = "3.0.1"
      val expectedScalaParams  = ScalaParameters(expectedScalaVersion)

      expect(scalaParams == expectedScalaParams)
    }
  }
```

lwronski (approved):

PR #1034, opened 2022-05-23
Ensure the latest Scala / Scala.js versions are used in version listings in doc

alexarchambault:


lwronski (commented):
It makes sense for me, good approach.

IMO we should also update documentation for `scala-js.md` and `scala-version.md`, because now there is the hardcoded range of Scala CLI versions like this:
```| 0.1.4 - 0.1.6 | 1.10.0 | ```

Documentation should show that ScalaCLI supports versions of Scala.js (`1.10.0`) from version `0.1.4` to the latest.

alexarchambault:
> IMO we should also update documentation for scala-js.md and scala-version.md, because now there is the hardcoded range of Scala CLI versions like this:

Oh right, I forgot.

lwronski (approved):

PR #1040, opened 2022-05-24
Add more interactive options

lwronski:


modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -1,44 +1,79 @@
...
+
+  def confirmOperation[E](
+    msg: String,
+    fallbackError: () => E
+  ): Either[E, Boolean] = ConfirmOperation(msg).run.toRight(fallbackError())
+
+  def chooseOne[E](
+    msg: String,
+    options: List[String],
+    fallbackError: E
```
alexarchambault: Maybe this could be by-name, just like `confirmOperation` above. That way, this could replace the override just below: users would just have to call `.merge` on the resulting `Either[Nothing, String]` to recover a `String`.

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -1,44 +1,79 @@
 package scala.build.interactive
-
-import scala.build.errors.BuildException
 import scala.build.options.BuildOptions
 import scala.io.StdIn.readLine
 
 case class Interactive(options: BuildOptions) {
```
alexarchambault: Should other values be read from `options` in the future? If not, maybe we could have a `trait Interactive`, without most implems, and a `InteractiveAsk` say, used when `options.internal.interactive.getOrElse(false)` is true, along with a `InteractiveNop`, whose methods would always return `None`. And we could create the `Interactive` instance in one place, and pass it around.

alexarchambault (commented):
Maybe most `fallbackError` params could be passed in a second parameter group, so that we could do things like
```scala
InteractiveFileOps.addEntryToFile(buildOptions, msg, scalafmtConfPath, entry) { () =>
  System.err.println(errorMsg)
  sys.exit(1)
}
```
rather than having callers defining `fallbackError` before they call `Interactive` methods.

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
alexarchambault: Actually, it seems we don't even need to pass the fallback error: these methods can return `Option[…]`, and callers would then call `.getOrElse { … }` on it.

modules/build/src/main/scala/scala/build/interactive/Interactive.scala
lwronski: I create `Interactive` instance in `BuildOptions`, but is some cases such as Update there is no available `BuildOptions `instance, so I created `Interactive` in `VerbosityOptionsUtil`.

modules/cli/src/main/scala/scala/cli/commands/util/VerbosityOptionsUtil.scala
```diff
@@ -0,0 +1,12 @@
+package scala.cli.commands
+package util
+
+object VerbosityOptionsUtil {
+  implicit class VerbosityOptionsOps(v: VerbosityOptions) {
+    import v._
+
+    val Interactive = if (interactive.getOrElse(false))
```
alexarchambault: No need to use upper case here. By not importing `v._`, it should be possible to keep `interactive` as a name I think:
```suggestion
    val interactive = if (v.interactive.getOrElse(false))
```

modules/options/src/main/scala/scala/build/interactive/InteractiveFileOps.scala
```diff
@@ -0,0 +1,32 @@
+package scala.build.interactive
+
+case object InteractiveFileOps {
```
alexarchambault: This doesn't need to be a `case object`. This is not really "data", so a simple `object` does.

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -553,6 +553,10 @@ final case class BuildOptions(
     BuildOptions.monoid.orElse(this, other)
 
   def validate: Seq[Diagnostic] = BuildOptionsRule.validateAll(this)
+
+  val Interactive = if (internal.interactive.getOrElse(false))
```
alexarchambault: Just like above, this can be kept as lower case.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -63,18 +62,17 @@ object Build {
...
+          options.Interactive.chooseOne(
+            "Found several main classes. Which would you like to run?",
+            foundMainClasses0.toList
+          ) match {
+            case Some(mainClass) => Right(mainClass)
+            case None => Left(new SeveralMainClassesFoundError(
+                ::(foundMainClasses0.head, foundMainClasses0.tail.toList),
+                Nil
+              ))
+          }
```
alexarchambault: `toRight` should allow to simplify this I think:
```suggestion
          ).toRight {
            new SeveralMainClassesFoundError(
              ::(foundMainClasses0.head, foundMainClasses0.tail.toList),
              Nil
            )
          }
```

modules/options/src/main/scala/scala/build/interactive/InteractiveFileOps.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object InteractiveFileOps {
+
+  def erasingPath(
+    interactive: Interactive,
+    printableDest: String,
+    destPath: os.Path
+  )(fallbackAction: () => Unit) = {
+    val msg = s"""|Destination path: $printableDest already exists.
+                  |Do you want to erasing $printableDest?""".stripMargin
```
alexarchambault:
```suggestion
                  |Do you want to erase $printableDest?""".stripMargin
```

modules/options/src/main/scala/scala/build/interactive/InteractiveFileOps.scala
```diff
@@ -0,0 +1,32 @@
+package scala.build.interactive
+
+object InteractiveFileOps {
+
+  def erasingPath(
+    interactive: Interactive,
+    printableDest: String,
+    destPath: os.Path
+  )(fallbackAction: () => Unit) = {
+    val msg = s"""|Destination path: $printableDest already exists.
```
alexarchambault: I'd phrase it this way:
```suggestion
    val msg = s"""|Destination $printableDest already exists.
```

alexarchambault: Or maybe even
```suggestion
    val msg = s"""|$printableDest already exists.
```

modules/options/src/main/scala/scala/build/interactive/InteractiveFileOps.scala
```diff
@@ -0,0 +1,32 @@
...
+                  |Do you want to erasing $printableDest?""".stripMargin
+    val response = interactive.confirmOperation(msg)
+    response match {
+      case Some(true) => os.remove.all(destPath)
+      case _          => fallbackAction()
+    }
+  }
+
+  def addEntryToFile(
```
alexarchambault: `appendToFile` sounds clearer to me, given what this does. 
```suggestion
  def appendToFile(
```

modules/cli/src/main/scala/scala/cli/commands/Update.scala
```diff
@@ -59,19 +59,15 @@ object Update extends ScalaCommand[UpdateOptions] {
...
-        }
-      }
-      else {
+    val interactive = options.verbosity.interactiveInstance
+    if (!options.force) {
+      val fallbackAction = () => {
         System.err.println(s"To update scala-cli to $newVersion pass -f or --force")
         sys.exit(1)
       }
+      val msg = s"Do you want to update scala-cli to version $newVersion"
```
Gedochao:
```suggestion
      val msg = s"Do you want to update scala-cli to version $newVersion?"
```

modules/options/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,77 @@
...
+      def action: Option[V]
+      final def run: Option[V] =
+        if (coursier.paths.Util.useAnsiOutput())
+          action
+        else None
+    }
+
+    case class ConfirmOperation(msg: String) extends Action[Boolean] {
+      override def action: Option[Boolean] = {
+        System.err.println(s"$msg [Y/n]")
```
Gedochao, line 27: Should we favor using a logger here, rather than just stderr?

Gedochao: nvm, I see [why not](https://github.com/VirtusLab/scala-cli/pull/1016#discussion_r875133009)

Gedochao (approved):

modules/cli/src/main/scala/scala/cli/commands/InstallCompletions.scala
```diff
@@ -43,12 +44,19 @@ object InstallCompletions extends ScalaCommand[InstallCompletionsOptions] {
...
+          )
+          System.err.println(s"  $name install completions --shell zsh")
+          System.err.println(s"  $name install completions --shell bash")
+          sys.exit(1)
+        }
+        val msg = "Cannot determine current shell. Which would you like to use?"
+        interactive.chooseOne(
+          msg,
+          List("zsh", "bash")
+        ).getOrElse(fallbackAction())
```
alexarchambault: NIT Isn't that clearer?
```suggestion
        val msg = "Cannot determine current shell. Which would you like to use?"
        interactive.chooseOne(msg, List("zsh", "bash")).getOrElse {
          System.err.println(
            "Cannot determine current shell, pass the shell you use with --shell, like"
          )
          System.err.println(s"  $name install completions --shell zsh")
          System.err.println(s"  $name install completions --shell bash")
          sys.exit(1)
        }
```

modules/options/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -0,0 +1,77 @@
...
+              None
+            }
+          case _ =>
+            System.err.println(
+              s"Unable to parse input: integer value from 0 to $range is expected."
+            )
+            None
+        }
+      }
+    }
```
alexarchambault: Couldn't these classes be private?

alexarchambault (approved):

PR #1050, opened 2022-05-25
Setup-ide use symlink if target is the same as current launcher path

### LINKED ISSUE #1048: `setup-ide` command shouldn't hardcode version in the path when scala-cli is installed with homebrew
**Version(s)**

`0.1.4` and newer

**Describe the bug**

Whenever I run `scala-cli setup-ide .` it creates `.bsp/scala-cli.json` file containing path with hardcoded version, eg.  `/usr/local/Cellar/scala-cli/0.1.5/bin/scala-cli`. Later, when homebrew updates scala-cli it removes entire `0.1.5` directory. That in order, results in metals throwing `No such file or directory` exception and generally not working.

**To Reproduce**
- Install scala-cli with homebrew.
- Run `scala-cli setup-ide .`

**Expected behaviour**
The path should point to the symlink (`/usr/local/bin/scala-cli` in my case)

lwronski:


modules/cli/src/main/scala/scala/cli/commands/CommandUtils.scala
```diff
@@ -25,6 +26,14 @@ object CommandUtils {
        */
       sys.props.get("coursier.mainJar")
         .map(Paths.get(_).toAbsolutePath.toString)
+        .orElse {
+          /*
+          To get the absolute path we secondly try to find scala-cli on PATH and used the first position.
+          It helps to find a symlink that can point to a directory that can be changed after updating scala-cli.
+          https://github.com/VirtusLab/scala-cli/issues/1048
+           */
+          ProcUtil.findApplicationPathsOnPATH(appName = "scala-cli").headOption
```
alexarchambault: I don't think we should always use the `scala-cli` found in `PATH` like this. The current launcher might be different than the one in `PATH`, and we would want to keep it rather than using the one in `PATH`.

I think we should rather look for the one in PATH, if it's a sym link and look at its target (recursively until it's not a sym link any more). Then we should compare that to the path of the current launcher, given by the `getProtectionDomain.getCodeSource.getLocation` call below. If both paths match, then it means we are the `scala-cli` from `PATH`, and we should use the path that is in `PATH`, not the sym link target.

We're resolving sym links in reverse somehow.

If the first `scala-cli` in `PATH` is not a sym link, we should keep doing what we were doing before.

alexarchambault: About the app name passed to `findApplicationPathsOnPATH`, we should pass `ScalaCli.progName` (just in case the launcher has a slightly different name, like `scala-cli-1.0.6` say), if it's not a relative or absolute path itself. In the latter case, we should keep doing what we were doing before too.

alexarchambault (approved):

PR #1052, opened 2022-05-26
Support for printing help from scala compiler without passing any sources

### LINKED ISSUE #753: Support for printing help from scala compiler

Scala CLI should print help message from scala compiler like `scalac -help`. It can be added new flag to print help from compiler.

Gedochao:
Fixes #753


modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
```diff
@@ -17,7 +17,6 @@ final case class ScalacOptions(
     scalacOption: List[String] = Nil
 )
 // format: on
-
```
alexarchambault: NIT Why remove this line?

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
```diff
@@ -33,29 +32,34 @@ object ScalacOptions {
...
         args: List[String],
         index: Int,
         acc: Option[List[String]],
         formatter: Formatter[Name]
-      ) =
+      ): Either[(Error, List[String]), Option[(Option[List[String]], List[String])]] =
         args match {
-          case h :: t
-              if scalacOptionsPrefixes.exists(h.startsWith) &&
-              !scalacOptionsPurePrefixes.contains(h) =>
```
alexarchambault, line 54: This shouldn't be removed. This prevents us from using `-V` / `-W` / `-X` / `-Y` ourselves (like `-X` as an alias for `--cross`)

modules/cli/src/main/scala/scala/cli/commands/ScalacLikeCommand.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.cli.commands.util.SharedOptionsUtil._
+
+/** Trait to mixin for commands which are intended to mimic `scalac` behaviour, allowing for
+  * printing scalac help & other options, even without passing input sources. Meant as an extension
+  * of [[scala.cli.commands.ScalaCommand]]
+  *
+  * @tparam T
+  *   command options type
+  */
+trait ScalacLikeCommand[T] { self: ScalaCommand[T] =>
```
alexarchambault: Instead of creating this trait, couldn't `maybePrintSimpleScalacOutput` be added directly to `ScalaCommand`, and have a signature like
```scala
def maybePrintSimpleScalacOutput(buildOptions: BuildOptions): Unit
```

modules/cli/src/main/scala/scala/cli/commands/ScalacLikeCommand.scala
```diff
@@ -0,0 +1,52 @@
...
+  def maybePrintSimpleScalacOutput(options: T): Unit =
+    for {
+      shared <- sharedOptions(options)
+      build         = buildOptions(options)
+      scalacOptions = shared.scalac.scalacOption.toSeq
+      updatedScalacOptions =
+        if (shared.scalacHelp && !scalacOptions.contains("-help"))
+          scalacOptions.appended("-help")
+        else scalacOptions
+      if (updatedScalacOptions intersect ScalacOptions.ScalacPrintOptions.toSeq).nonEmpty
```
alexarchambault:
```suggestion
      if updatedScalacOptions.exists(ScalacOptions.ScalacPrintOptions)
```
should be more efficient (iterates once on `updatedScalacOptions` and stops as soon as it finds a print option, rather than computing an intersection, which is more costly)

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
```diff
@@ -33,29 +32,34 @@ object ScalacOptions {
...
+    * message instead.
+    */
+  val ScalacPrintOptions: Set[String] =
+    scalacOptionsPurePrefixes ++ Set("-help", "-Xshow-phases", "-Vphases")
+
   private val scalacOptionsArgument: Argument[List[String]] =
     new Argument[List[String]] {
 
-      val underlying = StandardArgument[List[String]](scalacOptionsArg)
+      val underlying: StandardArgument[List[String]] = StandardArgument(scalacOptionsArg)
```
alexarchambault, line 46: Do we really need to add that type ascription? The type was pretty clear (it appears explicitly) in the former code.

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
```diff
@@ -33,29 +32,34 @@ object ScalacOptions {
     Set("-V", "-W", "-X", "-Y")
   private val scalacOptionsPrefixes =
     Set("-g", "-language", "-opt", "-P", "-target") ++ scalacOptionsPurePrefixes
+
+  /** This includes all the scalac options which disregard inputs and print a help and/or context
+    * message instead.
+    */
+  val ScalacPrintOptions: Set[String] =
+    scalacOptionsPurePrefixes ++ Set("-help", "-Xshow-phases", "-Vphases")
```
alexarchambault, line 40: Why is `scalacOptionsPurePrefixes` here?

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1698,4 +1699,51 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+      expect(res1.out.text().contains("scalac <options> <source files>"))
+
+      val res2 = os.proc(
+        TestUtil.cli,
+        extraOptions,
+        "--scalac-option",
+        "-help"
+      )
+        .call(cwd = root, mergeErrIntoOut = true)
+      expect(res2.exitCode == 0)
```
alexarchambault: No need to check this (`.call` would fail if that wasn't the case)

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1698,4 +1699,51 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+          else Seq.empty
+        )
+      printOptionsToTest.foreach { printOption =>
+        val res = os.proc(
+          TestUtil.cli,
+          extraOptions,
+          printOption
+        )
+          .call(cwd = root, mergeErrIntoOut = true)
+        expect(res.exitCode == 0)
```
alexarchambault: Same as above (unnecessary)

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
alexarchambault: Or we can keep it, and remove the current use of `-X` in the code base (look for `"X"`)

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
Gedochao: image

We don't really need to, but IntelliJ warns you about it by default.
IMO the codebase is more welcoming for new contributors if we keep the yellow highlights in their IDEs to the bare minimum.

alexarchambault: That warning looks spurious to me, that field isn't public (it's an anonymous class, so it doesn't appear in the public API of this module).

Gedochao: I don't think IntelliJ is smart enough to notice the difference 🧠
Still, doesn't hurt us to keep the type explicit here, I think

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
Gedochao: ended up removing the current use of `-X`

modules/cli-options/src/main/scala/scala/cli/commands/ScalacOptions.scala
Gedochao: accidental, I reversed this change

alexarchambault (approved):

PR #1053, opened 2022-05-30
Add more docs for piping sources

Gedochao:


lwronski (approved):

website/docs/commands/basics.md
```diff
@@ -174,12 +174,23 @@ scala-cli hello.zip
 
 ## Piping
 
-You can also pipe Scala code to `scala-cli` for execution:
-
-```bash
-echo 'println("Hello")' | scala-cli -
-# Hello
-```
+You can also pipe code to `scala-cli` for execution:
```
romanowski, line 177: `echo 'println("Hello")' | scala-cli -` also works so we should include it in the scripts section

website/docs/commands/basics.md
```diff
@@ -174,12 +174,23 @@ scala-cli hello.zip
...
+- scripts
+  ```bash
+  echo 'println("Hello")' | scala-cli _.sc
+  # Hello
+  ```
+- Scala code
+  ```bash
+  echo "@main def hello() = println(\"Hello\")" | scala-cli _.scala
```
romanowski: We can mention that `echo "@main def hello() = println(\"Hello\")" | scala-cli _`

romanowski (commented):
I've requested few clarifications, beside looks good

website/docs/commands/basics.md
Gedochao: there's actually 3 different aliases for the scripts wildcard (`-`, `_.sc` & `-.sc`), all of which are mentioned in the `Piping` guide. IMO it's enough to just give one example per section here.
I favored the underscore aliases for all 3 sections in `Basics` for consistency.

website/docs/commands/basics.md
Gedochao: same as with the scripts wildcard, there's also 3 aliases for the Scala code wildcard (`_`,`_.scala` and `-.scala`). I think such details should be left for the separate guide linked below.

alexarchambault: Single quotes would make that simpler, wouldn't it?
```suggestion
  echo '@main def hello() = println("Hello")' | scala-cli _.scala
```

website/docs/commands/basics.md
```diff
@@ -174,12 +174,23 @@ scala-cli hello.zip
...
+  # Hello
+  ```
+- Java code
+  ```bash
+  echo "class Hello { public static void main(String args[]) { System.out.println(\"Hello\"); } }" | scala-cli _.java
```
alexarchambault: Same as above, single quotes makes that more readable IMO
```suggestion
  echo 'class Hello { public static void main(String args[]) { System.out.println("Hello"); } }' | scala-cli _.java
```

website/docs/guides/piping.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Piping
+sidebar_position: 23
+---
+
+# Piping
+
+Instead of passing paths to your sources, you can also pipe your code via standard input:
+```bash
+echo "@main def hello() = println(\"Hello\")" | scala-cli _
```
alexarchambault: We could use single quotes here too.

website/docs/guides/piping.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Examples
+- scripts
+  ```bash
+  echo 'println("Hello")' | scala-cli _.sc
+  # Hello
+  ```
+- Scala code
+  ```bash
+  echo "@main def hello() = println(\"Hello\")" | scala-cli _.scala
```
alexarchambault: Same as above.

website/docs/guides/piping.md
```diff
@@ -0,0 +1,45 @@
...
+  # Hello
+  ```
+- Java code
+  ```bash
+  echo "class Hello { public static void main(String args[]) { System.out.println(\"Hello\"); } }" | scala-cli _.java
```
alexarchambault: Same as above.

website/docs/guides/piping.md
```diff
@@ -0,0 +1,45 @@
...
+  # Hello
+  ```
+
+## Mixing piped sources with on-disk ones
+It is also possible to pipe some code via standard input, while the rest of your code is on-disk.
+```bash
+echo "case class HelloMessage(msg: String)" > HelloMessage.scala
+echo "@main def hello() = println(HelloMessage(msg = \"Hello\").msg)" | scala-cli _ HelloMessage.scala
```
alexarchambault: Same as above.

alexarchambault (approved):

PR #1055, opened 2022-05-30
Add dependency update command

lwronski:
The actionable diagnostics logging is disable, because not there is no exists way to disable hints in ScalaCLI. 
After #1056 will be merged, then I will add entry to disable hints in ScalaCLI
```
hints.dependency-updates
```
so now, it os only possible to update all dependency in project running
```
scala-cli dependency-update --all Hello.scala
```

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
@@ -247,8 +247,8 @@ case object ScalaPreprocessor extends Preprocessor {
       val toFilePos = Position.Raw.filePos(path, content)
       val deps = value {
         dependencyTrees
-          .map { t => // add 6 to start position to skip ivy syntax ($ivy.`) and dep ($dep.`)
-            val pos      = toFilePos(Position.Raw(t.start + 6, t.end))
+          .map { t => /// skip ivy ($ivy.`) or dep syntax ($dep.`)
+            val pos      = toFilePos(Position.Raw(t.start + "$ivy.`".length, t.end))
             val strDep   = t.prefix.drop(1).mkString(".")
             val maybeDep = parseDependency(strDep, pos)
             maybeDep.map(dep => Positioned(Seq(pos), dep))
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
@@ -247,8 +247,8 @@ case object ScalaPreprocessor extends Preprocessor {
       val toFilePos = Position.Raw.filePos(path, content)
       val deps = value {
         dependencyTrees
-          .map { t =>
-            val pos      = toFilePos(Position.Raw(t.start, t.end))
+          .map { t => // add 6 to start position to skip ivy syntax ($ivy.`) and dep ($dep.`)
+            val pos      = toFilePos(Position.Raw(t.start + 6, t.end))
             val strDep   = t.prefix.drop(1).mkString(".")
             val maybeDep = parseDependency(strDep, pos)
             maybeDep.map(dep => Positioned(Seq(pos), dep))
--- a/modules/cli-options/src/main/scala/scala/cli/commands/DependencyUpdateOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/DependencyUpdateOptions.scala
@@ -1,0 +1,20 @@
+package scala.cli.commands
+
+import caseapp._
+import caseapp.core.help.Help
+
+// format: off
+@HelpMessage("Update dependencies in project")
+final case class DependencyUpdateOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
+  @Group("DependencyUpdate")
+  @HelpMessage("Update all dependency")
+    all: Boolean = false,
+)
+  // format: on
+
+object DependencyUpdateOptions {
+  implicit lazy val parser: Parser[DependencyUpdateOptions] = Parser.derive
+  implicit lazy val help: Help[DependencyUpdateOptions]     = Help.derive
+}
--- a/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
@@ -2,7 +2,8 @@ package scala.cli.commands
 
 import caseapp._
 
-import scala.build.actionable.{ActionableDependencyHandler, ActionableDiagnostic}
+import scala.build.actionable.ActionableDependencyHandler
+import scala.build.actionable.ActionableDiagnostic.ActionableDependencyUpdateDiagnostic
 import scala.build.internal.CustomCodeWrapper
 import scala.build.options.Scope
 import scala.build.{CrossSources, Logger, Position, Sources}
@@ -25,7 +26,9 @@ object DependencyUpdate extends ScalaCommand[DependencyUpdateOptions] {
       CrossSources.forInputs(
         inputs,
         Sources.defaultPreprocessors(
-          buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper)
+          buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper),
+          buildOptions.archiveCache,
+          buildOptions.internal.javaClassNameVersionOpt
         ),
         logger
       ).orExit(logger)
@@ -50,19 +53,44 @@ object DependencyUpdate extends ScalaCommand[DependencyUpdateOptions] {
   }
 
   private def updateDependencies(
-    actionableUpdateDiagnostic: Seq[ActionableDiagnostic],
+    actionableUpdateDiagnostic: Seq[ActionableDependencyUpdateDiagnostic],
     logger: Logger
   ): Unit = {
-    actionableUpdateDiagnostic.map { diagnostic =>
-      diagnostic.positions.collect {
-        case Position.File(Right(path), _, _) => path
-      }.map { file =>
-        logger.message(s"Updating dependency    ${diagnostic.from}")
-        val content           = os.read(file)
-        val appliedDiagnostic = content.replace(diagnostic.from, diagnostic.to)
-        os.write.over(file, appliedDiagnostic)
-        logger.message(s"Updated dependency to: ${diagnostic.to}")
-      }
+    for {
+      diagnostic                                  <- actionableUpdateDiagnostic
+      Position.File(Right(filePath), startPos, _) <- diagnostic.positions
+    } {
+      val lineIndex = startPos._1
+      val appliedDiagnostic = os.read.lines(filePath).zipWithIndex.map {
+        case (line, index) if index == lineIndex =>
+          val updatedDependency = updateDependency(line, startPos, diagnostic)
+          updatedDependency
+        case (line, _) => line
+      }.mkString(System.lineSeparator())
+      os.write.over(filePath, appliedDiagnostic)
+      logger.message(s"Updated dependency to: ${diagnostic.to}")
+    }
+  }
+
+  private def updateDependency(
+    line: String,
+    pos: (Int, Int),
+    diagnostic: ActionableDependencyUpdateDiagnostic
+  ) = {
+    val depColumnIndex = pos._2
+    val head           = line.take(depColumnIndex)
+    val tail           = line.drop(depColumnIndex)
+    if (tail.startsWith("$ivy.`")) {
+      val last = tail.stripPrefix("$ivy.`").dropWhile(_ != '`')
+      s"$head$$ivy.`${diagnostic.to}$last"
+    }
+    else if (tail.startsWith("$dep.`")) {
+      val last = tail.stripPrefix("$dep.`").dropWhile(_ != '`')
+      s"$head$$dep.`${diagnostic.to}$last"
+    }
+    else {
+      val last = line.drop(depColumnIndex).dropWhile(_ != '\"')
+      s"$head${diagnostic.to}$last"
     }
   }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
@@ -1,9 +1,10 @@
 package scala.cli.commands
 
 import caseapp._
+import os.Path
 
+import scala.build.actionable.ActionableDependencyHandler
 import scala.build.actionable.ActionableDiagnostic.ActionableDependencyUpdateDiagnostic
-import scala.build.actionable.ActionablePreprocessor
 import scala.build.internal.CustomCodeWrapper
 import scala.build.options.Scope
 import scala.build.{CrossSources, Logger, Position, Sources}
@@ -67,57 +68,54 @@ object DependencyUpdate extends ScalaCommand[DependencyUpdateOptions] {
     actionableUpdateDiagnostics: Seq[ActionableDependencyUpdateDiagnostic],
     logger: Logger
   ): Unit = {
-    actionableUpdateDiagnostics.foreach { diagnostic =>
-      diagnostic.positions
-        .collect { case file: Position.File => file }
-        .foreach {
-          case Position.File(Right(filePath), startPos, _) =>
-            val lineIndex = startPos._1
-            val appliedDiagnostic = splitFileContentByLine(filePath).map {
-              case (line, index) if index == lineIndex + 1 =>
-                updateDependency(line, startPos, diagnostic)
-              case (line, _) => line
-            }.mkString
-            os.write.over(filePath, appliedDiagnostic)
-            logger.message(s"Updated dependency to: ${diagnostic.to}")
-          case Position.File(Left(file), _, _) =>
-            logger.message(
-              s"scala-cli can't update dependency to:${diagnostic.to} in virtual file: $file"
-            )
+    val groupedByFileDiagnostics =
+      actionableUpdateDiagnostics.flatMap {
+        diagnostic =>
+          diagnostic.positions.collect {
+            case file: Position.File =>
+              file.path -> (file, diagnostic)
+          }
+      }.groupMap(_._1)(_._2)
+
+    groupedByFileDiagnostics.foreach {
+      case (Right(file), diagnostics) =>
+        val sortedByLine       = diagnostics.sortBy(_._1.startPos._1).reverse
+        val appliedDiagnostics = updateDependencies(file, sortedByLine)
+        os.write.over(file, appliedDiagnostics)
+        diagnostics.foreach(diagnostic =>
+          logger.message(s"Updated dependency to: ${diagnostic._2.to}")
+        )
+      case (Left(file), diagnostics) =>
+        diagnostics.foreach {
+          diagnostic =>
+            logger.message(s"Warning: Scala CLI can't update ${diagnostic._2.to} in $file")
         }
     }
   }
 
-  private def splitFileContentByLine(file: os.Path) = {
-    val content       = os.read(file)
-    val startIndicies = Position.Raw.lineStartIndices(content).toList
+  private def updateDependencies(
+    file: Path,
+    diagnostics: Seq[(Position.File, ActionableDependencyUpdateDiagnostic)]
+  ): String = {
+    val fileContent   = os.read(file)
+    val startIndicies = Position.Raw.lineStartIndices(fileContent)
 
-    def lsplit(pos: List[Int], s: String): List[String] =
-      pos match {
-        case x :: rest => s.substring(0, x) :: lsplit(rest.map(_ - x), s.substring(x))
-        case Nil       => List(s)
-      }
-    lsplit(startIndicies, content).zipWithIndex
-  }
+    diagnostics.foldLeft(fileContent) {
+      case (fileContent, (file, diagnostic)) =>
+        val (line, column) = (file.startPos._1, file.startPos._2)
+        val startIndex = {
+          val index = startIndicies(line) + column
+          val skipIvySyntax =
+            fileContent.slice(index, index + 6) match {
+              case "$ivy.`" | "$dep.`" => 6
+              case _                   => 0
+            }
+          index + skipIvySyntax
+        }
+        val endIndex = startIndex + diagnostic.oldDependency.render.length()
 
-  private def updateDependency(
-    line: String,
-    pos: (Int, Int),
-    diagnostic: ActionableDependencyUpdateDiagnostic
-  ) = {
-    val depColumnIndex = pos._2
-    val (head, tail)   = line.splitAt(depColumnIndex)
-    if (tail.startsWith("$ivy.`")) {
-      val last = tail.stripPrefix("$ivy.`").dropWhile(_ != '`')
-      s"$head$$ivy.`${diagnostic.to}$last"
-    }
-    else if (tail.startsWith("$dep.`")) {
-      val last = tail.stripPrefix("$dep.`").dropWhile(_ != '`')
-      s"$head$$dep.`${diagnostic.to}$last"
-    }
-    else {
-      val last = line.drop(depColumnIndex).dropWhile(_ != '\"')
-      s"$head${diagnostic.to}$last"
+        val newDependency = diagnostic.to
+        s"${fileContent.slice(0, startIndex)}$newDependency${fileContent.drop(endIndex)}"
     }
   }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
@@ -58,6 +58,14 @@ object DependencyUpdate extends ScalaCommand[DependencyUpdateOptions] {
 
     if (options.all)
       updateDependencies(actionableUpdateDiagnostics, logger)
+    else {
+      println("Updates")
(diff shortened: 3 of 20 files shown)


modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+import coursier.core.{Module, ModuleName, Organization}
+import dependency._
+
+import scala.build.Positioned
+import scala.build.options.BuildOptions
+import scala.concurrent.duration.DurationInt
+
+case object ActionableDependencyHandler extends ActionableHandler[AnyDependency] {
+
+  val cache = FileCache()
```
alexarchambault: This should be `options.finalCache` instead (where it's used below), so cache-related options passed on the command-line are taken into account.

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+  override def extractValues(options: BuildOptions): Seq[Positioned[AnyDependency]] =
+    options.classPathOptions.extraDependencies.toSeq
+
+  override def createActionableDiagnostic(
+    value: Positioned[AnyDependency],
+    options: BuildOptions
+  ): ActionableDiagnostic = {
+
+    val dependency         = value.value
+    val scalaParams        = options.scalaParams.getOrElse(sys.error("sys.err")).get
```
alexarchambault: We shouldn't keep the `sys.error` here. Better return a `Either[BuildException, ActionableDiagnostic]` above, and propagate the error.

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+    val scalaBinaryVersion = scalaParams.scalaBinaryVersion
+
+    val organization = Organization(dependency.organization)
+    val moduleName   = ModuleName(s"${dependency.name}_$scalaBinaryVersion")
+    val csModule     = Module(organization, moduleName, Map.empty)
```
alexarchambault: That's not the correct way to go from a `dependency.AnyDependency` to a coursier module. You can call `.module.applyParams(scalaParams)` on the `AnyDependency`, with `scalaParams` coming from `options.scalaParams`. That way, the Scala suffix will always be correct (it can be the full Scala version, there can be no Scala suffix, etc.)

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    val res = cache.withTtl(0.seconds).logger.use {
+      Versions(cache)
+        .withModule(csModule)
+        .result()
+        .unsafeRun()(cache.ec)
+    }
+
+    val dependencyLatestVersion = res.versions.latest(coursier.core.Latest.Release).get
```
alexarchambault: Is `get` here `Option#get`? If yes, maybe we should just do nothing if it's empty (and log that in debug mode maybe).

modules/cli-options/src/main/scala/scala/cli/commands/DependencyUpdateOptions.scala
```diff
@@ -0,0 +1,20 @@
+package scala.cli.commands
+
+import caseapp._
+import caseapp.core.help.Help
+
+// format: off
+@HelpMessage("Update dependencies in project")
+final case class DependencyUpdateOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
```
alexarchambault, line 10: Maybe we could only copy some fields of `SharedOptions` here. Some options are going to be unused for sure here (like the `SharedCompilationServerOptions` ones).

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,69 @@
...
+      diagnostic.positions.collect {
+        case Position.File(Right(path), _, _) => path
+      }.map { file =>
+        logger.message(s"Updating dependency    ${diagnostic.from}")
+        val content           = os.read(file)
+        val appliedDiagnostic = content.replace(diagnostic.from, diagnostic.to)
+        os.write.over(file, appliedDiagnostic)
+        logger.message(s"Updated dependency to: ${diagnostic.to}")
+      }
+    }
```
alexarchambault: No need to map if this returns `Unit`. A simple `for` would do:
```suggestion
    for {
      diagnostic <- actionableUpdateDiagnostic
      Position.File(Right(file), _, _) <- diagnostic.positions
    } {
      logger.message(s"Updating dependency    ${diagnostic.from}")
      val content           = os.read(file)
      val appliedDiagnostic = content.replace(diagnostic.from, diagnostic.to)
      os.write.over(file, appliedDiagnostic)
      logger.message(s"Updated dependency to: ${diagnostic.to}")
    }
```

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,69 @@
...
+    actionableUpdateDiagnostic: Seq[ActionableDiagnostic],
+    logger: Logger
+  ): Unit = {
+    actionableUpdateDiagnostic.map { diagnostic =>
+      diagnostic.positions.collect {
+        case Position.File(Right(path), _, _) => path
+      }.map { file =>
+        logger.message(s"Updating dependency    ${diagnostic.from}")
+        val content           = os.read(file)
+        val appliedDiagnostic = content.replace(diagnostic.from, diagnostic.to)
```
alexarchambault: This shouldn't be a blind `String#replace` (which might replace things at a wrong place by accident). We should use the position from the `Positioned[AnyDependency]`.

modules/cli-options/src/main/scala/scala/cli/commands/DependencyUpdateOptions.scala
alexarchambault: But if too many of them are needed, maybe don't bother for now…

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+        Sources.defaultPreprocessors(
+          buildOptions.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper),
+          buildOptions.archiveCache,
+          buildOptions.internal.javaClassNameVersionOpt
+        ),
+        logger
+      ).orExit(logger)
+
+    val scopedSources = crossSources.scopedSources(buildOptions).orExit(logger)
+    val sources       = scopedSources.sources(Scope.Main, crossSources.sharedOptions(buildOptions))
```
alexarchambault: We might want to handle the test scope too, here. Maybe we should extract diagnostics for both, merge both diagnostic lists, and drop duplicates (`.distinct`).

modules/options/src/main/scala/scala/build/actionable/ActionablePreprocessor.scala
```diff
@@ -0,0 +1,27 @@
...
+    options: BuildOptions,
+    logger: Logger
+  ): Seq[ActionableDiagnostic] =
+    actionableHandlers.flatMap { handler =>
+      val values = handler.extractPositionedOptions(options)
+      val (errors, actionableDiagnostic) =
+        values.map(v => handler.createActionableDiagnostic(v, options)).partitionMap(identity)
+      errors.map(logger.debug(_))
+      actionableDiagnostic.flatten
+    }
```
alexarchambault: Not sure we should discard errors this way here… Better accumulate them and keep them. Something like this ought to do:
```suggestion
  ): Either[BuildException, Seq[ActionableDiagnostic]] =
    actionableHandlers
      .map { handler =>
        handler
          .extractPositionedOptions(options)
          .map(v => handler.createActionableDiagnostic(v, options))
          .sequence
          .left.map(CompositeBuildException(_))
      }
      .sequence
      .left.map(CompositeBuildException(_))
      .map(_.flatten)
```

alexarchambault: This might need `import scala.build.Ops._`

The errors should then be propagated up to `scala.cli.commands.DependencyUpdate`, where we can call `.orExit(logger)`, or `.orReport(logger)` (in watch mode).

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+      updateDependencies(actionableUpdateDiagnostics.flatten, logger)
+  }
+
+  private def updateDependencies(
+    actionableUpdateDiagnostic: Seq[ActionableDependencyUpdateDiagnostic],
+    logger: Logger
+  ): Unit = {
+    for {
+      diagnostic                                  <- actionableUpdateDiagnostic
+      Position.File(Right(filePath), startPos, _) <- diagnostic.positions
```
alexarchambault: If the path is left, maybe we should log that we can't update things there.

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+    } {
+      val lineIndex = startPos._1
+      val appliedDiagnostic = os.read.lines(filePath).zipWithIndex.map {
+        case (line, index) if index == lineIndex =>
+          val updatedDependency = updateDependency(line, startPos, diagnostic)
+          updatedDependency
```
alexarchambault: NIT
```suggestion
          updateDependency(line, startPos, diagnostic)
```

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+        case (line, _) => line
+      }.mkString(System.lineSeparator())
```
alexarchambault: This is going to force the file to use `System.lineSeparator()` as line separator, rather than what it was already using… Maybe the lines shouldn't be split upfront like this, but rather edited from the whole file content, with the help of things like `scala.build.Position.Raw.lineStartIndices`.

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private def updateDependency(
+    line: String,
+    pos: (Int, Int),
+    diagnostic: ActionableDependencyUpdateDiagnostic
+  ) = {
+    val depColumnIndex = pos._2
+    val head           = line.take(depColumnIndex)
+    val tail           = line.drop(depColumnIndex)
```
alexarchambault: NIT
```suggestion
    val (head, tail) = line.splitAt(depColumnIndex)
```

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
```diff
@@ -0,0 +1,29 @@
+package scala.build.actionable
+import scala.build.Positioned
+import scala.build.errors.BuildException
+import scala.build.options.BuildOptions
+
+trait ActionableHandler[V] {
+
```
alexarchambault: NIT Maybe the type param could be changed to a type member, as it's only used to pass things around its various methods, like
```suggestion
trait ActionableHandler {
  type V
```

Things ought to compile down-the-line with this.

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
```diff
@@ -0,0 +1,29 @@
...
+  /** Extract options on the basis of which actionable diagnostics will be generated
+    *
+    * @param options
+    *   the Build Options to extract positioned options
+    * @return
+    *   the list of positioned options
+    */
+  def extractPositionedOptions(options: BuildOptions): Seq[Positioned[V]]
```
alexarchambault: NIT Do we need the `Positioned[…]` here? Couldn't it be put in the type `V` itself? That way, more complex handlers could use things like `(Positioned[A], Positioned[B])` as `V` for example.

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,121 @@
...
+        case ad: ActionableDependencyUpdateDiagnostic => ad
+      }
+    }
+
+    val actionableMainUpdateDiagnostics = generateActionableUpdateDiagnostic(Scope.Main)
+    val actionableTestUpdateDiagnostics = generateActionableUpdateDiagnostic(Scope.Test)
+    val actionableUpdateDiagnostics =
+      (actionableMainUpdateDiagnostics ++ actionableTestUpdateDiagnostics).distinct
+
+    if (options.all)
```
alexarchambault, line 55: Seems nothing happens if `--all` isn't passed (nothing is printed in particular). Is it what we want?

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  private def splitFileContentByLine(file: os.Path) = {
+    val content       = os.read(file)
+    val startIndicies = Position.Raw.lineStartIndices(content).toList
+
+    def lsplit(pos: List[Int], s: String): List[String] =
+      pos match {
+        case x :: rest => s.substring(0, x) :: lsplit(rest.map(_ - x), s.substring(x))
```
alexarchambault: This is not tail-recursive, it seems this can blow the stack for large files. But also, as we typically only edit a handful of lines, maybe we shouldn't split the whole file like this… Better keep its content (as a large string), then change a line in it using start and end indices.

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  private def updateDependencies(
+    actionableUpdateDiagnostics: Seq[ActionableDependencyUpdateDiagnostic],
+    logger: Logger
+  ): Unit = {
+    actionableUpdateDiagnostics.foreach { diagnostic =>
+      diagnostic.positions
+        .collect { case file: Position.File => file }
+        .foreach {
+          case Position.File(Right(filePath), startPos, _) =>
```
alexarchambault: Maybe we should group things by file here. Then sort the diagnostics by `startPos` in reverse order, and apply all the changes at once to the file content in memory.

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,121 @@
...
+            val appliedDiagnostic = splitFileContentByLine(filePath).map {
+              case (line, index) if index == lineIndex + 1 =>
+                updateDependency(line, startPos, diagnostic)
+              case (line, _) => line
+            }.mkString
+            os.write.over(filePath, appliedDiagnostic)
+            logger.message(s"Updated dependency to: ${diagnostic.to}")
+          case Position.File(Left(file), _, _) =>
+            logger.message(
+              s"scala-cli can't update dependency to:${diagnostic.to} in virtual file: $file"
```
alexarchambault: NIT This should do:
```suggestion
              s"Warning: Scala CLI can't update ${diagnostic.to} in $file"
```

modules/cli-options/src/main/scala/scala/cli/commands/DependencyUpdateOptions.scala
romanowski: I think this was the idea of SharedOption - so all/most commands has but maybe we can partition it even further to e.g. SharedOption and SharedBuildOptions. For sure, it is not a part of this PR

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
romanowski: Agree, shouldn't we accept also a org/name or a patter of deps to update? Also, I think we could have an interactive implementation, so scala-cli asks about all deps to upgrage

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,121 @@
...
+    def generateActionableUpdateDiagnostic(scope: Scope)
+      : Seq[ActionableDependencyUpdateDiagnostic] = {
+      val sources = scopedSources.sources(scope, crossSources.sharedOptions(buildOptions))
+
+      if (verbosity >= 3)
+        pprint.err.log(sources)
+
+      val options = buildOptions.orElse(sources.buildOptions)
+      val actionableDiagnostics =
+        ActionablePreprocessor.generateActionableDiagnostics(options).orExit(logger)
```
romanowski: I wonder if we should generate all of the diagnostics here, since there is nothing too specific to `generateActionableDiagnostics` method at this point

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
```diff
@@ -0,0 +1,30 @@
...
+  type V
+
+  /** Extract options on the basis of which actionable diagnostics will be generated
+    *
+    * @param options
+    *   the Build Options to extract options
+    * @return
+    *   the list of options on the basis of which actionable diagnostics will be generated
+    */
+  def extractOptions(options: BuildOptions): Seq[V]
```
romanowski: I wonder if we really need to split the API into 2 parts: `extractOptions` and `createActionableDiagnostic`. I would personally stick to single method that accepts `BuildOptions` and `Inputs` and return `Either[BuildException, Seq[ActionableDiagnostic]]`

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
lwronski: Yes, I think we should also accept some pattern deps to update dependencies, but I wanted to split it in more PR.  So I will add it later.

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
lwronski: From my side, splitting into two methods allows me, to implement the function `createActionableDiagnostics` and users don't have to implement logic to collect actionable diagnostic.

modules/options/src/main/scala/scala/build/actionable/ActionablePreprocessor.scala
```diff
@@ -0,0 +1,21 @@
+package scala.build.actionable
+
+import scala.build.Ops._
+import scala.build.errors.{BuildException, CompositeBuildException}
+import scala.build.options.BuildOptions
+
+case object ActionablePreprocessor {
```
alexarchambault: This shouldn't be a case object, as this isn't meant to be passed around as data.
```suggestion
object ActionablePreprocessor {
```

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+
+trait ActionableHandler {
+
+  /** Type of option used to generate actionable diagnostic
+    */
+  type V
+
+  /** Returned type of actionable diagnostics
+    */
+  type A
```
alexarchambault: Don't we need to add a type bound for the `A` instances returned by `actionableDiagnostic` and `createActionableDiagnostics` to be useful?
```suggestion
  type A <: ActionableDiagnostic
```

modules/options/src/main/scala/scala/build/actionable/ActionablePreprocessor.scala
```diff
@@ -0,0 +1,21 @@
...
+  val actionableHandlers = Seq(
```
alexarchambault: This should be typed as `Seq[ActionableHandler]`, so that the details of `ActionableDependencyHandler` don't leak here:
```suggestion
  val actionableHandlers = Seq[ActionableHandler](
```
(this should work fine with the type bound suggestion above)

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    val res = cache.withTtl(0.seconds).logger.use {
+      Versions(cache)
+        .withModule(csModule)
+        .result()
+        .unsafeRun()(cache.ec)
+    }
+
+    val latestVersion = value(res.versions.latest(coursier.core.Latest.Release).toRight(
+      new ActionableHandlerError(s"Not found dependency versions for ${dependency.render}")
```
alexarchambault:
```suggestion
      new ActionableHandlerError(s"No latest version found for ${dependency.render}")
```

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -0,0 +1,60 @@
...
+    ))
+
+    latestVersion
```
alexarchambault: No need to add that extra variable:
```suggestion
    value {
      res.versions.latest(coursier.core.Latest.Release).toRight {
        new ActionableHandlerError(s"Not found dependency versions for ${dependency.render}")
      }
    }
```

modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala
```diff
@@ -0,0 +1,122 @@
...
+      case (fileContent, (file, diagnostic)) =>
+        val (line, column) = (file.startPos._1, file.startPos._2)
+        val startIndex = {
+          val index = startIndicies(line) + column
+          val skipIvySyntax =
+            fileContent.slice(index, index + 6) match {
+              case "$ivy.`" | "$dep.`" => 6
+              case _                   => 0
+            }
+          index + skipIvySyntax
```
alexarchambault: The skip ivy stuff should be handled from `ActionableDependencyHandler` IMO. Downstream users of it shouldn't have to mind about it.

lwronski: In `ActionableDependencyHandler` I don't have information about type of dependency.  

For example, `Positioned[AnyDependency]` contains only this informations:
```
Positioned(List(File(Right(/scala-cli/Hello.scala),(0,13),(0,37))),com.lihaoyi::geny:0.6.5)
```

so I change resolving ivy dependencies and update start position in last commit.

modules/options/src/main/scala/scala/build/actionable/ActionableHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import scala.build.Ops._
+import scala.build.errors.{BuildException, CompositeBuildException}
+import scala.build.options.BuildOptions
+
+trait ActionableHandler {
+
+  /** Type of option used to generate actionable diagnostic
+    */
+  type V
```
romanowski: Why use type members not type parameters? Also, if we want to keep it as type member we should use more meaningful names

alexarchambault (approved):

romanowski:
I think we also should describe that we have a dependency update command in Dependency guide.

modules/options/src/main/scala/scala/build/actionable/ActionablePreprocessor.scala
```diff
@@ -0,0 +1,21 @@
+package scala.build.actionable
+
+import scala.build.Ops._
+import scala.build.errors.{BuildException, CompositeBuildException}
+import scala.build.options.BuildOptions
+
+object ActionablePreprocessor {
+  val actionableHandlers = Seq[ActionableHandler[_ <: ActionableDiagnostic]](
```
alexarchambault: NIT No need of the type constraint here (`ActionableHandler` adds it already)
```suggestion
  val actionableHandlers = Seq[ActionableHandler[_]](
```

modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
```diff
@@ -247,8 +247,8 @@ case object ScalaPreprocessor extends Preprocessor {
       val toFilePos = Position.Raw.filePos(path, content)
       val deps = value {
         dependencyTrees
-          .map { t =>
-            val pos      = toFilePos(Position.Raw(t.start, t.end))
+          .map { t => // add 6 to start position to skip ivy syntax ($ivy.`) and dep ($dep.`)
+            val pos      = toFilePos(Position.Raw(t.start + 6, t.end))
```
alexarchambault: NIT This can be made explicit:
```suggestion
          .map { t => // skip ivy ($ivy.`) or dep syntax ($dep.`)
            val pos      = toFilePos(Position.Raw(t.start + "$ivy.`".length, t.end))
```

modules/options/src/main/scala/scala/build/actionable/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,38 @@
+package scala.build.actionable
+
+import dependency._
+
+import scala.build.Position
+import scala.build.errors.{Diagnostic, Severity}
+
+sealed abstract class ActionableDiagnostic {
```
alexarchambault: I'd say this one doesn't need to be sealed (we shouldn't need to match `ActionableDiagnostic` on all their possible instance types). That should allow to move `ActionableDependencyUpdateDiagnostic` alongside its handler.

Gedochao (approved):

PR #1056, opened 2022-05-31
Add config command

alexarchambault:
This adds a `config` sub-command, that users can use to set and get values of pre-defined config keys, like
```text
$ scala-cli config user.name "Alex"
$ scala-cli config user.email "alex@alex.me"
$ scala-cli config github.token "file:$HOME/.secrets/scala-cli-gh-token"
$ scala-cli config pgp.public-key
…
```

These values are meant to be used in the upcoming `publish setup` sub-command.

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
@@ -1,0 +1,35 @@
+package scala.cli.commands.config
+
+import caseapp._
+
+import scala.cli.commands.{CoursierOptions, LoggingOptions, SharedDirectoriesOptions}
+
+// format: off
+final case class ConfigOptions(
+  @Recurse
+    logging: LoggingOptions = LoggingOptions(),
+  @Recurse
+    directories: SharedDirectoriesOptions = SharedDirectoriesOptions(),
+  @Recurse
+    coursier: CoursierOptions = CoursierOptions(),
+
+  @Group("Config")
+  @HelpMessage("Dump config DB as JSON")
+  @Hidden
+    dump: Boolean = false,
+  @Group("Config")
+  @HelpMessage("Create PGP key in config")
+    createPgpKey: Boolean = false,
+  @Group("Config")
+  @HelpMessage("If the entry is a password, print the password value rather than how to get the password")
+    password: Boolean = false,
+  @Group("Config")
+  @HelpMessage("Remove an entry from config")
+    unset: Boolean = false
+)
+// format: on
+
+object ConfigOptions {
+  implicit lazy val parser: Parser[ConfigOptions] = Parser.derive
+  implicit lazy val help: Help[ConfigOptions]     = Help.derive
+}
--- a/modules/cli/src/main/scala/scala/cli/config/ConfigDb.scala
+++ b/modules/cli/src/main/scala/scala/cli/config/ConfigDb.scala
@@ -1,0 +1,227 @@
+package scala.cli.config
+
+import com.github.plokhotnyuk.jsoniter_scala.core.{Key => _, _}
+import com.github.plokhotnyuk.jsoniter_scala.macros._
+import coursier.parse.RawJson
+
+import java.nio.file.attribute.PosixFilePermission
+
+import scala.build.Directories
+import scala.build.errors.BuildException
+import scala.collection.immutable.ListMap
+import scala.util.Properties
+
+/** In-memory representation of a configuration DB content.
+  *
+  * Use [[ConfigDb.apply]] or [[ConfigDb.open]] to create an instance of it.
+  *
+  * [[set]], [[setFromString]], and [[remove]] only change values in memory.
+  *
+  * Use [[save]] to persist values on disk.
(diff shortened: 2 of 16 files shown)


website/docs/commands/misc/config.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Config
+sidebar_position: 1
+---
+
+The `config` sub-command allows to get and set various configuration values, used by
+other Scala CLI sub-commands.
+
+Examples of use:
+```text
```
Gedochao, line 19:
```suggestion
```bash
```
imo it's better to use `bash` code blocks here, as it allows for some syntax coloring.
command outputs can then be put as comments in the code block.
this applies to the other code blocks in this doc as well.

Gedochao, line 10:
```suggestion
```bash
imo it's better to use `bash` code blocks here, as it allows for some syntax coloring.
command outputs can then be put as comments in the code block.
this applies to the other code blocks in this doc as well.

File: modules/cli/src/main/scala/scala/cli/config/ConfigDb.scala [hunk 2]
```diff
@@ -0,0 +1,163 @@
...
+import com.github.plokhotnyuk.jsoniter_scala.macros._
+import coursier.parse.RawJson
+
+import java.nio.file.attribute.PosixFilePermission
+
+import scala.build.Directories
+import scala.build.errors.BuildException
+import scala.collection.immutable.ListMap
+
+final class ConfigDb private (
```
Gedochao, line 13: wondering if adding a scaladoc here wouldn't be helpful long term (applies to most newly created source files in this PR as well)

Gedochao, line 22: wondering if adding a scaladoc here wouldn't be helpful long term (applies to most newly created source files in this PR as well)

modules/cli-options/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
```diff
@@ -0,0 +1,25 @@
...
+// format: off
+final case class ConfigOptions(
+  @Recurse
+    logging: LoggingOptions = LoggingOptions(),
+  @Recurse
+    directories: SharedDirectoriesOptions = SharedDirectoriesOptions(),
+  @Recurse
+    coursier: CoursierOptions = CoursierOptions(),
+  @Hidden
+    dump: Boolean = false,
```
lwronski, line 16: The name `show` would be more descriptive for this option.

lwronski (commented):
I also have one question for this functionality. Is there an option to remove entry from `ConfigDb`?

website/docs/commands/misc/config.md
alexarchambault: Changed it to `bash`

website/docs/commands/misc/config.md
```diff
@@ -0,0 +1,54 @@
...
+title: Config
+sidebar_position: 1
+---
+
+The `config` sub-command allows to get and set various configuration values, used by
+other Scala CLI sub-commands.
+
+Examples of use:
+```bash
+$ scala-cli config user.name "Alex Me"
```
Gedochao:
```suggestion
scala-cli config user.name "Alex Me"
```

website/docs/commands/misc/config.md
```diff
@@ -0,0 +1,54 @@
...
+$ scala-cli config user.name
```
Gedochao:
```suggestion
scala-cli config user.name
```

website/docs/commands/misc/config.md
```diff
@@ -0,0 +1,54 @@
...
+Alex Me
```
Gedochao, line 20:
```suggestion
# Alex Me
```

Gedochao (commented):
if lines of commands in `bash` command blocks aren't starting with `$` and the expected outputs are comments (starting with `#`), then it becomes possible to run the commands from the doc right in your terminal with a single click from the IDE, while browsing the `.md` file (screenshot below).

image

applies to the remaining code blocks as well.

website/docs/commands/misc/config.md
alexarchambault: The convention of using `#` for output feels really unnatural to me… I'd rather keep using the `$`.

modules/cli/src/main/scala/scala/cli/config/ConfigDb.scala
alexarchambault: Should be fixed.

website/docs/commands/misc/config.md
alexarchambault: I changed it to make the docs-test job pass, but the output it gives feels really ugly.

alexarchambault:
https://github.com/VirtusLab/scala-cli/pull/1068 should allow to make the CI green if needed (lowers memory needs to generate native launchers)

Gedochao (approved):

lwronski (approved):
The only thing is missing for me,  the `--unset-all` option, but that can be added in a separate PR

PR #1063, opened 2022-06-02
DO NOT MERGE

alexarchambault:
This reverts the use of a Java parser, to check if this makes native image generation need less memory.


alexarchambault:
Closing, https://github.com/VirtusLab/scala-cli/pull/1068 moves the java class name extraction stuff to an external binary.

PR #1068, opened 2022-06-02
Use external binary to extract class name from stdin Java sources

alexarchambault:


alexarchambault:
(Adding some more scaladoc right now…)

modules/build/src/main/scala/scala/build/internal/JavaParserProxy.scala
```diff
@@ -0,0 +1,65 @@
...
+    *   either some class name (if one was found) or none (if none was found), or a
+    *   [[BuildException]]
+    */
+  def className(
+    archiveCache: ArchiveCache[Task],
+    content: Array[Byte],
+    logger: Logger
+  ): Either[BuildException, Option[String]] = either {
+
+    val versionOpt = Some("0.1.0")
```
Gedochao: wouldn't this always be picked over `Constants.javaClassNameVersion`, given how it's coded atm?
Why not just do `val version = Constants.javaClassNameVersion`?

Gedochao:
I think it'd be useful to mention [java-class-name](https://github.com/scala-cli/java-class-name) among other libraries/forks we maintain ourselves in a separate `.md` doc, so that it's easier to look them up when contributing. 

(unless we already have a doc like that? in which case `java-class-name` should be mentioned there)

modules/build/src/main/scala/scala/build/internal/JavaParserProxy.scala
alexarchambault: Should be fixed.

alexarchambault:
> I think it'd be useful to mention [java-class-name](https://github.com/scala-cli/java-class-name) among other libraries/forks we maintain ourselves in a separate `.md` doc, so that it's easier to look them up when contributing. 

To me, that doc is https://github.com/scala-cli

Gedochao:
> To me, that doc is https://github.com/scala-cli

Okay... but do we mention the `scala-cli` org anywhere in our docs?
You may be right that it's probably sufficient to make the list of repos there visible to our contributors

alexarchambault:
> Okay... but do we mention the `scala-cli` org anywhere in our docs?

Just opened https://github.com/VirtusLab/scala-cli/pull/1075 for it.

Gedochao (approved):

Issue #670, opened 2022-02-21
Putting the installation method of scala-cli in an environment variable or a metadata file


ghost:
Related to issue #543  currently, if a user forgets the previous method they used for installing scala-cli on their system, they might use another one for reinstalling. This creates a duplicate in PATH; therefore, even when a newer version is installed later, through a second method, this would be the old version that might show up first on PATH and get executed. 

How it should behave:

The user types `scala-cli —version` or `scala-cli about` or `scala-cli update`.
Then the user receives information about the method with which they have installed scala-cli, for example Homebrew; and  instructions get shown to them about how they should update scala-cli using this installation method, such as Homebrew. 

How it can get Implemented :

We create a hidden command for `scala-cli` such as `installMethod` which then gets run by the installation tool, post installation. Upon receiving a command like `scala-cli installMethod brew`, Scala CLI creates an environment variable called `$SCALA_CLI_INSTALLATION_METHOD` and assigns the value `brew` to it. 

This way, each time the `about`, `version`, or `update` commands are called; scala-cli can just read the value of this environment variable for knowing how it is installed, or even update itself through the same method.

Alternatively, the installation method can be written by the installation method to a metadata file put in the same directory as the executable file of scala-cli. Then scala-cli can read the contents of this file for responding to the above mentioned commands. The downside of this method is that the metadata file would get lost upon an installation directory change.

ghost:
using Java Properties is another alternative for creating the metadata file:
https://docs.oracle.com/javase/tutorial/essential/environment/properties.html

Determining the location of this file remains an issue.

alexarchambault:
About this, I think a hackish way of doing could consist in appending a zip file at the end of the binaries right before putting them in packages. That zip file could contain an entry, like `META-INF/scala-cli/source.json` say, with a few metadata, like the package type, the typical command to update it, etc. Then the `update` command could try to open the binary path as a zip file, and check for that entry.

Gedochao:
Given the number of installation methods and platforms we support, I do not believe it likely for us to have the resources necessary to hack this in a stable manner.
If anyone has a sane idea on how to implement this, then by all means, feel free to reopen.
Otherwise, closing for now, we won't fix this.

PR #1069, opened 2022-06-03
Suggest user to update scala-cli by package manager in doctor

### LINKED ISSUE #1038: Confusing message about managing itself with doctor
**Version(s)**

0.1.6

**Describe the bug**

Not really a bug, but I think this might lead to a bit of confusion as it confused me a bit. When I try the `scala-cli doctor` command I see this message.

```
❯ scala-cli doctor
Your scala-cli version (0.1.6) is current.
scala-cli could update itself since it is correctly installed in only one location: Set(/Users/ckipp/Library/Application Support/Coursier/bin/scala-cli).
Your scala-cli is a native application.
```

I'm not really sure why it tells me the part about how scala-cli could update itself. I use `cs` to manage scala-cli but I was curious what would happen if I do a `scala-cli update`, since that's what the message makes it sound like I can do.

When I do that I get:

```
❯ scala-cli update
Scala CLI was not installed by the installation script, please use your package manager to update scala-cli.
```

These two messages seem at odds with one another. If I'm using a tool like `cs` to manage scala-cli I shouldn't see a message about using itself to update.

**To Reproduce**
Follow the steps above.

**Expected behaviour**
Clear messages.

lwronski:


modules/cli/src/main/scala/scala/cli/commands/Doctor.scala
```diff
@@ -67,10 +67,15 @@ object Doctor extends ScalaCommand[DoctorOptions] {
...
         s"scala-cli would not be able to update itself since it is installed in multiple directories: ${scalaCliPaths.mkString(", ")}."
       )
-    else
+    else if (Update.isScalaCLIInstalledByInstallationScript())
       println(
         s"scala-cli could update itself since it is correctly installed in only one location: $scalaCliPaths."
       )
+    else
+      println(
+        s"scala-cli could update by your package manager since it is correctly installed in only one location: $scalaCliPaths."
```
alexarchambault: NIT Phrasing, and don't leave `Seq(…)` or `List(…)` in the output.
```suggestion
        s"scala-cli can be updated by your package manager since it is correctly installed in only one location: ${scalaCliPaths.mkString}."
```

alexarchambault: Ideally, we should implement https://github.com/VirtusLab/scala-cli/issues/670#issuecomment-1123664191, and rely on it to know whether the `scala-cli` binary has been installed by a package manager.

alexarchambault (approved):

PR #1070, opened 2022-06-04
Add suffix to project name which contains a virtual files

### LINKED ISSUE #1054: Stale error reports are sometimes shown
**Version(s)**

`0.1.6`

**Describe the bug**

When I run `echo 'println("Hello")' | scala-cli _ ` and then `echo 'println("Hello")' | scala-cli _ .sc` I get:

```
> echo 'println("Hello")' | scala-cli _                                                                                                
Compiling project (Scala 3.1.2, JVM)
[error] <stdin>:1:1: Illegal start of toplevel definition
Error compiling project (Scala 3.1.2, JVM)
Compilation failed
> echo 'println("Hello")' | scala-cli _.sc 
[error] /Users/kromanowski/Library/Caches/ScalaCli/virtual-projects/7d/project-34b8e93c/.scala-build/project_103be31561_103be31561/src_generated/main/stdin.scala:1:1: Illegal start of toplevel definition
[error] ^^^^^^^
Hello
```

**To Reproduce**

Run:

```
echo 'println("Hello")' | scala-cli _   
echo 'println("Hello")' | scala-cli _.sc      
```

**Expected behaviour**
No error message in the 2nd run.

lwronski:
I think this is some kind of workaround,  the problem is that for 
```
echo 'println("Hello")' | scala-cli _  
```
 and 
```
echo 'println("Hello")' | scala-cli _.sc  
``` 
the same project name is generated.

When the project names are different, the problem reported in #1054 disappears.

IMO it still required additional work to find core of the problem.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -267,7 +267,7 @@ object Inputs {
           case elem: SourceFile => (elem.path / os.up, true, WorkspaceOrigin.SourcePaths)
           case _: Virtual =>
             val dir = homeWorkspace(validElems, directories)
-            (dir, true, WorkspaceOrigin.HomeDir)
+            (dir, false, WorkspaceOrigin.HomeDir)
           case r: ResourceDirectory =>
             // Makes us put .scala-build in a resource directory :/
             (r.path, true, WorkspaceOrigin.ResourcePaths)
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -186,7 +186,7 @@ object Inputs {
   }
 
   sealed abstract class VirtualSourceFile extends Virtual {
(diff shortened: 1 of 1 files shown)


Gedochao:
Can we have an integration test for this, to prevent regressions?

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -267,7 +267,7 @@ object Inputs {
           case elem: SourceFile => (elem.path / os.up, true, WorkspaceOrigin.SourcePaths)
           case _: Virtual =>
             val dir = homeWorkspace(validElems, directories)
-            (dir, false, WorkspaceOrigin.HomeDir)
+            (dir, true, WorkspaceOrigin.HomeDir)
```
alexarchambault: We shouldn't need to change this one. These projects already land under directories such as `~/Library/Caches/ScalaCli/virtual-projects/03/project-ad6a817e`, which already include the inputs hash (here, `03ad6a817e`).

alexarchambault (approved):

PR #1080, opened 2022-06-07
Convert relative path to absolute for coverage options

### LINKED ISSUE #1072: `scoverage.coverage` file not generated with `-coverage-out` flag for Scala 3
**Version(s)**

`0.1.6`

**Describe the bug**

Running the Scala 3 compiler with coverage enabled does not generate the `scoverage.coverage` file.

**To Reproduce**

```scala
//> using scala "3.2.0-RC1-bin-20220604-13ce496-NIGHTLY"
//> using options  "-Ycheck:instrumentCoverage", "-coverage-out:.", "-sourceroot:."

@main def main = ()
```

**Expected behaviour**
Compiling this should generate a `scoverage.coverage` file.

lwronski:


Gedochao (approved):

alexarchambault (commented):
Shouldn't we do that only when using Bloop, just in case? It could be handled around [here](https://github.com/VirtusLab/scala-cli/blob/fbb1d308bbfaf25e3a6f831a37e87de22650a850/modules/build/src/main/scala/scala/build/Project.scala#L49), right before we write a Bloop file.

(Like if any of the options starts with `-coverage-out:`, we parse the options again as a `ShadowingSeq[ScalacOpt]`, and update the `-coverage-out:…` one)

modules/build/src/main/scala/scala/build/Project.scala
```diff
@@ -72,6 +72,21 @@ final case class Project(
...
+    )
+    def maybeConvertRelativePathToAbsolute(scalacOption: String): String =
+      keysWhichAcceptRelativePath.find(scalacOption.startsWith) match {
+        case Some(scalacKey) =>
+          val maybeRelativePath = scalacOption.stripPrefix(scalacKey)
+          val absolutePath      = os.Path(maybeRelativePath, Os.pwd)
+          s"$scalacKey$absolutePath"
+        case None => scalacOption
+      }
+    scalacOptions.map(maybeConvertRelativePathToAbsolute(_))
```
alexarchambault: I think we should keep grouping the options with `ShadowingSeq` / `ScalacOpt` just in case, so that the `-coverage-out:` is expected to be interpreted as a coverage option by scalac, not as a value of some other argument. Something like
```scala
ShadowingSeq.from(opts.map(ScalacOpt(_))).values.map { l =>
  // only look at the head, the tail is only values passed to it
  if (l.head.value.startsWith("-coverage-out:"))
    // actual -coverage-out: option
    ScalacOpt("-coverage-out:" + *absolute path*) +: l.tail
  else
    // not a -coverage-out: option
    l
}
```

PR #1082, opened 2022-06-07
Update installation instructions for Nix

kubukoz:
It's already in the unstable / 22.05 channels so it shouldn't be necessary to specify a path to Nixpkgs master.


alexarchambault (approved):
Thanks!

PR #1084, opened 2022-06-07
Extract java home from env variable

lwronski:
` sys.props.get("java.home")` return `None` for native launcher, so we extract `JAVA_HOME ` from env variable.

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
@@ -183,8 +183,12 @@ final case class BuildOptions(
     javaOptions.javaHomeOpt
       .orElse {
         if (javaOptions.jvmIdOpt.isEmpty)
-          sys.props.get("java.home").map(p =>
-            Positioned(Position.Custom("java.home prop"), os.Path(p, Os.pwd))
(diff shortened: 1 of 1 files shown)


modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -183,8 +183,8 @@ final case class BuildOptions(
     javaOptions.javaHomeOpt
       .orElse {
         if (javaOptions.jvmIdOpt.isEmpty)
-          sys.props.get("java.home").map(p =>
-            Positioned(Position.Custom("java.home prop"), os.Path(p, Os.pwd))
+          Option(System.getenv("JAVA_HOME")).map(p =>
```
alexarchambault, line 186: Let's keep a fallback to `sys.props.get("java.home")`, if `JAVA_HOME` isn't set. On the JVM, this property should be set even when `JAVA_HOME` isn't (like if the `bin` directory is in the `PATH` and `JAVA_HOME` isn't set).

lwronski: Right, that makes sense

alexarchambault (approved):

PR #1086, opened 2022-06-07
Add --spark option to package sub-command

alexarchambault:


modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
```diff
@@ -39,6 +39,12 @@ final case class PackageOptions(
   @HelpMessage("Generate an assembly JAR")
     assembly: Boolean = false,
   @Group("Package")
+  @HelpMessage("For assembly JAR, whether to add a bash / bat preamble")
```
romanowski, line 42: Shouldn't we add using directives for those options? Especially for `provided` one

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
```diff
@@ -39,6 +39,12 @@ final case class PackageOptions(
...
+    preamble: Boolean = true,
+  @Group("Package")
+  @HelpMessage("Generate an assembly JAR for Spark")
```
romanowski: I think the doc should mention that it will exclude spark modules

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -608,12 +650,64 @@ object Package extends ScalaCommand[PackageOptions] {
...
+      sys.error("Internal error: expected resolution to have been kept")
+    }
+    val modules = value {
+      provided
+        .map(_.toCs(build.scalaParams))
+        .sequence
+        .left.map(CompositeBuildException(_))
+    }
+    val modulesSet   = modules.toSet
+    val providedDeps = res.dependencyArtifacts.map(_._1).filter(dep => modulesSet(dep.module))
```
romanowski:
```suggestion
    val providedDeps = res.dependencyArtifacts.map(_._1).filter(dep => modulesSet.contains(dep.module))
```

`apply` method on Set is a bit unintuitive

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -626,14 +720,28 @@ object Package extends ScalaCommand[PackageOptions] {
...
 
-    val preamble = Preamble()
-      .withOsKind(Properties.isWin)
-      .callsItself(Properties.isWin)
+    val allFiles = build.artifacts.artifacts.map(_._2)
+    val files =
+      if (provided.isEmpty) allFiles
+      else {
+        val providedFiles0 = value(providedFiles(build, provided, logger)).toSet
+        allFiles.filterNot(providedFiles0)
```
romanowski:
```suggestion
        allFiles.filterNot(providedFiles0.contains)
```

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -626,14 +720,28 @@ object Package extends ScalaCommand[PackageOptions] {
...
       }
 
-    val preamble = Preamble()
-      .withOsKind(Properties.isWin)
-      .callsItself(Properties.isWin)
+    val allFiles = build.artifacts.artifacts.map(_._2)
+    val files =
+      if (provided.isEmpty) allFiles
+      else {
+        val providedFiles0 = value(providedFiles(build, provided, logger)).toSet
```
romanowski:
```suggestion
        val providedFilesSet = value(providedFiles(build, provided, logger)).toSet
```

modules/cli/src/main/scala/scala/cli/commands/packaging/Spark.scala
```diff
@@ -0,0 +1,22 @@
+package scala.cli.commands.packaging
+
+import dependency._
+
+object Spark {
+
+  private def names = Seq(
+    // FIXME Add more? (see "cs complete-dependency org.apache.spark: | grep '_2\.12$'")
```
romanowski: @prolativ should we include more?

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -485,6 +485,7 @@ final case class BuildOptions(
       addJvmTestRunner = isTests && addJvmTestRunner,
       addJmhDependencies = jmhOptions.addJmhDependencies,
       extraRepositories = finalRepositories,
+      keepResolution = internal.keepResolution,
```
romanowski: is there a reason why we want to drop the resolution?

romanowski (changes requested):
I think we are missing:
- using directives versions of some settings
- changes in the package guide and maybe a short cookbook


Also, I recall that we spoke about scala-cli printing some information on how to deploy it spark - but I guess this can be done as a part of the next PR.

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
alexarchambault: I don't recall the exact numbers, but I remember `coursier.Resolution` instances can be quite heavy in memory (possibly in several MB, if not much more when spark is involved). Better not keep them around when they're not needed.

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
alexarchambault: Should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Package.scala
alexarchambault: Agreed, should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Package.scala
alexarchambault: Should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Package.scala
alexarchambault: Should be fixed.

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
alexarchambault: Should be fixed.

modules/cli/src/main/scala/scala/cli/commands/packaging/Spark.scala
prolativ: Do we need to list all existing spark modules here or only the most commonly used ones?

alexarchambault: These are the modules we exclude from the assemblies generated by `scala-cli package --spark`. Basically, we want to exclude modules whose JARs are already in the Spark distribution.

romanowski (approved):

PR #1087, opened 2022-06-07
Tweaking

alexarchambault:
Bits of https://github.com/VirtusLab/scala-cli/pull/926, moved here for easier review.


modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
```diff
@@ -618,22 +652,25 @@ object Publish extends ScalaCommand[PublishOptions] {
...
-              .getOrElse {
-                val url =
-                  if (repoStr.contains("://")) repoStr
-                  else os.Path(repoStr, Os.pwd).toNIO.toUri.toASCIIString
-                MavenRepository(url)
-              }
-              .withAuthentication(authOpt)
+            val repo0 = {
+              val r = RepositoryParser.repositoryOpt(repoStr)
+                .collect {
```
Gedochao, line 661: why the newline? 
imo it's readable enough to contain the `collect` as a single line:
```scala
.collect { case m: MavenRepository => m }
```

Gedochao (approved):
LGTM, the other comment is just a nitpick on formatting

modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
alexarchambault: It's kind of arbitrary… I tend to always put `case` clauses on new lines, it's more readable this way IMO. I also don't put `{ … }` blocks on a single line for the same reason (only `( … )` ones).

PR #1095, opened 2022-06-09
Favor non-script main classes & add an option for listing main classes in context

### LINKED ISSUE #1067: `.scala` main class should be prioritised over any scripts in the context
**Version(s)**

0.1.6

**Describe the bug**

When running a directory with a single main class defined in a `.scala` file and any scripts present, Scala CLI will return an error about the main class being ambigous (because technically speaking, scripts also have a main class in the wrapper). 

**To Reproduce**

image

```bash
$ scala-cli .
Compiling project (Scala 3.1.2, JVM)
Compiled project (Scala 3.1.2, JVM)
[error]  Found several main classes: Hello, scripts.Script_sc
```

To run successfully, the main class currently has to be chosen explicitly with the `--main-class` option.
```bash
$ scala-cli . --main-class Hello
Compiling project (Scala 3.1.2, JVM)
Compiled project (Scala 3.1.2, JVM)
Hello world
```

**Expected behaviour**
If there's a single main class defined in the `.scala` files in the context, it should be prioritised and run even if scripts are present as well.
```bash
$ scala-cli . --main-class Hello
Compiling project (Scala 3.1.2, JVM)
Compiled project (Scala 3.1.2, JVM)
Hello world
```

However, we should still maintain an option to print the list of available main classes, if a different behaviour was necessary.
Something like:
```bash
$ scala-cli . --main-class ?
Hello scripts.Script_sc
```

Gedochao:
Fixes #1067

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Run.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Run.scala
@@ -10,6 +10,7 @@ import scala.build.internal.{Constants, Runner, ScalaJsLinkerConfig}
 import scala.build.options.{BuildOptions, JavaOpt, Platform}
 import scala.build.{Build, BuildThreads, Inputs, Logger, Positioned}
 import scala.cli.CurrentParams
+import scala.cli.commands.util.MainClassOptionsUtil._
 import scala.cli.commands.util.SharedOptionsUtil._
 import scala.cli.internal.ProcUtil
 import scala.util.Properties
@@ -37,7 +38,6 @@ object Run extends ScalaCommand[RunOptions] {
     )
     baseOptions.copy(
       mainClass = mainClass.mainClass,
-      mainClassLs = mainClass.mainClassLs,
       javaOptions = baseOptions.javaOptions.copy(
         javaOpts =
           baseOptions.javaOptions.javaOpts ++
@@ -67,12 +67,15 @@ object Run extends ScalaCommand[RunOptions] {
       build: Build.Successful,
       allowTerminate: Boolean
     ): Either[BuildException, (Process, CompletableFuture[_])] = either {
+      val potentialMainClasses = build.foundMainClasses()
+      value(options.mainClass.maybePrintMainClasses(potentialMainClasses))
       val process = value(maybeRunOnce(
         build,
         programArgs,
         logger,
         allowExecve = allowTerminate,
-        jvmRunner = build.artifacts.hasJvmRunner
+        jvmRunner = build.artifacts.hasJvmRunner,
+        potentialMainClasses
       ))
 
       val onExitProcess = process.onExit().thenApply { p1 =>
@@ -169,7 +172,8 @@ object Run extends ScalaCommand[RunOptions] {
     args: Seq[String],
     logger: Logger,
     allowExecve: Boolean,
-    jvmRunner: Boolean
+    jvmRunner: Boolean,
+    potentialMainClasses: Seq[String]
   ): Either[BuildException, Process] = either {
 
     val mainClassOpt = build.options.mainClass.filter(_.nonEmpty) // trim it too?
@@ -179,7 +183,7 @@ object Run extends ScalaCommand[RunOptions] {
       }
     val mainClass = mainClassOpt match {
       case Some(cls) => cls
-      case None      => value(build.retainedMainClass(logger))
+      case None      => value(build.retainedMainClass(potentialMainClasses))
     }
     val verbosity = build.options.internal.verbosity.getOrElse(0).toString
 
--- a/modules/cli/src/main/scala/scala/cli/commands/Run.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Run.scala
@@ -37,6 +37,7 @@ object Run extends ScalaCommand[RunOptions] {
     )
     baseOptions.copy(
       mainClass = mainClass.mainClass,
+      mainClassLs = mainClass.mainClassLs,
       javaOptions = baseOptions.javaOptions.copy(
         javaOpts =
           baseOptions.javaOptions.javaOpts ++
--- a/modules/cli/src/main/scala/scala/cli/commands/Run.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Run.scala
@@ -66,34 +66,43 @@ object Run extends ScalaCommand[RunOptions] {
     def maybeRun(
       build: Build.Successful,
       allowTerminate: Boolean
-    ): Either[BuildException, (Process, CompletableFuture[_])] = either {
+    ): Either[BuildException, Option[(Process, CompletableFuture[_])]] = either {
       val potentialMainClasses = build.foundMainClasses()
-      value(options.mainClass.maybePrintMainClasses(potentialMainClasses))
-      val process = value(maybeRunOnce(
-        build,
-        programArgs,
-        logger,
-        allowExecve = allowTerminate,
-        jvmRunner = build.artifacts.hasJvmRunner,
-        potentialMainClasses
-      ))
+      if (options.mainClass.mainClassLs.contains(true))
+        value {
+          options.mainClass
+            .maybePrintMainClasses(potentialMainClasses, shouldExit = allowTerminate)
+            .map(_ => None)
+        }
+      else {
+        val process = value {
+          maybeRunOnce(
+            build,
+            programArgs,
+            logger,
+            allowExecve = allowTerminate,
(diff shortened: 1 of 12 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1744,4 +1744,55 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+        "run",
+        extraOptions,
+        "."
+      )
+        .call(cwd = root, mergeErrIntoOut = true, check = false)
+      expect(res.exitCode == 1)
+      val output = res.out.text().trim
+      expect(output.contains(scalaFile1))
+      expect(output.contains(scalaFile2))
+      expect(output.contains(s"${scriptName}_sc"))
```
alexarchambault, line 1844: Maybe these checks could be made slightly more specific… We could have the letters `A` or `B` in the output for other reasons.

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -33,6 +33,7 @@ final case class BuildOptions(
   scriptOptions: ScriptOptions = ScriptOptions(),
   internal: InternalOptions = InternalOptions(),
   mainClass: Option[String] = None,
+  mainClassLs: Option[Boolean] = None,
```
alexarchambault: I wouldn't put that option here. Listing main classes should only be a `cli` module concern, not a concern of the `build` module (which is kind of a library, so as much as possible, it shouldn't call `sys.exit` or basic `println` on its own).

Listing main classes could rely on `Build.Successful#foundMainClasses`, and be handled right after we get a successful build in commands that accept `--main-class-list`. Basically, instead of running / packaging / etc., we would print the main class list.

Gedochao: should be fixed now

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
Gedochao: fixed

Gedochao:
It seems there's a compat issue on Windows, converting back to draft temporarily

tgodzik:
Could we perhaps add a warning in this case. I think some users could assume a different approach is more valid. For example if they run `scala-cli run .` and they have a script and a main class, they might expect to run the script. I guess overall this a very niche case, but maybe when we detect more we could do:
```
Running MyAppClass, also detected scripts A.sc, B.sc, You can run them with option... 
```
What do you think?

Gedochao:
> Could we perhaps add a warning in this case. I think some users could assume a different approach is more valid. For example if they run `scala-cli run .` and they have a script and a main class, they might expect to run the script. I guess overall this a very niche case, but maybe when we detect more we could do:
> 
> ```
> Running MyAppClass, also detected scripts A.sc, B.sc, You can run them with option... 
> ```
> 
> What do you think?

I think this is a valid suggestion. `--list-main-classes` could also be suggested  in such a warning, to raise awareness of it.

I guess we'd output such a warning to `stderr`, not to mess up any piping done from `scala-cli` to other tools?

Also, I wonder about how the script main classes should be mentioned, so that there's no misunderstanding for the users. The actual main class coming from `A.sc` is `A_sc` (the actual script wrapper name). When `--list-main-classes` is called, `A_sc` would be printed, which should be consistent with this warning.

@alexarchambault any thoughts on this?

Gedochao:
Warning message added as per the suggestion, example output:

```bash
▶ tree -a
.
├── Hello.scala
└── scripts
    ├── AnotherScript.sc
    └── Script.sc

1 directory, 3 files

▶ scala-cli .
Compiling project (Scala 3.1.1, JVM)
Compiled project (Scala 3.1.1, JVM)
Running Hello. Also detected script main classes: scripts.AnotherScript_sc, scripts.Script_sc
You can run any one of them by passing option --main-class, i.e. --main-class scripts.AnotherScript_sc
All available main classes can always be listed by passing option --list-main-classes
Hello world

```

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -66,12 +67,15 @@ object Run extends ScalaCommand[RunOptions] {
       build: Build.Successful,
       allowTerminate: Boolean
     ): Either[BuildException, (Process, CompletableFuture[_])] = either {
+      val potentialMainClasses = build.foundMainClasses()
+      value(options.mainClass.maybePrintMainClasses(potentialMainClasses))
```
alexarchambault: Ideally, if `allowTerminate` is false (in watch mode basically), we'd rather not have `maybePrintMainClasses` call `sys.exit` here. In that case, we should print the main class list, and that's it. Maybe we should change the return type, so that when listing main classes, this can return a `Right(…)`, without providing a `Process` and a `CompletableFuture`.

alexarchambault (commented):
We may want to make main class listing work in all or most commands that accept `--main-class-ls`, in `package` in particular. So we'd need to call `maybePrintMainClasses` from there too.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: That way, we'd print the main class list upon every source change.

Gedochao: Hey, this is a great idea, didn't think of it.
It should now work with `--watch` mode after my last commit.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -154,7 +168,7 @@ object Run extends ScalaCommand[RunOptions] {
       builds.main match {
         case s: Build.Successful =>
           val (process, onExit) = maybeRun(s, allowTerminate = true)
-            .orExit(logger)
+            .orExit(logger).getOrElse(sys.exit(1))
```
alexarchambault, line 171: The `sys.exit` is for when we list main classes, right? A `0` exit code would work fine I think (users requested the main class list themselves, listing them isn't really an error).

alexarchambault: Oh, it seems it's not supposed to be called anyway (as `allowTerminate` is true). A `sys.error("Can't happen")` would have done the job then (and be more helpful if ever it ends up being called anyway), but it's no big deal.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1744,4 +1796,96 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+      val res = os.proc(
+        TestUtil.cli,
+        "run",
+        extraOptions,
+        "."
+      )
+        .call(cwd = root, mergeErrIntoOut = true, check = false)
+      expect(res.exitCode == 1)
+      val output = res.out.text().trim
+      expect(output.contains(scalaFile1))
```
alexarchambault, line 1842: I'd still rather have called `.linesIterator.toVector` on `output`, and checked that some lines are equal to `scalaFile1`, `scalaFile2`, etc. But it's no big deal…

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1744,4 +1796,96 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+      val res = os.proc(
+        TestUtil.cli,
+        "run",
+        extraOptions,
+        ".",
+        "--list-main-classes"
+      )
+        .call(cwd = root)
+      val output = res.out.text().trim
+      expect(output.contains(scalaFile1))
```
alexarchambault, line 1886: Same as above (look at whole lines)

alexarchambault (approved):
I left some minor comments, which can be addressed or discussed later I think.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
Gedochao: I'll fix it in another PR

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
Gedochao: same as above, will fix it in another PR

PR #1099, opened 2022-06-13
Update Scala to 2.12.16

alexarchambault:


lwronski:
Test fail, because latest supported scala version is set to `2.12.15`, therefore scala prefix `2.12` is resolved to `2.12.15`.

Maybe we should remove this test `use expected default scala version for prefix scala version: 2.12` and test it only from `conf file` as [here](https://github.com/VirtusLab/scala-cli/blob/18ac25d233c827c4f5220a7aa64ea26895d115cd/modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala#L306)

alexarchambault:
Merging https://github.com/VirtusLab/scala-cli-scala-versions/pull/4 should make that test pass (and proceeding this way overall - bumping in scala-cli-scala-versions, then here - looks fine I guess).

Gedochao (approved):

lwronski (approved):

PR #1100, opened 2022-06-13
Force interactive mode for update command

lwronski:
The IMO the `update` command should enable interactive mode by default.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/VerbosityOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/VerbosityOptionsUtil.scala
@@ -6,6 +6,6 @@ import scala.build.interactive.Interactive._
 object VerbosityOptionsUtil {
   implicit class VerbosityOptionsOps(v: VerbosityOptions) {
     def interactiveInstance(forceEnable: Boolean = false) =
-      if (v.interactive.getOrElse(false) || forceEnable) InteractiveAsk else InteractiveNop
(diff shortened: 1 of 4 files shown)


Gedochao (approved):
Makes sense.

modules/cli/src/main/scala/scala/cli/commands/util/VerbosityOptionsUtil.scala
```diff
@@ -5,6 +5,7 @@ import scala.build.interactive.Interactive._
 
 object VerbosityOptionsUtil {
   implicit class VerbosityOptionsOps(v: VerbosityOptions) {
-    val interactiveInstance = if (v.interactive.getOrElse(false)) InteractiveAsk else InteractiveNop
+    def interactiveInstance(forceEnable: Boolean = false) =
+      if (v.interactive.getOrElse(false) || forceEnable) InteractiveAsk else InteractiveNop
```
alexarchambault: Why not make the parameter the default value?
```suggestion
      if (v.interactive.getOrElse(forceEnable)) InteractiveAsk else InteractiveNop
```

alexarchambault: If forced, the command-line option is simply ignored, while this way, users can override the default.

lwronski: Makes sense, I applied your suggestion

PR #1107, opened 2022-06-15
Publish tweaks + documentation

alexarchambault:
This tweaks things in the `publish` sub-command (allowing to list licenses or checksum types, better defaults, accept `github` as repository, …), and adds documentation for it.

PR #1113, opened 2022-06-15
Stop compiling most stuff with Scala 2

alexarchambault:


alexarchambault:
Goal of this was to help in https://github.com/VirtusLab/scala-cli/pull/1107/commits/32662be1b6accaf588a1518b736bf4772e6e8624. This commit adds a dependency between `cli-options` and `options`, and as `cli-options` is compiled with Scala 2, this was bringing the Scala 2 version of `options`, itself bringing the Scala 2 version of `build-macros`. Down-the-line, we had both the Scala 2 and Scala 3 versions of `build-macros`, which were conflicting (either at compile time or at runtime).

I thought disabling Scala 2 wherever we could was the way to go, but it turns out changing some `moduleDeps` to `compileModuleDeps` did the job in https://github.com/VirtusLab/scala-cli/pull/1107/commits/32662be1b6accaf588a1518b736bf4772e6e8624.

alexarchambault:
The CI compilation errors don't make any sense… 🤔 (everything compiles fine locally)

romanowski (approved):

PR #1115, opened 2022-06-16
Fix static linking of libsodium in Windows launcher

### LINKED ISSUE #1114: Missing libsodium.dll on windows
**Version(s)**

At least 0.1.7 and 0.1.8, the later tested both using coursier and the msi installer, on Windows.

**Describe the bug**

System error on launch, about missing libsodium.dll .

**To Reproduce**

On Windows (10), install, launch, say, `scala-cli version`, and notice the system popup.

**Expected behaviour**
Do not complain about libsodium.dll .

I added libsodium.dll which I found browsing github, and I get through then.

alexarchambault:
Fixes https://github.com/VirtusLab/scala-cli/issues/1114 ~(hopefully)~

### DIFF
--- a/project/settings.sc
+++ b/project/settings.sc
@@ -220,35 +220,15 @@ trait CliLaunchers extends SbtModule { self =>
         else "lib"
       os.copy.over(libPath, destDir / s"${prefix}sodiumjni.$ext")
     }
-    private def copyLibsodiumTo(cs: String, destDir: os.Path): Unit = {
-      val (osPart, suffix, relPath, ext) = sys.props.get("os.arch") match {
-        case Some("x86_64" | "amd64") =>
-          if (Properties.isWin)
-            ("win-64", "-h62dcd97_1", os.rel / "Library" / "lib" / "libsodium.lib", "lib")
-          else if (Properties.isLinux)
-            ("linux-64", "-h36c2ea0_1", os.rel / "lib" / "libsodium.a", "a")
-          else if (Properties.isMac) ("osx-64", "-hbcb3906_1", os.rel / "lib" / "libsodium.a", "a")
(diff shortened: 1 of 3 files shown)


alexarchambault:
I didn't try to reproduce https://github.com/VirtusLab/scala-cli/issues/1114 yet, but given its description, this *ought* to fix it.

alexarchambault:
Thinking about add a check using `dumpbin /dependents`, to check if the Windows launcher has unexpected DLL dependencies. But we can always merge this PR, as it fixes an important issue (Windows launcher not working).

project/settings.sc
```diff
@@ -220,35 +220,18 @@ trait CliLaunchers extends SbtModule { self =>
...
-        s"https://anaconda.org/conda-forge/libsodium/$libsodiumVersion/download/$osPart/libsodium-$libsodiumVersion$suffix.tar.bz2"
+        "https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip"
       ).call()
       val dir = os.Path(dirRes.out.text().trim, os.pwd)
-      val prefix =
-        if (Properties.isWin) ""
-        else "lib"
-      os.copy.over(dir / relPath, destDir / s"${prefix}sodium.$ext")
+      os.copy.over(
+        dir / "libsodium" / "x64" / "Release" / "v143" / "static" / "libsodium.lib",
```
Gedochao, line 232: always `x64`? we don't have to worry about `x86` compat?

alexarchambault: I don't think the launchers generated by native-image support x86 (native-image [uses the x64 native tools on Windows](https://www.graalvm.org/22.1/docs/getting-started/windows/)), so that shouldn't be a problem.

Gedochao (approved):

PR #1116, opened 2022-06-16
Add Scala Steward cron job

alexarchambault:


alexarchambault:
Not sure the token has the correct rights for Scala Steward (to open PRs and all).

lwronski:
> Not sure the token has the correct rights for Scala Steward (to open PRs and all).

`secrets.GITHUB_TOKEN` doesn't have right for trigger CI. IMO we should use deploy keys. 

Maybe something like this with using the internal scala-cli deploy key should work, I can test it:
```
steps:
       # Needed or else it won't find your repos file
       - uses: actions/checkout@v2
          with:
            ssh-key: ${{ secrets.SSH_PRIVATE_KEY_SCALA_CLI }}
```

PR #1122, opened 2022-06-22
Uninstall command

wleczny:


modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,35 @@
...
+
+object Uninstall extends ScalaCommand[UninstallOptions] {
+  def run(options: UninstallOptions, args: RemainingArgs): Unit = {
+    CurrentParams.verbosity = options.verbosity.verbosity
+    val binDirPath =
+      options.binDirPath.getOrElse(scala.build.Directories.default().binRepoDir / "scala-cli")
+    val destBinPath = binDirPath / options.binaryName
+
+    if (!options.force) {
+      println("Do you want to uninstall scala-cli [Y/n]")
```
lwronski: You can use interactive functions, it could be something similar as [here](https://github.com/VirtusLab/scala-cli/blob/d8af06401ab9aaf63eef5e95e0e1c0909d4ad4e5/modules/cli/src/main/scala/scala/cli/commands/Update.scala#L62)

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,35 @@
...
+      val response = readLine()
+      if (response != "Y") {
+        System.err.println("Abort")
+        sys.exit(1)
+      }
+    }
+    if (os.exists(destBinPath)) {
+      BloopExit.run(options.bloopExit, args)
+      os.remove.all(binDirPath)
+      println("Uninstalled sucessfully")
```
lwronski: I think, we should also remove cache from `scala-cli`.

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,35 @@
...
+
+    if (!options.force) {
+      println("Do you want to uninstall scala-cli [Y/n]")
+      val response = readLine()
+      if (response != "Y") {
+        System.err.println("Abort")
+        sys.exit(1)
+      }
+    }
+    if (os.exists(destBinPath)) {
```
lwronski, line 37: It can be useful to add some short description in comment what things are you removing:

```
// clean bloop server

// removing scala-cli launcher 

// removing scala-cli caches
```

modules/cli-options/src/main/scala/scala/cli/commands/BloopExitOptions.scala
```diff
@@ -17,6 +17,7 @@ final case class BloopExitOptions(
 // format: on
 
 object BloopExitOptions {
-  implicit lazy val parser: Parser[BloopExitOptions] = Parser.derive
-  implicit lazy val help: Help[BloopExitOptions]     = Help.derive
+  implicit lazy val parser: Parser[BloopExitOptions]                  = Parser.derive
+  implicit lazy val parserAux: Parser.Aux[BloopExitOptions, parser.D] = parser
```
romanowski, line 23: Why do we need `Aux` here?

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,47 @@
...
+      // exit bloop server
+      BloopExit.run(options.bloopExit, args)
+      // remove scala-cli launcher
+      os.remove.all(binDirPath)
+      // remove scala-cli caches
+      if (!options.skipCache) os.remove.all(cacheDir)
+      println("Uninstalled sucessfully")
+    }
+    else if (!Update.isScalaCLIInstalledByInstallationScript()) {
```
romanowski: Shouldn't we check that in the first place and do not uninstall scala-cli if installed any other way? Or only if `--force` or other option is provided?

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,47 @@
...
+      System.err.println(
+        "Scala CLI was not installed by the installation script, please use your package manager to uninstall scala-cli."
+      )
+      sys.exit(1)
+    }
+    else {
+      System.err.println(s"Could't find $destBinPath.")
```
romanowski:
```suggestion
      System.err.println(s"Could't find scala-cli binary at $destBinPath.")
```

romanowski (commented):
Do we also uninstall completions if installed?

modules/cli-options/src/main/scala/scala/cli/commands/BloopExitOptions.scala
Gedochao: I think it's because `BloopExitOptions` is being used with `@Recurse` in `UninstallOptions` (and `@Recurse` uses an implicit `Parser.Aux`)

website/docs/_advanced_install.mdx
```diff
@@ -291,6 +291,184 @@ sdk install scalacli
...
+
+If Scala CLI was installed manually you can uninstall it with:
+
+```bash
+rmdir "%USERPROFILE%/scala-cli"
+```
+</TabItem>
+<TabItem value="installer-u">
+
+If Scala CLI was installed via installer you can uninstall it in Cotrol Panel.
```
Gedochao:
```suggestion
If Scala CLI was installed via the installer you can uninstall it in the Control Panel.
```

website/docs/_advanced_install.mdx
```diff
@@ -291,6 +291,184 @@ sdk install scalacli
...
+{label: 'Apt', value: 'apt-u'},
+{label: 'Deb', value: 'deb-u'},
+{label: 'Yum', value: 'yum-u'},
+{label: 'Rpm', value: 'rpm-u'},
+{label: 'Alpine', value: 'alpine-u'},
+]}
+>
+<TabItem value="installation-script-u">
+
+If Scala CLI was installed via installation script you can uninstall it with:
```
Gedochao:
```suggestion
If Scala CLI was installed via the installation script, you can uninstall it with:
```

website/docs/_advanced_install.mdx
```diff
@@ -291,6 +291,184 @@ sdk install scalacli
...
+defaultValue="installation-script-u"
+values={[
+{label: 'Installation script', value: 'installation-script-u' },
+{label: 'Manual', value: 'manual-u'},
+{label: 'Brew', value: 'brew-u'},
+]}
+>
+<TabItem value="installation-script-u">
+
+If Scala CLI was installed via installation script you can uninstall it with:
```
Gedochao:
```suggestion
If Scala CLI was installed via the installation script, you can uninstall it with:
```

modules/cli-options/src/main/scala/scala/cli/commands/UninstallOptions.scala
```diff
@@ -0,0 +1,32 @@
+package scala.cli.commands
+
+import caseapp._
+import java.nio.file.Path
```
alexarchambault: NIT I'm surprised the organize imports scalafix rule didn't say anything, it should have required an empty line here:
```suggestion
import caseapp._

import java.nio.file.Path
```

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,49 @@
...
+      options.bloopExit.logging.verbosityOptions.interactiveInstance(forceEnable = true)
+
+    val binDirPath =
+      options.binDirPath.getOrElse(scala.build.Directories.default().binRepoDir / "scala-cli")
+    val destBinPath = binDirPath / options.binaryName
+    val cacheDir    = scala.build.Directories.default().cacheDir
+
+    if (
+      !Update.isScalaCLIInstalledByInstallationScript() && (options.binDir == None || (options.binDir != None && !options.force))
+    ) {
```
alexarchambault: NIT The condition can be simplified a bit, and calling `isEmpty` on an option is more idiomatic than comparing it to `None` (and hopefully, scalafmt is fine having everything on the same line with those changes):
```suggestion
    if (!Update.isScalaCLIInstalledByInstallationScript() && (options.binDir.isEmpty || !options.force)) {
```

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,49 @@
...
+      interactive.confirmOperation(msg).getOrElse(fallbackAction())
+    }
+    if (os.exists(destBinPath)) {
+      // exit bloop server
+      BloopExit.run(options.bloopExit, args)
+      // remove scala-cli launcher
+      os.remove.all(binDirPath)
+      // remove scala-cli caches
+      if (!options.skipCache) os.remove.all(cacheDir)
+      println("Uninstalled sucessfully")
```
alexarchambault: NIT Better use a logger here and call `logger.message`. That way, the message won't be printed in quiet mode;
```suggestion
      logger.message("Uninstalled sucessfully")
```
(needs `logger` to be created above, with `options.bloopExit.logging.logger` I think).

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,49 @@
...
+      val msg = s"Do you want to uninstall scala-cli?"
+      interactive.confirmOperation(msg).getOrElse(fallbackAction())
+    }
+    if (os.exists(destBinPath)) {
+      // exit bloop server
+      BloopExit.run(options.bloopExit, args)
+      // remove scala-cli launcher
+      os.remove.all(binDirPath)
+      // remove scala-cli caches
+      if (!options.skipCache) os.remove.all(cacheDir)
```
alexarchambault: NIT Maybe some `logger.debug` calls could be inserted before each action here, to help debug things if anything goes wrong.

modules/integration/src/test/scala/scala/cli/integration/InstallHomeTests.scala
```diff
@@ -61,6 +61,24 @@ class InstallHomeTests extends munit.FunSuite {
...
+    // format: off
+    val cmdUninstall = Seq[os.Shellable](
+      TestUtil.cli, "uninstall",
+      "--binary-name", dummyScalaCliBinName,
+      "--bin-dir", binDirPath
+    ) 
+    ++ (if(force) Seq[os.Shellable]("--force") else Seq.empty)
+    ++ (if(skipCache) Seq[os.Shellable]("--skip-cache") else Seq.empty)
+    // format: on
+    os.proc(cmdUninstall).call(cwd = root)
```
alexarchambault: NIT We don't have to always resort to always explicitly typing with `os.Shellable` here:
```suggestion
    // format: off
    val cmdUninstall = Seq[os.Shellable](
      TestUtil.cli, "uninstall",
      "--binary-name", dummyScalaCliBinName,
      "--bin-dir", binDirPath
    ) 
    // format: on
    val forceOpts = if (force) Seq("--force") else Seq.empty
    val skipCacheOpts = if (skipCache) Seq("--skip-cache") else Seq.empty
    os.proc(cmdUninstall, forceOpts, skipCacheOpts).call(cwd = root)
```

modules/cli-options/src/main/scala/scala/cli/commands/UninstallOptions.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import caseapp._
+
+import java.nio.file.Path
+
+// format: off
+@HelpMessage("Unistall scala-cli - only works when installed by the installation script")
+final case class UninstallOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
```
alexarchambault: We shouldn't need those. The `BloopExitOptions` should be enough to get a logger, with
```scala
import scala.cli.commands.util.CommonOps._
options.bloopExit.logging.logger
```

modules/cli-options/src/main/scala/scala/cli/commands/BloopExitOptions.scala
```diff
@@ -19,4 +19,6 @@ final case class BloopExitOptions(
 object BloopExitOptions {
   implicit lazy val parser: Parser[BloopExitOptions] = Parser.derive
   implicit lazy val help: Help[BloopExitOptions]     = Help.derive
+  // Parser.Aux for using BloopExitOptions with @Recourse in other options
```
Gedochao: NIT: typo
```suggestion
  // Parser.Aux for using BloopExitOptions with @Recurse in other options
```

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    val binDirPath =
+      options.binDirPath.getOrElse(scala.build.Directories.default().binRepoDir / "scala-cli")
+    val destBinPath = binDirPath / options.binaryName
+    val cacheDir    = scala.build.Directories.default().cacheDir
+
+    if (
+      !Update.isScalaCLIInstalledByInstallationScript() && (options.binDir.isEmpty || !options.force)
+    ) {
+      System.err.println(
```
Gedochao: IMO can favor using `logger.error` here

modules/cli/src/main/scala/scala/cli/commands/Uninstall.scala
```diff
@@ -0,0 +1,54 @@
...
+      // remove scala-cli launcher
+      logger.debug("Removing scala-cli binary...")
+      os.remove.all(binDirPath)
+      // remove scala-cli caches
+      logger.debug("Removing scala-cli cache directory...")
+      if (!options.skipCache) os.remove.all(cacheDir)
+      logger.message("Uninstalled sucessfully.")
+    }
+    else {
+      System.err.println(s"Could't find scala-cli binary at $destBinPath.")
```
Gedochao: `logger.error` once more

Gedochao (approved):

PR #1123, opened 2022-06-23
Export and import coursier caches

### LINKED ISSUE #1061: Allow users to export and import coursier caches…
…so that users with limited internet connections (like can happen in some corporate environments) can import pre-baked coursier caches, to run various Scala CLI commands.

We could add commands allowing to export (in a `.tar.gz` say) and import a whole coursier cache, either in the [coursier](https://github.com/coursier/coursier) CLI or in Scala CLI itself.

Users wanting to distribute such a cache would be expected to generate it with commands like
```bash
$ export COURSIER_CACHE="$(pwd)/cache"
$ scala-cli run … # run a few Scala CLI commands that you want to work fine with the distribution
$ scala-cli cache export -o cache.tar.gz 
```

To import the distribution, one would have to run things like
```bash
$ scala-cli cache import cache.tar.gz
```

An important point to pay attention to when implementing this would be not to blindly copy the files of a distribution in the local coursier cache, but rather group together files that are logically tied (if `foo/a` is in cache, then `foo/.a.*` is tied to it), and import each group as a whole (if `foo/a` is already there locally, compare `foo/.a.checked` to the one to be imported, and either don't import `foo/a` at all, or clear the local `foo/a` and linked files first).

lwronski:
To export cache items that are required on per-project basis in scala-cli users should run following commands:
```
export COURSIER_CACHE="$(pwd)/cache/coursier"
export COURSIER_ARCHIVE_CACHE="$(pwd)/cache/arch"
export XDG_CACHE_HOME="$(pwd)/cache/bloop"

scala-cli cache export -o cache.tar.gz ${options which will be used in scala-cli after importing cache, for example `-S 3 Hello.scala`}
```

and then, users have to run:
```
scala-cli cache import --cache-path cache.tar.gz
```
to import coursier and bloop caches

Cache export and import commands requires installed `tar` application.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/CacheExport.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/CacheExport.scala
@@ -1,0 +1,122 @@
+package scala.cli.commands
+
+import caseapp._
+import coursier.cache.FileCache
+import coursier.paths.CoursierPaths
+import coursier.util.Task
+
+import scala.build.internal.CsLoggerUtil.CsJavaHomeExtensions
+import scala.build.internal.{Constants, OsLibc}
+import scala.build.options.BuildOptions
+import scala.build.{Inputs, Logger, Os}
+import scala.cli.CurrentParams
+import scala.cli.commands.util.SharedOptionsUtil._
+import scala.cli.internal.BloopCache
+import scala.util.control.NonFatal
+
+object CacheExport extends ScalaCommand[CacheExportOptions] {
+  override def group = "Main"
+
+  override def names: List[List[String]] = List(
+    List("cache", "export")
+  )
+
+  def run(options: CacheExportOptions, args: RemainingArgs): Unit = {
+    maybePrintGroupHelp(options)
+    CurrentParams.verbosity = options.runOptions.shared.logging.verbosity
+
+    val inputs = options.runOptions.shared.inputsOrExit(args)
+    CurrentParams.workspaceOpt = Some(inputs.workspace)
+    val logger                 = options.runOptions.shared.logger
+    val cache: FileCache[Task] = options.runOptions.shared.coursierCache
+
+    val buildOptions = options.runOptions.shared.buildOptions()
+    val dest         = options.output.getOrElse("cache.tar.gz")
+    val destPath     = os.Path(dest, Os.pwd)
+
+    // clean workspace .scala-build
+    cleanWorkspace(logger, inputs)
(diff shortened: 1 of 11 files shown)


modules/cli/src/main/scala/scala/cli/commands/CacheExport.scala
```diff
@@ -0,0 +1,122 @@
...
+    CurrentParams.workspaceOpt = Some(inputs.workspace)
+    val logger                 = options.runOptions.shared.logger
+    val cache: FileCache[Task] = options.runOptions.shared.coursierCache
+
+    val buildOptions = options.runOptions.shared.buildOptions()
+    val dest         = options.output.getOrElse("cache.tar.gz")
+    val destPath     = os.Path(dest, Os.pwd)
+
+    // clean workspace .scala-build
+    cleanWorkspace(logger, inputs)
```
Gedochao, line 38: why is it necessary to clean it?

modules/cli/src/main/scala/scala/cli/commands/CacheExport.scala
```diff
@@ -0,0 +1,122 @@
...
+    // force to download java to arch cache dir
+    downloadJava(cache, buildOptions)
+
+    val csCachePath       = os.Path(CoursierPaths.cacheDirectory())
+    val csArchPath        = os.Path(CoursierPaths.archiveCacheDirectory())
+    val bloopCacheDirPath = os.Path(BloopCache.cacheDir())
+
+    val tmpCacheDir = os.temp.dir(prefix = "scala-cli-cache")
+
+    println(s"Starting exporting caches to $destPath")
```
Gedochao, line 52: shouldn't we favor using a logger here? no need to print to `stdout`, no?

modules/cli/src/main/scala/scala/cli/commands/CacheImport.scala
```diff
@@ -0,0 +1,64 @@
...
+  override def names: List[List[String]] = List(
+    List("cache", "import")
+  )
+
+  def run(options: CacheImportOptions, args: RemainingArgs): Unit = {
+    maybePrintGroupHelp(options)
+    CurrentParams.verbosity = options.logging.verbosity
+
+    val cache = options.cachePath.getOrElse {
+      System.err.println(
```
Gedochao, line 21: similar, `logger.error` should do

Gedochao:
Some docs for both commands would be useful, IMO

PR #1129, opened 2022-06-24
Add --spark / --standalone-spark / --hadoop to the run command

alexarchambault:


alexarchambault:
Includes https://github.com/VirtusLab/scala-cli/pull/1086 ~and https://github.com/VirtusLab/scala-cli/pull/1163~ for now.

ekrich:
This is a great idea to help Scala Spark adoption.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -327,33 +373,199 @@ object Run extends ScalaCommand[RunOptions] {
...
+                scratchDirOpt = scratchDirOpt
+              )
+              Right((proc, None))
+            }
+          case RunMode.SparkSubmit =>
+            // FIXME Get Spark.sparkModules via provided settings?
+            val providedModules = Spark.sparkModules
+            val providedFiles =
+              value(PackageCmd.providedFiles(build, providedModules, logger)).toSet
+            val customSubmitOptions: Seq[String] = Nil // ???
```
Gedochao: seems unnecessary (?)

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -327,33 +373,199 @@ object Run extends ScalaCommand[RunOptions] {
...
+                mainClass,
+                args,
+                logger,
+                allowExecve = allowExecve,
+                useManifest = build.options.notForBloopOptions.runWithManifest,
+                scratchDirOpt = scratchDirOpt
+              )
+              Right((proc, None))
+            }
+          case RunMode.SparkSubmit =>
```
Gedochao, line 406: NIT I wonder if it wouldn't make sense to extract the Spark-related logic to a trait, like `RunSpark` (so that we could keep `Run.scala` at a reasonable line count)

modules/cli/src/main/scala/scala/cli/commands/packaging/Spark.scala
```diff
@@ -23,4 +23,10 @@ object Spark {
 
   def sparkModules: Seq[AnyModule] =
     names.map(name => mod"org.apache.spark::spark-$name")
+
+  def hadoopModules: Seq[AnyModule] =
+    Seq(
+      // TODO Add more for Hadoop 2, maybe for 3 too
```
Gedochao, line 29: NIT might be worth it to document in a ticket and leave a reference in this comment

Gedochao (approved):
left some nitpicks, otherwise LGTM

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: Yep, should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: Good idea, done.

modules/cli/src/main/scala/scala/cli/commands/packaging/Spark.scala
alexarchambault: I'm not sure... Let's wait for feedback on those features in a first time.

PR #1133, opened 2022-06-27
Update Scala Native to 0.4.5

alexarchambault:


alexarchambault:
I had missed https://github.com/VirtusLab/scala-cli/pull/1132, but the changes here should be better.

Gedochao (approved):

PR #1142, opened 2022-06-29
Run and test ESModule scripts as ES Modules

### LINKED ISSUE #1104: `jsModuleKind "es"` does not work with "run" when importing a module
**Version(s)**

Latest release and nightly (via `scala-cli --cli-version nightly`)

**Describe the bug**

Working with and Importing ES modules in a scala.js cli script does not work when running with `scala-cli run`, but does work when packaging the file and running with `node ./file.mjs`.

**To Reproduce**

Repro repo: https://github.com/hugo-vrijswijk/scala-cli-esm-bug (also see the [failing action](https://github.com/hugo-vrijswijk/scala-cli-esm-bug/runs/6878487740?check_suite_focus=true))

Given an ES module (`foo.mjs)`:

```javascript
export function double(x) {
  return x * 2;
}
```

Importing the file in scala.js:

```scala
//> using jsModuleKind "es"

@js.native
@JSImport("./foo.mjs", "double")
def double(x: Int): Int = js.native

@main def main =
  val n = 4
  println(s"Twice $n is ${double(n)}")
```

Calling `scala-cli run .` gives this error:
```
(node:2019) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/tmp/main11106905602926452705.js:2
import * as $i_$002e$002ffoo$002emjs from "./foo.mjs";
^^^^^^
SyntaxError: Cannot use import statement outside a module
```

Running `scala-cli package -o index.mjs && node ./index.mjs` works fine. It does not matter if there's a package.json with `"type": "module"` present in the working directory

**Expected behaviour**

Output of `scala-cli run .` and packaging is the same when running ES modules

hugo-vrijswijk:
Fixes #1104

`Input.Script` loads scripts with `require`, which does not work with ES modules. With this change `Script.ESModule` is used for es modules, as well as the file extension is changed to let Nodejs know we're working with an ES Module

This probably needs some tests, but I am not sure where to start with them

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/Runner.scala
+++ b/modules/build/src/main/scala/scala/build/internal/Runner.scala
@@ -161,7 +161,8 @@ object Runner {
     logger: Logger,
     allowExecve: Boolean = false,
     jsDom: Boolean = false,
-    sourceMap: Boolean = false
+    sourceMap: Boolean = false,
+    esModule: Boolean = false
   ): Process = {
 
     import logger.{log, debug}
@@ -201,7 +202,10 @@ object Runner {
       sys.error("should not happen")
     }
     else {
-      val inputs = Seq(Input.Script(entrypoint.toPath))
+      val inputs = Seq(
+        if (esModule) Input.ESModule(entrypoint.toPath)
+        else Input.Script(entrypoint.toPath)
+      )
 
       val config    = RunConfig().withLogger(logger.scalaJsLogger)
       val processJs = envJs.start(inputs, config)
@@ -309,7 +313,8 @@ object Runner {
     args: Seq[String],
     testFrameworkOpt: Option[String],
     logger: Logger,
-    jsDom: Boolean
+    jsDom: Boolean,
+    esModule: Boolean
   ): Either[TestError, Int] = either {
     import org.scalajs.jsenv.Input
     import org.scalajs.jsenv.nodejs.NodeJSEnv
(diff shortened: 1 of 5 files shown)


lwronski:
> Fixes #1104
> 
> `Input.Script` loads scripts with `require`, which does not work with ES modules. With this change `Script.ESModule` is used for es modules, as well as the file extension is changed to let Nodejs know we're working with an ES Module
> 
> This probably needs some tests, but I am not sure where to start with them

@hugo-vrijswijk Thanks for contribution, 

If you want to add some integration test, you can do it in `RunTestDefiniotions.scala`, [here](https://github.com/VirtusLab/scala-cli/blob/28bedb3e6b0b9180ee2ba8f01402b1a192d64fe9/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala#L79) you can find another test for `Scala.js`

hugo-vrijswijk:
@lwronski thanks for the pointer. I've added two tests. If you could give this a review, that'd be great!

Gedochao (approved):

Gedochao:
Great contribution @hugo-vrijswijk!

alexarchambault (approved):
Nice, thanks @hugo-vrijswijk!

hugo-vrijswijk:
Just in time to miss the 0.1.9 release 😅

lwronski:
@hugo-vrijswijk You can try to run the nightly version of Scala CLI with your fix:
```
scala-cli --cli-version nightly hello.sc --js
```

about `cli-version` parameter you can find more detail [here](https://scala-cli.virtuslab.org/docs/commands/basics/#scala-cli-version).

Issue #1098, opened 2022-06-10
Explicit handling of paths in using directives


prolativ:
**Is your feature request related to a problem? Please describe.**

Some using directives take paths as arguments. These might be some directives specific to scala-cli itself or something that is passed further to some other tool (e.g. options passed to the compiler or a compiler plugin). When relative paths are used it might be unclear if the programmer's intent was to:
1) pass the relative path unmodified to the underlying tool
2a) treat the path as relative to the file in which the directive is located
2b) treat the path as relative to the root of the build (e.g. `foo/bar` for `scala-cli run foo/bar`)
2c) treat the path as relative to the current working directory (where `scala-cli` is run from)

**Describe the solution you'd like**
To avoid the confusion, file paths in using directives could be represented with some special syntax distinct from bare string literals. This would make it clear whether a user meant `1` or `2`. Potentially a distinction between the subcases of `2` could be made if we find it useful to support all of them, although IMO it would be enough for most use cases to support only `2a`.

Taking https://github.com/VirtusLab/scala-cli/issues/1072 as an example, the problematic directive
```
//> using options  "-Ycheck:instrumentCoverage", "-coverage-out:.", "-sourceroot:."
```
could be replaced with something like
```
//> using options  "-Ycheck:instrumentCoverage", "-coverage-out:"${.}, "-sourceroot:"${.}
```
although I'm not insisting on `${...}` as the proper syntax - this is just to show the concept.
Then `${.}` would get substituted by scala-cli with the absolute path of the directory containing the file with the using directive (assuming we're considering `2a`) and the option containing the absolute path would get passed to the compiler. Paths defined as quoted strings wouldn't be modified in any way.
The example also shows that we would need some way to concatenate the resolved paths with some string literals to form proper options for the compiler or some other tool.

lwronski:
@prolativ Thanks for reporting,
your proposed syntax makes sense, but I would slightly modify it.

```
"-coverage-out:"${.}
```
will be replaced to 
```
"-coverage-out:${.}"
```

and if you want to skip a special character, users should add and additional dollar `$`:
```
"-coverage-out:$${.}"
```
`scala-cli` convert it options to `"-coverage-out:${.}"`

prolativ:
I wouldn't like to go too far but we should probably keep in mind that using this exact syntax could cause some complications if we wanted to have some kind of more general variable substitution in strings. E.g. we could support something like
```scala
//> using value "fooVersion" "1.2.3"
[...]

PR #1157, opened 2022-07-06
Define dependency to scala files in using directives

### LINKED ISSUE #1146: Add possibility for scripts to define dependency on other scala files
### Is your feature request related to a problem? Please describe.

The problem occurred for me when I was migrating a big pile of simple bash scripts to scala-cli.

I have multiple `*.sc` files, each with scala-cli in shebang, so I can just call them the same way I call bash scripts (just `scripts/buildBase.sc`) . I also want to have `utils.scala` file with common definitions for all the scripts. It is not possible to include the utils without needing to specify them explicitly every time I call a script. 

`scala-cli scripts/buildBase.sc scripts/utils.scala` is rather long in comparison with `scripts/buildBase.sc`.

### Describe the solution you'd like

I would like to be able to specify something like `//> using file "utils.scala"` which would add utils file to the classpath every time I compile and run my script. 

### Describe alternatives you've considered

I'm now calling my scripts using `scala-cli scripts --main-class buildBase_sc`. This is not elegant. Also, It loads dozens of scripts on classpath which may result in some unexpected problems in rare cases.

### Additional context

In bash, the user is able to just `source` other scripts. The necessity of specifying each scala dependency on every script invocation may discourage people from migrating scripts from bash to scala.

lwronski:


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -135,6 +135,18 @@ object CrossSources {
...
         .left.map(CompositeBuildException(_))
         .map(_.flatten)
+
+    val preprocessedSources = {
+      val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
+
+      val sourcesFromDirectives =
+        preprocessedInputFromArgs
+          .flatMap(_.options)
+          .flatMap(_.classPathOptions.extraSourceFiles).distinct
```
alexarchambault: NIT (readability)
```suggestion
          .flatMap(_.classPathOptions.extraSourceFiles)
          .distinct
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -201,4 +213,22 @@ object CrossSources {
...
     CrossSources(paths, inMemory, defaultMainClassOpt, resourceDirs, buildOptions)
   }
+
+  private def resolveInputsFromSources(sources: Seq[os.Path]) =
+    sources.map { sourcePath =>
+      lazy val dir     = sourcePath / os.up
+      lazy val subPath = sourcePath.subRelativeTo(dir)
+      if (sourcePath.ext == "scala") Right(Inputs.ScalaFile(dir, subPath))
+      else if (sourcePath.ext == "sc") Right(Inputs.Script(dir, subPath))
+      else if (sourcePath.ext == "java") Right(Inputs.JavaFile(dir, subPath))
```
alexarchambault: Couldn't `sourcePath` be a directory too?

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -201,4 +213,22 @@ object CrossSources {
...
+      else {
+        val msg =
+          if (os.exists(sourcePath))
+            s"$sourcePath: unrecognized source type (expected .scala, .sc or .java extension) in using directive."
+          else s"$sourcePath: not found path defined in using directive."
+        Left(new MalformedDirectiveError(msg, Nil))
```
alexarchambault: Couldn't we keep the positions in `extraSourceFiles`, and pass them here?

modules/directives/src/main/scala/scala/build/errors/WrongSourcePathError.scala
```diff
@@ -0,0 +1,6 @@
+package scala.build.errors
+
+class WrongSourcePathError(exceptionMessage: String)
+    extends BuildException(message =
+      s"""The file path argument in the using directives at could not be found!
+         |$exceptionMessage""".stripMargin) {}
```
alexarchambault: NIT
```suggestion
         |$exceptionMessage""".stripMargin)
```

modules/directives/src/main/scala/scala/build/preprocessing/directives/UsingSourceDirectiveHandler.scala
```diff
@@ -0,0 +1,56 @@
...
+      ProcessedDirective(
+        Some(BuildOptions(
+          classPathOptions = ClassPathOptions(
+            extraSourceFiles = pathSequence
+          )
+        )),
+        Seq.empty
+      )
+
+    }
```
alexarchambault: NIT (readability)
```suggestion
    checkIfValuesAreExpected(scopedDirective)
      .flatMap { groupedScopedValuesContainer =>
        groupedScopedValuesContainer.scopedStringValues
          .map {
            case ScopedValue(positioned, _) =>
              val eitherRootPathOrBuildException =
                Directive.osRoot(scopedDirective.cwd, positioned.positions.headOption)
              eitherRootPathOrBuildException.flatMap { root =>
                Try(os.Path(positioned.value, root)).toEither.left.map { ex =>
                  new WrongSourcePathError(exception.getLocalizedMessage)
                }
              }
          }
          .sequence
          .left.map(CompositeBuildException(_))
      }
      .map { pathSequence =>
        ProcessedDirective(
          Some(BuildOptions(
            classPathOptions = ClassPathOptions(
              extraSourceFiles = pathSequence
            )
          )),
          Seq.empty
        )
      }
```

alexarchambault: Or even, with `either` / `value`
```suggestion
    val groupedScopedValuesContainer = value(checkIfValuesAreExpected(scopedDirective))
    val pathSequence = value {
      groupedScopedValuesContainer.scopedStringValues
        .map {
          case ScopedValue(positioned, _) =>
            val eitherRootPathOrBuildException =
              Directive.osRoot(scopedDirective.cwd, positioned.positions.headOption)
            eitherRootPathOrBuildException.flatMap { root =>
              Try(os.Path(positioned.value, root)).toEither.left.map { ex =>
                new WrongSourcePathError(ex.getLocalizedMessage)
              }
            }
        }
        .sequence
        .left.map(CompositeBuildException(_))
    }
      
    ProcessedDirective(
      Some(BuildOptions(
        classPathOptions = ClassPathOptions(
          extraSourceFiles = pathSequence
        )
      )),
      Seq.empty
    )
```

alexarchambault: We can also keep the cause of the underlying exception `ex` in `WrongSourcePathError` (just pass `cause` in its constructor to `BuildException`), that can help sometimes.

modules/options/src/main/scala/scala/build/options/ClassPathOptions.scala
```diff
@@ -7,6 +7,7 @@ import scala.build.Positioned
 final case class ClassPathOptions(
   extraRepositories: Seq[String] = Nil,
   extraClassPath: Seq[os.Path] = Nil,
+  extraSourceFiles: Seq[os.Path] = Nil,
```
alexarchambault: Not sure we should add that field here. Maybe it should be added in `PostBuildOptions` (even though it's more a "pre-build" thing here) or in `InternalOptions`, which are not taken into account when computing a hash of the options. But then, we have to be sure we use an `Inputs` instance that has the new inputs when computing a hash of it (in `Build.scala` IIRC).

alexarchambault (commented):
There are a few questions I'm wondering about the new inputs:
- are they taken into account when watching files on disk? (IIRC, we only watch things from the initial `Inputs` instance)
- do they work fine from BSP? (they should, but I'd add a test for that in `BspTestDefinitions` just in case)
- are relative paths computed fine? like what if the `using file` directive is in a source in a sub-directory, like in `src/foo/bar/Something.scala`?

website/docs/commands/run.md
```diff
@@ -64,6 +64,40 @@ JVMs are [managed by coursier](https://get-coursier.io/docs/cli-java#managed-jvm
 (New JVM versions are automatically checked daily, and updates for those are - manually - merged
 swiftly.)
 
+## Define source files in using directives
+
+You can also add sources files with using directive `//> using file`:
```
Gedochao:
```suggestion
You can also add source files with the using directive `//> using file`:
```

website/docs/commands/run.md
```diff
@@ -64,6 +64,40 @@ JVMs are [managed by coursier](https://get-coursier.io/docs/cli-java#managed-jvm
...
+
+```bash
+scala-cli Main.scala
+```
+
+<!-- Expected:
+Hello World
+-->
+
+It is also possible to pass multiple paths to source files in one using directive:
```
Gedochao:
```suggestion
It is also possible to pass multiple paths to source files in a single using directive:
```

website/docs/commands/run.md
```diff
@@ -64,6 +64,40 @@ JVMs are [managed by coursier](https://get-coursier.io/docs/cli-java#managed-jvm
...
+ ```
+ //> using files "Utils.scala" "Helper.scala"
+ ```
+
+ Note that `//> using file` using directive supports only `Java`, `Scala`, `ScalaScripts` files or directory.
```
Gedochao:
```suggestion
 Note that the `//> using file` using directive only supports `.java`, `.scala`, `.sc` files or a directory. 
```

Gedochao (commented):
I think it might be cool to highlight this feature in the `scripts.md` guide as well, since that's where this feature would probably be at its most useful.

lwronski:
> There are a few questions I'm wondering about the new inputs:
> 
> * are they taken into account when watching files on disk? (IIRC, we only watch things from the initial `Inputs` instance)

Now it should work, and upon every change, I re-compute actual `Inputs` instance adding sources from using directives. 

> * do they work fine from BSP? (they should, but I'd add a test for that in `BspTestDefinitions` just in case)

Yes, It works fine from BSP and I also added one `BSP` test.


> * are relative paths computed fine? like what if the `using file` directive is in a source in a sub-directory, like in `src/foo/bar/Something.scala`?

I think, it is out of  the scope of this PR, it should be fixed in #1098. Now, it works the same as path resolving for the `resourceDir` parameter in using directive.

Gedochao (approved):

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
```diff
@@ -29,10 +30,27 @@ object Bsp extends ScalaCommand[BspOptions] {
...
+              CrossSources.forInputs(
+                initialInputs,
+                Sources.defaultPreprocessors(
+                  buildOptions0.scriptOptions.codeWrapper.getOrElse(CustomCodeWrapper),
+                  buildOptions0.archiveCache,
+                  buildOptions0.internal.javaClassNameVersionOpt
+                ),
+                persistentLogger
+              ).left.map((_, Scope.Main))
+                .map(_._2).getOrElse(initialInputs)
```
alexarchambault: It seems we're discarding an error here (if what `forInputs` returns is left, the error on the left is simply discarded). Before the changes of this PR, maybe you should change `argsToInputs` to a `Seq[String] => Either[BuildException, Inputs]` (and change its use sites accordingly, so that they create / propagate `BuildException`s), so that if `forInputs` returns a `Left`, it gets passed to the caller of `argsToInputs`.

modules/directives/src/main/scala/scala/build/errors/WrongSourcePathError.scala
```diff
@@ -0,0 +1,8 @@
+package scala.build.errors
+
+class WrongSourcePathError(exceptionMessage: String, cause: Throwable)
+    extends BuildException(
+      message = s"""The file path argument in the using directives at could not be found!
+                   |$exceptionMessage""".stripMargin,
```
alexarchambault: NIT Just in case, to be safer, I wouldn't call `stripMargin` on `exceptionMessage` (in case it has some `|` at the beginning of some lines)
```suggestion
                   |""".stripMargin + exceptionMessage,
```

website/docs/guides/scripts.md
```diff
@@ -163,6 +163,32 @@ world
 
 </ChainedSnippets>
 
+### Define source files in using directives
+
+You can also add source files with the using directive `//> using file` in Scala Scripts:
```
alexarchambault: I wouldn't capitalize scripts:
```suggestion
You can also add source files with the using directive `//> using file` in Scala scripts:
```

website/docs/guides/scripts.md
```diff
@@ -163,6 +163,32 @@ world
...
+
+println(Utils.message)
+```
+```scala title=Utils.scala
+object Utils {
+  val message = "Hello World"
+}
+```
+
+`scala-cli` compiles and adds to the class path source `Utils.scala`.
```
alexarchambault: NIT Phrasing (the source itself isn't added to the class path per se, more its… "build products" I'd say)
```suggestion
`scala-cli` takes into account and compiles `Utils.scala`.
```

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
lwronski: ~~Sure,  so I changed returned type of `argsToInputs` in this [commit](https://github.com/VirtusLab/scala-cli/pull/1157/commits/64d3425b59c7a5ab2167e68707536af482cffe40).~~

lwronski: I don't think we're discarding an error because it is processed in `BspImpl` and then reported in diagnostics.

If using directives contains errors then we should start `Bsp` with `initialInputs` which contains inputs only from args, therefore I added `.getOrElse(initialInputs)`

modules/build/src/test/scala/scala/build/tests/TestInputs.scala
```diff
@@ -44,7 +45,9 @@ final case class TestInputs(
         forcedWorkspace = forcedWorkspaceOpt.map(_.resolveFrom(tmpDir))
       )
       res match {
-        case Left(err)     => sys.error(err)
+        case Left(err) =>
+          Util.printException(err)
+          sys.error(err.getMessage)
```
alexarchambault: NIT Why not just throw the exception? Optionally wrapping it in `new Exception(…)`, so that the stack trace here is included in it:
```suggestion
          throw new Exception(err)
```

alexarchambault (approved):
Apart from the minor commit (and maybe reworking the commits a bit so that we can do a merge commit), LGTM!

PR #1161, opened 2022-07-08
Tweak package

alexarchambault:
These are mainly bits from https://github.com/VirtusLab/scala-cli/pull/1086, that don't involve Spark.

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
@@ -63,6 +63,12 @@ final case class PackageOptions(
     docker: Boolean = false,
 
   @Group("Package")
+  @Hidden
+  @HelpMessage("Exclude modules *and their transitive dependencies* from the JAR to be packaged")
+  @ValueDescription("org:name")
+    provided: List[String] = Nil,
+
+  @Group("Package")
   @HelpMessage("Use default scaladoc options")
   @ExtraName("defaultScaladocOpts")
     defaultScaladocOptions: Option[Boolean] = None,
--- a/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
@@ -39,6 +39,9 @@ final case class PackageOptions(
   @HelpMessage("Generate an assembly JAR")
     assembly: Boolean = false,
   @Group("Package")
+  @HelpMessage("For assembly JAR, whether to add a bash / bat preamble")
+    preamble: Boolean = true,
+  @Group("Package")
   @HelpMessage("Package standalone JARs")
     standalone: Option[Boolean] = None,
   @Recurse
--- a/modules/cli/src/main/scala/scala/cli/commands/util/PackageOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/PackageOptionsUtil.scala
@@ -2,14 +2,17 @@ package scala.cli.commands
 package util
 
 import scala.build.BuildThreads
+import scala.build.EitherCps.{either, value}
+import scala.build.Ops._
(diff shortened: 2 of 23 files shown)


modules/directives/src/main/scala/scala/build/preprocessing/directives/UsingPackagingDirectiveHandler.scala
```diff
@@ -0,0 +1,67 @@
+package scala.build.preprocessing.directives
+
+import scala.build.EitherCps.{either, value}
+import scala.build.Logger
+import scala.build.errors.{BuildException, MalformedInputError, UnexpectedDirectiveError}
+import scala.build.options.{BuildOptions, PackageOptions, PackageType, PostBuildOptions}
+
+case object UsingPackagingDirectiveHandler extends UsingDirectiveHandler {
```
lwronski, line 15: Often the package command is used with a parameter `--option`, such like this:
```
scala-cli package --msi --output path.msi Hello.scala
```
maybe, we should also support this parameter in using directive.

lwronski: There is no mention in the documentation that scala-cli supports such a `package` directive

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
```diff
@@ -62,6 +62,12 @@ final case class PackageOptions(
   @HelpMessage("Build Docker image")
     docker: Boolean = false,
 
+  @Group("Package")
+  @Hidden
+  @HelpMessage("Exclude modules *and their transitive dependencies* from the JAR to be packaged")
+  @ValueDescription("org:name")
+    provided: List[String] = Nil,
```
lwronski, line 72: Why not name this parameter as `--exclude`?

modules/cli/src/main/scala/scala/cli/commands/util/PackageOptionsUtil.scala
```diff
@@ -92,6 +92,9 @@ object PackageOptionsUtil {
         internal = baseOptions.internal.copy(
           // computing the provided modules sub-graph needs the final Resolution instance
           keepResolution = provided.nonEmpty
+        ),
+        internalDependencies = baseOptions.internalDependencies.copy(
+          addRunnerDependencyOpt = Some(false)
```
lwronski, line 97: Why was `addRunnerDependencyOpt` added to classPath by default before? In some cases it is needed?

modules/directives/src/main/scala/scala/build/preprocessing/directives/UsingPackagingDirectiveHandler.scala
alexarchambault: More directives can be added later if needed, in subsequent PRs. And it can be mentioned in the docs when more options are supported.

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
alexarchambault: "Exclude" is already used in [sbt](https://www.scala-sbt.org/1.x/docs/Library-Management.html#Exclude+Transitive+Dependencies), Mill, coursier, …, to exclude a dependency, but not necessarily all of its transitive dependencies. The option here is slightly different.

That said, "provided" is also used in sbt, and has slightly different semantics there (it means "compile-time" somehow). I would argue that the semantics here are closer to what users expect (which would be compile-time too, but also assume the dependency is already around at runtime, including all of its transitive dependencies).

modules/cli/src/main/scala/scala/cli/commands/util/PackageOptionsUtil.scala
alexarchambault: We use it in the `run` sub-command mainly. It gives better stack-traces in particular. But I wouldn't add it by default in package (where as a user, I'd expect the main class I provide to be used as is, rather than the one of the runner being used).

lwronski (approved):

PR #1163, opened 2022-07-08
Add --command and --scratch-dir to run command

alexarchambault:
`scala-cli run --command` basically prints the command `scala-cli run` would have run (with one argument per line). This can be passed over to `xargs` for example, but users can also manually pre-process these arguments if they feel like it. This is also useful when debugging for example (it was for me when working on https://github.com/VirtusLab/scala-cli/pull/1129).

`--scratch-dir` makes `scala-cli run` write temporary stuff (launchers, `.js` files, …) inside the passed directory rather than in a temporary OS location, and makes Scala CLI not delete these temporary files. When used in conjunction with `--command`, that allows users to later run the command, even if it involves temporary files.

(Bits from https://github.com/VirtusLab/scala-cli/pull/1129, that don't involve Spark.)


mpkocher:
Perhaps it would be useful to consider renaming this to `--scratch-dir` to avoid confusion with the standard env vars`TMPDIR`, `TEMP` mechanism of defining  a temporary directory?

The current naming scheme (e.g., `--tmp-dir`) is perhaps a little bit confusing and I suspect the behavior would not be completely obvious.

lwronski (approved):

PR #1166, opened 2022-07-11
Add support for running snippets

### LINKED ISSUE #1150: scala-cli is missing expression evaluation
**Version(s)**

Latest release

**Describe the bug**

Scala-cli is unable to evaluate expressions for scala flag `-e`.

**To Reproduce**

`scala-cli --scalac-option -e "println(123)"`
Error expecting the path.

**Expected behaviour**

There should be either an command `expression` used like `scala-cli expression "println(123)"` or additional flag `-e` used as `scala-cli -e "println(123)"`. Maybe both ?

Gedochao:
Fixes #1150 

Additionally,  this allows to run Scala and Java snippets in a similar fashion as `-e` allows for scripts.
Context:

- `-e`/`--script-snippet`/`--execute-script` allows for running code as a `.sc` script; it's similar to piping code via `stdin` to `_.sc`
  - this is meant to be the equivalent of `scala -e`
```bash
$ scala-cli -e 'println("Hello")'
Hello
```
- `--scala-snippet`/`--execute-scala` allows for running raw `.scala` code; it's similar to piping code via `stdin` to `_.scala`
```bash
$ scala-cli --scala-snippet '@main def run() = println("Hello")' 
Hello
```
- `--java-snippet`/`--execute-java` allows for running raw `.java` code; it's similar to piping code via `stdin` to `_.java`
```bash
$ scala-cli --java-snippet 'public class Main { public static void main(String[] args) { System.out.println("Hello"); } }'
Hello
```


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1916,4 +1958,41 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
       expect(mainClasses == Set(scalaFile1, scalaFile2, s"$scriptsDir.${scriptName}_sc"))
     }
   }
+
+  test("correctly run a script snippet") {
+    emptyInputs.fromRoot { root =>
+      val msg = "123456" // FIXME: change this to a a non-numeric string when Windows encoding is handled properly
```
Gedochao: the reason for this is that when code is being passed from a Windows terminal, Windows encoding is being used, while our inputs logic assumes everything is always to be `UTF-8`.
Fixing this requires a bit of a refactor, so I'd rather do it in a separate PR.

Gedochao: (so as a result, for now this feature will be buggy on Windows)

website/docs/guides/snippets.md
```diff
@@ -0,0 +1,217 @@
...
+```bash ignore
+cat ondisk.sc
+```
+
+```scala
+def script = "on-disk-script"
+```
+
+```bash ignore
+echo 'def piped = "piped-script"'|scala-cli-prototype . _.sc --scala-snippet 'case class ScalaSnippet(value: String = "scala-snippet")' --java-snippet 'public class JavaSnippet { public static String data = "java-snippet"; }' --script-snippet 'def script = "script-snippet"'
```
alexarchambault: ?
```suggestion
echo 'def piped = "piped-script"'|scala-cli. _.sc --scala-snippet 'case class ScalaSnippet(value: String = "scala-snippet")' --java-snippet 'public class JavaSnippet { public static String data = "java-snippet"; }' --script-snippet 'def script = "script-snippet"'
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -325,6 +326,32 @@ object Inputs {
     }
   }
 
+  def validateSnippets(
+    scriptSnippetOpt: Option[String] = None,
+    scalaSnippetOpt: Option[String] = None,
+    javaSnippetOpt: Option[String] = None
+  ): Seq[Either[String, Seq[Element]]] = {
```
alexarchambault, line 333: It seems the `Either`s will always be `Right`s here, isn't it? So do we have to return an `Either` here?

alexarchambault: Unless some kind of validation should be added in the near future?

alexarchambault (approved):
Looks great!

Two remarks:
- it seems the `--script-snippet` option and all can only be specified a single time each. Is that intended? (to be inline with `scala -e` maybe?) Changing their types to `List[String]` would allow them to be specified multiple times, which could be handy.
- I think it might be possible to preserve the order of these arguments, and add them to the other inputs in the right order too (relative to the other inputs), by using `caseapp.core.Indexed`, like `scriptSnippet: List[Indexed[String]]`. We get the other inputs via `caseapp.RemainingArgs`, which also offers the indices of individual arguments. That way, both `--*-snippet` options and other inputs would be taken into account in the right order, when computing the default main class for example. (Beware that I didn't use `Indexed` much, I hope it works fine…)

These points can be addressed or discussed in later PRs if needed, IMO.

website/docs/guides/snippets.md
Gedochao: oops, typo. fixed already.

modules/build/src/main/scala/scala/build/Inputs.scala
Gedochao: Now that I think about it, passing an empty snippet here should produce an error. 
Either way, I'll fix this in another PR, as I'll have to adjust this part when I add support for multiple snippets.

Gedochao:
> * it seems the `--script-snippet` option and all can only be specified a single time each. Is that intended? (to be inline with `scala -e` maybe?) Changing their types to `List[String]` would allow them to be specified multiple times, which could be handy.

Initially I intended to have it as close to `scala`'s `-e` as possible, but later it became clear that it should at the very least support whatever we have for piped inputs. Supporting a list of snippets passed my mind, but I wanted to get some feedback first. I'll add it in a separate PR.

> * I think it might be possible to preserve the order of these arguments, and add them to the other inputs in the right order too (relative to the other inputs), by using `caseapp.core.Indexed`, like `scriptSnippet: List[Indexed[String]]`. We get the other inputs via `caseapp.RemainingArgs`, which also offers the indices of individual arguments. That way, both `--*-snippet` options and other inputs would be taken into account in the right order, when computing the default main class for example. (Beware that I didn't use `Indexed` much, I hope it works fine…)

Hm... the potential benefits of doing this are minimal, if I understand correctly... But I'll try it in a separate PR

PR #1168, opened 2022-07-11
Turn the sip mode also for `scala-cli-sip` binary

romanowski:
Export and Package for now are excluded from the 'sip' mode.


Gedochao (approved):

modules/cli/src/main/scala/scala/cli/commands/Package.scala
```diff
@@ -41,6 +41,7 @@ import scala.util.Properties
 object Package extends ScalaCommand[PackageOptions] {
   override def name                                                          = "package"
   override def group                                                         = "Main"
+  override def inSipScala                                                    = false
```
alexarchambault, line 44: Why remove this one from the SIP binary? Generating a JAR is a pretty basic task.

PR #1171, opened 2022-07-12
chore(ci): bump version of mill-github-dependency-graph

ckipp01:
This brings in a few bugfixes about "direct" vs "indirect", failing
loudly if it doesn't work, and small stats at the end when it submits.

### DIFF
--- a/.github/workflows/github-dependency-graph.yml
+++ b/.github/workflows/github-dependency-graph.yml
@@ -17,6 +17,6 @@ jobs:
 
     - name: Submit dependency graph
       run:
-        ./mill --import ivy:io.chris-kipp::mill-github-dependency-graph::0.0.11 io.kipp.mill.github.dependency.graph.Graph/submit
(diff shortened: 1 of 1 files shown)


lwronski (approved):

lwronski:
Let's try again,, thanks @ckipp01!

ckipp01:
Btw, still not displaying, but found the issue. We were in contact with GitHub about it and this was the answer we got:

> It seems that at least one of your dependencies has an invalid package URL (aka “PURL”). The spec is [documented here](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst). The one causing an error was
pkg:maven/com.google.code.gson/gson@[2.8.6,2.9)
The version string should be a specific version (PURLs don’t support ranges). If it’s necessary to use special characters, they should be percent-encoded according to the spec
This should have been an error on snapshot creation, not in the UI. I’ve documented this bug for us to fix. Thanks for reporting this!

I'll get this fixed in the plugin and then updated again.

PR #1174, opened 2022-07-13
chore(ci): use mill-dependency-submission action

ckipp01:
There is now an action for this which makes it a bit easier to use. It
also _I hope_ brings in a fix that was prohibiting the graph to actually
show for scala-cli.


lwronski (approved):
Sure, good look, thanks @ckipp01!

ckipp01:
\o/ Finally!

Screenshot 2022-07-13 at 20 24 28

PR #1182, opened 2022-07-15
Add support for running multiple snippets of the same kind

Gedochao:


BlackAnubis7:
`website/docs/guides/snippets.md`
```
This is similar to how you refer to code from piped scripts through their wrapper name (`stdin`), more on which can be
found in [the scripts guide](scripts.md).
```
Is it still applicable, after adding description of snippet numeration in lines 192-208?

website/docs/guides/snippets.md
```diff
@@ -174,6 +189,23 @@ Hello
...
+scala-cli --scala-snippet '@main def main() = println(s"${snippet.hello} ${snippet1.world}${snippet2.exclamation}")' --script-snippet 'def hello: String = "Hello"' --script-snippet 'def world: String = "world"' --script-snippet 'def exclamation: String = "!"'
+```
+
+```text
+Hello world!
+```
+
+</ChainedSnippets>
+
 This is similar to how you refer to code from piped scripts through their wrapper name (`stdin`), more on which can be
```
BlackAnubis7, line 209: Is it still applicable, after adding description of snippet numeration above?

Gedochao: IMO yes, since it uses the exact same mechanism as the one with scripts passed by `stdin` (as in, you also have to refer to a predefined wrapper name to get to the contents of the passed script).
Also, the way you refer to the first ("0th") passed script snippet hasn't changed, its wrapper will still be `snippet`.

lwronski (approved):

PR #1190, opened 2022-07-18
Fix running scala check in scala native

### LINKED ISSUE #1158: `scala.NotImplementedError: an implementation is missing` at `AsmTestRunner$.matchFingerprints$$anonfun$1`
**Version(s)**

`0.1.9`

**Describe the bug**

```
scala.NotImplementedError: an implementation is missing
  scala.Predef$.$qmark$qmark$qmark(Predef.scala:344)
  scala.build.testrunner.AsmTestRunner$.matchFingerprints$$anonfun$1(AsmTestRunner.scala:92)
  scala.collection.IterableOnceOps.find(IterableOnce.scala:622)
  scala.collection.IterableOnceOps.find$(IterableOnce.scala:618)
  scala.collection.AbstractIterable.find(Iterable.scala:926)
  scala.build.testrunner.AsmTestRunner$.matchFingerprints(AsmTestRunner.scala:93)
 ...
```

**To Reproduce**

```
scala-cli test bug.scala
```

```scala
//> using scala "2.13.8"
//> using platform "native"
//> using lib "org.typelevel::cats-kernel-laws::2.8.0"
//> using lib "eu.timepit::refined::0.10.1"

import cats.kernel._
import cats.kernel.laws.discipline._
import eu.timepit.refined._
import eu.timepit.refined.api._
import eu.timepit.refined.types.numeric._
import eu.timepit.refined.numeric.Positive

import org.scalacheck._

class BugSpec extends Properties("bugged") {

  implicit val posByteCommutativeSemigroup: CommutativeSemigroup[PosByte] =
    getPosIntegralCommutativeSemigroup[Byte]

  private def getPosIntegralCommutativeSemigroup[A: CommutativeSemigroup: NonNegShift](implicit
      integral: Integral[A],
      v: Validate[A, Positive]
  ): CommutativeSemigroup[A Refined Positive] =
    CommutativeSemigroup.instance { (x, y) =>
      val combined: A = Semigroup[A].combine(x.value, y.value)

      refineV[Positive](combined).getOrElse {
        val result: A =
          CommutativeSemigroup[A].combine(NonNegShift[A].shift(combined), integral.one)
        refineV[Positive].unsafeFrom(result)
      }
    }

  implicit def eq: Eq[PosByte] = Eq.by(_.value)

  implicit def arb: Arbitrary[PosByte] =
    Arbitrary(Gen.choose(0.toByte, Byte.MaxValue).map(refineV[Positive](_).toOption.get))

  property("propped") = CommutativeSemigroupTests[PosByte].commutativeSemigroup.props(0)._2

}


/**
 * Typeclass to shift Negative values to Non Negative values.
 */
trait NonNegShift[T] extends Serializable {
  def shift(t: T): T
}

object NonNegShift {
  def apply[T](implicit ev: NonNegShift[T]): NonNegShift[T] = ev

  def instance[T](function: T => T): NonNegShift[T] =
    new NonNegShift[T] {
      def shift(t: T): T = function(t)
    }

  // Instances
  implicit val byteNonNegShift: NonNegShift[Byte] = instance(t => (t & Byte.MaxValue).toByte)
  implicit val shortNonNegShift: NonNegShift[Short] = instance(t => (t & Short.MaxValue).toShort)
  implicit val intNonNegShift: NonNegShift[Int] = instance(t => t & Int.MaxValue)
  implicit val longNonNegShift: NonNegShift[Long] = instance(t => t & Long.MaxValue)
}
```

lwronski:


modules/integration/src/test/scala/scala/cli/integration/TestTestDefinitions.scala
```diff
@@ -103,6 +103,25 @@ abstract class TestTestDefinitions(val scalaVersionOpt: Option[String])
     )
   )
 
+  val successfulScalaCheckFromCatsNativeInputs = TestInputs(
+    Seq(
+      os.rel / "MyTests.scala" ->
+        """//> using scala "2.13.8"
+          |//> using platform "native"
+          |//> using lib "org.typelevel::cats-kernel-laws::2.8.0"
```
armanbilge, line 111:
```suggestion
          |//> using lib "org.scalacheck::scalacheck::1.16.0"
```

lwronski: But If I replace `cats-kernel-laws` dependency to `scalachek` it works in the actual version of Scala CLI. Therefore I added a test with `cats-kernel-laws` dependency which doesn't work in `0.1.10`

armanbilge: Thanks, I didn't realize it was necessary to reproduce the issue.

alexarchambault (approved):
Let's merge, as this can help for #1158, but this is not really a fix, more a workaround. I think we should investigate why the added test framework isn't found by default (it should), and also have a look at the Scala Native [junit compiler plugin](https://github.com/scala-native/scala-native/tree/main/junit-plugin/src/main/scala-3/scala/scalanative/junit/plugin), which apparently is necessary from sbt in order to run such tests (I tried manually adding it, but it seems it wasn't generating the expected junit-related classes).

PR #1192, opened 2022-07-19
Generate scalafmt conf file on fmt command

### LINKED ISSUE #1160: Fmt command should create config with Scalafmt version if not found
**Version(s)**

`0.1.9`

**Describe the bug**

`scala-cli fmt .` throws errors if version not found in `.scalafmt.conf` file:
```
Scalafmt requires explicitly specified version.
To configure the scalafmt version add the following line into .scalafmt.conf:
    version = 3.5.8 
```

**Expected behaviour**

Scala CLI should create configuration for Scalafmt which contains version

### LINKED ISSUE #1176: Defining `dialect` in `.scalafmt.conf` shouldn't be necessary when running the `fmt` command
**Version(s)**

0.1.9

**Describe the bug**

when the `runner.dialect` is not defined in `.scalafmt.conf`, Scala CLI throws an error on running the `fmt` command.

**To Reproduce**

```bash
▶ cat .scalafmt.conf
version = 3.5.8

▶ scala-cli fmt .
Default dialect is deprecated; use explicit: [sbt0137,sbt1,scala211,scala212,scala212source3,scala213,scala213source3,scala3]
Also see https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects"

```

passing `--dialect` as an option does not help, either.

```
▶ cat .scalafmt.conf
version = 3.5.8

▶ scala-cli fmt . --dialect scala3
Default dialect is deprecated; use explicit: [sbt0137,sbt1,scala211,scala212,scala212source3,scala213,scala213source3,scala3]
Also see https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects"
```

**Expected behaviour**
The user shouldn't have to define the `runner.dialect` field in the `.scalafmt.conf` manually.
The desired behaviour should reflect what we settled on with the `version` field, as described in #1160 : [relevant comment](https://github.com/VirtusLab/scala-cli/issues/1160#issuecomment-1184323549)

wleczny:


modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -39,14 +49,28 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
-      lazy val default: ScalafmtVersionConfig = ScalafmtVersionConfig()
-      implicit lazy val surface: metaconfig.generic.Surface[ScalafmtVersionConfig] =
-        metaconfig.generic.deriveSurface[ScalafmtVersionConfig]
-      implicit lazy val decoder: metaconfig.ConfDecoder[ScalafmtVersionConfig] =
-        metaconfig.generic.deriveDecoder[ScalafmtVersionConfig](default)
+  def readVersionAndDialectFromFile(
+    workspace: os.Path,
+    logger: Logger
+  ): (Option[String], Option[String], Option[os.Path]) = {
+    case class RunnerMetaconfig(dialect: String)
```
lwronski:
```suggestion
    case class RunnerMetaconfig(dialect: String = "")
```

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -63,17 +87,128 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
     }
-    (versionMaybe, pathMaybe.isDefined)
+    val dialectMaybe = confContentMaybe.collect {
+      case conf if conf.runner.dialect.trim.nonEmpty => conf.runner.dialect
+    }
+    (versionMaybe, dialectMaybe, pathMaybe)
+  }
+
+  // Based on https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ScalafmtDialect.scala
+  sealed abstract class ScalafmtDialect(val value: String)
```
lwronski: Can you move copied `object`/`classes` from `metals` repo to another file?

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -94,62 +229,46 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
+          val newPath =
+            if (options.saveScalafmtConf) path
+            else workspace / Constants.workspaceDirName / confName
+          val prevConf = os.read(path)
+          val entry    = updateScalafmtConfig(prevConf, Some(version), dialect)
+          os.write.over(newPath, entry)
+          newPath
+
+        pathMaybe.fold(newConfFilePath)(confFileBasedOnExistingPath(_))
+      }
```
lwronski:
```suggestion
      val entry = {
           val dialect  = ScalafmtDialect.fromString(dialectString)
           val maybePrevConf = pathMaybe.map(p => os.read(p))
            updateScalafmtConfig(maybePrevConf.gerOrElse(""), Some(version), dialect)
      }
      val scalaFmtConfPath = {
        val confFileName = ".scalafmt.conf"
        val path =
            if (options.saveScalafmtConf) workspace / confFileName
            else workspace / Constants.workspaceDirName / confFileName
        os.write.over(path, entry)
        path
      }
```

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -63,17 +87,128 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
+      case "scala212"        => Some(Scala212)
+      case "scala212source3" => Some(Scala212Source3)
+      case "scala213"        => Some(Scala213)
+      case "scala213source3" => Some(Scala213Source3)
+      case "scala3"          => Some(Scala3)
+      case _                 => None
+    }
+  }
+
+  // Based on https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```
lwronski: Did you check, if copied code has `Apache-2.0 license`?

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
wleczny: Shouldn't `path` be something like:
```
val path =
    if (options.saveScalafmtConf) pathMaybe.getOrElse(workspace / confFileName)
    else workspace / Constants.workspaceDirName / confFileName
```

lwronski: yes, you are right

modules/cli/src/main/scala/scala/cli/commands/util/FmtUtil.scala
```diff
@@ -0,0 +1,194 @@
...
+      case "scala212source3" => Some(Scala212Source3)
+      case "scala213"        => Some(Scala213)
+      case "scala213source3" => Some(Scala213Source3)
+      case "scala3"          => Some(Scala3)
+      case _                 => None
+    }
+  }
+
+  // Based on https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
+  def updateScalafmtConfig(
```
Gedochao: the `update` in the function name is confusing IMO, as it suggests there's a side-effect, while in fact, the function is pure.

hard to name it well, maybe smth like:
```suggestion
  def scalafmtConfigWithFields(
```

some scaladoc might help, maybe

modules/cli/src/main/scala/scala/cli/commands/util/FmtUtil.scala
```diff
@@ -0,0 +1,194 @@
...
+    configText: String,
```
Gedochao: NIT
```suggestion
    previousConfigText: String,
```

modules/cli/src/main/scala/scala/cli/commands/util/FmtUtil.scala
```diff
@@ -0,0 +1,194 @@
...
+        logger.log(
+          s"""Could not get root of the git repository.
+             |Cause: $e""".stripMargin
+        )
+        None
+    }
+
+  // Based on scalafmt comment:
+  // https://github.com/scalameta/scalafmt/blob/d0c11e98898334969f5f4dfc4bd511630cf00ab9/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+  // First we look for the file in the cwd.
```
Gedochao: these comments should be in scaladoc format, IMO

website/docs/commands/fmt.md
```diff
@@ -20,29 +20,26 @@ You can check formatting correctness using a `--check` flag:
...
 
-### Dialects
+### Scalafmt version and dialect
 
-Scala CLI also supports dialects that are passed to the formatter.
-This value is only used if there is no `.scalafmt.conf` file.
-However, if it exists, then all configuration should be placed there.
-For a list of all possible values, consult
-the [official Scala Dialects documentation](https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects):
+Configuration file with specified **version** and **dialect** is required to run `scalafmt`. Scala CLI `fmt` tries to find the `.scalafmt.conf` file in the **projects root** directory and **current workspace** directory. There are 3 possible cases:
```
Gedochao: Actually, the configuration file is no longer required, the way I understand it.
Instead:
```suggestion
Specifying the  **version** and **dialect** is required to run `scalafmt`. Those can be passed through the `scalafmt` configuration file or directly with Scala CLI. If none is specified explicitly, default values depending on the context will be used.

Scala CLI's `fmt` sub-command tries to find the `.scalafmt.conf` file in the **projects root** directory and **current workspace** directory. 

There are 3 possible cases:
```

website/docs/commands/fmt.md
```diff
@@ -20,29 +20,26 @@ You can check formatting correctness using a `--check` flag:
...
 
-```bash
-scala-cli fmt --dialect scala212
-```
-
-### Scalafmt version
+1. Configuration file with the specified version and dialect is found.
+2. Configuration file is found, but it doesn't have specified version or dialect.
```
Gedochao: NIT: technically, you missed the few extra cases when one is found, but not the other

website/docs/commands/fmt.md
wleczny: Scala CLI `fmt` command runs `scalafmt` under the hood and as far as I know `scalafmt` still needs `.scalafmt.conf` file.

website/docs/commands/fmt.md
```diff
@@ -20,29 +20,26 @@ You can check formatting correctness using a `--check` flag:
...
-To configure the Scalafmt version, add the following to `.scalafmt.conf`. For example, to set the version to `3.5.0`,
-add the following line:
+If the `--save-scalafmt-conf` option is passed, then `fmt` command behaves as follows:
+- In the **first** case `fmt` uses the found `.scalafmt.conf` file to run `scalafmt`.
+- In the **second** case `fmt` writes missing parameters directly into the previously found file and then uses it to run `scalafmt`.
+- In the **third** case `fmt` creates a `.scalafmt.conf` file in the current workspace directory, writes a default version and a dialect into it and uses it to run `scalafmt`.
 
-```
-version = "3.5.0"
+It is also possible to pass `--dialect` option to define which dialect should be written into configuration file, which will be used to run `scalafmt` command:
```
Gedochao: IMO, we should also enable passing `--scalafmt-version` in a similar fashion from the command line

Gedochao: whatever is passed should also take priority over the `.scalafmt.conf` contents, which IMO could be clarified further here. (I think it's already the case for `--dialect`)

website/docs/commands/fmt.md
wleczny: Maybe I should start with information that `fmt` command runs `scalafmt` under the hood, but this is how description of [fmt](https://scala-cli.virtuslab.org/docs/commands/fmt) starts.

Gedochao: we possibly could also clarify what do we mean by the project root here (since here we calculate it based on the existing `git` configuration, and not based on the directory of the first input, like in other sub-commands)

website/docs/commands/fmt.md
wleczny: Isn't subsection 2 covering it?

website/docs/commands/fmt.md
Gedochao: > Scala CLI fmt command runs scalafmt under the hood and as far as I know scalafmt still needs .scalafmt.conf file.

Ok, fair point.

What I mean is - from the user's perspective, the important part is the `fmt` sub-command, which now does not need the conf file. 

You're right that we can and should clarify how we do it under the hood (since one can peek into the `.scala-build` folder on how we do it), but IMO the reader of this doc should be clear that when using the `fmt` command, he doesn't need to specify anything, really (as long as he's fine with the defaults we infer)

website/docs/commands/fmt.md
Gedochao: it is, just pointing out there's actually more than 3 cases, which create a slight dissonance here

website/docs/commands/fmt.md
wleczny: Sounds good. I will take care of it.

modules/cli/src/main/scala/scala/cli/commands/util/FmtUtil.scala
wleczny: Should I actually change it since this function is based on metals function?

Gedochao: No reason to leave it unchanged once it's part of `scala-cli`'s codebase.

modules/cli-options/src/main/scala/scala/cli/commands/FmtOptions.scala
```diff
@@ -42,7 +46,9 @@ final case class FmtOptions(
     scalafmtArg: List[String] = Nil,
 
   @Group("Format")
-    dialect: Option[String] = None
+    scalafmtDialect: Option[String] = None,
```
Gedochao, line 51: since you're changing the option name here, might be worth adding an alias for the old name.
I'd add a help message, too, while I was at it.
```suggestion
  @HelpMessage("Pass a global dialect for scalafmt. This overrides whatever value is configured in the .scalafmt.conf file.")
  @Name("dialect")
    scalafmtDialect: Option[String] = None,
```

modules/cli-options/src/main/scala/scala/cli/commands/FmtOptions.scala
```diff
@@ -42,7 +46,9 @@ final case class FmtOptions(
...
+  @Group("Format")
```
Gedochao, line 52: NIT while you're at it, a help message would be cool here as well. and potentially a shorter alias.
```suggestion
  @Group("Format")
  @HelpMessage("Pass scalafmt version before running it. This overrides whatever value is configured in the .scalafmt.conf file.")
  @Name("fmtVersion")
```

website/docs/commands/fmt.md
```diff
@@ -20,29 +20,33 @@ You can check formatting correctness using a `--check` flag:
...
 
-### Dialects
+### Scalafmt version and dialect
 
-Scala CLI also supports dialects that are passed to the formatter.
-This value is only used if there is no `.scalafmt.conf` file.
-However, if it exists, then all configuration should be placed there.
-For a list of all possible values, consult
-the [official Scala Dialects documentation](https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects):
+Scala CLI `fmt` command runs `scalafmt` under the hood, which requires `.scalafmt.conf` configuration file with specified **version** and **dialect**.
```
Gedochao: I meant something more along the lines of:
```suggestion
Scala CLI `fmt` command supports passing the dialect and `scalafmt` version directly from the command line, using respectively the `--scalafmt-dialect` and `--scalafmt-version` options. You can even skip passing either of those, which will make Scala CLI to infer a default value. 

If a `.scalafmt.conf` file is present in the workspace, the values will be read from there, unless specified with Scala CLI options.

Otherwise, the default dialect will be inferred based on the configured Scala version (through a Scala CLI option or a using directive), while the default `scalafmt` version will be the latest one used by your Scala CLI version (so it is subject to change when updating Scala CLI).

The Scala CLI `fmt` command runs `scalafmt` under the hood, which *normally* requires `.scalafmt.conf` configuration file with explicitly specified **version** and **dialect** fields. 
The way it is handled by Scala CLI is as follows:
```

modules/cli/src/main/scala/scala/cli/commands/Fmt.scala
```diff
@@ -94,62 +37,39 @@ object Fmt extends ScalaCommand[FmtOptions] {
...
+      val dialectString = options.scalafmtDialect.getOrElse {
+        dialectMaybe.getOrElse {
+          version match
+            case v if v.startsWith("2.11.") => "scala211"
+            case v if v.startsWith("2.12.") => "scala212"
+            case v if v.startsWith("2.13.") => "scala213"
+            case v if v.startsWith("3.")    => "scala3"
+            case _                          => "default"
         }
       }
```
alexarchambault: NIT The nested `getOrElse` looks a bit ugly. This should do:
```suggestion
      val dialectString = options.scalafmtDialect.orElse(dialectMaybe).getOrElse {
        version match
          case v if v.startsWith("2.11.") => "scala211"
          case v if v.startsWith("2.12.") => "scala212"
          case v if v.startsWith("2.13.") => "scala213"
          case v if v.startsWith("3.")    => "scala3"
          case _                          => "default"
      }
```

modules/cli/src/main/scala/scala/cli/commands/util/FmtUtil.scala
```diff
@@ -0,0 +1,202 @@
...
+import com.typesafe.config.parser.ConfigDocumentFactory
+import scala.build.Logger
+import scala.util.control.NonFatal
+
+object FmtUtil {
+  private def getGitRoot(workspace: os.Path, logger: Logger): Option[String] =
+    try {
+      val result = os.proc("git", "rev-parse", "--show-toplevel").call(
+        cwd = workspace,
+        stderr = null
```
alexarchambault: I'm not sure passing `null` is fine here. If you just want to discard stderr, maybe something like this should do:
```suggestion
        stderr = os.ProcessOutput.ReadBytes((_, _) => ())
```

But if that command doesn't print garbage by default, `os.Inherit` should do. Else `os.Pipe` could work, that should put the error output in the exception below, if the command fails.

website/docs/commands/fmt.md
```diff
@@ -20,31 +20,45 @@ You can check formatting correctness using a `--check` flag:
...
-This value is only used if there is no `.scalafmt.conf` file.
-However, if it exists, then all configuration should be placed there.
-For a list of all possible values, consult
-the [official Scala Dialects documentation](https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects):
+Scala CLI `fmt` command supports passing the **dialect** and `scalafmt` **version** directly from the command line, using respectively the `--scalafmt-dialect` and `--scalafmt-version` options:
+```
+scala-cli fmt --scalafmt-dialect scala3 --scalafmt-version 3.5.8
+```
+You can skip passing either of those, which will make Scala CLI to infer a default value:
+- If a `.scalafmt.conf` file is present in the workspace and it has field defined, the value will be read from there, unless specified with Scala CLI options.
```
Gedochao:
```suggestion
- If a `.scalafmt.conf` file is present in the workspace and it has the field defined, the value will be read from there, unless explicitly specified with Scala CLI options.
```

website/docs/commands/fmt.md
```diff
@@ -20,31 +20,45 @@ You can check formatting correctness using a `--check` flag:
...
 ```
-version = "3.5.0"
+
+```bash
+scala-cli fmt --scalafmt-version 3.5.8
 ```
 
+For above setup `fmt` will use:
+- `version="3.5.8"`, because passed `--scalafmt-version` option overrides version from the file
+- `dialect=scala3`, because dialect is neither passed as an option nor is it present in the configuration file, so it is inferred based on the previously evaluated version (`3.5.8`)
```
Gedochao: uh, the dialect is inferred based on the Scala version, not the `scalafmt` version, no?
since `-S` wasn't passed, the default would be the newest release of dotty: `3.1.3`, as of writing this comment.

So it should be something more like:
```suggestion
- `dialect=scala3`, because dialect is neither passed as an option nor is it present in the configuration file, so it is inferred based on the Scala version; the Scala version wasn't explicitly specified in the command either, so it falls back to the default Scala version - the latest one, thus the resulting dialect is `scala3`. 
```

wleczny: Oh, thats right. My bad...

Gedochao (approved):

PR #1195, opened 2022-07-22
Prevent erroneous using directives from blocking the initial run of BSP

### LINKED ISSUE #1097: Incorrect using directive prevents proper bsp setup
**Version(s)**

0.1.7

**Describe the bug**

When running `scala-cli setup-ide` for the first time in a project, if an incorrect using directive is present (e.g. with missing quotes)  `.scala-build/.bloop` directory and its contents are not generated at all and there's no error about that and then the IDE sees no build targets (and as a consequence the error in the directive is not indicated in the IDE).

**To Reproduce**

```shell
mkdir /tmp/foo
cd /tmp/foo
code .      # open the directory in VSCode
echo '//> using scala 3.1.2' > Foo.scala
scala-cli setup-ide .
```

**Expected behaviour**

Either `scala-cli setup-ide` command should report some error or it should generate the build targets for bsp anyway so that the error is displayed in the IDE.

Gedochao:
Fixes #1097


modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -44,9 +62,19 @@ final class BspImpl(
...
 
+  /** Initial setup for the Bloop project.
+    *
+    * @param currentBloopSession
+    *   the current Bloop session
+    * @param reloadableOptions
+    *   options which may be reloaded on a bsp workspace/reload request
+    * @param isBspInitRun
+    *   a flag indicating if this is the BSP initial setup workflow
+    */
```
lwronski: The name of parameter `isBspInitRun` and description should be more descriptive. 

To understand this flag, I first need to know how BSP initial setup workflow works

modules/options/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -375,12 +380,19 @@ object Artifacts {
...
+    classifiersOpt: Option[Set[String]],
+    isBspInitRun: Boolean = false
   ): Either[BuildException, Fetch.Result] = either {
     val coursierDependenciesWithFallbacks = value {
       dependencies.value
         .map(Positioned(dependencies.positions, _))
         .map(dep => dep.toCs(paramsOpt).map(csDep => (dep.value, csDep.value)))
+        .filter {
+          case Left(l) if isBspInitRun =>
+            logger.error(l.message)
```
lwronski:
```suggestion
          case Left(e) if isBspInitRun =>
            logger.error(e.message)
```

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -1118,9 +1120,50 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+                new URI(targetUri).getQuery.stripPrefix("id=")
+              }
+
+              val bloopDir = root / Constants.workspaceDirName / ".bloop"
+              expect(os.isDir(bloopDir))
+
+              bloopProjectNames.foreach { bloopProjectName =>
+                val bloopProjectJsonPath = bloopDir / s"$bloopProjectName.json"
+                expect(os.isFile(bloopProjectJsonPath))
+              }
```
lwronski: We should also test, if `bsp` properly generates error diagnostic.

Gedochao: good call, testing for diagnostics made me realise those were broken on the initial run (recovering from errors prevented the silenced errors to be transferred as diagnostics).
ought to be fixed now.

alexarchambault (commented):
NIT About the commit structure, so that we have a clean(ish) history on the main branch: could you create a separate commit for the added scaladoc, and squash the other non-refacto commits?

Gedochao:
> NIT About the commit structure, so that we have a clean(ish) history on the main branch: could you create a separate commit for the added scaladoc, and squash the other non-refacto commits?

@alexarchambault should be fixed now.

lwronski (approved):

PR #1196, opened 2022-07-22
Put the release notes doc on the website

Gedochao:
The release notes for `v0.1.10` were missing in the doc, so I added them in here.
We have to be careful about using GitHub-idiomatic syntax, as that won't work for the release notes on our docs website (and we probably want it to be consistent). That includes `@` mentions of users, links to PRs, issues and such.


Gedochao:
I didn't use the `<ChainedSnippets>` component in the release notes doc, as that would make it slightly high-maintenance (you then wouldn't be able to copy the notes over unchanged after the release notes PR gets merged). However, if we decide the looks are worth the effort, we could do that as well.

lwronski (approved):

PR #1198, opened 2022-07-25
Use manifest JARs in "run" command if needed

alexarchambault:
When running `java -cp …` commands under-the-hood in `scala-cli run`, the class path passed to it may be too long on Windows. A common workaround for this consists in using "manifest JARs", that is JARs consisting only of a manifest file, with the class path encoded in that file.

This PR enables manifest JARs automatically when needed (and fixes how we generate them too).

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/ManifestJar.scala
+++ b/modules/build/src/main/scala/scala/build/internal/ManifestJar.scala
@@ -1,0 +1,84 @@
+package scala.build.internal
+
+import java.io.OutputStream
+
+object ManifestJar {
+
+  /** Creates a manifest JAR, in a temporary directory or in the passed scratched directory
+    *
+    * @param classPath
+    *   Entries that should be put in the manifest class path
+    * @param wrongSimplePaths
+    *   Write paths slightly differently in manifest, so that tools such as native-image accept them
+    *   (but the manifest JAR can't be passed to 'java -cp' any more on Windows)
+    * @param scratchDirOpt
+    *   an optional scratch directory to write the manifest JAR under
+    */
+  def create(
+    classPath: Seq[os.Path],
+    wrongSimplePaths: Boolean = false,
+    scratchDirOpt: Option[os.Path] = None
+  ): os.Path = {
+    import java.util.jar._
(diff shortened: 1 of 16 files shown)


modules/build/src/main/scala/scala/build/internal/ManifestJar.scala
```diff
@@ -0,0 +1,70 @@
+package scala.build.internal
+
+import java.io.OutputStream
+
+object ManifestJar {
+
+  def create(
```
Gedochao, line 17: A scaladoc would be useful here, i.e. it wasn't immediately obvious to me what the `wrongSimplePaths` arg stands for

modules/cli-options/src/main/scala/scala/cli/commands/RunOptions.scala
```diff
@@ -29,7 +29,11 @@ final case class RunOptions(
     command: Boolean = false,
   @Group("Run")
   @HelpMessage("Temporary / working directory where to write generated launchers")
-    scratchDir: Option[String] = None
+    scratchDir: Option[String] = None,
+  @Group("Run")
+  @Hidden
```
Gedochao, line 34: eh, same thing on making it hidden as with `PackageOptions.mainClassInManifest`
maybe we could mark these obscure, border-arcane options for niche usecases somehow?
so we remember to document them later in a guide/cookbook/some specialised help command.

treat this comment as thinking aloud rather than a suggestion for a change in this PR

Gedochao (approved):
LGTM, just needs more documentation IMO

modules/build/src/main/scala/scala/build/internal/ManifestJar.scala
alexarchambault: Should be fixed.

modules/cli-options/src/main/scala/scala/cli/commands/RunOptions.scala
alexarchambault: I left it hidden just like `PackageOptions.mainClassInManifest`… This can be changed / perfected later.

PR #1200, opened 2022-07-25
Make main class optional in preamble-less assemblies

alexarchambault:
These changes allow one to generate an assembly, even for code that has no main class, when `--preamble=false` is passed. This can be useful for libraries, if users want to pass the assembly to tools such as proguard. This also accepts a (hidden) `--main-class-in-manifest=false` option if users want not only no preamble, but also no mention of main class in the assembly manifest (`META-INF/MANIFEST.MF` in the assembly JAR). The latter option is useful for tools, such as the `hadoop jar` command, that behave differently depending on the presence or not of a main class in the manifest.

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
@@ -45,6 +45,10 @@ final case class PackageOptions(
     preamble: Boolean = true,
   @Group("Package")
   @Hidden
+  @HelpMessage("For assembly JAR, whether to specify a main class in the JAR manifest")
+    mainClassInManifest: Option[Boolean] = None,
+  @Group("Package")
+  @Hidden
   @HelpMessage("Generate an assembly JAR for Spark (assembly that doesn't contain Spark, nor any of its dependencies)")
     spark: Boolean = false,
   @Group("Package")
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -125,6 +125,21 @@ object Build {
       else
         Left(mainClasses)
     }
+    def retainedMainClassOpt(
+      mainClasses: Seq[String],
+      logger: Logger
+    ): Option[String] = {
+      val defaultMainClassOpt = sources.defaultMainClass
+        .filter(name => mainClasses.contains(name))
+      def foundMainClass =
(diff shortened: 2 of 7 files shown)


modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
```diff
@@ -45,6 +45,10 @@ final case class PackageOptions(
     preamble: Boolean = true,
   @Group("Package")
   @Hidden
+  @HelpMessage("For assembly JAR, whether to specify a main class in the JAR manifest")
+    mainClassInManifest: Option[Boolean] = None,
+  @Group("Package")
+  @Hidden
```
Gedochao, line 51: Why make it hidden, though?

alexarchambault: Because it feels kind of arcane to me… I could un-hide though.

Gedochao: IMO hiding it only makes it more arcane, while it might be useful to users (even if obscure until we cover it in a cookbook

Gedochao: I think we should only hide the stuff that's too unstable/experimental for users to rely on or it's something targeting the Scala CLI devs

alexarchambault: I think we hide much more stuff than this. Many options are hidden so that we don't overflow users when they pass `--help`.

Gedochao: we could consider specialised `--help` sections in a few spots, I suppose.

Gedochao: I'd still un-hide it, otherwise it won't become useful to anyone, I think.
but yeah, your points are fair as well, so I'm leaving the decision to you.

Gedochao (approved):

modules/cli-options/src/main/scala/scala/cli/commands/PackageOptions.scala
alexarchambault: Merging so that I don't have to wait for the CI again. We can always revisit that later, or when adding specialised help options.

PR #1202, opened 2022-07-26
Split integration tests

alexarchambault:
These often take more than an hour to run on the CI… Let's see if splitting them helps here.

This will probably make the workflow definition more verbose. Indeed, keeping the OS-based build matrix and naively splitting the `native-tests` job into `generate-launchers` (to generate the native images), and `native-tests-1` and `native-tests-2` (running each half the ITs with the launchers generated by the previous job) might be a problem: for example, it makes the Windows image generation block starting the ITs on Linux and macOS, even if there launchers are done being generated. And if the Windows launcher generation fails, no IT is run at all. That is, no ITs start running until launchers are successfully generated for all 3 OSes.

That means, we should replace the matrix-based `native-tests` by 12 non matrix-based jobs:
- `generate-linux-launcher`
- `generate-macos-launcher`
- `generate-windows-launcher`
- `native-linux-tests-1`
- `native-linux-tests-2`
- `native-linux-tests-3`
- `native-macos-tests-1`
- `native-macos-tests-2`
- `native-macos-tests-3`
- `native-windows-tests-1`
- `native-windows-tests-2`
- `native-windows-tests-3`

And `jvm-tests` by:
- `jvm-tests-1`
- `jvm-tests-2`
- `jvm-tests-3`

And `native-static-tests` and `native-mostly-static-tests` by:
- `generate-static-launcher`
- `generate-mostly-static-launcher`
- `native-static-tests-1`
- `native-static-tests-2`
- `native-static-tests-3`
- `native-mostly-static-tests-1`
- `native-mostly-static-tests-2`
- `native-mostly-static-tests-3`

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -698,7 +698,7 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
       mainArgs = Seq(cache.toNIO.toString, classpath),
       workingDir = os.pwd
     )
-    val cp = res.out.text.trim
+    val cp = res.out.text().trim
     cp.split(File.pathSeparator).toSeq.map(p => mill.PathRef(os.Path(p)))
   }
 
--- a/build.sc
+++ b/build.sc
@@ -859,6 +859,49 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
     def test(args: String*) =
       jvm(args: _*)
 
+    def forcedLauncher = T.persistent {
+      val ext      = if (Properties.isWin) ".exe" else ""
+      val launcher = T.dest / s"scala-cli$ext"
+      if (!os.exists(launcher)) {
+        val dir = Option(System.getenv("SCALA_CLI_IT_FORCED_LAUNCHER_DIRECTORY")).getOrElse {
+          sys.error("SCALA_CLI_IT_FORCED_LAUNCHER_DIRECTORY not set")
+        }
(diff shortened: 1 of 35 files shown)


build.sc
```diff
@@ -1034,6 +1034,39 @@ def writeShortPackageVersionTo(dest: os.Path) = T.command {
...
+      import java.util.zip.GZIPOutputStream
+
+      val compressed = os.read.bytes(from)
+      val baos       = new ByteArrayOutputStream
+      Using.resource(new GZIPOutputStream(baos)) { gzos =>
+        gzos.write(compressed)
+      }
+      baos.toByteArray
+    }
+  os.write(nativeLauncher, content, createFolders = true)
```
lwronski: Instead of copying native luncher from artifacts, maybe we can unzip `scala-cli` native launchers from artifacts and then use it in integration test, just modify `TestUtil.cliPath` and read Scala CLI native launcher path from env path.

alexarchambault: I'll probably end up doing something like that, yeah. For now, the native-linux-tests job doesn't pick the imported launcher, and [keeps re-generating it](https://github.com/VirtusLab/scala-cli/runs/7523392761?check_suite_focus=true#step:7:2449).

alexarchambault:
Ideally, it'd be cool to generate that workflow definition from a more abstract definition…

lwronski (approved):

PR #1207, opened 2022-07-27
Use more lightweight Spark distribs in spark tests

alexarchambault:
This uses Spark distributions downloaded from the releases of [`scala-cli/lightweight-spark-distrib`](https://github.com/scala-cli/lightweight-spark-distrib), which weight around 16 MB, and download the JARs they need via coursier. That should allow to work around transient issues seen on CI when downloading the standard Spark distributions.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/SparkTests212.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SparkTests212.scala
@@ -8,34 +8,69 @@ import scala.util.Properties
 
 object SparkTests212 {
 
-  private final case class Spark(sparkVersion: String, scalaVersion: String) {
-    def sparkHome(): os.Path = {
-      val url =
-        s"https://archive.apache.org/dist/spark/spark-$sparkVersion/spark-$sparkVersion-bin-hadoop2.7.tgz"
-      val dirName = url.drop(url.lastIndexOf('/') + 1).stripSuffix(".tgz")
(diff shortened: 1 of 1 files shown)


modules/integration/src/test/scala/scala/cli/integration/SparkTests212.scala
```diff
@@ -8,34 +8,69 @@ import scala.util.Properties
...
+  import SparkTests212._
+
+  private val spark30 = new Spark(
     "3.0.3",
     // The spark distribution actually ships with Scala 2.12.10, but we run into #1092 if we use it here
     "2.12.15"
   )
 
-  private val spark24 = Spark(
+  private val spark24 = new Spark(
```
Gedochao, line 64: NIT: the `new` keyword is actually redundant now, since we compile integration tests with Scala 3 now, no?

Gedochao (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/SparkTests212.scala
alexarchambault: Right. Fixed it, although I'm not fond of dropping `new` for non-case classes…

alexarchambault: Actually, I added back the `new`, it seems [they're still required](https://github.com/VirtusLab/scala-cli/runs/7539539013?check_suite_focus=true#step:7:957).

PR #1216, opened 2022-07-29
Overview of Markdown support

BlackAnubis7:
This PR is meant to show the overview of the changes made, to facilitate discussion and further development. It's not meant to be merged as-is, because it contains some more or less makeshift or temporary solutions

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/MarkdownPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/MarkdownPreprocessor.scala
@@ -1,0 +1,124 @@
+package scala.build.preprocessing
+
+import java.nio.charset.StandardCharsets
+
+import scala.build.EitherCps.{either, value}
+import scala.build.errors.BuildException
+import scala.build.internal.{AmmUtil, CodeWrapper, CustomCodeWrapper, Name}
+import scala.build.options.{BuildOptions, BuildRequirements}
+import scala.build.preprocessing.ScalaPreprocessor.ProcessingOutput
+import scala.build.{Inputs, Logger}
+
+import scala.build.preprocessing.mdsandbox.SnippetPackager
+
+case object MarkdownPreprocessor extends Preprocessor {
+  def preprocess(
+    input: Inputs.SingleElement,
+    logger: Logger
+  ): Option[Either[BuildException, Seq[PreprocessedSource]]] =
+    input match {
+      case markdown: Inputs.MarkdownFile =>
+        val res = either {
+          val content = value(PreprocessingUtil.maybeRead(markdown.path))
+          val preprocessed = value {
+            MarkdownPreprocessor.preprocess(
+              Right(markdown.path),
+              content,
+              markdown.subPath,
+              ScopePath.fromPath(markdown.path),
+              logger
+            )
+          }
+          preprocessed
+        }
+        Some(res)
+
+      case _ =>
+        None
+    }
+  
+  private def preprocess(
+    reportingPath: Either[String, os.Path],
+    content: String,
+    subPath: os.SubPath,
+    scopePath: ScopePath,
+    logger: Logger
+  ): Either[BuildException, List[PreprocessedSource.InMemory]] = either {
+
+    val packager: SnippetPackager = new SnippetPackager(subPath.toString, content)
+    val topWrapperLen = 0
+
+    val parsedMain: String = packager.buildScalaMain()
+    val mainProcessingOutput =
+      value(ScalaPreprocessor.process(
+        parsedMain,
+        reportingPath,
+        scopePath / os.up,
+        logger
+      ))
+        .getOrElse(ProcessingOutput(BuildRequirements(), Nil, BuildOptions(), None))
+    val mainCode = mainProcessingOutput.updatedContent.getOrElse(parsedMain)
+    val mainClassName = s"Markdown_${subPath.last}"
+    val mainRelPath   = os.rel / (subPath / os.up) / s"${subPath.last.stripSuffix(".md")}_mainmd.scala"
+    val mainFile = PreprocessedSource.InMemory(
+      reportingPath.map((subPath, _)),
+      mainRelPath,
+      mainCode,
+      topWrapperLen,
+      Some(mainProcessingOutput.opts),
+      Some(mainProcessingOutput.globalReqs),
+      mainProcessingOutput.scopedReqs,
+      Some(CustomCodeWrapper.mainClassObject(Name(mainClassName)).backticked),
+      scopePath
+    )
+
+    val parsedTest: String = packager.buildScalaTest()
+    val testProcessingOutput =
+      value(ScalaPreprocessor.process(
+        parsedTest,
+        reportingPath,
+        scopePath / os.up,
+        logger
+      ))
+        .getOrElse(ProcessingOutput(BuildRequirements(), Nil, BuildOptions(), None))
+    val testCode = testProcessingOutput.updatedContent.getOrElse(parsedTest)
+    val testRelPath   = os.rel / (subPath / os.up) / s"${subPath.last.stripSuffix(".md")}_testmd.scala"
+    val testFile = PreprocessedSource.InMemory(
+      reportingPath.map((subPath, _)),
+      testRelPath,
+      testCode,
+      topWrapperLen,
+      Some(testProcessingOutput.opts),
+      Some(testProcessingOutput.globalReqs),
+      testProcessingOutput.scopedReqs,
+      None,
+      scopePath
+    )
+
+    val parsedRaw: String = packager.buildScalaRaw()
+    val rawProcessingOutput =
+      value(ScalaPreprocessor.process(
+        parsedRaw,
+        reportingPath,
+        scopePath / os.up,
+        logger
+      ))
+        .getOrElse(ProcessingOutput(BuildRequirements(), Nil, BuildOptions(), None))
+    val rawCode = rawProcessingOutput.updatedContent.getOrElse(parsedRaw)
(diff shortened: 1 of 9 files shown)


modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -185,10 +185,20 @@ object Run extends ScalaCommand[RunOptions] {
       finally watcher.dispose()
     }
     else {
+      val inputsMd: Inputs = inputs.copy(
+        elements = inputs.elements :+ scala.build.preprocessing.mdsandbox.MdRunner.generateRunnerFile(inputs)
+      )
+
+      val MDMODE = true
```
BlackAnubis7, line 192: `MDMODE` is a temporarily implemented mode in which a class that runs Markdown snippets overrides main class defined in the compiled project. It's hardcoded, so it can't remain in this form

modules/build/src/main/scala/scala/build/preprocessing/MarkdownPreprocessor.scala
```diff
@@ -0,0 +1,127 @@
...
+      Some(rawProcessingOutput.opts),
+      Some(rawProcessingOutput.globalReqs),
+      rawProcessingOutput.scopedReqs,
+      None,
+      scopePath
+    )
+
+    // System.out.println(s"<<< MAIN\n${mainCode}\nMAIN >>>")
+    // System.out.println(s"<<< TEST\n${testCode}\nTEST >>>")
+    // System.out.println(s"<<< RAW\n${rawCode}\nRAW >>>")
```
Gedochao: lines like this need to go.
(just marking things that need to change before we'll be able to consider merging the PoC)

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/MdRunner.scala
```diff
@@ -0,0 +1,30 @@
+package scala.build.preprocessing.mdsandbox
+
+import scala.build.Inputs
+
+object MdRunner {
```
Gedochao, line 5: Hmm... `MdRunner` doesn't seem like a fortunate name, as far as I understand the code, this seems to be just a handler for wrapping the Scala snippets from markdown. So maybe:
```suggestion
object MarkdownScalaSnippetWrapper {
```
This also makes it clearer, that we will need the equivalent of this class to handle Java snippets in the future.

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/MdRunner.scala
```diff
@@ -0,0 +1,30 @@
+package scala.build.preprocessing.mdsandbox
```
Gedochao, line 1:
```suggestion
package scala.build.internal
```
I'd change the package to internal, to put it next to the script `CustomCodeWrapper` (and adjust the file location accordingly)

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
```diff
@@ -0,0 +1,147 @@
+package scala.build.preprocessing.mdsandbox
+
+import MarkdownSnippet.Fence
+
+class SnippetPackager(val fileName: String, val snippets: Seq[Fence]) {
```
Gedochao, line 5: naming it as a `packager` seems unfortunate, since it's got nothing to do with how we package things with the `package` command.
It's really just a tool for extracting the snippets from markdown, so maybe something like:
```suggestion
class MarkdownSnippetUtil(val fileName: String, val snippets: Seq[Fence]) {
```
`-Util` files in `scala-cli` are usually objects which don't require instantiating, so maybe we need an even better name

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
```diff
@@ -0,0 +1,147 @@
...
+  def this(fileName: String, content: String) = this(fileName, MarkdownSnippet.findFences(content))
```
Gedochao, line 6: it'd make sense to make this constructor into an `apply` method on the companion object

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
Gedochao: actually, this could be a `-Util` class, since I don't think any of the methods truly require a state. the `fileName` and `snippets` could instead be passed around to static methods on an object.

in other words:
```suggestion
object MarkdownSnippetUtil {
```

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  /** Returns Scala snippets packed into classes and glued together into an object */
+  def buildScalaMain(): String = {
+    if (runSnippets.isEmpty) s"object $runObjectIdentifier {def execute(): Unit = {}}"  // no snippets
+    else (0 until runSnippets.length).foldLeft(
+      s"object $runObjectIdentifier {def execute(): Unit = {"
+    ) (
+      (sum, index) => 
+        if (runSnippets(index).resetScope || index == 0) sum :++ s"new ${runClassName(index)}; "
+        else sum  // that class hasn't been created
```
Gedochao, line 28: ?

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /** Returns test snippets packed into classes and glued together into an object */
+  def buildScalaTest(): String = {
+    if (testSnippets.isEmpty) s"object $testObjectIdentifier {}"  // no snippets
+    else buildScalaTest(0, 0) :++ "}"
+  }
+
+  /*
+   * TESTING APPROACH A - RAW TESTING
```
Gedochao, line 58: seems like we'll be going with approach A, so the code needs to solidify on that

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/MarkdownSnippet.scala
```diff
@@ -0,0 +1,121 @@
...
+
+    /** 
+      * @return `true` if this snippet should be ignored, `false` otherwise 
+      */
+    def shouldIgnore: Boolean = info(0) != "scala" || info.contains("ignore")
+
+    /**
+      * @return `true` if this snippet should have its scope reset, `false` otherwise
+      */
+    def resetScope: Boolean = info.contains("reset")
```
Gedochao, line 24: I don't particularly like using `reset` as the keyword for separating scopes. 
something like `--fresh-scope`, `--markdown-flush`, `--fresh-wrapper` (or anything less generalizing, really) would be better, imo.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -185,10 +185,20 @@ object Run extends ScalaCommand[RunOptions] {
       finally watcher.dispose()
     }
     else {
+      val inputsMd: Inputs = inputs.copy(
+        elements = inputs.elements :+ scala.build.preprocessing.mdsandbox.MdRunner.generateRunnerFile(inputs)
```
Gedochao, line 189: This should be detected and included in an `Inputs` instance for all commands, instead of being just added here. (with proper handling for allowing  markdown files in place)

modules/cli/src/main/scala/scala/cli/commands/Run.scala
Gedochao: so I guess for the PoC we'll need proper detection for this mode.
for as long as this is experimental, we'll probably settle on something like:
- we could have a `--markdown` flag to enable detecting `.md` files in directories passed as inputs
- if a `*.md` file is passed explicitly as an input, `--markdown` should probably get automatically enabled 
- however, if a `*.md` file is passed explicitly, I'm not sure if we should by default only include that single one (and any others passed explicitly), or should `--markdown` then be enabled for the whole command (so `.md` files included in subdirectories passed as sources would also be included)
  - cc @alexarchambault @lwronski @romanowski
- we might want to warn about Markdown support being experimental when `--markdown` is being passed (or maybe even name it `--markdown-experimental`, since after it's fully implemented, the flag will no longer be necessary)
- ofc, to make it fully stable, we'll need a separate scope for markdown files, probably (alongside `test` and `main`)

Gedochao:
I don't think we'll need the `mdsandbox` package at all, the contents should go to `scala.build.internal` probably (but that can be decided later, since there's stuff to be fixed in the code first)

Gedochao:
We'll need a lot more scaladocs on the new classes & functions.
plus a proper documentation for users to play with it while it's experimental.

Gedochao:
We'll need to cover all the expected use cases with tests, by the way.

modules/build/src/main/scala/scala/build/preprocessing/mdsandbox/SnippetPackager.scala
BlackAnubis7: If we have following snippets:
- `scala`
- `scala`
- `scala reset`

the wrapper will create `Snippet0` and `Snippet2` classes for them. The middle snippet, as its scope has not been reset, does not have its own class, and so `Snippet1` class is being omitted here not to try to execute a non-existent class (thus raising an exception)

Gedochao:
Closing this in favor of the reworked PR: https://github.com/VirtusLab/scala-cli/pull/1268

PR #1227, opened 2022-08-02
Add option to specify custom location of `.scalafmt.conf` for `fmt` command.

### LINKED ISSUE #1206: Enable passing custom location of `.scalafmt.conf` for `fmt` command
I would like to be able to pass custom location of `.scalafmt.conf` file when running `fmt` command.

`fmt` command option, which will allow me to pass path to `.scalafmt.conf` file.

**Additional context**
`--config` option in `scalafmt` is example how the new option should work.

wleczny:


modules/cli-options/src/main/scala/scala/cli/commands/FmtOptions.scala
```diff
@@ -46,6 +46,9 @@ final case class FmtOptions(
     scalafmtArg: List[String] = Nil,
 
   @Group("Format")
+  @HelpMessage("Custom path to the directory with .scalafmt.conf file.")
+    configDir: Option[String] = None,
```
Gedochao: why pass the path to the directory containing the file and not the file itself?
It'd make a lot more sense to be able to pass a path to any text file containing the configuration. 
Someone might even keep a couple of different configurations in one directory, to run for different setups.

Moreover, this should be as close to `scalafmt`'s native options as our API allows.

AFAIK there are 2 options allowing to pass the configuration from a custom location to `scalafmt` itself:

- the `--config` (or `-c`) option, which allows to pass exactly what I mentioned - a path to any file (and not its surrounding directory
- the ` --config-str` option, which allows to pass a configuration as a string from the command line (similarly to how `scala-cli` allows to pass snippets)

(more on that in [scalafmt's own docs](https://scalameta.org/scalafmt/docs/installation.html))

Now, we probably should still rename these, as they will be too ambigous in `scala-cli` context.
Some suggestions:
- `--config` could become `--scalafmt-conf` (maybe with an alias for `--scalafmt-config`)
- `--config-str` similarly could be `--scalafmt-conf-str `(with aliases like `--scalafmt-config-str` and `--scalafmt-conf-snippet`)

Gedochao: (support for an equivalent of `--config-str` is out of scope of the ticket btw, but since you're touching this anyway... 😉 )

wleczny: What if `--scalafmt-config-str` doesn't have `version` or `dialect` specified. Should `fmt` add it under the hood or throw an error?

Gedochao: add it under the hood, definitely.
whatever would be passed with `--scalafmt-conf-str` should be treated the same as any other configuration file.

website/docs/commands/fmt.md
```diff
@@ -167,3 +169,7 @@ If the `--save-scalafmt-conf` option is passed, then `fmt` command behaves as fo
 - In the **first** case `fmt` uses the found `.scalafmt.conf` file to run `scalafmt`.
 - In the **second** case `fmt` [infers](/docs/commands/fmt#scalafmt-version-and-dialect) missing parameters, writes them directly into the previously found file and then uses this file to run `scalafmt`.
 - In the **third** case `fmt` creates a `.scalafmt.conf` file in the current workspace directory, writes [inferred](/docs/commands/fmt#scalafmt-version-and-dialect) version and dialect into it and uses it to run `scalafmt`.
+
+:::note
+If the configuration is passed in the `--scalafmt-conf-str` option, Scala CLI behaves exactly the same as it would find `.scalafmt.conf` with specified configuration in the workspace.
```
Gedochao:
```suggestion
If the configuration is passed in the `--scalafmt-conf-str` option, Scala CLI will behave exactly the same as if it would find the specified configuration in a `.scalafmt.conf` file in the workspace.
```

website/docs/commands/fmt.md
```diff
@@ -30,6 +30,8 @@ You can skip passing either of those, which will make Scala CLI to infer a defau
 - If a `.scalafmt.conf` file is present in the workspace and it has the field defined, the value will be read from there, unless explicitly specified with Scala CLI options.
 - Otherwise, the default `scalafmt` **version** will be the latest one used by your Scala CLI version (so it is subject to change when updating Scala CLI). The default **dialect** will be inferred based on Scala version (defined explicitly by `-S` option, or default version if option would not be passed).
 
+It is possible to pass configuration as a string directly from the command line, using `--scalafmt-conf-str` option. If the configuration is passed this way, Scala CLI behaves exactly the same as it would find `.scalafmt.conf` with specified configuration in the workspace.
```
Gedochao:
```suggestion
It is possible to pass the configuration as a string directly from the command line, using `--scalafmt-conf-str` option. If the configuration is passed this way, Scala CLI will behave exactly the same as if it would find the specified configuration in a `.scalafmt.conf` file in the workspace.
```

lwronski (approved):

Gedochao (approved):

PR #1228, opened 2022-08-02
Add 0.1.11 release notes to release_notes.md

BlackAnubis7:


website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+### Add support for running multiple snippets of the same kind
+Scala CLI now allows passing and running multiple snippets. Each snippet is then treated as a
+separate input by Scala CLI.
+
+```bash
+scala-cli --scala-snippet '@main def main() = println(Messages.hello)' --scala-snippet 'object Messages { def hello = "Hello" }'
+```
+
+```text
+Hello
```
Gedochao: when putting the output of a bash example in a doc, these chained snippets should be wrapped in a `ChainedSnippets` component to render properly on our docs page.
You can refer [here](https://github.com/VirtusLab/scala-cli/blob/cad549fbc0de22e61588b8ec7d3394404724a3f8/website/docs/getting_started.md?plain=1#L16) for an example.

Now, the downside of using `ChainedSnippets` here is that it's our component and it will not work on the GitHub releases page when you copy the doc back there after the review is done as-is. 
So you might notice that in this particular doc we are fine with pasting examples such as this one in a plain text snippet, prefixing the bash prompt lines with a `$` ([here](https://github.com/VirtusLab/scala-cli/blob/cad549fbc0de22e61588b8ec7d3394404724a3f8/website/docs/release_notes.md?plain=1#L46) is an example for that).

Your call which way to fix that, really - but I think the latter is a better convention here, for simplicity's sake

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
 
 # Release notes
 
+## [v0.1.11](https://github.com/VirtusLab/scala-cli/releases/tag/v0.1.11)
+
+### Add support for running multiple snippets of the same kind
+Scala CLI now allows passing and running multiple snippets. Each snippet is then treated as a
```
Gedochao:
```suggestion
Scala CLI now allows to pass multiple snippets of the same kind. 

It was previously possible to mix different kinds (so to pass a Java snippet alongside a Scala one), but not for example 2 separate Scala snippets. That limitation no longer applies

When passed this way, each snippet is then treated as a
```

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+```bash
+scala-cli --scala-snippet '@main def main() = println(Messages.hello)' --scala-snippet 'object Messages { def hello = "Hello" }'
+```
+
+```text
+Hello
+```
+Added in [#1182](https://github.com/VirtusLab/scala-cli/pull/1182) by [Gedochao](https://github.com/Gedochao)
+
+### Make main class optional in preamble-less assemblies
```
Gedochao: this is a bit of a niche feature (not to say obscure, which is why the `--main-class-in-manifest` option is hidden).
We can still highlight it in this section, but I think it should be mentioned at the end.

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+Added in [#1199](https://github.com/VirtusLab/scala-cli/pull/1199) by [alexarchambault](https://github.com/alexarchambault)
+
+### Define `output` option for `package` command with using directives
+It is now possible to pass some of the `package` options with [using directives](/docs/guides/using-directives) instead of passing them directly from bash.
+
+Added in [#1213](https://github.com/VirtusLab/scala-cli/pull/1213) by [wleczny](https://github.com/wleczny)
+
+### Important fixes & enhancements
+#### Prevent erroneous using directives from blocking the initial run of BSP
+Erroneous `using` directives are now prevented from blocking the initial BSP setup (fixes [#1097](https://github.com/VirtusLab/scala-cli/issues/1097))
```
Gedochao: I think this fix can be described in more detail.
```suggestion
Up till now, running the `setup-ide` sub-command on sources containing `using directives` with syntax errors or pointing to dependencies which could not be fetched would create a `BSP` setup which could not be imported correctly by IDEs. This is no longer the case and such a `BSP` connection should now import correctly, so that it's possible to fix the faulty code within the comfort of one's IDE of choice.

This fixes [#1097](https://github.com/VirtusLab/scala-cli/issues/1097)
```

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+
+Added in [#1195](https://github.com/VirtusLab/scala-cli/pull/1195) by [Gedochao](https://github.com/Gedochao)
+
+### Work in progress
+#### Allow to global turn off/on actionable diagnostic
```
Gedochao:
```suggestion
#### Allow to globally turn actionable diagnostics on or off
```

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+`actionable diagnostics` can now be globally enabled/disabled in config db using
```
Gedochao:
```suggestion
It is now possible to globally enable or disable actionable diagnostics using the `config` sub-command.
The relevant configuration is under the `actions` key.
```

website/docs/release_notes.md
```diff
@@ -5,6 +5,97 @@ sidebar_position: 99
...
+```text
+scala-cli config actions true
+```
+and
+```text
+scala-cli config actions false
```
Gedochao: I think the example for turning it on is sufficient

website/docs/release_notes.md
BlackAnubis7: Second approach implemented

website/docs/release_notes.md
```diff
@@ -5,6 +5,94 @@ sidebar_position: 99
...
+
+Added in [#1192](https://github.com/VirtusLab/scala-cli/pull/1192) by [wleczny](https://github.com/wleczny)
+
+### Add bloop sub-command
+Scala CLI now has a (hidden for now) bloop sub-command, that runs a command using the Scala CLI Bloop server (while the mainline Bloop bloop CLI uses its default Bloop server). This is handy when debugging things on Scala CLI for example, allowing one to manually run scala-cli bloop projects or scala-cli bloop compile.
+
+Added in [#1199](https://github.com/VirtusLab/scala-cli/pull/1199) by [alexarchambault](https://github.com/alexarchambault)
+
+### Define `output` option for `package` command with using directives
+It is now possible to pass some of the `package` options with [using directives](/docs/guides/using-directives) instead of passing them directly from bash.
```
wleczny: I would rather say that it is possible to pass `output` option (passing other options was possible before and only documentation has been added for them).

website/docs/release_notes.md
BlackAnubis7: Changed

website/docs/release_notes.md
BlackAnubis7: Marked PR notes are now moved to the end of the section

website/docs/release_notes.md
BlackAnubis7: Changed

website/docs/release_notes.md
BlackAnubis7: Changed

website/docs/release_notes.md
BlackAnubis7: Changed

website/docs/release_notes.md
BlackAnubis7: Changed

website/docs/release_notes.md
```diff
@@ -5,6 +5,94 @@ sidebar_position: 99
...
+Hello
+```
+Added in [#1182](https://github.com/VirtusLab/scala-cli/pull/1182) by [Gedochao](https://github.com/Gedochao)
+
+### Make main class optional in preamble-less assemblies
+It is now allowed to generate an assembly, even for code that has no main class, when `--preamble=false` is passed. This can be useful for libraries, if users want to pass the assembly to tools such as proguard. This also accepts a (hidden) `--main-class-in-manifest=false` option if users want not only no preamble, but also no mention of main class in the assembly manifest (`META-INF/MANIFEST.MF` in the assembly JAR). The latter option is useful for tools, such as the hadoop jar command, that behave differently depending on the presence or not of a main class in the manifest.
+
+Added in [#1200](https://github.com/VirtusLab/scala-cli/pull/1200) by [alexarchambault](https://github.com/alexarchambault)
+
+### Generate scalafmt conf file on fmt command
```
romanowski: IMHO this should be on top

website/docs/release_notes.md
```diff
@@ -5,6 +5,94 @@ sidebar_position: 99
...
+Scala CLI now generates a default `.scalafmt.conf` file using `fmt`, `format` or `scalafmt` command if such file is absent. Previously, a `Scalafmt requires explicitly specified version.` error would be raised instead.
+
+Added in [#1192](https://github.com/VirtusLab/scala-cli/pull/1192) by [wleczny](https://github.com/wleczny)
+
+### Add bloop sub-command
+Scala CLI now has a (hidden for now) bloop sub-command, that runs a command using the Scala CLI Bloop server (while the mainline Bloop bloop CLI uses its default Bloop server). This is handy when debugging things on Scala CLI for example, allowing one to manually run scala-cli bloop projects or scala-cli bloop compile.
+
+Added in [#1199](https://github.com/VirtusLab/scala-cli/pull/1199) by [alexarchambault](https://github.com/alexarchambault)
+
+### Define `output` option for `package` command with using directives
```
romanowski: Same here

website/docs/release_notes.md
Gedochao: by the way, I wouldn't showcase this to users as generating the `.scalafmt.conf` file.
Yes, that's what is happening under the hood, but what really matters to the users is that they can skip creating the configuration file and just pass the version & dialect with scala-cli (or even rely on defaults).

So maybe something like:
```suggestion
### Make `.scalafmt.conf` optional when running the `fmt` sub-command
```
(and change the description accordingly)

Gedochao: some example for how the version & dialect can be passed (and how the defaults are inferred) might be a good idea, too.

Gedochao: ...but yes, we still can mention that generating the configuration (and persisting it in the workspace) is possible with the `--save-scalafmt-conf` option.

BlackAnubis7: Moved to top, changed and expanded the description

website/docs/release_notes.md
BlackAnubis7: Moved to the 2nd topmost position

website/docs/release_notes.md
BlackAnubis7: Description has been changed accordingly

website/docs/release_notes.md
```diff
@@ -5,6 +5,106 @@ sidebar_position: 99
...
+## [v0.1.11](https://github.com/VirtusLab/scala-cli/releases/tag/v0.1.11)
+
+### Make `.scalafmt.conf` optional when running the `fmt` command
+Scala CLI can now run the `fmt` command without a `.scalafmt.conf` file present. Previously, if such a file was absent, a `Scalafmt requires explicitly specified version.` error was raised while using the `fmt` command.
+
+The Scala CLI `fmt` command now supports passing the `scalafmt` version and dialect directly from the command line, using the `--scalafmt-dialect` and `--scalafmt-version` options respectively:
+```
+scala-cli fmt --scalafmt-dialect scala3 --scalafmt-version 3.5.8
+```
+Either of those can be skipped, which will make Scala CLI infer a default value:
```
Gedochao: NIT: I wonder if explaining it in this much detail is necessary in the release notes, looks slightly verbose.
I'd just end with this sentence and cut the following bullet points.
```suggestion
Either of those can be skipped, which will make Scala CLI infer a default value.
```
...but it can also stay the way it is, I suppose.

Gedochao (approved):
LGTM

website/docs/release_notes.md
BlackAnubis7: I agree, this much information could be confusing in Release Notes. I cut the bullet points

PR #1229, opened 2022-08-03
Actionable diagnostics

lwronski:
The next part of changes that are necessary before the integration with Metals.

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/VerbosityOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/VerbosityOptions.scala
@@ -12,7 +12,9 @@ final case class VerbosityOptions(
     verbose: Int @@ Counter = Tag.of(0),
   @HelpMessage("Interactive mode")
   @Name("i")
-    interactive: Option[Boolean] = None
+    interactive: Option[Boolean] = None,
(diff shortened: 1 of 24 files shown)


modules/cli-options/src/main/scala/scala/cli/commands/VerbosityOptions.scala
```diff
@@ -12,7 +12,9 @@ final case class VerbosityOptions(
     verbose: Int @@ Counter = Tag.of(0),
   @HelpMessage("Interactive mode")
   @Name("i")
-    interactive: Option[Boolean] = None
+    interactive: Option[Boolean] = None,
+  @HelpMessage("Enable actionable diagnostics")
+    actions: Option[Boolean] = None
```
Gedochao, line 17: any reason why we don't want those on by default?

lwronski: IMO we shouldn't enable it by default now, it still require a work, for example integration with Metals.

Gedochao (approved):

alexarchambault (approved):
Ok for either squash-and-merge, or adjust format / reference doc in the first 6 commits, so that the "Fix scalafmt" one isn't necessary.

PR #1253, opened 2022-08-16
Update org name to VirtusLab for downloading scalafmt-native-image

lwronski:
`scalafmt-native-image` is now downloaded from virtuslab organisation. Native launcher is built in fork [scalafmt-native-image](https://github.com/VirtusLab/scalafmt-native-image)

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
@@ -5,15 +5,22 @@ import scala.cli.commands.FmtOptions
 import scala.cli.commands.util.SharedOptionsUtil._
 import scala.util.Properties
 
+import coursier.core.Version
+
 object FmtOptionsUtil {
   implicit class FmtOptionsOps(v: FmtOptions) {
     import v._
     def binaryUrl(version: String): (String, Boolean) = {
       val osArchSuffix0 = osArchSuffix.map(_.trim).filter(_.nonEmpty)
         .getOrElse(FetchExternalBinary.platformSuffix())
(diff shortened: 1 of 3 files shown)


lwronski:
The `docs-tests` is failed, because we use there latest stable `scala-cli` version, which already trying to download `scalafmt-native-image` with the wrong URL.

IMO we can merge it and disable check if CI `docs-tests` is passed. After release, we enable again it.

modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
```diff
@@ -12,7 +12,7 @@ object FmtOptionsUtil {
       val osArchSuffix0 = osArchSuffix.map(_.trim).filter(_.nonEmpty)
         .getOrElse(FetchExternalBinary.platformSuffix())
       val tag0           = scalafmtTag.getOrElse("v" + version)
-      val gitHubOrgName0 = scalafmtGithubOrgName.getOrElse("alexarchambault/scalafmt-native-image")
+      val gitHubOrgName0 = scalafmtGithubOrgName.getOrElse("virtuslab/scalafmt-native-image")
```
alexarchambault: I think we should still use the former URL, for versions lower than `3.5.9`. Unless these versions' artifacts can be copied over to the new repository too?

lwronski: I didn't find any way, how to copy artifacts from old repository to new

alexarchambault: I guess you'd need to write a script to do it (using the GitHub API to create releases and upload things to the new repo).

alexarchambault (approved):

PR #1259, opened 2022-08-17
Restrict directives based on command used

romanowski:


modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveHandler.scala
```diff
@@ -24,6 +24,9 @@ trait DirectiveHandler[T] {
   def usageMd: String       = s"`$usage`"
   def examples: Seq[String] = Nil
 
+  /** Is this directive an advanved feature, that will not be accessible when running scala-cli as `scala` */
```
alexarchambault: It seems you used that with the opposite meaning down below (`isRestricted` is true for non-advanced directives).

Gedochao (approved):

PR #1260, opened 2022-08-17
Project settings file

### LINKED ISSUE #646: Scala CLI does not compute root project properly

### LINKED ISSUE #934: Using directive to configure sources / source directories

In some cases, one may want to configure sources to compile a given source within a source (e.g. script with she-bang or define a non-breakable dependency or to create sbt/mill like project). 

Additionally, we should have a default file name (e.g. `project.scala`) that will be a recommended place to put most using directives (dependencies, etc). When no inputs are provided, scala-cli should assume this file is an input if present in CWD, so it would be possible to run `scala-cli compile` or `scala-cli test` for projects that wants to have more predefined structure (e.g. libraries)>

wleczny:


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -227,6 +227,8 @@ object CrossSources {
       lazy val dir     = sourcePath / os.up
       lazy val subPath = sourcePath.subRelativeTo(dir)
       if (os.isDir(sourcePath)) Right(Inputs.singleFilesFromDirectory(Inputs.Directory(sourcePath)))
+      else if (sourcePath.endsWith(os.SubPath("project.settings.scala")))
```
lwronski: Do we want accept also files such as `Hello.project.settings.scala`?
```suggestion
      else if (sourcePath == os.SubPath("project.settings.scala"))
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -224,6 +226,8 @@ object Inputs {
       .collect {
         case p if p.last.endsWith(".java") =>
           Inputs.JavaFile(d.path, p.subRelativeTo(d.path))
+        case p if p.last.endsWith("project.settings.scala") =>
```
lwronski: The same as above.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -271,22 +281,51 @@ object Inputs {
 
     assert(validElems.nonEmpty)
 
-    val (inferredWorkspace, inferredNeedsHash, workspaceOrigin) = validElems
-      .collectFirst {
-        case d: Directory => (d.path, true, WorkspaceOrigin.SourcePaths)
+    val (inferredWorkspace, inferredNeedsHash, workspaceOrigin) = {
+      val settingsFiles = validElems.flatMap {
```
lwronski: Does the `Seq` of `settingsFiles` contain different configuration files? What happened in this case?

```
$ ls src
Hello.scala
project.settings.scala
$ scala-cli src src/project.settings.scala
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -271,22 +281,51 @@ object Inputs {
...
+          case d: Directory =>
+            if (dirsAndFiles.length > 1)
+              System.err.println(
+                s"Warning: setting ${d.path} as the project root directory for this run."
+              )
+            (d.path, true, WorkspaceOrigin.SourcePaths)
+          case f: SourceFile =>
+            if (dirsAndFiles.length > 1)
+              System.err.println(
+                s"Warning: setting ${f.path / os.up} as the project root directory for this run."
```
lwronski, line 333: Does that mean every run of: 
```
scala-cli Hello.scala
```
will print this warning?

website/docs/reference/root-dir.md
```diff
@@ -0,0 +1,60 @@
...
+  - to write class files
+  - for Bloop
+
+## Setting root directory
+
+First of all Scala CLI checks every passed input for the `project.settings.scala` file:
+- If the `project.settings.scala` file is passed explicitly as a **source**, Scala CLI sets its parent directory as the root directory.
+- If the input is a **directory**, Scala CLI looks for the `project.settings.scala` inside this directory. If the file is found, Scala CLI sets the passed directory as the root directory.
+
+If more than one `project.settings.scala` file is found, Scala CLI uses only **the first one** to set the root directory and raises **warning** saying which one was used.
```
lwronski, line 19: > **the first one**

We should mention how order is determined.

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -552,6 +552,59 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+           |}
+           |""".stripMargin,
+      os.rel / "bar.scala" -> ""
+    )
+    inputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "custom-dir", "bar.scala", extraOptions).call(cwd = root)
+      assert(os.exists(root / "custom-dir" / ".scala-build"))
+      assert(!os.exists(root / ".scala-build"))
+    }
+  }
```
lwronski: It seems that these tests can be moved to unit tests, for example into `InputsTests`.

One integration test is enough, e.g. the one test with gist you can keep in `RunTestsDefinitions`.

modules/build/src/test/scala/scala/build/tests/InputsTests.scala
```diff
@@ -1,8 +1,30 @@
...
+  val directories     = Directories.under(extraRepoTmpDir)
+  def bloopConfigOpt  = Some(BloopServer.bloopConfig)
+  val baseOptions = BuildOptions(
+    internal = InternalOptions(
+      localRepository = LocalRepo.localRepo(directories.localRepoDir),
+      keepDiagnostics = true
+    )
+  )
+  def sv2 = "2.13.5"
+  val defaultOptions = baseOptions.copy(
```
lwronski: It will be good enough if you use `baseOptions` in your tests. You don't have to specify Scala version.

lwronski (approved):

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -190,10 +191,11 @@ object Inputs {
       extends OnDisk with SourceFile with AnyScalaFile with AnyScript {
     lazy val path: os.Path = base / subPath
   }
-  final case class ScalaFile(base: os.Path, subPath: os.SubPath)
+  case class ScalaFile(base: os.Path, subPath: os.SubPath)
```
alexarchambault: Inheriting from case classes is kind of an anti-pattern… They should be considered as effectively final. Better introduce a super-type if you really need `ScalaFile` and `ProjectSettings` to implement some common type in some contexts.

alexarchambault: Alternatively, adding a method such as `def isProjectSettings: Boolean` to `ScalaFile` would be fine for example, if it can work for you here, rather than adding a new `ProjectSettings` class.

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -233,6 +237,18 @@ object Inputs {
...
+  def projectSettingsFiles(elements: Seq[Inputs.Element]): Seq[Inputs.ProjectSettings] =
+    elements.flatMap {
+      case f: ProjectSettings => Seq(f)
+      case d: Directory       => Inputs.configFileFromDirectory(d)
+      case _                  => Nil
+    }.distinct
+
+  def configFileFromDirectory(d: Inputs.Directory): Seq[Inputs.ProjectSettings] =
+    if (os.exists(d.path / "project.settings.scala"))
+      Seq(Inputs.ProjectSettings(d.path, os.SubPath("project.settings.scala")))
```
alexarchambault: NIT I believe this works (and is easier to read):
```suggestion
      Seq(Inputs.ProjectSettings(d.path, os.sub / "project.settings.scala"))
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -271,22 +288,37 @@ object Inputs {
...
+                s"Warning: setting ${d.path} as the project root directory for this run."
+              )
+            (d.path, true, WorkspaceOrigin.SourcePaths)
+          case f: SourceFile =>
+            if (dirsAndFiles.length > 1)
+              System.err.println(
+                s"Warning: setting ${f.path / os.up} as the project root directory for this run."
+              )
+            (f.path / os.up, true, WorkspaceOrigin.SourcePaths)
+        }.getOrElse((os.pwd, true, WorkspaceOrigin.Forced))
```
alexarchambault: Having several nested `getOrElse` is kind of an anti-pattern too. Better use `orElse`, and a single `getOrElse` at the end, like
```scala
settingsFiles
  .collectFirst { s =>
    …
  }
  .orElse {
    dirsAndFiles.collectFirst {
      …
    }
  }
  .getOrElse {
    (os.pwd, true, WorkspaceOrigin.Forced)
  }
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -404,6 +436,7 @@ object Inputs {
             List(resolve(url, content))
         }
       }
+      else if (arg.endsWith("project.settings.scala")) Right(Seq(ProjectSettings(dir, subPath)))
```
alexarchambault: Not sure about the `arg.endsWith` here. Wouldn't this work better?
```suggestion
      else if (path.last == "project.settings.scala") Right(Seq(ProjectSettings(dir, subPath)))
```

modules/build/src/main/scala/scala/build/Inputs.scala
```diff
@@ -271,22 +288,37 @@ object Inputs {
...
-          case _: Virtual =>
-            val dir = homeWorkspace(validElems, directories)
-            (dir, false, WorkspaceOrigin.HomeDir)
-          case r: ResourceDirectory =>
-            // Makes us put .scala-build in a resource directory :/
-            (r.path, true, WorkspaceOrigin.ResourcePaths)
-          case _: Directory => sys.error("Can't happen")
-        }
+
+      settingsFiles.collectFirst { s =>
```
alexarchambault: That use of `collectFirst` looks odd. Wouldn't something like this work better?
```suggestion
      settingsFiles.headOption.map { s =>
```

modules/build/src/test/scala/scala/build/tests/InputsTests.scala
```diff
@@ -21,4 +36,70 @@ class InputsTests extends munit.FunSuite {
...
+             |    println("Foo")
+             |}
+             |""".stripMargin,
+        os.rel / "bar.scala" -> ""
+      ),
+      inputArgs = Seq("custom-dir", "bar.scala")
+    )
+    testInputs.withBuild(buildOptions, buildThreads, bloopConfigOpt) {
+      (root, _, _) =>
+        assert(os.exists(root / "custom-dir" / ".scala-build"))
```
Gedochao:
```suggestion
        assert(os.exists(root / "custom-dir" / Constants.workspaceDirName))
```

modules/build/src/test/scala/scala/build/tests/InputsTests.scala
```diff
@@ -21,4 +36,70 @@ class InputsTests extends munit.FunSuite {
...
+        assert(!os.exists(root / ".scala-build"))
```
Gedochao:
```suggestion
        assert(!os.exists(root / Constants.workspaceDirName))
```

modules/build/src/test/scala/scala/build/tests/InputsTests.scala
```diff
@@ -21,4 +36,70 @@ class InputsTests extends munit.FunSuite {
...
+             |    println("Foo")
+             |}
+             |""".stripMargin,
+        os.rel / "project.settings.scala" -> ""
+      ),
+      inputArgs = Seq(".", "project.settings.scala")
+    )
+    testInputs.withBuild(buildOptions, buildThreads, bloopConfigOpt) {
+      (root, inputs, _) =>
+        assert(os.exists(root / ".scala-build"))
```
Gedochao:
```suggestion
        assert(os.exists(root / Constants.workspaceDirName))
```

website/docs/reference/root-dir.md
```diff
@@ -0,0 +1,60 @@
...
+## Usage
+
+Scala CLI needs a root directory:
+  - to write mapped sources
+  - to write class files
+  - for Bloop
+
+## Setting root directory
+
+First of all Scala CLI checks every passed input (in the same order in which inputs were passed) for the `project.settings.scala` file:
```
Gedochao:
```suggestion
First of all, Scala CLI checks every passed input (in the same order in which inputs were passed) for the `project.settings.scala` file:
```

website/docs/reference/root-dir.md
```diff
@@ -0,0 +1,60 @@
...
+
+If more than one `project.settings.scala` file is found, Scala CLI uses only **the first one** to set the root directory and raises **warning** saying which one was used.
+
+If no `project.settings.scala` files are found, Scala CLI sets the root directory based on the first file/directory input:
+- If the input is a **directory**, it is set as the root directory. 
+- If the input is a **file**, Scala CLI sets its parent directory as the root directory. 
+
+If more then one file/directory input has ben passed Scala CLI raises the warning saying which directory has been set as the project root directory.
+
+If no `project.settings.scala` files are found and no file/directory inputs have ben passed, Scala CLI sets **cwd** as the project root directory.
```
Gedochao: I think `cwd` isn't known widely enough as an abbreviation to be used in our docs, let's state this more descriptively:
```suggestion
If no `project.settings.scala` files are found and no file/directory inputs have ben passed, Scala CLI sets the current working directory (where Scala CLI was invoked from) as the project root directory.
```

lwronski (approved):

PR #1262, opened 2022-08-17
Add version of Scala to the output of version command

### LINKED ISSUE #1250: Add version of Scala to the output of `scala-cli --version`

Getting the version of the launcher is important mainly for advanced users; beginners would expect to see a version of Scala used.

See more in the @julienrf comment: https://github.com/scala/improvement-proposals/pull/46#discussion_r943420835

When running `scala-cli --version` the default version of Scala is printed as well as the version of the launcher. With a provided argument (e.g. `scala-cli --version .`), the version of Scala as configured in the argument should be provided.

A dedicated command/argument to print Scala version.

lwronski:
I added support for printing info about version of Scala CLI and Scala, the following commands return output:

```
$ scala-cli --version
Scala CLI version 0.1.11
Default Scala version: 3.1.3
$ scala-cli -version
Scala CLI version 0.1.11
Default Scala version: 3.1.3
$ scala-cli version
Scala CLI version 0.1.11
Default Scala version: 3.1.3
```

I skipped added support for printing version for the following command:
```
$ scala-cli -v
```

now `-v` allows user to increase verbosity, to avoid double meaning of `-v`, I skip adding support for this case

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/VersionOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/VersionOptions.scala
@@ -6,7 +6,13 @@ import caseapp._
 @HelpMessage("Print `scala-cli` version")
 final case class VersionOptions(
   @Recurse
-    verbosity: VerbosityOptions = VerbosityOptions()
+    verbosity: VerbosityOptions = VerbosityOptions(),
+  @HelpMessage("Show only plain scala-cli version")
+  @Name("cli")
+    cliVersion: Boolean = false,
+  @HelpMessage("Show only plain scala version") 
+  @Name("scala")
+    scalaVersion: Boolean = false
 )
 // format: on
 
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -16,12 +16,20 @@ import scala.util.Properties
 
 object ScalaCli {
 
+  if (Properties.isWin && isGraalvmNativeImage)
+    // have to be initialized before running (new Argv0).get because Argv0SubstWindows uses csjniutils library
+    // The DLL loaded by LoadWindowsLibrary is statically linke/d in
+    // the Scala CLI native image, no need to manually load it.
+    coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
+
   val progName = (new Argv0).get("scala-cli")
 
-  private def checkName(name: String) =
-    progName == name ||
-    progName.endsWith(s"/$name") ||
-    progName.endsWith(File.separator + name)
+  private def checkName(name: String) = {
+    val baseProgName = if (Properties.isWin) progName.stripSuffix(".exe") else progName
+    baseProgName == name ||
+    baseProgName.endsWith(s"/$name") ||
+    baseProgName.endsWith(File.separator + name)
+  }
 
   private var isSipScala = checkName("scala") || checkName("scala-cli-sip")
 
@@ -170,11 +178,6 @@ object ScalaCli {
     // just make some read / write calls return -1).
     if (!Properties.isWin && isGraalvmNativeImage)
       ignoreSigpipe()
-
-    if (Properties.isWin && isGraalvmNativeImage)
-      // The DLL loaded by LoadWindowsLibrary is statically linked in
-      // the Scala CLI native image, no need to manually load it.
-      coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
 
     if (Properties.isWin && System.console() != null && coursier.paths.Util.useJni())
       // Enable ANSI output in Windows terminal
--- a/modules/cli/src/main/scala/scala/cli/commands/Version.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Version.scala
@@ -5,10 +5,26 @@ import caseapp._
 import scala.build.internal.Constants
 import scala.cli.CurrentParams
 
-object Version extends ScalaCommand[VersionOptions] {
+class Version(isSipScala: Boolean) extends ScalaCommand[VersionOptions] {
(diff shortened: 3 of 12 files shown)


Gedochao (commented):
We don't have tests for the `version` sub-command, do we?
I think adding one for SIP mode and one for vanilla Scala CLI might be worth it.
Otherwise, LGTM

modules/cli/src/main/scala/scala/cli/commands/Version.scala
```diff
@@ -5,10 +5,26 @@ import caseapp._
...
+
+object Version {
+  def versionInfo(isSipScala: Boolean) =
+    val version            = Constants.version
+    val detailedVersionOpt = Constants.detailedVersion.filter(_ != version).fold("")(" (" + _ + ")")
+    val appName =
+      if (isSipScala) "Scala code runner"
+      else "Scala CLI"
+    s"""$appName version: $version$detailedVersionOpt
+       |Default Scala version: ${Constants.defaultScalaVersion}""".stripMargin
```
alexarchambault: What do you think of printing it this way?
```suggestion
       |Scala version (default): ${Constants.defaultScalaVersion}""".stripMargin
```

modules/cli-options/src/main/scala/scala/cli/commands/VersionOptions.scala
```diff
@@ -6,7 +6,11 @@ import caseapp._
 @HelpMessage("Print `scala-cli` version")
 final case class VersionOptions(
   @Recurse
-    verbosity: VerbosityOptions = VerbosityOptions()
+    verbosity: VerbosityOptions = VerbosityOptions(),
+  @HelpMessage("Show only plain scala-cli version")
+    cliVersion: Boolean = false,
+  @HelpMessage("Show only plain scala version") 
+    scalaVersion: Boolean = false
```
alexarchambault: `--cli` and `--scala` aliases would be nice too IMO.

tanishiking:
~Does this PR change the output of `scala-cli version` too?~ (edit: never mind, it looks like so 😅 )
If, yes, Metals's ScalaCli support have to update the way to parse the scala-cli's version
https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/scalacli/ScalaCli.scala#L192-L205 FYI @tgodzik

Gedochao (approved):

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -39,4 +39,33 @@ class SipScalaTests extends ScalaCliSuite {
...
+          if (binaryName == "scala") "Scala code runner version:"
+          else "Scala CLI version:"
+        expect(version.out.text().contains(expectedLauncherVersion))
+        expect(version.out.text().contains(s"Scala version (default): ${Constants.defaultScala}"))
+      }
+    }
+
+  if (TestUtil.isNativeCli) {
+    test("version command print detailed info run as scala") {
+      runVersionCommand("scala")
```
alexarchambault, line 66: The name should be `scala.exe` on Windows.

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -39,4 +39,33 @@ class SipScalaTests extends ScalaCliSuite {
...
+    }
+
+    test("version command print detailed info run as scala-cli") {
+      runVersionCommand("scala-cli")
```
alexarchambault, line 70: Like above, the name should have a `.exe` extension on Windows too.

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -16,6 +16,12 @@ import scala.util.Properties
 
 object ScalaCli {
 
+  if (Properties.isWin && isGraalvmNativeImage)
+    // have to be initialized before running (new Argv0).get because Argv0SubstWindows uses csjniutils library
+    // The DLL loaded by LoadWindowsLibrary is statically linke/d in
+    // the Scala CLI native image, no need to manually load it.
+    coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
+
```
alexarchambault, line 24: On Windows, a `.exe` extension might need to be stripped from `progName`.

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -16,12 +16,20 @@ import scala.util.Properties
...
   val progName = (new Argv0).get("scala-cli")
 
-  private def checkName(name: String) =
-    progName == name ||
-    progName.endsWith(s"/$name") ||
-    progName.endsWith(File.separator + name)
+  private def checkName(name: String) = {
+    val baseProgName = progName.stripSuffix(".exe")
```
alexarchambault: I think it'd be safer to only do that (stripping `.exe`) on Windows.

alexarchambault (approved):

tgodzik:
@tanishiking good catch! We need to adjust to it in Metals before the next release.

tgodzik:
Does this change makes sense: https://github.com/scalameta/metals/pull/4288 ?

lwronski:
> Does this change makes sense: [scalameta/metals#4288](https://github.com/scalameta/metals/pull/4288) ?

Or you can use `--cli-version` to print plain Scala CLI version:

```
scala-cli version --cli-version
0.1.11
```

tgodzik:
Was `--cli-version` available in previous versions also?

lwronski:
> Was `--cli-version` available in previous versions also?

No, it will be available from `0.1.12`

tgodzik:
So it's probably better to just use `version` and parse it correctly. Otherwise, it would be harder to support older versions.

Issue #1093, opened 2022-06-08
Enable defining and running tests inside same Scala file


jvican:
**Is your feature request related to a problem? Please describe.**

Languages such as Rust enable you to define tests (usually unit tests) together with the code one is testing. This is useful and a productivity booster:
* It allows readers to see how an API is immediately used and what tests are present; and,
* It allows writers to avoid the tax to create an independent test file for everything you want to test and benefit from the locality of having the tests right next to the tested source.

All in all, the following `a.scala` source should be valid:

```scala
//> using scala "3.1"
//> using lib "org.scalameta::munit::0.7.27"

class NumAPI:
  def foo(): Int = 3

class NumAPISpec extends munit.FunSuite:
  test("check foo returns 3") {
    assertEquals(NumAPI().foo(), 3)
  }
```

**Describe the solution you'd like**

`scala-cli` could enable us to define and run tests in the same source code. This would involve making `scala-cli test a.scala` compile and run the tests defined in `a.scala`, regardless of whether `//> using target.scope "test"` has been defined or not.

**Additional context**

Here's a few design considerations to take into account just from the user and semantics POV (not discussing implementation details).

Should test code defined in `a.scala` be available/importable (when https://github.com/VirtusLab/scala-cli/issues/948 is implemented)? I don't think it should. Tests defined with the tested code should be treated as private. The benefit of this is that we can also drop the test dependencies from being depended on.

What should we do with test dependencies? The unit tests might need to import additional test dependencies aside from the test framework. We should have a way that these test dependencies are not leaked to the classpath of other modules depending on the source.

alexarchambault:
I like this idea.

Implementation-wise, I think we would likely compile everything together (main scope + tests), then split things in a second time. So we would have to:
- split test-specific classes from the rest (test suites, that we can detect like the `test` command does, would automatically go in the test scope, plus… maybe classes with a specific annotation, like `@scala.testOnly`, would be put there too)
- check that no class in the main scope references a class in the test scope
- split transitive dependencies (resulting in main ones and test ones), and maybe warn if some dependencies in the main scope were bumped by dependencies in test
- check that no class in the main scope references test-specific dependencies

Gedochao:
@antholeole if you're indeed interested in implementing this, let me know if you need any help with setting up.
I think having this would be cool, Markdown sources could also benefit from this.

antholeole:
[...]

PR #1268, opened 2022-08-22
Experimental support for Markdown

### LINKED ISSUE #1231: Initial Markdown support
The initial Markdown support PoC is on PR https://github.com/VirtusLab/scala-cli/pull/1216
We need to clean and improve this solution to bring it to the `main` branch.

Gedochao:
Fixes #1231 
This is a reworked version of https://github.com/VirtusLab/scala-cli/pull/1216, kudos to @BlackAnubis7 for the initial PoC

Issues to be addressed separately
- properly supporting using directives
  - currently only supported in the first `scala raw` and `scala test` snippets in a given `.md` file, given to the way they are being processed
  - not supported at all in plain `scala` snippets.
  - fixing those will require a further refactor of the code wrapping & preprocessing logic
  - separate scopes will probably have to be supported as separate Scala files after all
  - I initially thought we could maybe even limit the number of generated sources to just 1 (after https://github.com/VirtusLab/scala-cli/issues/1093 is addressed), but this will likely prevent that
- currently there's only support for on-disk sources (piped and remote Markdown inputs will be addressed separately)
- another refactor will be necessary to get rid of the empty lines hack in the generated code (currently, a lot of empty lines are added in the Scala sources generated from Markdown to keep the line numbers consistent)
- we currently only support `scala` snippets, but supporting `java` shouldn't be hard to add


modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+    *
+    * @param md
+    *   Markdown file in a `String` format
+    * @return
+    *   list of all found snippets
+    */
+  def findFences(md: String): Seq[MarkdownFence] = {
+    var startedFenceOpt: Option[StartedFence] = None
+    val fences                                = mutable.ListBuffer.empty[MarkdownFence]
+    val lines: Array[String]                  = md.split("\n\r?")
```
lwronski: It would be safer, if you use `Position.Raw.lineStartIndices(fileContent)` to split content by lines - [here](https://github.com/VirtusLab/scala-cli/blob/2cf64ad2f16f70c720fa878261381dd5c7dadd70/modules/cli/src/main/scala/scala/cli/commands/DependencyUpdate.scala#L102)

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  /** Finds all code snippets in given input
+    *
+    * @param md
+    *   Markdown file in a `String` format
+    * @return
+    *   list of all found snippets
+    */
+  def findFences(md: String): Seq[MarkdownFence] = {
+    var startedFenceOpt: Option[StartedFence] = None
```
lwronski: You can try to rewrite this function as recursive method, I suppose that it can help to avoid using `var`

Gedochao: right, missed this spot

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+    *   end line on which the code block was closed
+    */
+  case class MarkdownFence(
+    info: Seq[String],
+    body: String,
+    startLine: Int, // start of fenced body EXCLUDING backticks
+    endLine: Int    // same as above
+  ) {
+    override def toString: String =
+      s"Fence[$info, lines $startLine-$endLine]{${body.replace("\n", "\\n")}}"
```
lwronski: does it work on windows as well?

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+
+    /** @return
+      *   `true` if this snippet should be ignored, `false` otherwise
+      */
+    def shouldIgnore: Boolean = info.head != "scala" || info.contains("ignore")
```
lwronski: There is any validation of info input? 
What happened if user pass?
```
scala test-reset raw
```

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
```diff
@@ -0,0 +1,156 @@
...
+    val maybePkgString =
+      if pkg.isEmpty then None else Some(s"package ${AmmUtil.encodeScalaSourcePath(pkg)}")
+    val allSnippets                      = MarkdownCodeBlockUtil.findFences(content)
+    val (rawSnippets, processedSnippets) = allSnippets.partition(_.isRaw)
+    val (testSnippets, mainSnippets)     = processedSnippets.partition(_.isTest)
+    val wrapperName                      = s"${wrapper.raw}_md"
+    (
+      wrapScalaCode(mainSnippets, wrapperName, maybePkgString),
+      rawScalaCode(rawSnippets),
+      rawScalaCode(testSnippets)
```
lwronski, line 37: Maybe it should be wrapped in class.

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
```diff
@@ -0,0 +1,156 @@
...
+    )
+  }
+
+  /** Scope object name for a given index
```
lwronski, line 41: Can you explain more, how works index? Is this a scope ID that can be reset by `reset` keyword?

Gedochao: yep, it's the scope id

Gedochao: BTW we might have to get rid of the scopes completely when fixing the using directives issues in a later PR

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
Gedochao: What do you mean?

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
Gedochao: I ended up rewriting a big chunk of the code in this source file, this should be addressed already

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
Gedochao: Should be fixed now

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
Gedochao: Added some more scaladocs, should be clearer now

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
Gedochao: I don't think this override is even necessary anymore, removed it

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlockUtil.scala
Gedochao: Better error handling will be addressed in a separate PR

Gedochao: IMO this one is large enough already 🤡

modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeBlock.scala
```diff
@@ -0,0 +1,73 @@
...
+    *   list of all found snippets
+    */
+  def findCodeBlocks(md: String): Seq[MarkdownCodeBlock] = {
+    val lines = md
+      .lines()
+      .toList
+      .asScala
+    lines
+      .zipWithIndex
+      .foldLeft((0, Seq.empty[MarkdownCodeBlock], Option.empty[MarkdownOpenFence])) {
```
lwronski: It would be better, if you move logic from `foldLeft` to tail recursive function.

Gedochao: done

lwronski (approved):

PR #1270, opened 2022-08-22
Add --python

alexarchambault:


kiendang:
ScalaPy native for windows doesn't work right now but I'll work on it once scalapy/scalapy#301 is settled which would make testing a lot easier.

You could also try making it work here right now by adding `"-l" + nativeLibrary.get` to ldflags and appending `nativeLibraryPaths` to `PATH`.

The problem currently with `python-native-libs` currently is I want to add `nativeLibraryPaths` to ldflags too so that it works for setting up sbt projects since you can't really manipulate env vars in sbt scala native (right?), but on windows paths usually have spaces in them, *e.g.* `-LC:\Program Files\...` and I'm not sure how to properly escape them. Spaces should work fine with the `PATH` env var though so I suspect the approach above would work.

kiendang:
Oh I saw that you alr added `nativeLibraryPaths` to `PATH`, so maybe on windows adding `"-l" + nativeLibrary` to linking options instead of using `python.ldflags` for now would work.

kiendang:
So looks like that was not the issue at all. Adding support for ScalaPy native on Windows turned out to be more complicated than that. I put up a summary at scalapy/scalapy#307.

kiendang:
For native the `SCALAPY_PYTHON_PROGRAMNAME` env var should be set to `python.executable` too so that virtualenv works (see scalapy/scalapy#200). JVM doesn't need that coz it's alr handled via system properties.

alexarchambault:
Closing as most of the stuff initially pushed here has been merged. Will open a new PR for the remaining things.

PR #1278, opened 2022-08-23
Correct using directives on configuration.md 

megri:
I came across the `//> using`-style directive in this page and it didn't work. The syntax on https://scala-cli.virtuslab.org/docs/guides/using-directives/ seems to work so I updated them here.


Gedochao:
Hey, good catch, thanks for contributing!

Gedochao (approved):

PR #1282, opened 2022-08-24
Detect ARM64 macs when downloading coursier launcher

keynmol:
This has knock off effects on the JDK that will be downloaded, and subsequently ability to build native image.

With this change I was able to built native image using `./.github/scripts/generate-native-image.sh` and it seems to have the correct architecture.


lwronski (approved):
LGTM, @keynmol Thanks for contributing

PR #1295, opened 2022-08-25
Add --python to run command

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -360,30 +364,82 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
+              val python       = Python()
+              val pathsOrError = python.nativeLibraryPaths
+              logger.debug(s"Python native library paths: $pathsOrError")
+              pathsOrError
+                .toEither
+                .left.map(new Run.PythonDetectionError(_))
+            }
+          else
+            Nil
+        // seems conda doesn't add the lib directory to LD_LIBRARY_PATH, which prevents
```
Gedochao: NIT shouldn't we check if it isn't there before we prepend it, to avoid duplicates?
conda might fix it in the future.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -360,30 +364,82 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
+            val setupPython   = build.options.notForBloopOptions.doSetupPython.getOrElse(false)
+            val pythonJavaProps =
+              if (setupPython) {
+                val scalapyProps = value {
+                  val python       = Python()
+                  val propsOrError = python.scalapyProperties
+                  logger.debug(s"Python Java properties: $propsOrError")
+                  propsOrError
+                    .toEither
+                    .left.map(new PythonDetectionError(_))
```
Gedochao: NIT this shows up in a couple places, might want to add an extension like
```scala
  extension[T] (t: Try[T])
    def orPythonDetectionError: Either[PythonDetectionError, T] =
      t.toEither.left.map(new PythonDetectionError(_))
```

Gedochao (approved):
commented on some nitpicks, otherwise LGTM

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: It seemed unlikely to me that conda would fix it (I'm adding a mention to https://github.com/conda/conda/issues/308 in the PR, but I recall other issues where conda maintainers were more… assertive in their refusal to update LD_LIBRARY_PATH). But users might have updated it themselves.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: Should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
alexarchambault: Should be fixed.

PR #1297, opened 2022-08-26
Add mechanism to restrict options based on description and generate reference docs for SIP

romanowski:
This PR adds mechanism to restrict visibiltity of some of the options based on the fact if its description contains '[restricted]' or '[experimental]'.

I've marked some of the spark and hadoop options as experimental.

I've refactor some of sip-related mentods to be meaningfull.

Additional, I created a pice of reference docs that would be needed for SIP


lwronski (approved):

romanowski:
Adding `@tag` to case app does not seem to work I think we should stick to text markers for a short while for marking options as `experimental` or `restricted`

PR #1305, opened 2022-08-29
Default to the `repl` sub-command when no args are passed

### LINKED ISSUE #1249: Running Scala CLI without any arguments should start a REPL

This issue was created based on the [comments in Scala CLI SIP](https://github.com/scala/improvement-proposals/pull/46#discussion_r925787060). 

The current `scala` command starts repl with no arguments provided but with scala-cli `repl` subcommand is needed. I think we should change the behaviour to match what `scala` is doing now: running repl when no commands are provided.

When running `scala-cli` without any command, a reply should be started.

**Additional context**

We would need to find a way to display help for the whole `scala-cli' tool. With the default changed, user running `scala-cli --help` may expect either help for the whole `scala-cli` or the `repl` command. I can see 2 options:

- `scala-cli --help` prints the help for the whole tool with the mention that to see a help dedicated for `repl` command run `scala-cli repl --help`
- `scala-cli --help` prints the help for `repl` command with the mention that to see the whole help run e.g. `scala-cli --help --all`

I am personally prefer the former mechanism.

### LINKED ISSUE #1304: Passing an invalid option without an explicit sub-command skips suggesting `help`
**Version(s)**
0.1.12

**Describe the bug**
Passing an invalid option (i.e. erroneously passing an option belonging to another sub-command) to an explicitly chosen sub-command suggests looking at the relevant help.
```bash
▶ scala-cli run . --invalid-option
Unrecognized argument: --invalid-option

  scala-cli run --help
```
However, doing the same for the default command skips that.
```bash
▶ scala-cli . --invalid-option
Unrecognized argument: --invalid-option
```

**Expected behaviour**
I would expect the following:
```bash
▶ scala-cli . --invalid-option
Unrecognized argument: --invalid-option

  scala-cli --help
```

Gedochao:
Fixes #1304 
Fixes #1249


wleczny (commented):
LGTM

wleczny (approved):

PR #1313, opened 2022-08-31
Fix backwards compat for `-e` and `--scalac-help`

Gedochao:
As we now default to the `repl` sub-command when no args are passed, the `-e` option requires special handling.

Normally, adding a script snippet for the `repl` just adds it on the classpath.
```text
▶ scala-cli repl -e 'println("Hello")'
Welcome to Scala 3.1.3 (17.0.2, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                             
scala> snippet_sc.main(Array.empty)
Hello
                                                                                                                                                             
scala> 
```

However, the `scala` command's `-e` option's behaviour is closer to the `run` sub-command of Scala CLI.
And so, when `scala-cli` is being run without the `repl` sub-command passed explicitly, passing `-e` will now default to `run`.
```text
▶ scala-cli -e 'println("Hello")'
Hello
```
Using the `--script-snippet` alias instead of `-e` will still pass the snippet to the `repl`.
```text
▶ scala-cli --script-snippet 'println("Hello")'
Welcome to Scala 3.1.3 (17.0.2, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                             
scala> snippet_sc.main(Array.empty)
Hello
```

Also, this PR fixes `--scalac-help` (and other help options passed to the compiler through `--scalac-option`) and adds `--help-scalac` as an alias).

### DIFF
--- a/modules/cli-options/src/main/scala/scala/cli/commands/SharedOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/SharedOptions.scala
@@ -42,14 +42,15 @@ final case class SharedOptions(
   @Name("scalaBin")
   @Name("B")
     scalaBinaryVersion: Option[String] = None,
-  
+
   @Group("Scala")
   @HelpMessage("Show help for scalac. This is an alias for --scalac-option -help")
+  @Name("helpScalac")
     scalacHelp: Boolean = false,
 
   @Recurse
     snippet: SnippetOptions = SnippetOptions(),
-  
+
   @Recurse
     markdown: MarkdownOptions = MarkdownOptions(),
 
--- a/modules/cli-options/src/main/scala/scala/cli/commands/SnippetOptions.scala
+++ b/modules/cli-options/src/main/scala/scala/cli/commands/SnippetOptions.scala
@@ -6,21 +6,32 @@ import caseapp._
 final case class SnippetOptions(
   @Group("Scala")
   @HelpMessage("Allows to execute a passed string as a Scala script")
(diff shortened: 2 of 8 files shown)


mpkocher:
Isn't this creating a bunch of entropy by creating all of these special cases and duplication with `scala-cli run` and `scala-cli repl`? 

At a minimum this will create a bit of cognitive overhead and confusion for new users  when examples/docs/people are using different styles?

Gedochao:
@mpkocher Our aim is to maximize backwards compatibility with the old `scala` runner command (to fulfill the requirements of [SIP-46](https://github.com/scala/improvement-proposals/pull/46)) while at the same time minimizing changes to the current `scala-cli` behaviour.

`scala`'s `-e` option is a bit of a niche, barely documented feature. 
it's not even mentioned in `dotty`'s `scala` runner's help (the only mention I know of is the [3.1.2's release notes](https://www.scala-lang.org/blog/2022/04/12/scala-3.1.2-released.html))
So AFAIK it's only properly documented in [Scala 2's help](https://github.com/scala/scala/blob/af56abcbe03571c92897ebfaebdfe52cf8275304/src/compiler/scala/tools/nsc/GenericRunnerSettings.scala#L55)
Still, as it may have some people relying on it regardless, we want the transition seemless.

For `scala-cli`, `-e` is merely an alias for `--script-snippet`.
I understand your concerns, but IMO the confusion should be minimal.

lwronski (approved):

PR #1314, opened 2022-08-31
 Update scala3 to 3.2.0

lwronski:
- [Fix compilation error after dump scala to 3.2.0](https://github.com/VirtusLab/scala-cli/pull/1314/commits/f3ccb8d35c531a48fc1876351a8197ec24a6c2f6) - Fix compiler bug which is reported in https://github.com/lampepfl/dotty/issues/15940
- [Set explicitly version of Scala for ammonite](https://github.com/VirtusLab/scala-cli/pull/1314/commits/dbd98d8644b95ac03ef5de00a09499f710fe66a6) - `mill` throws the following error, because ammonite is not yet published for `3.2.0` - 
```
1 targets failed
dummy.amm[3.2.0].resolvedRunIvyDeps 
Resolution failed for 1 modules:
--------------------------------------------
  com.lihaoyi:ammonite_3.2.0:2.5.4-19-cd76521f 
        not found: /Users/lwronski/.ivy2/local/com.lihaoyi/ammonite_3.2.0/2.5.4-19-cd76521f/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/com/lihaoyi/ammonite_3.2.0/2.5.4-19-cd76521f/ammonite_3.2.0-2.5.4-19-cd76521f.pom

```
- [Remove options compile deps in cli-options module](https://github.com/VirtusLab/scala-cli/pull/1314/commits/5983709ac34df3ece2409c62cf7ec658d42365b0) - I have to remove compile deps to `options` because `cli-options` module is compiled using scala `2.13.8` which can not read TASTy files from module `options` which is built by `3.2.0`. It should work when `2.13.9` will be published with adding support for reading TASTy files, but IMO we should remove compile deps.


modules/tasty-lib/src/main/scala/scala/build/tastylib/TastyVersions.scala
```diff
@@ -1,15 +1,13 @@
...
+
 object TastyVersions {
 
   // Every time tasty version is updated, please update LatestSupportedScala as well!
-  final val MajorVersion: Int        = 28
-  final val MinorVersion: Int        = 1
-  final val ExperimentalVersion: Int = 0
-
   object LatestSupportedScala {
     final val MajorVersion: Int = 3
```
Gedochao, line 9: NIT it's not like Scala 4 is coming out anytime soon, but this could be derived from `latestSupportedScala` as well, IMO
```suggestion
    final val MajorVersion: Int = Constants.latestSupportedScala.split('.').head.toInt
```

modules/tasty-lib/src/main/scala/scala/build/tastylib/TastyVersions.scala
```diff
@@ -1,15 +1,13 @@
...
-    final val MinorVersion: Int = 1
+    final val MinorVersion: Int = Constants.latestSupportedScala.split('.')(1).toInt
```
Gedochao, line 10: NIT actually, I'm thinking you could even do both versions in one line.
```suggestion
    final val Array(MajorVersion: Int, MinorVersion: Int, _) = Constants.latestSupportedScala.split('.').map(_.toInt)
```

Gedochao (approved):
LGTM

PR #1323, opened 2022-09-05
Improve the error message for when a build's main class is ambiguous

### LINKED ISSUE #1315: give hint on what to do when [error]  Found several main classes
Write meaningful help text explaining how to use the `--main-class` option when
```
[error]  Found several main classes:
```
and also explain that you can
```
scala-cli config interactive true
```

Gedochao:
Fixes #1315 

Before:
```text
$ scala-cli run .
[error]  Found several main classes: ScalaMainClass2, ScalaMainClass1, scripts.ScalaScript_sc
```

After these changes:
```text
$ scala-cli run .
[error]  Found several main classes: ScalaMainClass1, ScalaMainClass2, scripts.ScalaScript_sc
You can run one of them by passing it with the --main-class option, e.g.
  scala-cli . --main-class ScalaMainClass1

You can pick the main class interactively by passing the --interactive option.
  scala-cli . --interactive
```

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -59,6 +59,7 @@ object Build {
     def foundMainClasses(): Seq[String]   = MainClass.find(output)
     def retainedMainClass(
       mainClasses: Seq[String],
+      commandString: String,
       logger: Logger
     ): Either[MainClassError, String] = {
       val defaultMainClassOpt = sources.defaultMainClass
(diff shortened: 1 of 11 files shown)


alexarchambault (approved):
Ok to merge, but just so you know, initially, for command suggestions such as these, my plan was to handle those like [this one](https://github.com/VirtusLab/scala-cli/blob/6b35e2c933d1afefd38eb8d7bf066bd9a0c9c440/modules/cli/src/main/scala/scala/cli/ScalaCli.scala#L122), that is catching specific exceptions in `ScalaCli`, and print suggestions for them.

That way, CLI-specific concerns don't leak in the `build` module. Originally, my intent was to maybe allow the build modules to be used as a library, where the CLI isn't around. So these suggestions don't make sense there. That said, I'm not sure the build modules are going to be used this way at some point, so it's no big deal...

PR #1327, opened 2022-09-07
Improve the error message for unsupported Scala version with Ammonite

Gedochao:
Before
```
▶ scala-cli --ammonite -S 3.2.0
Downloading Ammonite 2.5.4-15-f4a8969b
Downloading Ammonite 2.5.4-15-f4a8969b sources
[error]  Error downloading com.lihaoyi:ammonite_3.2.0:2.5.4-15-f4a8969b
  not found: /Users/pchabelski/.ivy2/local/com.lihaoyi/ammonite_3.2.0/2.5.4-15-f4a8969b/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/com/lihaoyi/ammonite_3.2.0/2.5.4-15-f4a8969b/ammonite_3.2.0-2.5.4-15-f4a8969b.pom
  not found: /Users/pchabelski/Library/Caches/ScalaCli/local-repo/v0.1.12/com.lihaoyi/ammonite_3.2.0/2.5.4-15-f4a8969b/ivys/ivy.xml
  No fallback URL found
```

After
```
▶ scala-cli --ammonite -S 3.2.0
Downloading Ammonite 2.5.4-19-cd76521f
Downloading Ammonite 2.5.4-19-cd76521f sources
[error]  Can't download Ammonite 2.5.4-19-cd76521f for Scala 3.2.0.
Ammonite with this Scala version might not yet be supported.
Try passing a different Scala version with the -S option.
```

### DIFF
--- a/modules/core/src/main/scala/scala/build/errors/CantDownloadAmmoniteError.scala
+++ b/modules/core/src/main/scala/scala/build/errors/CantDownloadAmmoniteError.scala
@@ -1,0 +1,28 @@
+package scala.build.errors
+
+import coursier.error.ResolutionError
+
+import scala.build.Position
+
+final class CantDownloadAmmoniteError(
+  ammoniteVersion: String,
+  scalaVersion: String,
+  underlying: ResolutionError.CantDownloadModule,
+  override val positions: Seq[Position]
+) extends BuildException(
+      s"""Can't download Ammonite $ammoniteVersion for Scala $scalaVersion.
+         |Ammonite with this Scala version might not yet be supported.
+         |Try passing a different Scala version with the -S option.""".stripMargin,
+      positions,
+      underlying
+    )
+
+object CantDownloadAmmoniteError {
+  def apply(
+    ammoniteVersion: String,
+    scalaVersion: String,
+    underlying: ResolutionError.CantDownloadModule,
+    positions: Seq[Position]
+  ): CantDownloadAmmoniteError =
+    new CantDownloadAmmoniteError(ammoniteVersion, scalaVersion, underlying, positions)
+}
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/scala/build/errors/CantDownloadAmmoniteError.scala
```diff
@@ -0,0 +1,18 @@
+package scala.build.errors
+
+import coursier.error.ResolutionError
+
+import scala.build.Position
+
+case class CantDownloadAmmoniteError(
```
alexarchambault: Exceptions shouldn't be case classes (because they hold a stack trace of the location where they've been instantiated, so that calling `copy()` on an instance returns an exception with a different stack trace - also, the `equals` method is wrong, making equal two exceptions with the same parameters but different stack traces)
```suggestion
final class CantDownloadAmmoniteError(
```

modules/core/src/main/scala/scala/build/errors/FetchingDependenciesError.scala
```diff
@@ -4,9 +4,9 @@ import coursier.error.CoursierError
 
 import scala.build.Position
 
-final class FetchingDependenciesError(
+final case class FetchingDependenciesError(
```
alexarchambault: Same as above, this shouldn't be a case class.
```suggestion
final class FetchingDependenciesError(
```

modules/core/src/main/scala/scala/build/errors/CantDownloadAmmoniteError.scala
Gedochao: fixed

modules/core/src/main/scala/scala/build/errors/FetchingDependenciesError.scala
Gedochao: fixed

alexarchambault (approved):

PR #1335, opened 2022-09-08
Disable mill-scala-cli for now

alexarchambault:
Let's see how it goes. See https://github.com/com-lihaoyi/mill/issues/2003 for more context.


alexarchambault:
We can totally remove it from the build in a second time, if it goes well (after we've all used that for a few days).

lwronski (approved):

PR #1336, opened 2022-09-08
Add --python to repl command

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -182,44 +211,58 @@ object Repl extends ScalaCommand[ReplOptions] {
     directories: scala.build.Directories,
     logger: Logger,
     allowExit: Boolean,
-    dryRun: Boolean
+    dryRun: Boolean,
+    buildOpt: Option[Build.Successful]
   ): Either[BuildException, Unit] = either {
 
+    val setupPython = options.notForBloopOptions.python.getOrElse(false)
+    val python      = options.notForBloopOptions.python.getOrElse(false)
```
Gedochao: `python` and `setupPython` are the same, no?

alexarchambault: Yep, should be fixed.

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -96,7 +101,8 @@ object Repl extends ScalaCommand[ReplOptions] {
       buildOptions: BuildOptions,
       artifacts: Artifacts,
       classDir: Option[os.Path],
-      allowExit: Boolean
+      allowExit: Boolean,
+      buildOpt: Option[Build.Successful]
```
Gedochao, line 107: NIT given that in a couple invocations, `buildOptions`, `artifacts` and `classDir` are taken from the `Build.Successful` anyway, I'm thinking another `def` for `doRunRepl` would simplify the code by a bit.

As in, we could add something like:
```scala
def doRunRepl(build: Build.Successful, allowExit: Boolean): Unit =
  doRunRepl(
            build.options,
            build.artifacts,
            build.outputOpt,
            allowExit,
            buildOpt = Some(build)
           ) 
```


```scala
doRunRepl(
            s.options,
            s.artifacts,
            s.outputOpt,
            allowExit = true,
            buildOpt = Some(s)
          )
```
could then be simplified down to
```scala
doRunRepl(s, allowExit = true)
```

Gedochao: We are also in a desparate need for scaladocs and/or a refactor of the naming of the functions here, given that we have `Repl.run`, `Repl.runRepl`, `Repl.maybeRunRepl` and `Repl.doRunRepl` and it takes more effort than it should to get the feel of what is going on in the code... But that's maybe for a future PR, just noting.

Gedochao (approved):
LGTM, commented on some stuff that could potentially be refactored (but can be done separately)

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
alexarchambault: Just added a `doRunRepl` helper method, accepting a build as input.

About the `Repl.*` methods, only two of them are top-level (`run` and `runRepl`), the others are helper internal methods of them (`doRunRepl` and `maybeRunRepl`). But yeah, maybe a few things could be clarified a little…

Issue #1310, opened 2022-08-30
Consider making Scala CLI backwards compatible with current scala commands


romanowski:
A list of incompatible options that makes a lot of sense:

- `-classpath` does not work even if prefixed with `-O` (how to pass agrs in such case?)
- `-d` conflicts with Scala CLI dependencies - do we really need to emit classes from REPL?
- `-color` - we have it on by default if I am not mistaken
- `-encoding`
- `-explain`, `-explain-types` (`-explaintypes` from Scala 2)
- `-feature`
- `-help` - conflicts with our help
- `-language` (`-language:<features>` from Scala 2)
- `-nowarn`
- `-print-lines`
- `-release` (`-target` from Scala 2)
- `-source`
- `-sourcepath` and `-sourceroot` (`-rootdir` from Scala 2) but I have mixed feeling about those
- `-unchecked`
- `@<file> ` - provide a file with arguments and options
- 

A list of options that I can see some usage but proabaly are not critical (so we can just fallback to require `-O` prefix)
- `-old-syntax`
- `-new-syntax`
- `-no-indent`
- `-from-tasty`
- `-javabootclasspath`, `-bootclasspath`  (`-nobootcp` and `-toolcp` from Scala 2)- not sure if we can support both
- `-scalajs`, `-scalajs-genStaticForwardersForNonTopLevelObjects`, `-scalajs-mapSourceURI` - I was under impression that REPL does not support Scala JS but even if so I think `--js` and related flags has superior UX
- `-uniqid`, `-verbose` - debugging REPL is instead for powerusers and I think we can expect them to prefix these options with `-O`
- `-dependencyfile` - not supported in Scala 3
- `-no-specialization`- not supported in Scala 3
- `-opt:<optimizations>`, `-opt-inline-from:<patterns>`, `-opt-warnings:<warnings>` - not supported in Scala 3
- `-print` - not supported in Scala 3
- `-usemanifestcp` - not supported in Scala 3

A list of options that I do not see a use case in REPL:
- `-indent` - do not print indent based version of the provided code also when running
- `-print-tasty` - do not print TASTY from provided code
- `-project-url`
- `-rewrite` - no effect also when running code from file
- `-semanticdb-target` - does not seem to have any effect

A list of options I have no idea what they do:
- `-extdirs`
- `-javaextdirs`

Gedochao:
relevant: https://github.com/VirtusLab/scala-cli/pull/1313

julienrf:
[...]

PR #1340, opened 2022-09-09
Make `-d` and `-classpath` options backwards compatible with the `scala` command

Gedochao:
Context: #1310

Summary of changes (most of them, breaking changes):
- the `compile` sub-command's `--output` option has been renamed to `--compilation-output`
  - the `--compilation-output` option is now also available from the `run` and `package` sub-commands
  - the rename is necessary to prevent a clash with the `--output` option of the `package` sub-command
- the `-d` option is no longer an alias for `--dependency`, but for `--compilation-output`
  - the aim of that change is to be backwards compatible with `scala` command's `-d`
- `-O -d -O path/to/compilation/output` now defaults to `-d path/to/compilation/output`
- the old `--classpath` option has been renamed to `--print-classpath`, to better reflect what it actually does
  - the rename is necessary to prevent the near-clash between `scala-cli`'s `--classpath` and `scala` command's `-classpath` options (which have radically different meanings, BTW)
-  `--classpath`, `--class-path` and `-classpath` options are now aliases for the `--extra jars` option
 - `-O -classpath -O path/to/classpath` now defaults to `--extra-jars path/to/classpath`

Reference docs have been generated for each of the commits, separately.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
@@ -1,10 +1,11 @@
 package scala.cli.commands.util
 
 import scala.build.errors.MainClassError
-import scala.build.{Build, Logger}
-import scala.cli.commands.ScalaCommand
+import scala.build.{Build, Logger, Os}
+import scala.cli.commands.{ScalaCommand, SharedOptions}
 
 trait BuildCommandHelpers { self: ScalaCommand[_] =>
+  import scala.cli.commands.util.ScalacOptionsUtil.*
   extension (successfulBuild: Build.Successful) {
     def retainedMainClass(
       logger: Logger,
@@ -15,5 +16,15 @@ trait BuildCommandHelpers { self: ScalaCommand[_] =>
         self.argvOpt.map(_.mkString(" ")).getOrElse(actualFullCommand),
         logger
       )
+
+    /** -O -d defaults to --compile-output; if both are defined, --compile-output takes precedence
+      */
+    def copyOutput(sharedOptions: SharedOptions): Unit =
+      sharedOptions.compilationOutput.filter(_.nonEmpty)
+        .orElse(sharedOptions.scalac.scalacOption.toScalacOptShadowingSeq.getScalacOption("-d"))
+        .filter(_.nonEmpty)
+        .map(os.Path(_, Os.pwd)).foreach(output =>
+          os.copy.over(successfulBuild.output, output, createFolders = true)
+        )
   }
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
@@ -7,13 +7,15 @@ object ScalacOptionsUtil {
   extension (opts: List[String]) {
     def toScalacOptShadowingSeq: ShadowingSeq[ScalacOpt] =
       ShadowingSeq.from(opts.filter(_.nonEmpty).map(ScalacOpt(_)))
+
   }
 
   extension (opts: ShadowingSeq[ScalacOpt]) {
-    def getScalacOption(key: String): Option[String] =
-      opts.get(ScalacOpt(key)).headOption.map(_.value)
-
     def filterScalacOptionKeys(f: String => Boolean): ShadowingSeq[ScalacOpt] =
       opts.filterKeys(_.key.exists(f))
+    def filterNonRedirected: ShadowingSeq[ScalacOpt] =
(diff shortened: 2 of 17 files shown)


modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
```diff
@@ -15,5 +17,13 @@ trait BuildCommandHelpers { self: ScalaCommand[_] =>
...
         logger
       )
+
+    /** -O -d defaults to --compile-output; if both are defined, --compile-output takes precedence
+      */
+    def copyOutput(sharedOptions: SharedOptions): Unit =
+      sharedOptions.compilationOutput.filter(_.nonEmpty)
+        .orElse(sharedOptions.scalac.getForKey("-d").headOption)
+        .filter(_.nonEmpty)
+        .map(os.Path(_, Os.pwd)).foreach(output => os.copy.over(successfulBuild.output, output))
```
lwronski: If the output path doesn't exist, will it create directories recursively?

modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
```diff
@@ -0,0 +1,29 @@
+package scala.cli.commands.util
+
+import scala.build.options.{ScalacOpt, ShadowingSeq}
+import scala.cli.commands.ScalacOptions
+
+object ScalacOptionsUtil {
```
lwronski, line 6: I think these features are already implemented in `ShadowingSeq`, expect the only one - `get` function.

IMO in `ShadowingSeq` you should add the `get` function.

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -169,7 +170,12 @@ object SharedOptionsUtil extends CommandHelpers {
           addScalaLibrary = scalaLibrary.orElse(java.map(!_)),
           generateSemanticDbs = semanticDb,
           scalacOptions = ShadowingSeq.from(
-            scalac.scalacOption
+            scalac
+              .scalacOptionKeyFilter(
+                // -O -classpath should be redirected as --extra-jars instead
+                // similarly for -O -d, it should be redirected as --compilation-output
+                key => key != "-classpath" && key != "-d"
```
lwronski: It seems that these functions should be omitted everywhere, maybe you should add this filter [here](https://github.com/VirtusLab/scala-cli/blob/2e0b4b4c6816dba6004999db2be36e31f57c89d1/modules/options/src/main/scala/scala/build/options/ScalacOpt.scala#L7).

modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
Gedochao: good catch. should be fixed now, now it will.

modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
Gedochao: done.

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
Gedochao: still, the values for these keys have to be extracted first, so that they are redirected to the appropriate Scala CLI options.
I added a function in `ScalacOptionsUtil` for filtering them out, let me know if it's better now.

modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
```diff
@@ -0,0 +1,21 @@
...
+  extension (opts: List[String]) {
+    def toShadowingSeq: ShadowingSeq[ScalacOpt] =
```
lwronski: This looks much better for me, I have one suggestion, we should name better this `extension`, because it can be used for any `List[String]`.
```suggestion
    def toScalacOptShadowingSeq: ShadowingSeq[ScalacOpt] =
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
lwronski: It's good enough for me, thanks.

lwronski (approved):

modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
Gedochao: fixed

PR #1347, opened 2022-09-12
Add platform option

### LINKED ISSUE #1214: `--jvm` flag
Sometimes I have a script with `//> using platform "js"` but I would like quickly run it on JVM to compare behavior (I work a lot on cross-compiling libraries to have identical cross-platform semantics).

There are already flags `--js` and `--native`, so it seems fitting to have `--jvm`.

I could not specify the platform in a directive, and always pass by command-line. But directives are very convenient :)

wleczny:
This adds `--platform` option, which can be used to specify, which platform (Scala.js | Scala Native | JVM) should be used to compile and run application.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
@@ -1,6 +1,8 @@
 package scala.cli.commands.util
 import coursier.core.Version
 
+import scala.build.EitherCps.{either, value}
+import scala.build.errors.BuildException
 import scala.build.internal.FetchExternalBinary
 import scala.build.options.BuildOptions
 import scala.cli.commands.FmtOptions
@@ -26,7 +28,7 @@ object FmtOptionsUtil {
       (url, !tag0.startsWith("v"))
     }
 
-    def buildOptions: BuildOptions = shared.buildOptions()
+    def buildOptions: Either[BuildException, BuildOptions] = shared.buildOptions()
 
     def scalafmtCliOptions: List[String] =
       scalafmtArg :::
--- a/modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
@@ -9,10 +9,11 @@ import dependency.parser.DependencyParser
 
 import java.io.{File, InputStream}
 
+import scala.build.EitherCps.{either, value}
 import scala.build.*
 import scala.build.blooprifle.BloopRifleConfig
 import scala.build.compiler.{BloopCompilerMaker, ScalaCompilerMaker, SimpleScalaCompilerMaker}
-import scala.build.errors.BuildException
+import scala.build.errors.{AmbiguousPlatformError, BuildException}
 import scala.build.interactive.Interactive
 import scala.build.interactive.Interactive.{InteractiveAsk, InteractiveNop}
 import scala.build.internal.CsLoggerUtil.*
@@ -155,11 +156,25 @@ object SharedOptionsUtil extends CommandHelpers {
       enableJmh: Boolean = false,
       jmhVersion: Option[String] = None,
       ignoreErrors: Boolean = false
-    ): bo.BuildOptions = {
-      val platformOpt =
-        if (js.js) Some(Platform.JS)
-        else if (native.native) Some(Platform.Native)
-        else None
+    ): Either[BuildException, bo.BuildOptions] = either {
+      val parsedPlatform = platform.map(Platform.normalize).flatMap(Platform.parse)
+      val platformOpt = value {
+        (parsedPlatform, js.js, native.native) match {
+          case (Some(p: Platform.JS.type), _, false)      => Right(Some(p))
+          case (Some(p: Platform.Native.type), false, _)  => Right(Some(p))
+          case (Some(p: Platform.JVM.type), false, false) => Right(Some(p))
+          case (Some(p), _, _) =>
+            val jsSeq        = if (js.js) Seq(Platform.JS) else Seq.empty
+            val nativeSeq    = if (native.native) Seq(Platform.Native) else Seq.empty
+            val platformsSeq = Seq(p) ++ jsSeq ++ nativeSeq
+            Left(new AmbiguousPlatformError(platformsSeq.distinct.map(_.toString)))
+          case (_, true, true) =>
+            Left(new AmbiguousPlatformError(Seq(Platform.JS.toString, Platform.Native.toString)))
+          case (_, true, _) => Right(Some(Platform.JS))
+          case (_, _, true) => Right(Some(Platform.Native))
+          case _            => Right(None)
+        }
+      }
       bo.BuildOptions(
         scalaOptions = bo.ScalaOptions(
           scalaVersion = scalaVersion
@@ -292,9 +307,8 @@ object SharedOptionsUtil extends CommandHelpers {
       javaCmd
     }
 
-    def bloopRifleConfig(): BloopRifleConfig = {
-
-      val options = buildOptions(false, None)
+    def bloopRifleConfig(): Either[BuildException, BloopRifleConfig] = either {
+      val options = value(buildOptions(false, None))
       lazy val defaultJvmCmd =
         downloadJvm(OsLibc.baseDefaultJvm(OsLibc.jvmIndexOs, "17"), options)
       val javaCmd = compilationServer.bloopJvm.map(downloadJvm(_, options)).orElse {
@@ -315,17 +329,21 @@ object SharedOptionsUtil extends CommandHelpers {
       )
     }
 
-    def compilerMaker(threads: BuildThreads, scaladoc: Boolean = false): ScalaCompilerMaker =
+    def compilerMaker(
+      threads: BuildThreads,
+      scaladoc: Boolean = false
+    ): Either[BuildException, ScalaCompilerMaker] = either {
       if (scaladoc)
         SimpleScalaCompilerMaker("java", Nil, scaladoc = true)
       else if (compilationServer.server.getOrElse(true))
         new BloopCompilerMaker(
-          bloopRifleConfig(),
+          value(bloopRifleConfig()),
           threads.bloop,
           strictBloopJsonCheckOrDefault
         )
       else
         SimpleScalaCompilerMaker("java", Nil)
+    }
 
     def coursierCache = cached(v)(coursier.coursierCache(logging.logger.coursierLogger("")))
 
--- a/modules/core/src/main/scala/scala/build/errors/AmbiguousPlatformError.scala
+++ b/modules/core/src/main/scala/scala/build/errors/AmbiguousPlatformError.scala
@@ -1,0 +1,5 @@
+package scala.build.errors
+
+final class AmbiguousPlatformError(passedPlatformTypes: Seq[String]) extends BuildException(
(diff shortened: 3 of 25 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -206,6 +206,23 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  test("simple script JS via platform option") {
+    val message = "Hello"
+    val inputs = TestInputs(
+      os.rel / "simple.sc" ->
+        s"""import scala.scalajs.js
```
lwronski: To be sure, that `--platform` flag overrides options from using directives, you can add this one directive to `simple.sc`.

```
//> using platform "jvm"
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -156,10 +156,16 @@ object SharedOptionsUtil extends CommandHelpers {
...
-        else None
+      val platformStr = Platform.normalize(platform.getOrElse("None"))
+      val platformOpt = Platform.parse(platformStr) match {
+        case Some(p) => Some(p)
+        case None =>
+          if (native.native && js.js) Some(Platform.Native)
+          else if (js.js) Some(Platform.JS)
+          else if (native.native) Some(Platform.Native)
+          else None
+      }
```
lwronski: To avoid executing the following expression `platform.getOrElse("None")`, you can do something like this:

```suggestion
      val platformStr = platform.map(Platform.normalize)
      val platformOpt = platformStr.map(Platform.parse) match {
        case Some(p) => Some(p)
        case None =>
          if (native.native && js.js) Some(Platform.Native)
          else if (js.js) Some(Platform.JS)
          else if (native.native) Some(Platform.Native)
          else None
      }
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
wleczny: Do you mean `"scala-native"`?

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -156,10 +156,16 @@ object SharedOptionsUtil extends CommandHelpers {
...
     ): bo.BuildOptions = {
-      val platformOpt =
-        if (js.js) Some(Platform.JS)
-        else if (native.native) Some(Platform.Native)
-        else None
+      val platformStr = platform.map(Platform.normalize)
+      val platformOpt = platformStr.map(Platform.parse) match {
+        case Some(p) => p
+        case None =>
+          if (native.native && js.js) Some(Platform.Native)
```
Gedochao: Instead of the complex `if (..) else if (..) else` clause, you could solve the whole thing with asingle `match` clause.

So something like:
```scala
val platformOpt = (platformStr.map(Platform.parse), native.native, js.js) match {
        case (Some(p), _, _) => p
        case (_, true, _) => Some(Platform.Native)
        case (_, _, true) => Some(Platform.JS)
        case _ => None
      }
```

Gedochao: also, I'm thinking if we shouldn't throw an error if conflicting platforms are picked, instead of guessing that one of them should have priority.

website/docs/commands/run.md
```diff
@@ -146,8 +152,27 @@ Note that the [Scala Native requirements](https://scala-native.readthedocs.io/en
...
+
+The `--platform` option can be used to choose the platform, which should be used to compile and run application. Available platforms are:
+* JVM (`jvm`)
+* Scala.js (`scala.js` | `scala-js` | `scalajs` | `js`)
+* Scala Native (`scala-native` | `scalanative` | `native`)
+
+Please note that `--platform` option is prioritized over `--js` and `--native` options. For example, the following command will run using Scala Native:
+
+```bash
+scala-cli Hello.scala --js --platform scala-native
```
Gedochao: I am thinking this command should actually throw an error, instead of prioritising `scala-native`
Similarly if `--js` and `--native` are passed together.

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
wleczny: I tried to raise a warning saying about platform ambiguity, but buildOptions function is triggered 3 times and the warning was raised every single time. I didn't have idea how to do it properly and that is why I gave up with it.

wleczny: But error is quite a good idea, since it will finish scala-cli on the first call of buildOptions.

Gedochao: yeah, I don't think a warning is sufficient here. 
IMO specifying conflicting platforms should not be treated as a valid command and fail instead.

modules/cli/src/main/scala/scala/cli/commands/util/FmtOptionsUtil.scala
```diff
@@ -26,7 +28,7 @@ object FmtOptionsUtil {
       (url, !tag0.startsWith("v"))
     }
 
-    def buildOptions: BuildOptions = shared.buildOptions()
+    def buildOptions: Either[BuildException, BuildOptions] = either(value(shared.buildOptions()))
```
Gedochao:
```suggestion
    def buildOptions: Either[BuildException, BuildOptions] = shared.buildOptions()
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
-    ): bo.BuildOptions = {
-      val platformOpt =
-        if (js.js) Some(Platform.JS)
-        else if (native.native) Some(Platform.Native)
-        else None
+    ): Either[BuildException, bo.BuildOptions] = either {
+      val parsedPlatform = platform.map(Platform.normalize).flatMap(Platform.parse)
+      val platformOpt = value {
+        (parsedPlatform, js.js, native.native) match {
+          case (Some(p: Platform.JS.type), true, false)     => Right(Some(p))
```
Gedochao: this and the fix below will let you get rid of `case (Some(p), _, _)`
```suggestion
          case (Some(p: Platform.JS.type), _, false)     => Right(Some(p))
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
+          case (Some(p: Platform.Native.type), false, true) => Right(Some(p))
```
Gedochao:
```suggestion
          case (Some(p: Platform.Native.type), false, _) => Right(Some(p))
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
+          case (Some(p), _, _) if native.native || js.js    => Left(new AmbiguousPlatformError)
+          case (Some(p), _, _)                              => Right(Some(p))
+          case (None, true, true)                           => Left(new AmbiguousPlatformError)
+          case (None, true, false)                          => Right(Some(Platform.JS))
```
Gedochao:
```suggestion
          case (_, true, _)                          => Right(Some(Platform.JS))
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
+          case (None, false, true)                          => Right(Some(Platform.Native))
```
Gedochao:
```suggestion
          case (_, _, true)                          => Right(Some(Platform.Native))
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
+          case (None, false, false)                         => Right(None)
```
Gedochao:
```suggestion
          case _                        => Right(None)
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
-        else None
+    ): Either[BuildException, bo.BuildOptions] = either {
+      val parsedPlatform = platform.map(Platform.normalize).flatMap(Platform.parse)
+      val platformOpt = value {
+        (parsedPlatform, js.js, native.native) match {
+          case (Some(p: Platform.JS.type), true, false)     => Right(Some(p))
+          case (Some(p: Platform.Native.type), false, true) => Right(Some(p))
+          case (Some(p), _, _) if native.native || js.js    => Left(new AmbiguousPlatformError)
+          case (Some(p), _, _)                              => Right(Some(p))
+          case (None, true, true)                           => Left(new AmbiguousPlatformError)
```
Gedochao:
```suggestion
          case (_, true, true)                           => Left(new AmbiguousPlatformError)
```

modules/cli/src/main/scala/scala/cli/commands/util/SharedOptionsUtil.scala
```diff
@@ -155,11 +156,20 @@ object SharedOptionsUtil extends CommandHelpers {
...
-        else if (native.native) Some(Platform.Native)
-        else None
+    ): Either[BuildException, bo.BuildOptions] = either {
+      val parsedPlatform = platform.map(Platform.normalize).flatMap(Platform.parse)
+      val platformOpt = value {
+        (parsedPlatform, js.js, native.native) match {
+          case (Some(p: Platform.JS.type), true, false)     => Right(Some(p))
+          case (Some(p: Platform.Native.type), false, true) => Right(Some(p))
+          case (Some(p), _, _) if native.native || js.js    => Left(new AmbiguousPlatformError)
+          case (Some(p), _, _)                              => Right(Some(p))
```
Gedochao: this case should be skippable with the other fixes
```suggestion
```

modules/core/src/main/scala/scala/build/errors/AmbiguousPlatformError.scala
```diff
@@ -0,0 +1,5 @@
+package scala.build.errors
+
+final class AmbiguousPlatformError extends BuildException(
+      "Ambiguous platform: more than one type of platform has been passed."
```
Gedochao: optimally, we could check what particular types of platforms were passed and then mention them in the error, so that the user knows what to look for in his implementation.
```suggestion
      s"Ambiguous platform: more than one type of platform has been passed: ${passedPlatformTypes.mkString(", "}."
```

Gedochao (approved):

lwronski (approved):

PR #1360, opened 2022-09-13
Remove mill-scalafix customization

alexarchambault:
Removal of the mill-scalafix customization is made possible thanks to the better sourceroot setting, changed in the first commit. (The customization was basically about setting the scalafix sourceroot setting.)


.gitignore
```diff
@@ -6,5 +6,3 @@ out/
 
 .bsp
 .scala-build
-
-*.semanticdb
```
Gedochao, line 10: shouldn't we keep this line?

alexarchambault: I'm checking that again locally, but I think those changes make that unnecessary anymore.

alexarchambault: I confirm that. From clean sources, compiling everything and importing / compiling the project in / from Metals doesn't create spurious `*.semanticdb` files alongside sources.

lwronski (approved):

Gedochao (approved):

PR #1367, opened 2022-09-14
Split config db stuff to a separate config module

alexarchambault:
Aimed at being re-used from coursier in particular, so that Scala CLI and coursier can shared their config.


modules/config/src/main/scala/scala/cli/config/ConfigDb.scala
```diff
@@ -44,19 +41,6 @@ final class ConfigDb private (
           .map(Some(_))
     }
 
-  /** Gets an entry.
-    *
-    * If the value cannot be decoded or the key isn't in the DB, None is returned.
-    *
-    * Otherwise, the value is returned wrapped in Some.
-    */
-  def getOrNone[T](key: Key[T], logger: Logger): Option[T] = get[T](key) match {
```
Gedochao, line 53: why remove this? you end up having to copy duplicates of this match clause everywhere

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
```diff
@@ -77,7 +78,8 @@ object Compile extends ScalaCommand[CompileOptions] with BuildCommandHelpers {
     val threads      = BuildThreads.create()
 
     val compilerMaker = options.shared.compilerMaker(threads).orExit(logger)
-    val configDb = ConfigDb.open(options.shared.directories.directories)
+    val configDb = ConfigDb.open(options.shared.directories.directories.dbPath.toNIO)
```
Gedochao: i`ConfigDb.open(options.shared.directories.directories.dbPath.toNIO)` could be extracted to `SharedOptionsUtil` to avoid repeating it in every command relying on the `configDb`

Gedochao: (or `ConfigUtil`)

modules/config/src/main/scala/scala/cli/config/ConfigDb.scala
alexarchambault: It's removed because it references `Logger`, which isn't around anymore once the `config` module has (almost) no dependencies.

alexarchambault: It's only called twice anyway IIRC.

modules/cli/src/main/scala/scala/cli/commands/Compile.scala
alexarchambault: Good point, should be fixed.

Gedochao (approved):

PR #1369, opened 2022-09-15
Make inputs optional when `-classpath` and `--main-class` are passed

Gedochao:
## Context

The `scala` command allows to do the following:
```
▶ ls
Main.scala
▶ mkdir out
▶ scalac Main.scala -d out
▶ scala Main -classpath out
Hello
```

Meanwhile, Scala CLI:

  - passing the `-classpath` option out of the blue will just start the repl with the relevant classpath. this still allows to run whatever you want in the repl. We are keeping this behaviour.
```
▶ scala-cli -classpath out
Welcome to Scala 3.2.0 (17.0.2, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                         
scala> Main.main(Array.empty)
Hello
```

  - it becomes more confusing if you try to pass the main class with the `--main-class` option, similarly to how the `scala` command allows to pass it as an input.
```
▶ scala-cli --main-class Main -classpath out
Unrecognized argument: --main-class

  scala-cli --help
```
This fails, as `--main-class` isn't an option for the repl, and we are defaulting to the repl.
Still, if we pass the `run` sub-command explicitly, it fails as well.

```
▶ scala-cli run --main-class Main -classpath out
[error]  No inputs provided (expected files with .scala, .sc, .java or .md extensions, and / or directories).
```
Even though the main class is specified and it is indeed passed on the classpath, `scala-cli` fails because it expects inputs.

# After the changes
TL;DR the following syntax becomes available:
  -  if the `run` sub-command is passed explicitly, it's sufficient to have a main class on the classpath, inputs aren't necessary then 
```
▶ scala-cli run -classpath out          
Hello
```
  - when using the default command, `--main-class` has to be passed explicitly (even if there's only one main class on the classpath) to indicate the intent of running it instead of defaulting to the repl
```
▶ scala-cli --main-class Main -classpath out
Hello
```
  - just passing the `-classpath` still defaults to the `repl`
```
▶ scala-cli -classpath out                  
Welcome to Scala 3.1.3 (17.0.2, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                         
scala> Main.main(Array.empty)
Hello
```

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -57,7 +57,8 @@ object Build {
     def outputOpt: Some[os.Path]          = Some(output)
     def dependencyClassPath: Seq[os.Path] = sources.resourceDirs ++ artifacts.classPath
     def fullClassPath: Seq[os.Path]       = Seq(output) ++ dependencyClassPath
-    def foundMainClasses(): Seq[String]   = MainClass.find(output)
+    def foundMainClasses(): Seq[String] =
+      MainClass.find(output) ++ options.classPathOptions.extraClassPath.flatMap(MainClass.find)
     def retainedMainClass(
       mainClasses: Seq[String],
       commandString: String,
--- a/modules/build/src/main/scala/scala/build/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/Inputs.scala
@@ -302,10 +302,11 @@ object Inputs {
     directories: Directories,
     forcedWorkspace: Option[os.Path],
     enableMarkdown: Boolean,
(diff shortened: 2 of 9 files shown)


mpkocher:
It looks like the root behavior has/is become quite complex.  How are all of these different cases communicated in `--help`?

Gedochao:
> It looks like the root behavior has/is become quite complex. How are all of these different cases communicated in `--help`?

@mpkocher they aren't just yet, but we are planning some docs and `-help` improvements tied to the `SIP` changes down the line.

EDIT: I created a catch-all ticket to track those - #1370

alexarchambault (approved):

PR #1377, opened 2022-09-20
Fix calculation of Scala version and turn off the `-release` flag for 2.12.x < 2.12.5

### LINKED ISSUE #1372: Incorrect version of scala repl started when 2.12.1 is requested
**Version(s)**
```
$ scala-cli -version
Scala CLI version: 0.1.14
Scala version (default): 3.2.0
```

**Describe the bug**
When I request a scala repl for v2.12.1 then v2.12.16 gets started instead.
All other versions I managed to check (v2.12.0, v.2.12.2, v2.12.3, v2.13.0, v2.13.1) gets started correctly.

**To Reproduce**
```
$ scala-cli repl -S 2.12.1
Welcome to Scala 2.12.16 (OpenJDK 64-Bit Server VM, Java 18.0.2.1).
```
or
```
$ scala-cli repl -j 8 -S 2.12.1
Welcome to Scala 2.12.16 (OpenJDK 64-Bit Server VM, Java 1.8.0_345).
```
I.e. it does not depend on JVM version either.

**Expected behaviour**
The requested version of scala repl should be started (v2.12.1 in this case).

### LINKED ISSUE #1373: Scala code won't compile on Scala 2.12.x < 2.12.5 because of the `-release` compiler option
**Version(s)**
0.1.14

**To Reproduce**
```bash
▶ scala-cli -S 2.12.2 -e 'println("Hello")'                            
Compiling project (Scala 2.12.2, JVM)
Error: bad option: '-release'
Error compiling project (Scala 2.12.2, JVM)
Compilation failed
```

Gedochao:
Fixes #1372 
Fixes #1373 

However, code still doesn't compile on Scala `2.12.0` and compiles infinitely on `2.12.4`, which should be addressed separately.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -18,13 +18,14 @@ import scala.build.compiler.{ScalaCompiler, ScalaCompilerMaker}
 import scala.build.errors.*
 import scala.build.internal.resource.ResourceMapper
 import scala.build.internal.{Constants, CustomCodeWrapper, MainClass, Util}
+import scala.build.options.ScalaVersionUtil.maybeScalaPatchVersion
 import scala.build.options.*
 import scala.build.options.validation.ValidationException
 import scala.build.postprocessing.*
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.duration.DurationInt
-import scala.util.Properties
 import scala.util.control.NonFatal
+import scala.util.{Properties, Try}
 
 trait Build {
   def inputs: Inputs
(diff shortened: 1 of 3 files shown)


modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -825,9 +825,19 @@ object Build {
             Seq(ScalacOpt("-scalajs"))
           else Nil
 
-        val scalacReleaseV = releaseFlagVersion
-          .map(v => List("-release", v).map(ScalacOpt(_)))
-          .getOrElse(Nil)
+        val scalacReleaseV = {
+          val maybeScalaPatchVersion =
```
lwronski: IMO we should move this function for extracting patch version to `ScalaParameters`, it can be useful in another place.

Gedochao: `ScalaParameters` is a part of [coursier/dependency](https://github.com/coursier/dependency)
I extracted the logic to `scala.build.options.ScalaVersionUtil` instead.

lwronski (approved):

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -825,9 +826,15 @@ object Build {
             Seq(ScalacOpt("-scalajs"))
           else Nil
 
-        val scalacReleaseV = releaseFlagVersion
-          .map(v => List("-release", v).map(ScalacOpt(_)))
-          .getOrElse(Nil)
+        val scalacReleaseV =
+          // the -release flag is not supported for Scala 2.12.x < 2.12.5
+          if params.scalaVersion.startsWith("2.12") &&
+            params.scalaVersion.maybeScalaPatchVersion.exists(_ < 5)
```
alexarchambault, line 832: No big deal, but version comparison with `coursier.core.Version` could have worked here, it seems, like
```scala
val sv = coursier.core.Version(params.scalaVersion)
if params.scalaVersion.startsWith("2.12.") && sv.compareTo(coursier.core.Version("2.12.15")) < 0
```

PR #1378, opened 2022-09-20
Add --debug option for integration tests

wleczny:
This PR adds `--debug` option for integration tests, which should improve DX for Scala CLI contributors.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -855,11 +855,24 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
       def test(args: String*) = {
         val argsTask = T.task {
           val launcher = launcherTask().path
+          val debugReg = "^--debug$|^--debug:([0-9]+)$".r
+          val debugPortOpt = args.find(debugReg.matches).flatMap {
+            case debugReg(port) => Option(port).orElse(Some("5005"))
+            case _              => None
+          }
+          val debugArgs = debugPortOpt match {
+            case Some(port) =>
+              System.err.println(
+                s"--debug option has been passed. Listening for transport dt_socket at address: $port"
+              )
+              Seq(s"-Dtest.scala-cli.debug.port=$port")
+            case _ => Seq.empty
+          }
           val extraArgs = Seq(
             s"-Dtest.scala-cli.path=$launcher",
             s"-Dtest.scala-cli.kind=$cliKind"
           )
-          args ++ extraArgs
+          args ++ extraArgs ++ debugArgs
         }
         T.command {
           val res            = testTask(argsTask, T.task(Seq.empty[String]))()
--- a/modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
@@ -12,18 +12,29 @@ import scala.util.Properties
 
 object TestUtil {
 
-  val cliKind: String      = sys.props("test.scala-cli.kind")
-  val isNativeCli: Boolean = cliKind.startsWith("native")
-  val isCI: Boolean        = System.getenv("CI") != null
-  val cliPath: String      = sys.props("test.scala-cli.path")
-  val detectCliPath        = if (TestUtil.isNativeCli) TestUtil.cliPath else "scala-cli"
-  val cli: Seq[String]     = cliCommand(cliPath)
+  val cliKind: String              = sys.props("test.scala-cli.kind")
+  val isNativeCli: Boolean         = cliKind.startsWith("native")
+  val isCI: Boolean                = System.getenv("CI") != null
+  val cliPath: String              = sys.props("test.scala-cli.path")
+  val debugPortOpt: Option[String] = sys.props.get("test.scala-cli.debug.port")
+  val detectCliPath                = if (TestUtil.isNativeCli) TestUtil.cliPath else "scala-cli"
+  val cli: Seq[String]             = cliCommand(cliPath)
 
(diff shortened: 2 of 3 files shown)


wleczny:
Unfortunately tests, which assert equality between expected output and output of tested command might fail, because Java remote debugging adds for example `Listening for transport dt_socket at address: 5005`. It is possible to filter out unwanted content, but it would be necessary to do it for a lot of tests and I'm worried that it might be more confusing than helpful.

wleczny:
Unfortunately tests, which assert equality between expected output and output of tested command might fail, because Java remote debugging adds for example `Listening for transport dt_socket at address: 5005`. It is possible to filter out unwanted content, but I'm worried that it might be more confusing than helpful when writing tests in the future.

build.sc
```diff
@@ -844,9 +844,23 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
       def test(args: String*) = {
         val argsTask = T.task {
           val launcher = launcherTask().path
+          val debugReg = "--debug(:[0-9]+)?".r
+          val debugOpt = args.filter(a => debugReg.pattern.matcher(a).matches).headOption
+          val (isDebug, debugPort) = debugOpt match {
+            case Some(opt: String) => opt.split(":", 2) match {
```
alexarchambault: NIT Better not type things in match cases, unless it's a type check (that is, the object might be something of an other instance sometimes, and we check that it has the right type at runtime). If ever the type of `debugOpt` unexpectedly becomes `Option[Any]`, we may not detect it, as the pattern makes it look as if it contains a string (while it may not anymore).

If you really want to write some explicit types here, these can be added to some of the `val`s (like `val debugOpt`, but not `val (isDebug, debugPort)`, as it's a pattern too).

build.sc
```diff
@@ -844,9 +844,23 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
       def test(args: String*) = {
         val argsTask = T.task {
           val launcher = launcherTask().path
+          val debugReg = "--debug(:[0-9]+)?".r
+          val debugOpt = args.filter(a => debugReg.pattern.matcher(a).matches).headOption
+          val (isDebug, debugPort) = debugOpt match {
```
alexarchambault: It seems we could have only one variable here: `debugPortOpt` say (typed as `Option[String]` I guess).

When it's `None`, debugging is disabled, when it contains a value, it's the debug port. And we only set `test.scala-cli.debug.port` for debugging (no need of `test.scala-cli.debug.is-debug`), when `debugPortOpt` is non empty.

build.sc
```diff
@@ -844,9 +844,23 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
...
+            case Some(opt: String) => opt.split(":", 2) match {
+                case Array(_, p) => (true, p)
+                case _           => (true, "5005")
+              }
+            case _ => (false, "")
+          }
+          if (isDebug) System.err.println(
+            "--debug option has been passed. Listening for transport dt_socket at address: 5005"
+          )
```
alexarchambault: With `debugPortOpt` suggested above, this could be something like
```scala
for (port <- debugPortOpt)
  System.err.println(s"… $port")
```

modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
```diff
@@ -15,15 +15,24 @@ object TestUtil {
   val cliKind: String      = sys.props("test.scala-cli.kind")
   val isNativeCli: Boolean = cliKind.startsWith("native")
   val isCI: Boolean        = System.getenv("CI") != null
+  val isDebug: Boolean     = sys.props("test.scala-cli.debug.is-debug") == "true"
+  val debugPort: String    = sys.props("test.scala-cli.debug.port")
```
alexarchambault: Like above, something like `debugPortOpt: Option[String]` should be enough, no need of two variables.

build.sc
```diff
@@ -844,9 +844,23 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
       def test(args: String*) = {
         val argsTask = T.task {
           val launcher = launcherTask().path
+          val debugReg = "--debug(:[0-9]+)?".r
```
Gedochao: NIT, the regex could be more precise.
```suggestion
          val debugReg = "^--debug$|^--debug:([0-9]+)$".r
```
how it works in the repl:
```
scala> val regex = "^--debug$|^--debug:([0-9]+)$".r
val regex: scala.util.matching.Regex = ^--debug$|^--debug:([0-9]+)$                                                                                                                                                                      
scala> "--debug:5005" match { case regex(a) => println(a)}
5005                                                                                                                                                                      
scala> "--debug" match { case regex(a) => println(a)}
null
```

build.sc
```diff
@@ -844,9 +844,23 @@ trait CliIntegration extends SbtModule with ScalaCliPublishModule with HasTests
...
+          val debugOpt = args.filter(a => debugReg.pattern.matcher(a).matches).headOption
```
Gedochao: `filter(...).headOption` is the same as `find(...)`
also, you can figure out the port in a single chain.
```suggestion
          val maybeDebugPort = args.find(regex.matches).flatMap { case regex(port) => Option(port).orElse(Some("5005")); case _ => None }
```
BTW there's no need for the `isDebug` boolean value, after the line above you need only check `maybeDebugPort.isDefined` (since we already fallback to `5005`)

Gedochao (approved):

alexarchambault (approved):

PR #1380, opened 2022-09-20
Fix finding main classes in external jars

Gedochao:
This fixes finding main classes in external jars, enabling the following syntax:
```
$ scala-cli run -cp Main.jar --list-main-classes
Main
$ scala-cli run -cp Main.jar
Hello
$ scala-cli --main-class Main -cp Main.jar
Hello
$ ls jarDir
Main.jar
$ scala-cli run -cp jarDir
Hello
$ scala-cli --main-class Main -cp jarDir
Hello
```

BTW do note that this enables to pass `.jar` files in a directory, which formerly wasn't being handled.
Also, `-cp` was added as an alias to `-classpath`, as part of the `SIP` backwards compat.

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/MainClass.scala
+++ b/modules/build/src/main/scala/scala/build/internal/MainClass.scala
@@ -3,6 +3,12 @@ package scala.build.internal
 import org.objectweb.asm
 import org.objectweb.asm.ClassReader
 
+import java.io.{ByteArrayInputStream, InputStream}
+import java.util.zip.ZipEntry
+
+import scala.build.Inputs.{Element, resolve}
+import scala.build.internal.zip.WrappedZipInputStream
+
 object MainClass {
 
   private def stringArrayDescriptor = "([Ljava/lang/String;)V"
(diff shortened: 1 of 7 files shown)


modules/build/src/main/scala/scala/build/internal/MainClass.scala
```diff
@@ -37,26 +43,47 @@ object MainClass {
...
       reader.accept(checker, 0)
       checker.mainClassOpt.iterator
     }
     finally is.close()
+
+  def findInJar(path: os.Path): Iterator[String] = {
+    val content        = os.read.bytes(path)
+    val jarInputStream = WrappedZipInputStream.create(new ByteArrayInputStream(content))
+    val classesInputStreams: Seq[InputStream] =
+      jarInputStream.entries().foldLeft(List.empty[InputStream]) {
```
alexarchambault: NIT Why do we need a `foldLeft` here? Wouldn't a `flatMap` do?

alexarchambault: Plus we could keep it as an `Iterator` right from the start it seems.

alexarchambault (approved):

modules/build/src/main/scala/scala/build/internal/MainClass.scala
Gedochao: ah, good catch.
should be fixed now.

alexarchambault (approved):

PR #1387, opened 2022-09-22
Make config module compatible with Java 8

alexarchambault:


Gedochao (approved):

lwronski:
What about the `scala-steward`? will ignore the dependency `jsoniterCoreJava8` and `jsoniterMacrosJava8`?

alexarchambault:
> What about the scala-steward? will ignore the dependency jsoniterCoreJava8 and jsoniterMacrosJava8?

I'm not sure what Scala Steward will do here...

PR #1389, opened 2022-09-22
Enable debugging for run and test commands

### LINKED ISSUE #1212: Options to make remote debugging easier

Providing options for remote debugging (that is really powerful) is not always intuitive or even may not be possible (if you want to debug a compiler). 

I would like that Scala CLI expose `--debug` and `--debug-compiler` that accepts a port to debug (or use 8080 or some other default one if no argument is provided). It should also accept a flag (e.g. `8080:listen`) to indicate if listen/attached mode should be used.

Providing full debug option (e.g. `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=8000`) as jvm arguments.

wleczny:


modules/cli-options/src/main/scala/scala/cli/commands/SharedDebugOptions.scala
```diff
@@ -0,0 +1,28 @@
...
+final case class SharedDebugOptions(
+  @Group("Debug")
+  @HelpMessage("Turn debugging on")
+    debug: Boolean = false,
+  @Group("Debug")
+  @HelpMessage("Debug port (5005 by default)")
+    debugPort: Option[String] = None,
+  @Group("Debug")
+  @HelpMessage("Debug mode (attach by default)")
+  @ValueDescription("attach|listen")
```
lwronski: It should be added here, that there is available another shorter value, such like this:
```
att|a|lis|l
```
but IMO, we can drop alias `att` for attach mode and `lis`.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -103,11 +103,36 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
+    opts: RunOptions,
     inputArgs: Seq[String],
     programArgs: Seq[String],
     defaultInputs: () => Option[Inputs]
   ): Unit = {
+    val options =
+      if (!opts.sharedRun.sharedDebug.debug)
+        opts
+      else
+        val server = opts.sharedRun.sharedDebug.debugMode match {
```
lwronski: If you move debug options to `SharedJvmOptions`, and then resolve debug java opts in `JvmUtils`. Support `--debug` flag for `test` command is for free.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -103,11 +103,36 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
+            )
+            "y"
+        }
+        val port = opts.sharedRun.sharedDebug.debugPort.getOrElse("5005")
+        opts.copy(
+          sharedRun = opts.sharedRun.copy(
+            sharedJava = opts.sharedRun.sharedJava.copy(
+              javaOpt = opts.sharedRun.sharedJava.javaOpt ++ List(
+                "-Xmx512m",
+                "-Xms128m",
```
lwronski: Scala CLI shouldn't apply this java options by default. I think, we can skip it.

modules/cli/src/main/scala/scala/cli/commands/Run.scala
```diff
@@ -103,11 +103,36 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
 
   def run(
-    options: RunOptions,
+    opts: RunOptions,
     inputArgs: Seq[String],
     programArgs: Seq[String],
     defaultInputs: () => Option[Inputs]
   ): Unit = {
+    val options =
+      if (!opts.sharedRun.sharedDebug.debug)
```
Gedochao: IMO we can make it slightly smarter.
if the user passes `--debug-port 5005` or `--debug-mode listen`, but he didn't pass the `--debug` flag, we probably can assume he still intended to run in debug mode.

so I think the following syntax variants should all be accepted (and any permutations thereof):
```
scala-cli run . --debug
scala-cli run . --debug-port 5005
scala-cli run . --debug-mode listen
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -520,6 +520,39 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+        TestUtil.cli,
+        "run",
+        extraOptions,
+        ".",
+        "--debug-port",
+        "5006",
+        "--debug-mode",
+        "listen",
+        "--command"
+      ).call(cwd = root).out.trim().split("\n").toList
```
Gedochao: the following should fix this test for Windows on the CI (splitting by `\n` won't cut it there)
```suggestion
      ).call(cwd = root).out.trim.lines.toList.asScala
```

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -520,6 +520,39 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+
+      def debugString(server: String, port: String) =
+        s"-agentlib:jdwp=transport=dt_socket,server=$server,suspend=y,address=$port"
+
+      assert(out1.filter(_ == debugString("y", "5005")).nonEmpty)
```
Gedochao:
```suggestion
      assert(out1.exists(_ == debugString("y", "5005")))
```

Gedochao (approved):

PR #1396, opened 2022-09-27
Generate native launcher of scala-cli for M1

lwronski:
I disable only one test for m1 - `building pkg package`,  it throws the following error:
```
installer: The install failed. (The Installer encountered an error that caused the installation to fail. Contact the software manufacturer for assistance. An error occurred while running scripts from the package “helloworldscalacli.pkg”.)
```
I will investigate it later.

### DIFF
--- a/modules/core/src/main/scala/scala/build/Os.scala
+++ b/modules/core/src/main/scala/scala/build/Os.scala
@@ -1,5 +1,7 @@
 package scala.build
 
+import java.util.Locale
+
 import scala.util.Properties
 
 object Os {
@@ -8,4 +10,6 @@ object Os {
       os.Path(os.pwd.toIO.getCanonicalFile)
     else
       os.pwd
+  lazy val isArmArchitecture: Boolean =
+    sys.props.getOrElse("os.arch", "").toLowerCase(Locale.ROOT) == "aarch64"
 }
--- a/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
+++ b/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
@@ -5,6 +5,7 @@ import coursier.jvm.{JavaHome, JvmIndex}
 import java.io.IOException
 import java.nio.charset.Charset
 
+import scala.build.Os
 import scala.build.blooprifle.VersionUtil.parseJavaVersion
 import scala.util.{Properties, Try}
 
@@ -68,7 +69,8 @@ object OsLibc {
       .forall(_ >= 17)
     if (os == "linux-musl") s"liberica:$jvmVersion" // zulu could work too
     else if (java17OrHigher) s"temurin:$jvmVersion"
-    else s"adopt:$jvmVersion"
+    else if (Os.isArmArchitecture) s"zulu:$jvmVersion" // adopt doesn't support Java 8 on macOS arm
+    else s"temurin:$jvmVersion"
   }
 
   def defaultJvm(os: String): String = {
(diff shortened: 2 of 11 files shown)


modules/core/src/main/scala/scala/build/Os.scala
```diff
@@ -8,4 +8,6 @@ object Os {
       os.Path(os.pwd.toIO.getCanonicalFile)
     else
       os.pwd
+  def isArmArchitecture: Boolean =
+    os.proc("uname", "-p").call(cwd = os.pwd).out.trim() == "arm"
```
alexarchambault: Wouldn't it work by checking some Java property (like `"os.arch"`)? It seems this is going to make even the x86_64 version of Scala CLI run or load M1 binaries, which might not work when loading libraries, but also may not be what users expect also (the x86_64 version should run and load x86_64 too under the hood, I'd say).

project/deps.sc
```diff
@@ -166,7 +167,10 @@ object BuildDeps {
 
 def graalVmVersion     = "22.2.0"
 def graalVmJavaVersion = 17
-def graalVmJvmId       = s"graalvm-java$graalVmJavaVersion:$graalVmVersion"
+def graalVmJvmId =
+  if (isArmArchitecture)
+    s"https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.2.0/graalvm-ce-java17-darwin-aarch64-22.2.0.tar.gz"
```
alexarchambault: By using `cs` binaries from [VirtusLab/coursier-m1](https://github.com/VirtusLab/coursier-m1/releases) on M1, we shouldn't need to special case this I think.

alexarchambault: The `cs` binary used in the build is pulled [here](https://github.com/VirtusLab/scala-cli/blob/7013b86b349e72ed6d6c0d63aee5ddbf1cb223dc/project/settings.sc#L50) (and an M1 entry is missing).

modules/core/src/main/scala/scala/build/Os.scala
```diff
@@ -8,4 +10,6 @@ object Os {
       os.Path(os.pwd.toIO.getCanonicalFile)
     else
       os.pwd
+  def isArmArchitecture: Boolean =
```
alexarchambault: NIT Maybe this could be made a `lazy val`, as this is called multiple times.

modules/core/src/main/scala/scala/build/internals/OsLibc.scala
```diff
@@ -68,7 +69,8 @@ object OsLibc {
       .forall(_ >= 17)
     if (os == "linux-musl") s"liberica:$jvmVersion" // zulu could work too
     else if (java17OrHigher) s"temurin:$jvmVersion"
-    else s"adopt:$jvmVersion"
+    else if (Os.isArmArchitecture) s"zulu:$jvmVersion" // adopt don't support arm for java 8
```
alexarchambault:
```suggestion
    else if (Os.isArmArchitecture) s"zulu:$jvmVersion" // adopt doesn't support Java 8 on macOS arm
```

alexarchambault (approved):

modules/options/src/main/scala/scala/build/options/JavaOptions.scala
```diff
@@ -75,9 +75,14 @@ final case class JavaOptions(
             val enforceLiberica =
               finalJvmIndexOs == "linux-musl" &&
               jvmId.forall(c => c.isDigit || c == '.' || c == '-')
+            val enforceZulu =
+              Os.isArmArchitecture &&
+              jvmId.forall(c => c.isDigit || c == '.' || c == '-')
             val jvmId0 =
               if (enforceLiberica)
                 s"liberica:$jvmId" // FIXME Workaround, until this is automatically handled by coursier-jvm
+              else if (enforceZulu) // default jvmId adopto doesn't support java 8 for M1
```
alexarchambault:
```suggestion
              else if (enforceZulu) // default jvmId adoptium doesn't support java 8 for M1
```

PR #1397, opened 2022-09-28
Add --scalapy-version option

alexarchambault:
Allowing users to change the ScalaPy version when --python is passed.


Gedochao:
Maybe we could add some simple test?
otherwise, LGTM

alexarchambault:
> Maybe we could add some simple test?

Yeah. Initially I had integration tests in mind for that, but I was a bit reluctant to add ITs just for that. I just went for some more unitary tests (along with an extra commit tweaking the ScalaPy organization).

Gedochao:
> > Maybe we could add some simple test?
> 
> Yeah. Initially I had integration tests in mind for that, but I was a bit reluctant to add ITs just for that. I just went for some more unitary tests (along with an extra commit tweaking the ScalaPy organization).

Yeah, that works

Gedochao (approved):

PR #1399, opened 2022-09-28
Make run --python work with virtualenv for Scala Native

kiendang:
Make `run --python` work inside virtualenv for Scala Native by setting the `SCALAPY_PYTHON_PROGRAMNAME` to the path to the Python interpreter. This already works for JVM. For more details see https://github.com/VirtusLab/scala-cli/pull/1270#issuecomment-1240749730 and scalapy/scalapy#200.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/Run.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Run.scala
@@ -370,19 +370,22 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
         value(res)
       case Platform.Native =>
         val setupPython = build.options.notForBloopOptions.doSetupPython.getOrElse(false)
-        val pythonLibraryPaths =
+        val (pythonExecutable, pythonLibraryPaths) =
           if (setupPython)
             value {
-              val python       = Python()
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Thanks @kiendang!

Wondering if there would be a way to add non-regression tests for that… (likely in [`RunTestDefinitions`](https://github.com/VirtusLab/scala-cli/blob/cd5c6166bcdaa98688cef8fc4f3aca62361af57f/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala)) Maybe we could setup Python the same way ScalaPy does on its CI?

kiendang:
Currently virtualenv test in ScalaPy is WIP, part of scalapy/scalapy#303. I'm still not sure what's the best way to do this. The way I used to do it is to 

- create a venv
- install a dummy package into a venv
- activate the venv
- test if scalapy can import the package

something like this script https://github.com/scalapy/sbt-scalapy/blob/e8bb194d8277ccb60b6b42545be047702c7380db/scripts/virtualenv_test.sh
I'm not sure if we can do something like that here.

alexarchambault:
Merging this so that it's part of `0.1.16`. We can always add non regression tests for it later…

PR #1403, opened 2022-09-28
Turn off running tests in PR for M1 runner

lwronski:


Gedochao (approved):

alexarchambault (commented):
Shouldn't we restrict it to branches living on the `VirtusLab` organization? That way, we could run the M1 CI on our own PRs by pushing branches to `VirtusLab/scala-cli` rather than our own forks.

alexarchambault (approved):

alexarchambault:
Wondering if users wouldn't be able to run things on the M1 machine by simply removing these two lines… 🤔

PR #1406, opened 2022-09-29
Ensure package doesn't trigger a re-compilation after run

alexarchambault:


alexarchambault:
I need to add non-regression tests for those changes…

alexarchambault:
In the last commit, I should rather move them to `PostBuildOptions` I think 🤔

Gedochao (approved):

PR #1410, opened 2022-09-30
Generate pkg package for m1

lwronski:


modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
```diff
@@ -21,9 +20,6 @@ object TestUtil {
   val detectCliPath                = if (TestUtil.isNativeCli) TestUtil.cliPath else "scala-cli"
   val cli: Seq[String]             = cliCommand(cliPath)
 
-  lazy val isArmArchitecture: Boolean =
```
Gedochao, line 24: NIT I'd leave this here, since it's just a `TestUtil`
I know it's dead code atm and that's usually bad, but I think this one's bound to be useful in some test later on.

Gedochao (approved):

modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
lwronski: I don't think we should leave this code here. M1 and x86_64 should work the same way, so we don't need function to handle special cases.

Gedochao: ...fair enough

PR #1416, opened 2022-10-04
Add Scoop to Windows installation methods

### LINKED ISSUE #355: add scoop support on windows
Scoop is just like brew.

nightscape:
With https://github.com/ScoopInstaller/Main/pull/3994 scala-cli can now be installed via Scoop. The version will always be the most current one, due to Scoop's updating mechanism.

### DIFF
--- a/website/docs/_advanced_install.mdx
+++ b/website/docs/_advanced_install.mdx
@@ -165,6 +165,7 @@ values={[
 {label: 'Installer', value: 'installer'},
 {label: 'SDKMAN', value: 'sdkman'},
 {label: 'Chocolatey', value: 'choco'},
+{label: 'Scoop', value: 'scoop'},
 ]}
 >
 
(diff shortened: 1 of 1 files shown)


lwronski:
> With [ScoopInstaller/Main#3994](https://github.com/ScoopInstaller/Main/pull/3994) scala-cli can now be installed via Scoop. The version will always be the most current one, due to Scoop's updating mechanism.

@nightscape Thanks for your great contribution, it will be nice to have the next way to install scala-cli for Windows. LGTM

lwronski (approved):

nightscape:
@lwronski I think I did sth. wrong in the PR, at least I don't see the changes showing up in the live docs...

romanowski:
@nightscape, we do not deploy docs from `main` branch but rather from `stable` - a branch that gets updated when we do a release. Your contribution will be visible once we do the next release - those are quite often: ~ every 2 weeks.

We are doing it, so our documentation does not show bits and pieces that are not yet released.

PR #1423, opened 2022-10-06
Add missing resource for jgit

romanowski:
Without it, publish local fails when locale are set en_pl


alexarchambault (approved):

alexarchambault:
IIRC, this wasn't making things fail, this was only printing garbage in the console, then publishing keeps going.

PR #1424, opened 2022-10-06
Compute version for nested projects

romanowski:
Previously .git needs to be placed exactly in project root.

This change allow to host multiple libraries inside single repository.


romanowski:
@alexarchambault I think this may be really usable, but it may be too optimistic - please let me know what you think.

PR #1427, opened 2022-10-06
Update case-app_2.13 to 2.1.0-M17 & scala-cli-signing to v0.1.10

lwronski:


Gedochao (approved):
I adjusted the PR title, as I was initially confused by why this works now.
LGTM

alexarchambault (approved):

PR #1431, opened 2022-10-07
Add docs on how to install scala-cli for M1

lwronski:


website/docs/_advanced_install.mdx
```diff
@@ -299,6 +302,42 @@ sdk install scalacli
...
+```
+
+Check that it runs fine by running its `about` command:
+```bash
+scala-cli about
+```
+</TabItem>
+<TabItem value="installer">
+
+Download PKG installer with Scala CLI for MacOS/M1
```
Gedochao:
```suggestion
Download the PKG installer with Scala CLI for MacOS/M1
```

Gedochao (approved):

PR #1432, opened 2022-10-07
Publish / publish setup tweaks

alexarchambault:
Things I tweaked / fixed for the demo of my Scalacon talk.


alexarchambault:
Just checking what the CI says for now, I'll of course squash things, and maybe even split things in different PRs.

alexarchambault:
About this PR: I've been wanting to write non regression tests for it, but I don't recall exactly which scenarios it's fixing. But from what I recall, it's definitely useful, and the publish command should have been written this way initially IMO. And as it doesn't break any existing test, my opinion is that it can be merged as is…

MaciejG604 (approved):
LGTM!

MaciejG604:
The scenario you're fixing here is part of #1899 - artifacts won't get signed when secret keys are in config, and no warning will be presented to the user.
I will add tests for this in a follow up PR that will fully resolve #1899.

PR #1435, opened 2022-10-10
Run doc tests from munit test suites

alexarchambault:


modules/docs-tests/src/test/scala/sclicheck/GifTests.scala
```diff
@@ -0,0 +1,176 @@
...
+  val scenarioScripts = os.list(scenariosDir)
+    .filter(!_.last.startsWith("."))
+    .filter(_.last.endsWith(".sh"))
+    .filter(os.isFile(_))
+
+  lazy val hasTty =
+    os.proc("tty").call(stdin = os.Inherit, stdout = os.Inherit, check = false).exitCode == 0
+  lazy val ttyOpts = if (hasTty) Seq("-it") else Nil
+
+  def buildImages      = true
```
lwronski, line 31: Is there any way to change value this definition from `mill`?

alexarchambault: No, but it shouldn't be a problem: the images are built only once. If they're already built, this doesn't re-build them.

lwronski (approved):

Issue #1407, opened 2022-09-29
--help-native and --help-js drop into REPL confusingly


keynmol:
**Version(s)**
0.1.16-SNAPSHOT, built from 7a3011e584495d2266a85678ba69c04c58f8a43d

**Describe the bug**
Using some dedicated help flags drops into repl if no command is specificed:

```
~/p/scala-cli (main) > scli --help-native
Welcome to Scala 3.2.0 (17.0.4, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala>

~/p/scala-cli (main) > scli --help-js
Welcome to Scala 3.2.0 (17.0.4, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala>

~/p/scala-cli (main) > scli --help-full
Usage: scala-cli <COMMAND>
Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code.

Other commands:
  add-path                                      (hidden)
...
```

Passing a command fixes it:

```
~/p/scala-cli (main) > scli run --help-native
Usage: run [options]
Compile and run Scala code.

To pass arguments to the application, just add them after `--`, like:
```

**Expected behaviour**

Help is displayed without specifiyng the command the command.

PR #1440, opened 2022-10-11
Refactor `ScalaCommand` to enforce respecting help options

Gedochao:
The following sub-commands will now respect help options (`--help-native`, `--help-js`, `--scalac-help`, other compiler options printing an ouptut regardless of inputs, etc):
- `metabrowse`
- `export`
- `fmt`
- `setup-ide`
- `doc`

The idea is to prevent us from having to fix bugs like #1407 over and over again in the future.
It is now a lot easier to centralise certain pieces of logic in `ScalaCommand`, if they apply for each and every `sub-command`.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
@@ -6,7 +6,7 @@ import caseapp.core.complete.{Completer, CompletionItem}
 import caseapp.core.help.{Help, HelpFormat}
 import caseapp.core.parser.Parser
 import caseapp.core.util.Formatter
-import caseapp.core.{Arg, Error}
+import caseapp.core.{Arg, Error, RemainingArgs}
 
 import scala.annotation.tailrec
 import scala.build.compiler.SimpleScalaCompiler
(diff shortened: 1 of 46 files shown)


wleczny (approved):
LGTM

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -202,4 +202,56 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+        sharedOptions(options).foreach(_.logger.debug("build options could not be initialized"))
+        sys.exit(1)
+    }
+
+  /** @param options
+    *   command-specific [[T]] options
+    * @return
+    *   by default [[sharedOptions]].logging, override to adjust.
+    */
+  def loggingOptions(options: T): Option[LoggingOptions] =
```
lwronski, line 228: I think `loggingOptions` should always be set, why does this function return `Option`?
```suggestion
  def loggingOptions(options: T): LoggingOptions =
```
This should force everyone to implement `loggingOptions ` for new commands.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -202,4 +202,56 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+    loggingOptions(options).map(_.verbosity)
+
+  /** This should be overridden instead of [[run]] when extending [[ScalaCommand]].
+    *
+    * @param options
+    *   the command's specific set of options
+    * @param remainingArgs
+    *   arguments remaining after parsing options
+    */
+  def scalaCliRun(options: T, remainingArgs: RemainingArgs): Unit
```
lwronski:
```suggestion
  def runCommand(options: T, remainingArgs: RemainingArgs): Unit
```

What do you think about changing name of this function to `runCommand`?

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -220,6 +220,22 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+    */
+  def loggingOptions(options: T): Option[LoggingOptions] =
+    sharedOptions(options).map(_.logging)
+
+  /** @param options
+    *   command-specific [[T]] options
+    * @return
+    *   by default [[loggingOptions]].verbosity, override to adjust.
+    */
+  def verbosity(options: T): Option[Int] =
```
lwronski, line 236: I think `verbosity ` should always be set, why does this function return `Option`?
```suggestion
  def verbosity(options: T): Int =
```

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: I considered if it's by design, actually, but we do have sub-commands which do not support `LoggingOptions`. An example is the `about` sub-command.
If we made all those have `LoggingOptions`, then this wouldn't have to be an option, strictly.
then a definition for easy access to the logger could be added as well.

However, the benefit of leaving it as an option is that it can depend on `sharedOptions` by default.

Gedochao: However, if we decide to enforce all options to have `loggingOptions`, I'd rather do that in a separate PR

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: actually, that's not bad. Lemme rename it.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: verbosity indeed could always be set... but then we couldn't make it dependent on `loggingOptions` (chain dependent on `sharedOptions`) and it would have to be overriden from the perspective of each sub-command.
which I guess is a lot of code duplication... but I could extract it to a trait, maybe. 🤔

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: done

lwronski (approved):
LGTM, but lest's enforce all options to have `loggingOptions` in a separate PR.

PR #1451, opened 2022-10-13
Re-run things when users press enter in watch mode

alexarchambault:
Like how Mill does.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/WatchUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/WatchUtil.scala
@@ -17,7 +17,14 @@ object WatchUtil {
   def printWatchMessage(): Unit =
     System.err.println(waitMessage("Watching sources"))
 
-  def waitForCtrlC(): Unit =
-    while (System.in.read() != -1) {}
+  def waitForCtrlC(onPressEnter: () => Unit = () => ()): Unit = {
+    var readKey = -1
+    while ({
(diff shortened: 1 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/WatchUtil.scala
```diff
@@ -17,7 +17,14 @@ object WatchUtil {
...
 
-  def waitForCtrlC(): Unit =
-    while (System.in.read() != -1) {}
+  def waitForCtrlC(onPressEnter: () => Unit = () => ()): Unit = {
+    var readKey = -1
+    while ({
+      readKey = System.in.read()
+      readKey != -1
+    })
+      if (readKey == '\n')
```
Gedochao, line 26: Did you test it on Windows? Since it'd be `\r\n` there...
I think it should still catch the `\n`, but asking to be sure, since it's hard to add a test for it...

alexarchambault: Just tried it there, it seems to work (in the default command terminal).

lwronski (approved):

romanowski:
TODO: add a small section in the watch doc and a prompt when comment eds

alexarchambault:
https://github.com/VirtusLab/scala-cli/pull/1465 mentions pressing Enter in the prompt.

PR #1455, opened 2022-10-13
Accept spark-submit arguments on the command-line

alexarchambault:


modules/integration/src/test/scala/scala/cli/integration/SparkTestDefinitions.scala
```diff
@@ -129,7 +129,7 @@ abstract class SparkTestDefinitions(val scalaVersionOpt: Option[String]) extends
       val extraEnv =
         if (needsWinUtils) maybeHadoopHomeForWinutils(root / "hadoop-home")
         else Map.empty[String, String]
-      val res = os.proc(TestUtil.cli, "run", extraOptions, "--spark-standalone", "--jvm", "8", ".")
+      val res = os.proc(TestUtil.cli, "run", extraOptions, "--spark-standalone", ".")
```
lwronski, line 139: Does the default of scala-cli use java 8 even if I have on path java 17? If yes, I think it should be warning what happened.

Gedochao (approved):

modules/integration/src/test/scala/scala/cli/integration/SparkTestDefinitions.scala
alexarchambault: Yeah, if the system JVM is at least 8, maybe we should use it instead… 🤔

alexarchambault: I'll give it a try in another PR.

PR #1466, opened 2022-10-19
Add repositories.credentials config key

alexarchambault:


alexarchambault:
Includes https://github.com/VirtusLab/scala-cli/pull/1460 for now.

alexarchambault:
Going to add some docs, but the code changes can already be reviewed (and merged too).

modules/integration/src/test/scala/scala/cli/integration/ConfigTests.scala
```diff
@@ -100,7 +103,30 @@ class ConfigTests extends ScalaCliSuite {
...
+        else {
+          val script =
+            s"""#!/usr/bin/env bash
+               |exec "${TestUtil.cs}" "$$@"
+               |""".stripMargin
+          os.write(binDir / "cs", script, "rwxr-xr-x", createFolders = true)
+        }
+        Map(
+          "SCALA_CLI_CONFIG" -> confFile.toString,
+          pathVarName        -> s"$binDir${File.pathSeparator}$currentPath"
```
lwronski: Why you need to enforce using the version of `cs` from `deps.sc`?  If I used `config` command with older version of `cs` installed of my computer, it should this command work or not?

modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
```diff
@@ -111,23 +113,94 @@ object Config extends ScalaCommand[ConfigOptions] {
...
-                  if (options.passwordValue && entry.isPasswordOption)
-                    values.map { input =>
-                      PasswordOption.parse(input) match {
-                        case Left(err) =>
-                          System.err.println(err)
+                def parseSecret(input: String): Either[BuildException, Option[PasswordOption]] =
+                  if (input.trim.isEmpty) Right(None)
+                  else
+                    PasswordOption.parse(input)
+                      .left.map(err => ??? : BuildException)
```
lwronski: `PasswordOption.parse(input)` should return `BuildException` instead of `String`.

modules/config/src/main/scala/scala/cli/config/Keys.scala
```diff
@@ -54,4 +60,104 @@ object Keys {
...
+        cred = cred.copy(realm = Some(r))
+      for (opt <- optional)
+        cred = cred.copy(optional = Some(opt))
+      for (v <- matchHost)
+        cred = cred.copy(matchHost = Some(v))
+      for (v <- httpsOnly)
+        cred = cred.copy(httpsOnly = Some(v))
+      for (v <- passOnRedirect)
+        cred = cred.copy(passOnRedirect = Some(v))
+      cred
```
lwronski: Maybe it would be good enough to create the `RepositoryCredentials` at the end instead of using `var`

modules/config/src/main/scala/scala/cli/config/Keys.scala
```diff
@@ -54,4 +60,104 @@ object Keys {
...
+    matchHost: Option[Boolean] = None,
+    httpsOnly: Option[Boolean] = None,
+    passOnRedirect: Option[Boolean] = None
+  ) {
+    def credentials: RepositoryCredentials = {
+      var cred = RepositoryCredentials(host = host)
+      for (u <- user)
+        PasswordOption.parse(u) match {
+          case Left(error) =>
+            throw new Exception(
```
lwronski: Instead of throwing new Exception maybe it should be better to return `Either`.

modules/integration/src/test/scala/scala/cli/integration/ConfigTests.scala
alexarchambault: Yeah, when running the Scala CLI JVM tests, the test CLI launcher is a coursier "bootstrap" (generated from Mill), that tries to run `cs config` commands when the conf file seems to contain proxy details. And `cs config` only works with newer versions of coursier. Better use the one downloaded from the Scala CLI build.

lwronski: Shouldn't we add some warning that installed `cs` is outdated and `config` command won't work properly?

alexarchambault: That should be handled from coursier (in [this file](https://github.com/coursier/coursier/blob/ec619c43ab46a6425396a9ccd01394d65e4a4ea1/modules/bootstrap-launcher/src/main/java/coursier/bootstrap/launcher/Config.java)), the code running the `cs` commands lives there. I need to make the error messages better there.

modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
alexarchambault: Should be fixed.

modules/config/src/main/scala/scala/cli/config/Keys.scala
alexarchambault: Good point, should be fixed.

alexarchambault: Good point, should be fixed.

modules/config/src/main/scala/scala/cli/config/Keys.scala
```diff
@@ -49,4 +60,124 @@ object Keys {
...
+                lines += s"$prefix.https-only=$b"
+              for (b <- cred.matchHost)
+                lines += s"$prefix.auto=$b"
+              for (b <- cred.passOnRedirect)
+                lines += s"$prefix.pass-on-redirect=$b"
+              // seems cred.optional can't be changed from properties…
+              lines.map(_ + System.lineSeparator()).mkString
+          }
+      def fromString(values: Seq[String]): Either[Key.MalformedValue, List[RepositoryCredentials]] =
+        sys.error("Inline credentials not accepted, please manually edit the config file")
```
lwronski: Why not return error `Key.MalformedValue` instead of running `sys.error`?

lwronski (approved):
IMO we can marge it, but we have to remember to add warning in `cs` before next release of scala-cli. Creating a ticket for this will also be helpful.

PR #1485, opened 2022-10-21
remove paragraph about bug for coursier install

bishabosha:
0.1.16 shipped with a fix to recognise the way coursier installs it.


alexarchambault (approved):
Thanks!

PR #1487, opened 2022-10-24
Add help for repl, scalafmt and scaladoc

### LINKED ISSUE #1308: Scala CLI help should provide a way to display help from compiler, REPL and other undelaying tools

Scala CLI is intended to be a replacement for many tools like compiler, REPL and scalafmt. That is why we should allow our users a way to inspect the help from underlying tools as well. Also a help message for each such tool should mention about different such options.

A dedicated flag e.g. `--help-repl` or `--help-scalac` to display a help from underlying tool.

wleczny:


modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -159,6 +163,74 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+      artifacts      <- buildOptions.artifacts(logger, Scope.Main).toOption
+      scalaArtifacts <- artifacts.scalaOpt
+      scalaParams = scalaArtifacts.params
+      if shared.helpGroups.helpScaladoc || shared.helpGroups.helpRepl || shared.helpGroups.helpScalafmt
+    } {
+      val exitCode: Either[BuildException, Int] = either {
+        val (classPath: Seq[os.Path], mainClass: String) =
+          if (shared.helpGroups.helpScaladoc) {
+            val docArtifacts = value {
+              Artifacts.fetch(
```
lwronski, line 179: If I'm right, you can use `Artifacts.artifacts`, I think, it can be better because this method contains the logging fetched artifacts.

lwronski (commented):
I think, that some integrations should be added, for every added option to print help.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: isn't the logging potentially helpful, though?

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -159,6 +163,74 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+                Positioned.none(Seq(dep"org.scala-lang::scaladoc:${scalaParams.scalaVersion}")),
+                buildOptions.finalRepositories,
+                Some(scalaParams),
+                logger,
+                buildOptions.finalCache,
+                None
+              )
+            }
+            (docArtifacts.files.map(os.Path(_, os.pwd)), "dotty.tools.scaladoc.Main")
```
Gedochao: NIT I prefer the `->` syntax if there's a lot of brackets for improved readability
```suggestion
            docArtifacts.files.map(os.Path(_, os.pwd)) -> "dotty.tools.scaladoc.Main"
```

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -159,6 +163,74 @@ abstract class ScalaCommand[T](implicit myParser: Parser[T], help: Help[T])
...
+                  dep"${Constants.scalafmtOrganization}:${Constants.scalafmtName}:${Constants.defaultScalafmtVersion}"
+                )),
+                buildOptions.finalRepositories,
+                Some(scalaParams),
+                logger,
+                buildOptions.finalCache,
+                None
+              )
+            }
+            (fmtArtifacts.files.map(os.Path(_, os.pwd)), "org.scalafmt.cli.Cli")
```
Gedochao: NIT as above
```suggestion
            fmtArtifacts.files.map(os.Path(_, os.pwd)) -> "org.scalafmt.cli.Cli"
```

Gedochao (approved):

Issue #1349, opened 2022-09-12
Dynamically detect the default Scala version to be run with `ammonite`


Gedochao:
**Version(s)**
0.1.12

**Describe what needs to be done and why**
We needed #1348 to get rid the following problem after bumping to `3.2.0`
```
▶ scala-cli --ammonite
Downloading Ammonite 2.5.4-19-cd76521f
Downloading Ammonite 2.5.4-19-cd76521f sources
[error]  Can't download Ammonite 2.5.4-19-cd76521f for Scala 3.2.0.
Ammonite with this Scala version might not yet be supported.
Try passing a different Scala version with the -S option.
```
We need an automated mechanism to detect what is the Scala version to be run by default along with the `--ammonite` repl, so that we don't need similar hotfixes whenever Scala is bumped.

**Is your feature request related to a past ticket or discussion?**
#1348
#1327 
https://github.com/com-lihaoyi/Ammonite/pull/1286

**Describe alternatives you've considered**
Don't get me wrong, the hotfix does its job, but it's less than an ideal solution.
We could also just accept that `scala-cli --ammonite` will fail if `ammonite` doesn't support the Scala version `scala-cli` runs on.
Still I think even if explicitly chosen Scala version are ok to fail (i.e. `ammonite` will always fail if a nightly Scala version is passed), the default at the very least should fallback to the last supported version for as long as we support running the repl with `ammonite` at all.

Gedochao:
As described in #1450, an alternative solution:

We could have a separate Scala 2.12.x, 2.13.x and 3.x.y versions matching the last ones supported by Ammonite.
This would let us update Scala CLI's and Ammonite's Scala separately.
Additionally, when the last supported Ammonite Scala version and scala-cli Scala version don't match, a warning should be printed when running it with the default settings (when -S and --ammonite-version options are not passed; if they are, it's ok to just fail starting the repl).

Gedochao:
Ammonite support has been dropped, closing this.

PR #1493, opened 2022-10-24
Downgrade Scala version in 'scala-cli repl --amm' if needed

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -31,8 +31,34 @@ object Repl extends ScalaCommand[ReplOptions] {
...
+              def needsDowngradeForAmmonite = {
+                import coursier.core.Version
+                Version(maxScalaVer) < Version(defaultScalaVer)
+              }
+              ammonite.contains(true) &&
+              ammoniteVersionOpt.isEmpty &&
+              needsDowngradeForAmmonite
+            }
+            if (shouldDowngrade) {
+              System.err.println(
```
Gedochao: NIT might want to use `shared.logger` here instead of just `System.err`

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
```diff
@@ -31,8 +31,34 @@ object Repl extends ScalaCommand[ReplOptions] {
...
     import ops.sharedRepl.*
-    val baseOptions = shared.buildOptions().orExit(ops.shared.logger)
+
+    val ammoniteVersionOpt = ammoniteVersion.map(_.trim).filter(_.nonEmpty)
+    val baseOptions        = shared.buildOptions().orExit(ops.shared.logger)
     baseOptions.copy(
+      scalaOptions = baseOptions.scalaOptions.copy(
+        scalaVersion = baseOptions.scalaOptions.scalaVersion
+          .orElse {
+            val maxScalaVer     = scala.cli.internal.Constants.maxAmmoniteScala3Version
```
Gedochao: if I'm not mistaken, this will only work for running `--amm` without specifying the Scala version explicitly.
it would be neat if we had the auto-downgrade kicking in when the user passes a version with `-S`, but without an explicit patch version, like:
```scala
scala-cli --amm -S 3
scala-cli --amm -S 3.2
scala-cli --amm -S 2.13
scala-cli --amm -S 2.12
```

Gedochao:
This is neat btw, and potentially sufficient to satisfy #1349 
I am wondering, could we have integration tests with a dry run for the `repl --amm` with `-S 2.12`, `-S 2.13` and `-S 3`, so that we're certain it gets downgraded correctly whenever we bump Scala?

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
alexarchambault: I started to think a bit about that, but this needs more work. I htink this can be done in a subsequent PR.

modules/cli/src/main/scala/scala/cli/commands/Repl.scala
alexarchambault: Should be fixed.

Gedochao (approved):

lwronski (approved):

PR #1499, opened 2022-10-26
Enforce logging options for all scala commands

Gedochao:
Spinned off of https://github.com/VirtusLab/scala-cli/pull/1440#discussion_r993386283

These changes effectively force all sub-commands to have access to a logger (and all relevant options).
The only exception are the pgp signing sub-commands, which are now treated separately.


modules/cli/src/main/scala/scala/cli/commands/Doctor.scala
```diff
@@ -2,6 +2,7 @@ package scala.cli.commands
 
 import caseapp.core.RemainingArgs
 
+import scala.build.Logger
```
alexarchambault, line 5: Seems that's unused for now, isn't it?

Gedochao: the logger isn't being used, strictly, yeah.
but the import is necessary for `runCommand`'s signature on line 33.

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
```diff
@@ -44,8 +44,8 @@ object Bsp extends ScalaCommand[BspOptions] {
             pprint.err.log(initialInputs)
 
           val buildOptions0    = buildOptions(sharedOptions)
-          val logger           = sharedOptions.logging.logger
-          val persistentLogger = new PersistentDiagnosticLogger(logger)
+          val latestLogger     = sharedOptions.logging.logger
+          val persistentLogger = new PersistentDiagnosticLogger(latestLogger)
```
wleczny, line 48: Why can't you use `logger` instead?

```
val persistentLogger = new PersistentDiagnosticLogger(logger)
```

modules/cli/src/main/scala/scala/cli/commands/RestrictableCommand.scala
```diff
@@ -0,0 +1,10 @@
+package scala.cli.commands
+
+import caseapp.core.app.Command
+import caseapp.core.parser.Parser
+
+trait RestrictableCommand[T](implicit myParser: Parser[T]) {
```
wleczny, line 6: What is advantage of `RestrictableCommand` over `ScalaCommand`?

modules/cli/src/main/scala/scala/cli/commands/Bsp.scala
Gedochao: Because `BSP` refreshes its options based on the latest connection details.
The logger coming from the latest `SharedOptions` instance is used here in case i.e. the verbosity was changed since the last `BSP` `workspace/reload`.

modules/cli/src/main/scala/scala/cli/commands/RestrictableCommand.scala
Gedochao: I extracted the command-restricting logic to the `RestrictableCommand` trait to skip enforcing `LoggingOptions` in pgp signing sub-commands (https://github.com/VirtusLab/scala-cli/pull/1499/commits/047ec417aa2900bf08bd4a105ead995db15a1606).
Those sub-commands come from a dependency ([scala-cli#scala-cli-signing](https://github.com/scala-cli/scala-cli-signing)) and thus can't depend on `LoggingOptions`. 
They can't easily inherit from `ScalaCommand` now, since their sets of options are defined in the dependency. However, it's easy to add the `RestrictableCommand` trait for them, so that we can still restrict them.

We might want to enforce logging options in those sub-commands in a similar manner in the future, but that would require extracting `LoggingOptions` and `HasLoggingOptions` to a separate dependency and make both `scala-cli-signing` and `scala-cli` depend on them. IMO not worth the effort for now.

wleczny (approved):

lwronski (approved):

Issue #1285, opened 2022-08-25
`//> using option "--explain"` does nothing


Kordyjan:
**Version(s)**
`0.1.12`

**Describe the bug**
`//> using option "--explain"` doesn't change the compilation output. Probably it has the same root cause as #1023. The problem goes away if we run scala-cli with `--server=false`. 

**To Reproduce**
explain.scala:
``` scala
//> using option "--explain"

class A

val i: Int = A()
```

Run `scala-cli compile explain.scala`. The output will be:

```
[error] ./explain.scala:5:14: Found:    A
[error] Required: Int
[error] val i: Int = A()
[error]              ^^^
Error compiling project (Scala 3.1.3, JVM)
Compilation failed
```

**Expected behaviour**

The output should look like this:
```
-- [E007] Type Mismatch Error: /Users/pmarks/projs/rep/explain.scala:5:14 ------
5 |val i: Int = A()
  |             ^^^
  |             Found:    A
  |             Required: Int
  |-----------------------------------------------------------------------------
  | Explanation (enabled by `-explain`)
  |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
  |
  | Tree: new A()
  | I tried to show that
  |   A
  | conforms to
  |   Int
  | but the comparison trace ended with `false`:
  |
  |   ==> A  <:  Int
  |   <== A  <:  Int = false
  |
  | The tests were made under the empty constraint
   -----------------------------------------------------------------------------
   ```

lwronski:
@Kordyjan Thanks for reporting. I want to summarize this issue. 

First of all, the `explain` message which you mention is generated [here](https://github.com/lampepfl/dotty/blob/8de6d88450f89cb3d2dc779b880dae1a92deef70/sbt-bridge/src/dotty/tools/xsbt/DelegatingReporter.java#L40) in `scalac` and set to a  field `rendered` in `Problem` class.

From the `bloop` side, the `rendered` message are available and ready to use but it is skipped. In `BSP` there is no reasonable field in [Diagnostic](https://github.com/lwronski/build-server-protocol/blob/8281ddc9f45f4f725ec867cbbf559c94e8ac6fa5/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/Diagnostic.java#L10) class to send `rendered` message. The `rendered` content isn't so important to the IDE and is probably for this reason the message isn't sent anywhere from `bloop`.

[...]

Issue #1023, opened 2022-05-19
Compiler output passed wrongly from Bloop


prolativ:
**Version(s)**

0.1.4, 0.1.5, possibly earlier ones as well

**Describe the bug**

Compiler options passed after `-O` seem to be ignored when compiling with scala 2

**To Reproduce**

Foo.scala:
```scala
trait Foo
```

Running
```shell
scala-cli compile -S 2.13.8 -O -Vprint:typer Foo.scala
```
prints just
```
Compiling project (Scala 2.13.8, JVM)
Compiled project (Scala 2.13.8, JVM)
```

**Expected behaviour**
Compiler option passed in the command line after `-O` should be correctly passed to the compiler.

In the example above this should print the program tree after the typer phase similarly to

```shell
sdk use scala 2.13.8
scalac -Vprint:typer Foo.scala
```

printing
```
[[syntax trees at end of                     typer]] // Foo.scala
package <empty> {
  abstract trait Foo extends scala.AnyRef
}
```

lwronski:
Hi @prolativ, Thanks for reporting,
it seems that is bloop issue that output from scalac is not properly handled.

There is exists a workaround and you have to use plain `scalac` without `bloop`. To disable using `bloop` you should use `--server=false` flag.

```
$ scala-cli compile -S 2.13.8 -O -Vprint:typer Foo.scala --server=false
[[syntax trees at end of                     typer]] // Foo.scala
package <empty> {
  abstract trait Foo extends scala.AnyRef
}

```

SethTisue:
Could this be considered a blocker for promoting `scala-cli` to be `scala`? I feel (rather strongly) that it should be.

tgodzik:
Makes sense, I will need to take a look at it.

tgodzik:
Fixing it in https://github.com/scalacenter/bloop/pull/2361

tgodzik:
@SethTisue do you know if the output from "-Vprint" is being printed only to standard output? This might not be possible to sensibly fix as we would need to forward Bloop stdout and that could bring in more than we wanted, since it is running multiple compilations.

SethTisue:
where else would or could it be printed?

tgodzik:
> where else would or could it be printed?

I was thinking that it could go via info diagnostics for example? Since, I guess there is no sensible way to pass a custom output stream to the compiler

Gedochao:
bump @tgodzik

tgodzik:
[...]

PR #1503, opened 2022-10-27
Improve the docs on using Scala compiler options

Gedochao:
This should cover all the recent changes on handling compiler options.

### DIFF
--- a/website/docs/commands/compile.md
+++ b/website/docs/commands/compile.md
@@ -121,8 +121,50 @@ See the [Dependency management](../guides/dependencies.md) guide for more detail
 
 ## Scala compiler options
 
-Most [Scala compiler options](https://docs.scala-lang.org/overviews/compiler-options) can be passed as-is to `scala-cli`:
+### Passing compiler options with `-O`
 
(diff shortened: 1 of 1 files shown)


website/docs/commands/compile.md
```diff
@@ -121,7 +133,60 @@ See the [Dependency management](../guides/dependencies.md) guide for more detail
...
+```
+
+Passing a value to a compiler option requires another `-O`:
+
+```bash
+scala-cli -O -Xshow-phases -O -color -O never
+```
+
+:::note
+Scala CLI uses `bloop` by default, which at times gets in the way of getting the full compiler output.
```
tgodzik, line 155: Is that only the `-explain` issue? I think @lwronski found the origin of the issue.

Gedochao: Yeah, it's the `-explain` issue. 
We seem to have 2 related open tickets (#1023 and #1285), so I figured putting the workaround in the doc for the time being won't hurt.

tgodzik: Sure, makes sense. I was just worried there are more issues :sweat_smile:

lwronski (approved):

PR #1504, opened 2022-10-27
Tweak documentation verification tests

Gedochao:
- fixed outdated instructions for running documentation verification tests in the readme
- the tests should now also be run on guides, which seem to have been skipped
- fixed a few guides documents to make them pass tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,21 +185,32 @@ yarn run start
 
 ### Verifying the documentation
 
-We have a built-in tool to validate `.md` files called [Sclicheck](/sclicheck/Readme.md). To check all douments (and this is what we run on CI) run:
-
-```.github/scripts/check_docs.sh```
-
-You can also check single documents or directories using
-
-
-```
-.github/scripts/check_docs.sh <file> <dir>
+We have a built-in tool to validate `.md` files called [Sclicheck](/sclicheck/Readme.md). 
+All `Sclicheck` tests can be run with `Mill` + `munit`: (and this is what we run on the CI, too)
+```bash
+./mill -i docs-tests.test
 ```
 
-To debug failing document, Sclicheck has build-in following options: `--step` (stop after each command) or `--stopAtFailure` (to stop after a failure). To debug  getting started guide run following command:
-
+The former also includes testing gifs and `Sclicheck` itself.
+To just check the documents, run:
+```bash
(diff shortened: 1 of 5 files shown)


modules/docs-tests/src/test/scala/sclicheck/DocTests.scala
```diff
@@ -1,17 +1,16 @@
 package sclicheck
 
-class DocTests extends munit.FunSuite {
+import os.Path
```
alexarchambault: NIT Let's not import that one? It doesn't save many characters, and makes it clear that those paths originate from os-lib, rather than NIO.

Gedochao: yep, yep, wasn't intentional, my bad
should be fixed.

README.md
```diff
@@ -185,21 +185,31 @@ yarn run start
...
 
-We have a built-in tool to validate `.md` files called [Sclicheck](/sclicheck/Readme.md). To check all douments (and this is what we run on CI) run:
-
-```.github/scripts/check_docs.sh```
-
-You can also check single documents or directories using
-
-
+We have a built-in tool to validate `.md` files called [Sclicheck](/sclicheck/Readme.md). 
+All `Sclicheck` tests can be run with `Mill` + `munit`: (and this is what we run on the CI, too)
```
lwronski, line 189: I think it should describe what kind of `tests` is run.   


What is the difference between running `./mill -i docs-tests.test` and `/mill -i docs-tests.test 'sclicheck.DocTests.*'`. I suppose, that the first command also checks all documents.

website/docs/guides/configuration.md
```diff
@@ -23,12 +23,12 @@ scala-cli package --help
 ```
 
 As an example of command line configuration, one thing you can do with `scala-cli` command line options is to specify the Scala version:
-```bash
+```bash ignore
```
lwronski, line 26: We shouldn't have so many `ignore` in the documentation, could you create a issue to remove the `ignore` where possible?

Gedochao: #1523

README.md
Gedochao: hm... the former also tests gifs and the tool itself.
I added an extra line of explanation.

lwronski (approved):

PR #1518, opened 2022-11-03
Return exit code 1 when build fails for test

### LINKED ISSUE #1512: When running tests, compile error results in error code 0
**Version(s)**
0.1.16

**Describe the bug**
If I run `scala-cli test` and the code does not compile, the error messages get displayed but the command returns 0, i.e. success. This is problematic for example in CI pipelines.

**To Reproduce**
1. Create an empty folder and in it create `tests/SomeTest.scala`:
```Scala
//> using lib "org.scalameta::munit::0.7.29"

class  SomeTest extends munit.FunSuite:
	test("failig") {
		val s: String = 1
		assert(true == true)
	}
```
2. run `scala-cli test . && echo "tests OK" || echo "tests failed"`
3. This results in
```
[error] ./test/SomeTest.scala:5:19: Found:    (1 : Int)
[error] Required: String
[error] 
[error] The following import might make progress towards fixing the problem:
[error] 
[error]   import munit.Clue.generate
[error] 
[error] 		val s: String = 1
[error]                   ^
Error compiling project (test, Scala 3.2.0, JVM)
tests OK
```
**Expected behaviour**
The error code should be non-zero, and thus the output of the above command should be
```
[error] ./test/SomeTest.scala:5:19: Found:    (1 : Int)
[error] Required: String
[error] 
[error] The following import might make progress towards fixing the problem:
[error] 
[error]   import munit.Clue.generate
[error] 
[error] 		val s: String = 1
[error]                   ^
Error compiling project (test, Scala 3.2.0, JVM)
tests failed
```

lwronski:
Fixes #1512

### DIFF
--- a/modules/build/src/main/scala/scala/build/Builds.scala
+++ b/modules/build/src/main/scala/scala/build/Builds.scala
@@ -14,6 +14,7 @@ final case class Builds(
     }
   def get(scope: Scope): Option[Build] =
     builds.find(_.scope == scope)
+  def anyFailed: Boolean = !all.forall(_.success)
 
   def all: Seq[Build] =
     builds ++ crossBuilds.flatten
--- a/modules/cli/src/main/scala/scala/cli/commands/Test.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/Test.scala
@@ -153,6 +153,10 @@ object Test extends ScalaCommand[TestOptions] {
           actionableDiagnostics = actionableDiagnostics
         )
           .orExit(logger)
+      if (builds.anyFailed) {
+        System.err.println("Compilation failed")
+        sys.exit(1)
+      }
       maybeTest(builds, allowExit = true)
     }
   }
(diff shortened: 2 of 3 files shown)


modules/build/src/main/scala/scala/build/Builds.scala
```diff
@@ -14,6 +14,7 @@ final case class Builds(
     }
   def get(scope: Scope): Option[Build] =
     builds.find(_.scope == scope)
+  def isFail: Boolean = !builds.forall(_.success)
```
Gedochao:
```suggestion
  def anyFailed: Boolean = !builds.forall(_.success)
```
also, shouldn't we check all the builds here?
if so, then:
```suggestion
  def anyFailed: Boolean = !all.forall(_.success)
```

lwronski: good catch! thanks

Gedochao (approved):

PR #1519, opened 2022-11-03
Support `jvmRunEnvironment` and `jvmTestEnvironment` for BSP

### LINKED ISSUE #1516: ScalaCLI doesn't support jvmRunEnvironment
**Version(s)**
0.1.16

**Describe the bug**
ScalaCLI announces support of jvmRunEnviroment, but doesn't actually forward the requests to Bloop. 

**To Reproduce**
We had to special case scala-cli

https://github.com/scalameta/metals/pull/4566/files

**Expected behaviour**
Either the capability is announced as not supported or more preferably the support is added

Gedochao:
Fixes #1516

### DIFF
--- a/modules/bloop-rifle/src/main/scala/scala/build/bloop/BuildServer.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/bloop/BuildServer.scala
@@ -2,4 +2,4 @@ package scala.build.bloop
 
 import ch.epfl.scala.bsp4j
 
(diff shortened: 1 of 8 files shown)


lwronski (commented):
I think we need to add a few entries to reflect-config.json, I'm afraid it won't work for the native launcher.

Gedochao:
> I think we need to add a few entries to reflect-config.json, I'm afraid it won't work for the native launcher.

should be fixed now, I think...

tgodzik (approved):
LGTM from me! Thanks for looking into this!

lwronski (approved):

PR #1529, opened 2022-11-04
Decompose RunTestDefinitions

Gedochao:
This is an attempt at decomposing the source file containing tests for the `run` sub-command.
2.7k lines is a bit overkill


modules/integration/src/test/scala/scala/cli/integration/MarkdownTests.scala
```diff
@@ -0,0 +1,142 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+class MarkdownTests extends ScalaCliSuite {
```
wleczny, line 5: Maybe change this class to trait named RunMarkdownTests, since it tests `run` command?

lwronski: I like this convention that each test name will contain a prefix that points to name of command which already test.

Gedochao: if we change it to a trait, then we'll run it separately for Scala 2.12.x, 2.13.x and 3.x
Which I suppose isn't totally useless, but I wonder if it's worth the extra time on the CI...
but then, I guess it won't hurt all that much.
Should I?

lwronski (approved):

modules/integration/src/test/scala/scala/cli/integration/MarkdownTests.scala
lwronski: Right, I didn't think about it, so I approved this PR

PR #1531, opened 2022-11-04
Make package --python work

alexarchambault:
First commit is a refactoring, useful now that most commands support `--python` (after this PR, at least).

Second commit adds a non regression test for `scala-cli compile --python` (`compile` accepting `--python` thanks to the first commit).

Third commit adds support for `scala-cli package --python`, using the same setup as https://github.com/coursier/coursier/pull/2532.


modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -708,12 +718,41 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
...
-    val params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
+    var params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
       .withDeterministic(true)
       .withPreamble(preamble)
 
+    val params0 =
+      if (build.options.notForBloopOptions.doSetupPython.getOrElse(false)) {
+        val res = value {
+          Artifacts.fetch(
+            Positioned.none(Seq(dep"io.github.alexarchambault.python:interface:0.1.0")),
```
lwronski: The dependency version should be saved in `deps.sc`, this will allow scala-steward to automatically bump this version.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -708,12 +718,41 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
       .withOsKind(Properties.isWin)
       .callsItself(Properties.isWin)
       .withJavaOpts(build.options.javaOptions.javaOpts.toSeq.map(_.value.value))
-    val params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
+    var params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
       .withDeterministic(true)
       .withPreamble(preamble)
 
+    val params0 =
```
lwronski: Rename variable to name without 0 suffix.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -708,12 +718,41 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
       .withOsKind(Properties.isWin)
       .callsItself(Properties.isWin)
       .withJavaOpts(build.options.javaOptions.javaOpts.toSeq.map(_.value.value))
-    val params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
+    var params = Parameters.Bootstrap(Seq(loaderContent), mainClass)
```
lwronski: IMO, `var` is not required in this code snippet.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
alexarchambault: Adding it in `deps.sc` isn't enough, but I enabled it, yeah.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
alexarchambault: Should be fixed.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
alexarchambault: Yep, that's an oversight.

lwronski (approved):
LGTM

PR #1533, opened 2022-11-04
Add publish.credentials config key, use it to publish

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
```diff
@@ -672,150 +662,83 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
...
+    def authOpt(repo: String): Either[BuildException, Option[Authentication]] = either {
+      val isHttps = {
+        val uri = new URI(repo)
+        uri.getScheme == "https"
+      }
+      val hostOpt = {
+        val uri = new URI(repo)
+        if (uri.getScheme == "https") Some(uri.getHost)
+        else None
+      }
```
Gedochao:
```suggestion
      val hostOpt = if isHttps then Some(uri.getHost) else None
```

alexarchambault: Yep, should be fixed.

Gedochao (approved):

PR #1552, opened 2022-11-08
Remove cli-options module

alexarchambault:


Gedochao (approved):
LGTM, but:
- NIT: middle commit needs a better message or a NIT descriptor
- `SharedOptions` is a huge source file now, might be worth a look later (but not in this PR)
- great work tidying all this up, btw

romanowski (approved):

PR #1558, opened 2022-11-09
Move commands under distinct namespaces

alexarchambault:
This moves most commands in the `cli` module in specific namespaces, like `scala.cli.commands.run` for the `run` command, `scala.cli.commands.test` for the `test` command, etc.

Later on, when folding `cli-options` back into `cli`, this allows moving each command options to the command namespace, so that merging both modules creates less cluttering overall.


alexarchambault:
(Extracted from https://github.com/VirtusLab/scala-cli/pull/1552.)

Gedochao (approved):

PR #1565, opened 2022-11-10
Decompose & refactor `Inputs`

Gedochao:


modules/build/src/test/scala/scala/build/tests/InputsTests.scala
```diff
@@ -1,8 +1,10 @@
 package scala.build.tests
 
 import com.eed3si9n.expecty.Expecty.expect
+import os.Path
```
alexarchambault: Let's not import that one, so that we don't confuse it with NIO paths. (And importing it doesn't save much characters anyway…)

alexarchambault (commented):
Apart from the minor comment, LGTM!

lwronski (approved):

modules/build/src/test/scala/scala/build/tests/InputsTests.scala
Gedochao: Should be fixed

PR #1567, opened 2022-11-10
Remove superfluous annotation

alexarchambault:
Seems it doesn't matter when using case-app from Scala 2, but it does when using from Scala 3 apparently (subsequent PRs handle that).


alexarchambault:
Handling that from https://github.com/VirtusLab/scala-cli/pull/1568 actually.

alexarchambault:
Re-opened it, as it's more likely to be green earlier than https://github.com/VirtusLab/scala-cli/pull/1568.

lwronski (approved):

PR #1569, opened 2022-11-10
Update scala-cli-signing to 0.1.13

alexarchambault:


lwronski (changes requested):
It seems that the errors on CI are related to changes in this PR

lwronski (approved):

PR #1588, opened 2022-11-16
Disable create PGP key test on Windows CI

alexarchambault:
Downloading some of its dependencies have been too flaky there.


Gedochao (approved):

alexarchambault:
(Once https://github.com/VirtusLab/scala-cli/pull/1586 is merged, squash-and-merge here should only retain the last commit changes.)

PR #1589, opened 2022-11-16
Enforce stablity of our command line / using directives using Specification levels

romanowski:


modules/cli-options/src/main/scala/scala/cli/commands/SpecificationLevel.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import caseapp.annotation.Tag
+import os.copy
+
+sealed trait SpecificationLevel extends Product with Serializable {
+  def md = toString() + " have"
+}
+
+/** Specification levels in the context of Scala CLI runner specification. For more refer to
+  * [SIP-46](https://github.com/scala/improvement-proposals/pull/46)
```
julienrf, line 11: Since the PR has been merged, what do you think of linking to the website version? https://docs.scala-lang.org/sips/scala-cli.html

modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
```diff
@@ -8,9 +8,20 @@ import caseapp.core.{Arg, Error}
...
-    )
+    a.tags.exists(_.name == tags.restricted) || a.tags.exists(_.name == tags.experimental)
+
+  def level(a: Arg) =
+    a.tags.flatMap(t => tags.levelFor(t.name)).headOption match
+      case Some(opt) =>
+        if (
+          a.noHelp && opt != SpecificationLevel.EXPERIMENTAL && opt != SpecificationLevel.RESTRICTED
+        )
+          println(s"$a defines level as ${a.name.name} but is hidden")
```
julienrf, line 19: Is `println` intended?

Issue #1489, opened 2022-10-24
Set default name for settings file as `project.scala`


tgodzik:
**Describe what needs to be done and why**
In case of multiple files it might be useful to allow the user to specify all the using directives in a central file as one would expect similarly to build tools.

There was multiple discussions about what the name should and it seems the winning vote was on `project.scala`

We need to probably do a number of things:
- [x] warn users about multiple using directives and suggest using `project.scala`
- [x] automatically detect project as ScalaCLI in Metals when `project.scala` is present
- [ ] ~create a g8 template with `project.scala` and also add it to new project wizard in Metals~ (now tracked under #2108)
- [ ] ~add an automatic action to move all directives into `project.scala`~ (now tracked under #2104)
- [ ] ~fix tapir templates to use `project.scala`~ (now tracked under #2109)
- [x] update all existing references in our docs from `project.settings.scala` to `project.scala`

This issue can act as a central point coordinate the work being done.

Nice to have:
-  ~automatically detect project as ScalaCLI in Intellij when `project.scala` is present~ (now tracked under #2111)
- [ ] ~the existing docs on `project.settings.scala` seem to have been confusing to some users, so probably improve the docs a bit~ (now tracked under #2110)
  - [ ] ~we probably need a dedicated, small doc on this in the `Guides` section~
  - [ ] ~the [Project root directory doc](https://scala-cli.virtuslab.org/docs/reference/root-dir) should link to the aforementioned guide~
- [ ] ~Once all else is done, we probably should enable migrating directives to `project.scala` automatically with a sub-command, as [our docs are implying we've been planning to do that for a long while](https://scala-cli.virtuslab.org/docs/guides/using-directives#using-directives-in-the-scala-cli).~ (now tracked under #2104)

tgodzik:
Anything is missing?

Gedochao:
- [ ] update all existing references in our docs from `project.settings.scala` to `project.scala`

Nice to have
- [ ] the existing docs on `project.settings.scala` seem to have been confusing to some users, so probably improve the docs a bit
  - [ ] we probably need a dedicated, small doc on this in the `Guides` section
  - [ ] the [Project root directory doc](https://scala-cli.virtuslab.org/docs/reference/root-dir) should link to the aforementioned guide

Another nice to have: (this should be a separate ticket)
[...]

PR #1591, opened 2022-11-17
Warning for multiple files with using directives

wleczny:
One of the requirements of https://github.com/VirtusLab/scala-cli/issues/1489

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -6,6 +6,7 @@ import scala.build.Positioned
 import scala.build.errors.{BuildException, CompositeBuildException, MalformedDirectiveError}
 import scala.build.input.ElementsUtils.*
 import scala.build.input.*
+import scala.build.internal.Constants
 import scala.build.options.{
   BuildOptions,
   BuildRequirements,
@@ -166,6 +167,29 @@ object CrossSources {
     val preprocessedSources =
       (preprocessedInputFromArgs ++ preprocessedSourcesFromDirectives).distinct
 
+    val preprocessedWithUsingDirs = preprocessedSources.filter(_.directivesPositions.isDefined)
+    if (preprocessedWithUsingDirs.length > 1) {
+      val projectFilePath = inputs.elements.projectSettingsFiles.headOption match
+        case Some(s) => s.path
+        case _       => inputs.workspace / Constants.projectFileName
+      preprocessedWithUsingDirs
+        .filter(_.scopePath != ScopePath.fromPath(projectFilePath))
+        .foreach { source =>
+          source.directivesPositions match
+            case Some(positions) =>
+              val pos = Seq(
+                positions.codeDirectives,
+                positions.specialCommentDirectives,
+                positions.plainCommentDirectives
+              ).maxBy(p => p.endPos._1)
+              logger.diagnostic(
+                s"Using directives detected in multiple files. It is recommended to keep them centralized in the $projectFilePath file.",
+                positions = Seq(pos)
+              )
+            case _ => ()
+        }
+    }
+
     val scopedRequirements       = preprocessedSources.flatMap(_.scopedRequirements)
     val scopedRequirementsByRoot = scopedRequirements.groupBy(_.path.root)
     def baseReqs(path: ScopePath): BuildRequirements = {
@@ -238,7 +262,7 @@ object CrossSources {
       lazy val subPath = sourcePath.subRelativeTo(dir)
       if (os.isDir(sourcePath))
         Right(Directory(sourcePath).singleFilesFromDirectory(enableMarkdown))
-      else if (sourcePath == os.sub / "project.scala")
+      else if (sourcePath == os.sub / Constants.projectFileName)
         Right(Seq(ProjectScalaFile(dir, subPath)))
       else if (sourcePath.ext == "scala") Right(Seq(SourceScalaFile(dir, subPath)))
       else if (sourcePath.ext == "sc") Right(Seq(Script(dir, subPath)))
--- a/modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
@@ -2,9 +2,12 @@ package scala.build.preprocessing
 
 import java.nio.charset.StandardCharsets
 
+import scala.build.EitherCps.{either, value}
 import scala.build.Logger
 import scala.build.errors.BuildException
 import scala.build.input.{Inputs, SingleElement, VirtualData}
+import scala.build.options.BuildRequirements
+import scala.build.preprocessing.PreprocessingUtil.optionsAndPositionsFromDirectives
 
 case object DataPreprocessor extends Preprocessor {
   def preprocess(
@@ -15,23 +18,36 @@ case object DataPreprocessor extends Preprocessor {
   ): Option[Either[BuildException, Seq[PreprocessedSource]]] =
     input match {
       case file: VirtualData =>
-        val content = new String(file.content, StandardCharsets.UTF_8)
+        val res = either {
+          val content = new String(file.content, StandardCharsets.UTF_8)
+          val (updatedOptions, directivesPositions) = value {
+            optionsAndPositionsFromDirectives(
+              content,
+              file.scopePath,
+              Left(file.subPath.toString),
+              logger,
+              maybeRecoverOnError,
+              allowRestrictedFeatures
+            )
+          }
 
-        val inMemory = Seq(
-          PreprocessedSource.InMemory(
-            Left(file.source),
-            file.subPath,
-            content,
-            0,
-            None,
-            None,
-            Nil,
-            None,
-            file.scopePath
+          val inMemory = Seq(
+            PreprocessedSource.InMemory(
+              Left(file.source),
+              file.subPath,
+              content,
+              0,
+              Some(updatedOptions.global),
+              Some(BuildRequirements()),
+              Nil,
+              None,
+              file.scopePath,
+              directivesPositions
+            )
           )
-        )
-
-        Some(Right(inMemory))
+          inMemory
+        }
+        Some(res)
       case _ =>
         None
     }
--- a/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
@@ -18,16 +18,23 @@ import scala.jdk.CollectionConverters.*
 
 case class ExtractedDirectives(
   offset: Int,
-  directives: Seq[StrictDirective]
+  directives: Seq[StrictDirective],
+  positions: Option[DirectivesPositions]
 ) {
   @targetName("append")
   def ++(other: ExtractedDirectives): ExtractedDirectives =
-    ExtractedDirectives(offset, directives ++ other.directives)
+    ExtractedDirectives(offset, directives ++ other.directives, positions)
 }
+
+case class DirectivesPositions(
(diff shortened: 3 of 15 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,12 @@ object CrossSources {
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
 
+    val preprocessedWithUsingDirs =
+      preprocessedInputFromArgs.filter(p => p.options != Some(BuildOptions()) && p.options != None)
+    if (preprocessedWithUsingDirs.length > 1) println(
+      "Warning: using directives detected in multiple files. It is recommended to keep them centralized in the project.scala file."
+    )
```
lwronski:
```suggestion
    if (preprocessedWithUsingDirs.length > 1) logger.message(
      "Warning: using directives detected in multiple files. It is recommended to keep them centralized in the project.scala file."
    )
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,12 @@ object CrossSources {
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
 
+    val preprocessedWithUsingDirs =
+      preprocessedInputFromArgs.filter(p => p.options != Some(BuildOptions()) && p.options != None)
```
lwronski:
```suggestion
      preprocessedInputFromArgs.filter(p => p.options.exists(_ != BuildOptions()))
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,12 @@ object CrossSources {
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
 
+    val preprocessedWithUsingDirs =
+      preprocessedInputFromArgs.filter(p => p.options.exists(_ != BuildOptions()))
+    if (preprocessedWithUsingDirs.length > 1) logger.message(
+      "Warning: using directives detected in multiple files. It is recommended to keep them centralized in the project.scala file."
```
Gedochao: can we have the current location of the `project.scala` file here?
as in, if there is a `project.scala` within the scope of the project, I think it'd be useful to point the user to its path in this warning.
(and if there is none, we could still suggest the default project root as its path)

Gedochao (approved):

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,18 @@ object CrossSources {
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
 
+    val preprocessedWithUsingDirs =
+      preprocessedInputFromArgs.filter(p => p.options.exists(_ != BuildOptions()))
```
romanowski: shouldn't we compare it to `BuildOptions.configMonoid.zero`? We may change at some point the `default` empty BuildOptions.

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,18 @@ object CrossSources {
...
+    if (preprocessedWithUsingDirs.length > 1)
+      inputs.elements.projectSettingsFiles.headOption match {
+        case Some(p) => logger.message(
+            s"Warning: using directives detected in multiple files. It is recommended to keep them centralized in the ${p.path} file."
```
romanowski: Shouldn't we report this problem as warnings where the directives are  defined excluding project file itself?

romanowski (changes requested):
Some usability comments but can be done as a followup

modules/build/src/main/scala/scala/build/CrossSources.scala
wleczny: Good point, didn't think about it. I will change that.

wleczny: We should always raise warning, if there is more than one file with using directives (no matter if some of them are project files). 

As far as I know, we shouldn't raise warning if all using directives are at most in one file (even if it is not project file).

wleczny: Are you sure we should exclude project file from the warning?

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -153,6 +153,29 @@ object CrossSources {
...
+      val projectFilePath = inputs.elements.projectSettingsFiles.headOption match
+        case Some(s) => s.path
+        case _       => inputs.workspace / "project.scala"
+
+      preprocessedWithUsingDirs
+        .filter(p => os.Path(p.scopePath.path, os.root) != projectFilePath)
+        .foreach { file =>
+          file.scopePath.root match
+            case Right(base: os.Path) =>
+              val path = os.Path(file.scopePath.path, base)
```
romanowski: I do not like this approach. When we process directives we know their positions, can we just somehow reuse it?

Also following code:

```
// Use Scala 3
//> using scala 3
```

is a valid directive but will not be picked up.

romanowski (changes requested):
We should rework the way how we find where directives are

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -152,6 +152,31 @@ object CrossSources {
...
+            case Some(location) =>
+              val dirsScope = Seq(
+                location.codeDirectives.line,
+                location.specialCommentDirectives.line,
+                location.plainCommentDirectives.line
+              ).max
+              val path = source.scopePath.root match
+                case Right(base: os.Path) => os.Path(source.scopePath.path, base).toString
+                case Left(str: String)    => str + source.scopePath.path.toString
+              logger.diagnostic(
```
lwronski: Try using the `positions` arguments in `logger.diagnostic` and check how it is displayed in Metals. If you set the `positions` arguments correctly, the metals should show a warning message in the file.

It would be helpful in the future, if we decided here to add actionable diagnostic.

modules/build/src/main/scala/scala/build/CrossSources.scala
wleczny: I just added new approach, which uses positions extracted from files.

modules/build/src/main/scala/scala/build/CrossSources.scala
wleczny: I just committed changes, which use positions in logger.diagnostics. It works well with metals ide.

romanowski (approved):

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -152,6 +152,41 @@ object CrossSources {
         .map(_.flatten)
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
+    val preprocessedWithUsingDirs = preprocessedInputFromArgs.filter(_.directivesLocation.isDefined)
+
+    if (preprocessedWithUsingDirs.length > 1) {
+      val projectFilePath = inputs.elements.projectSettingsFiles.headOption match
+        case Some(s) => s.path
+        case _       => inputs.workspace / "project.scala"
```
lwronski: The project settings file name `"project.scala"` should be added to `Constants` file, instead of hardcoded in code.

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -152,6 +152,41 @@ object CrossSources {
...
+                case Left(str: String) =>
+                  val path = str + source.scopePath.path
+                  (Left(path), path)
+              val position = Position.File(
+                pathEither,
+                (0, 0),
+                (latestLocation.line, latestLocation.column)
+              )
+              logger.diagnostic(
+                s"Using directives detected in $pathStr. It is recommended to keep them centralized in the $projectFilePath file.",
```
lwronski: In metals, I don't think it's useful to print the path in the message alert.

And in the command line I think the path is printed twice because by default the warning is printed with the file path if the `position` is set.

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -18,16 +18,29 @@ import scala.jdk.CollectionConverters.*
...
+    ExtractedDirectives(offset, directives ++ other.directives, location)
 }
 
+case class DirectivesLocation(
+  codeDirectives: Location,
+  specialCommentDirectives: Location,
+  plainCommentDirectives: Location
+)
+
+case class Location(
```
lwronski: Why not use `Position.File` instead of `Location`?

modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
```diff
@@ -11,4 +19,35 @@ object PreprocessingUtil {
...
+    val ExtractedDirectives(_, directives0, dirsLocation) =
+      value(from(
+        content.toCharArray,
+        path,
+        logger,
+        Array(UsingDirectiveKind.PlainComment, UsingDirectiveKind.SpecialComment),
+        scopePath,
+        maybeRecoverOnError
+      ))
+    val updatedOptions = value(DirectivesProcessor.process(
```
lwronski, line 43: Do I understand correctly that now we process the directives twice for each file? I also see, that `DirectivesProcessor.process` is executed in `ScalaPreprocessor:352`

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -152,6 +152,41 @@ object CrossSources {
         .map(_.flatten)
 
     val preprocessedInputFromArgs = value(preprocessSources(inputs.flattened()))
+    val preprocessedWithUsingDirs = preprocessedInputFromArgs.filter(_.directivesLocation.isDefined)
```
lwronski: Do you test your mechanism with `file` directives?
```
//> using file "Utils.scala"
```

modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
```diff
@@ -15,23 +18,36 @@ case object DataPreprocessor extends Preprocessor {
   ): Option[Either[BuildException, Seq[PreprocessedSource]]] =
     input match {
       case file: VirtualData =>
-        val content = new String(file.content, StandardCharsets.UTF_8)
+        val res = either {
+          val content = new String(file.content, StandardCharsets.UTF_8)
+          val (updatedOptions, dirsLocation) = value {
```
lwronski:
```suggestion
          val (updatedOptions, directivesLocation) = value {
```

At first glance I thought it was a variable containing directories. Replace it in another piece of code as well.

modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
wleczny: I don't think so - I'm not using this method for ScalaPreprocessor.

lwronski: You have right!

modules/build/src/main/scala/scala/build/CrossSources.scala
wleczny: No I do not. I will change that.

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
wleczny: I didn't think about using `Position.File`. It seems logical. I will use it then.

lwronski (approved):

PR #1597, opened 2022-11-21
Enable to force using jvm signing launcher for native launcher of scala-cli

lwronski:
It is addressed to fix bug in CI for M1 - [download error](https://github.com/VirtusLab/scala-cli/actions/runs/3491103659/jobs/5843398632#step:6:8757) and add regression test to download JVM launcher of scala cli signing.


modules/cli/src/main/scala/scala/cli/commands/pgp/PgpScalaSigningOptions.scala
```diff
@@ -0,0 +1,30 @@
...
+// format: off
+final case class PgpScalaSigningOptions(
+  @Hidden
+    signingCliVersion: Option[String] = None,
+  @ValueDescription("option")
+  @Hidden
+    signingCliJavaArg: List[String] = Nil,
+  @HelpMessage("Whether to run the Scala Signing CLI on the JVM or using a native executable")
+  @Hidden
+    signingCliOnJvm: Option[Boolean] = None
```
Gedochao: NIT Mixed feelings on the current name for this flag.
How about:
```suggestion
  @HelpMessage("Specify the platform to run Scala Signing CLI (defaults to running the native executable)")
  @ValueDescription("native|jvm")
  @Hidden
    signingCliPlatform: Option[String] = None
```
or if we want to keep it as a boolean flag:
```suggestion
  @HelpMessage("Forces the JVM platform for Scala Signing CLI")
  @Hidden
  @Name("forceJvmSigning")
  @Name("jvmSigningCli")
  @Name("jvmSigning")
    forceJvmSigningCli: Option[Boolean] = None
```

modules/cli/src/main/scala/scala/cli/commands/publish/PublishOptions.scala
```diff
@@ -22,6 +23,8 @@ final case class PublishOptions(
     sharedPublish: SharedPublishOptions = SharedPublishOptions(),
   @Recurse
     sharedPython: SharedPythonOptions = SharedPythonOptions(),
+  @Recurse
+    singlingCli: PgpScalaSigningOptions = PgpScalaSigningOptions(),
```
Gedochao:
```suggestion
    signingCli: PgpScalaSigningOptions = PgpScalaSigningOptions(),
```

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpScalaSigningOptions.scala
lwronski: Right, `forceJvmSigningCli` it much better than `signingCliOnJvm`.  But I wouldn't add any alias, this flag is added only for test purpose.

Gedochao (approved):

PR #1598, opened 2022-11-22
Fix `using` directives for Markdown inputs

### LINKED ISSUE #1288: Fix using directives for multiple Scala snippets in a Markdown file and for plain `scala` code blocks
**Version(s)**
nightly

**Describe the bug**
Using directives are only correctly handled in the first `scala raw` or `scala test` snippet.
Also, plain `scala` code blocks don't handle using directives correctly at all.

**Known workarounds**
Just put all the `using` directives in a single `scala raw` code block

**To Reproduce**
````markdown
# Faulty snippet
```scala
//> using lib com.lihaoyi::os-lib:0.7.8
println(os.pwd)
```
````

**Expected behaviour**
`using` directives should be handled correctly in each snippet, even when declared within the bounds of a single Markdown file.

Gedochao:
Fixes #1288


lwronski (approved):
LGTM

PR #1605, opened 2022-11-24
Enable by default actionable diagnostic

lwronski:
To disable printing actionable diagnostic messages in command line, users have to run:
```
scala-cli config action false 
```

If someone wants to disable actionable diagnostic with Metals should run `setup-ide` with `--action false` flag:

```
scala-cli setup-ide --actions false . 
```


Gedochao (approved):
LGTM, but
- it'd be great for our docs to have some reference for the keys available in the `config` subcommand
- we could also use a dedicated doc for actionable diagnostics where these changes could also be mentioned

PR #1614, opened 2022-11-28
Better revolver output

alexarchambault:


modules/cli/src/main/scala/scala/cli/commands/WatchUtil.scala
```diff
@@ -9,18 +11,47 @@ object WatchUtil {
...
+    val reset = Console.RESET
+    s"$gray$message$reset"
   }
 
   def printWatchMessage(): Unit =
     System.err.println(waitMessage("Watching sources"))
 
-  def waitForCtrlC(onPressEnter: () => Unit = () => ()): Unit = {
+  def printWatchWhileRunningMessage(): Unit =
+    System.err.println(gray("Watching sources while your program is running."))
```
Gedochao, line 27: NIT I'm thinking the logger could be passed here and used to print the watch messages, to respect verbosity... 
But it's not a big deal, can be done separately.

Gedochao (approved):
LGTM

PR #1616, opened 2022-11-28
Put workspace and / or current directory in PYTHONPATH in the run and repl commands

alexarchambault:
I guess it's something that ScalaPy ought to do on its own (going to open an issue for it…). In the mean time, this PR makes things work for the `run` command at least.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -194,6 +194,13 @@ object dummy extends Module {
       Deps.pythonInterface
     )
   }
+  object scalaPy extends ScalaModule with Bloop.Module {
+    def skipBloop    = true
+    def scalaVersion = Scala.defaultInternal
+    def ivyDeps = Agg(
+      Deps.scalaPy
+    )
+  }
 }
 
 trait BuildMacros extends ScalaCliSbtModule
--- a/modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
@@ -14,7 +14,11 @@ import scala.build.internal.{Constants, Runner}
 import scala.build.options.{BuildOptions, JavaOpt, MaybeScalaVersion, Scope}
 import scala.cli.CurrentParams
 import scala.cli.commands.publish.ConfigUtil.*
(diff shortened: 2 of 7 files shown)


kiendang:
I think a more correct solution that makes this behaves more like `python` would be
- for `repl`, adding the current directory
- for `run`, adding the directory of the main script

and shoud be added to the beginning of `PYTHONPATH`.

For `run`, this lets you run the script from outside the directory where it's located, `scala-cli run --python ./projectdir/main.sc` if `main.sc` import from some `script.py` inside `projectdir`.

You can test this with `python`.

```py
// projectdir/lib.py
s = 'Hello'
```

```py
// projectdir/main.py
import lib
print(lib.s)
```

```sh
python ./projectdir/main.py
Hello
```

It's documented here https://docs.python.org/3/c-api/init_config.html#c.PyConfig.safe_path.

Initially I also thought we should change this in ScalaPy (the init process in ScalaPy indeed could be improved since the approach currently in use is deprecated), but now I'm not so sure since ScalaPy doesn't know whether it's running in repl or script mode.

alexarchambault:
@kiendang Good point! I went for the "workspace" that Scala CLI defines, that corresponds to the sources' root most of the time. And also made `scala-cli repl --python`  update `PYTHONPATH` (with the current directory, and optionally the workspace dir alongside it, when doing things like `scala-cli repl --python ./my-project`).

kiendang:
Just read about a potential security risk with this https://twitter.com/marekgibney/status/1598706464583028736.

Python introduced the [`PYTHONSAFEPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONSAFEPATH) env var to let users disable this.

Maybe we should check if `PYTHONSAFEPATH` is set to a non-empty string? If so don't add to `PYTHONPATH`.

kiendang:
See https://www.reddit.com/r/Python/comments/zapv47/looking_at_the_python_documentation_can_be/iynk7xb/

Gedochao (approved):
LGTM, although I think this could use some docs
(well, the whole ScalaPy feature would, so it's fine to do that later)

PR #1617, opened 2022-11-29
Reduce ignore tags in the docs snippets

### LINKED ISSUE #1523: Reduce `ignore` tags in the docs snippets
> We shouldn't have so many `ignore` in the documentation

_Originally posted by @lwronski in https://github.com/VirtusLab/scala-cli/pull/1504#discussion_r1013772503_

We use the `ignore` tag in numerous scripts all over our docs. In most cases, it's because an input is used and it's not explicitly created in the relevant snippet.
It'd be good to be able to test those snippets, optimally without cluttering the docs too much in the process.

Gedochao:
Fixes #1523


website/docs/commands/compile.md
```diff
@@ -28,31 +30,47 @@ For a full list of options, run `scala-cli compile --help`, or check the options
 
 `--test` makes `scala-cli` compile main and test scopes:
 
-```bash ignore
-scala-cli compile --test Hello.scala
+```scala title=Sample.test.scala
+//> using lib "org.scalameta::munit:0.7.29"
```
lwronski, line 34: IMO we miss a mechanism to update libraries into a documentation. 

Perhaps `dependency-update` command will work fine with markdown files, but it is out of the scope this PR.

lwronski (approved):
LGTM

website/docs/commands/compile.md
lwronski: https://github.com/VirtusLab/scala-cli/issues/1619

PR #1626, opened 2022-11-30
Add scalaCompilerPlugins to Mill project export

carlosedp:
This PR adds the export of scala compiler plugins to Mill projects.

A file like:

```scala
//> using scala "2.13.10"
//> using lib "edu.berkeley.cs::chisel3::3.5.5"
//> using plugin "edu.berkeley.cs:::chisel3-plugin::3.5.5"

import chisel3._
import circt.stage.ChiselStage
...
```

Generates a `build.sc`:

```scala
import mill._
import mill.scalalib._

object memmask extends ScalaModule {
  def scalaVersion = "2.13.10"
  
  def ivyDeps = super.ivyDeps() ++ Seq(
  ivy"edu.berkeley.cs::chisel3::3.5.5",
)

  def scalacPluginIvyDeps = super.scalacPluginIvyDeps() ++ Seq(
  ivy"edu.berkeley.cs:::chisel3-plugin::3.5.5"
)
...
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
+++ b/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
@@ -48,7 +48,9 @@ final case class Mill(
 
   private def scalaCompilerPlugins(buildOptions: BuildOptions): MillProject =
     MillProject(scalaCompilerPlugins =
-      buildOptions.scalaOptions.compilerPlugins.toSeq.map(_.value.render)
+      buildOptions.scalaOptions.compilerPlugins.toSeq.map { p =>
+        s"${p.value.organization}:::${p.value.name}:${p.value.version}"
+      }
     )
 
   private def scalacOptionsSettings(buildOptions: BuildOptions): MillProject =
--- a/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
+++ b/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
@@ -48,9 +48,7 @@ final case class Mill(
 
   private def scalaCompilerPlugins(buildOptions: BuildOptions): MillProject =
     MillProject(scalaCompilerPlugins =
-      buildOptions.scalaOptions.compilerPlugins.toSeq.map { p =>
-        s"${p.value.organization}:::${p.value.name}:${p.value.version}"
-      }
+      buildOptions.scalaOptions.compilerPlugins.toSeq.map(_.value.render)
     )
 
   private def scalacOptionsSettings(buildOptions: BuildOptions): MillProject =
--- a/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
+++ b/modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
@@ -46,6 +46,11 @@ final case class Mill(
       MillProject(scalaVersion = Some(sv))
   }
 
+  private def scalaCompilerPlugins(buildOptions: BuildOptions): MillProject =
+    MillProject(scalaCompilerPlugins =
+      buildOptions.scalaOptions.compilerPlugins.toSeq.map(_.value.render)
+    )
+
   private def scalacOptionsSettings(buildOptions: BuildOptions): MillProject =
     MillProject(scalacOptions = buildOptions.scalaOptions.scalacOptions.toSeq.map(_.value.value))
 
@@ -187,6 +192,7 @@ final case class Mill(
       sourcesSettings(sourcesMain, sourcesTest),
       scalaVersionSettings(optionsMain, sourcesMain),
       scalacOptionsSettings(optionsMain),
+      scalaCompilerPlugins(optionsMain),
       dependencySettings(optionsMain, optionsTest),
       repositorySettings(optionsMain),
       if (optionsMain.platform.value == Platform.JS) scalaJsSettings(optionsMain.scalaJsOptions)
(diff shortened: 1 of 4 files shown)


modules/cli/src/main/scala/scala/cli/exportCmd/Mill.scala
```diff
@@ -46,6 +46,11 @@ final case class Mill(
       MillProject(scalaVersion = Some(sv))
   }
 
+  private def scalaCompilerPlugins(buildOptions: BuildOptions): MillProject =
+    MillProject(scalaCompilerPlugins =
+      buildOptions.scalaOptions.compilerPlugins.toSeq.map(_.value.render)
```
lwronski, line 51: @carlosedp  Thanks for your contribution, 

~~it seems that the `.reder` method returns the wrong number of colons for the compiler plugin.~~

~~Therefore `mill` fail with downloading plugin dependency:~~
```
newproject.enablePluginScalacOptions 
Resolution failed for 1 modules:--------------------------------------------
  com.olegpy:better-monadic-for_3:0.3.1 
	not found: /home/runner/.ivy2/local/com.olegpy/better-monadic-for_3/0.3.1/ivys/ivy.xml
	not found: https://repo1.maven.org/maven2/com/olegpy/better-monadic-for_3/0.3.1/better-monadic-for_3-0.3.1.pom
```

carlosedp: Thanks for picking that up @lwronski! I haven't noticed that this was the problem.
Fixed on latest push.

lwronski: @carlosedp  It wasn't a problem, sorry for pointing the problem wrong.

The problem is, that `ExportMillTestDefinitions` tests are running with every version of Scala - [2.12, 2.13, 3]. But `better-monadic-for` is published only for `2.12` and `2.13`. Therefore the test which used Scala 3 fail in CI, because mill can not download `better-monadic-for` for Scala 3.

You can run Mill export tests with Scala 3 in this way:
```
 ./mill integration.test.jvm 'scala.cli.integration.ExportMillTestsDefault.JVM custom project name'
```

I think you should revert the last commit because the `render` works fine, but this PR needs to update tests. Maybe it would be better to add separate tests and test only export scala compiler plugins with specific scala versions. It could be added small tests in `ExportMillTestsDefault` and in the test you can hardcode the scala version, for example, 3.1.0.

lwronski: `ExportMillTestsDefault` tests contains test which are running with Scala 3. If you specify scala versions in test via `-S` it should be good enough, sth like this:
```
      os.proc(TestUtil.cli, "export", extraOptions, "--mill", setProject, "-o", "mill-proj", ".", "-S", "3.1.0")
        .call(cwd = root, stdout = os.Inherit)
```

carlosedp: Got it... Thanks a lot for the tips.
I'm figuring out a way to test it locally since I'm developing on a Mac and these tests only run on Linux.

lwronski: I think, you can temporary change definition of `ExportMillTestDefinitions.runExportTests` to:
```
  protected def runExportTests: Boolean = true
```
to local development. And before pushing the commit back to the previous implementation of `ExportMillTestDefinitions.runExportTests`.

Maybe, scala-cli should turn on ExportMillTest to run on Mac CI. I will investigate it.

carlosedp: Nice, setting it to `true` worked (dunno why it didn't work the last time I tried). I filtered my test with `if (runExportTests && !actualScalaVersion.startsWith("3."))`.

I'm running it as `./mill integration.test.jvm 'scala.cli.integration.ExportMillTestsDefault.*'` but it's not running the tests for Scala 2... is there a way to pass the scala version as an argument?

```
scala.cli.integration.ExportMillTestsDefault:
  + JVM 22.97s
  + JVM custom project name 19.303s
  + Scala.js 20.15s
```

lwronski: If you want to run test for Scala 2 you can do it in this way:
```
./mill integration.test.jvm 'scala.cli.integration.ExportMillTests213.*' 
./mill integration.test.jvm 'scala.cli.integration.ExportMillTests212.*' 
```

Eg `ExportMillTests212` runs tests that are defined in `ExportMillTestDefinitions` using Scala 2.12. And ExportMillTestsDefault uses Scala 3.

carlosedp: Oh, got it... my bad I didn't see it before!
Running right now and will push in a couple minutes.

carlosedp: Yay, worked!

```
scala.cli.integration.ExportMillTests213:
  + JVM 28.302s
  + JVM custom project name 24.978s
  + JVM with compiler plugin 26.33s
  + Scala.js 25.377s
  + Scala Native 39.551s
  + scalac options 20.64s
  + pure java 9.633s
  + custom JAR 21.0s
```

Gedochao (approved):

lwronski (approved):

lwronski (commented):
LGTM, thanks for your contribution.

PR #1627, opened 2022-11-30
Add fix command

wleczny:


modules/cli/src/main/scala/scala/cli/commands/fix/FixOptions.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import caseapp.*
+import caseapp.core.help.Help
+
+import scala.build.internal.Constants
+import scala.cli.commands.migratedirectives.MigrateDirectivesOptions
+import scala.cli.commands.shared.{HasSharedOptions, SharedOptions}
+
+// format: off
+@HelpMessage("Run fixes for Scala CLI project")
```
Gedochao:
```suggestion
@HelpMessage("Run fixes for a Scala CLI project")
```

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Fix
+sidebar_position: 20
+---
+
+The fix commands can be used to perform fixes on the Scala CLI project.
```
Gedochao:
```suggestion
The fix command can be used to perform fixes on a Scala CLI project.
```

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Migrating directives
+It is recommended to keep all using directives centralized in one file. Fix command along with `--migrate-directives` option will migrate all directives of given inputs to the `project.scala` file (if the file does not exist, Scala CLI will create it).
```
Gedochao:
```suggestion
It is recommended to keep all `using` directives centralised in one file. Fix command along with the `--migrate-directives` option will migrate all directives of given inputs to the `project.scala` configuration file (if the file does not exist, Scala CLI will create it in the project root directory).
```

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
...
+
+Following command will move all directives inside current directory to `project.scala` file:
```
Gedochao:
```suggestion
The following command will move all directives inside the current working directory to the `project.scala` file:
```

modules/cli/src/main/scala/scala/cli/commands/fix/FixOptions.scala
```diff
@@ -0,0 +1,25 @@
...
+final case class FixOptions(
+  @Recurse
+    migrateDirectivesOptions: MigrateDirectivesOptions = MigrateDirectivesOptions(),
+  @HelpMessage(s"Move all using directives to ${Constants.projectFileName} file")
```
Gedochao:
```suggestion
  @HelpMessage(s"Move all using directives to the ${Constants.projectFileName} file")
```

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+        )
+      val code       = indexedCode.map(_._1)
+      val directives = indexedDirectives.map(_._1)
+      (code.mkString("\n"), directives.mkString("\n"))
+    }
+
+    preprocessedWithUsingDirs.length match
+      case 0 => logger.message("No using directives have been found")
+      case 1 => logger.message("All using directives are already in one file")
+      case _ =>
```
Gedochao:
```suggestion
      case n =>
        logger.message(s"Using directives found in $n files. Migrating...")
```

Gedochao (commented):
Can we add some integration tests?

Gedochao:
I think this won't work with markdown inputs for now, although that can be fixed in a separate PR

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+                  case Some(positions) =>
+                    val (code, directives) = splitCodeAndDirectives(contentLines, positions)
+                    logger.message(s"Deleting directives from ${file.path}...")
+                    os.write.over(file.path, code.getBytes(StandardCharsets.UTF_8))
+                    Seq(directives)
+                  case _ => Nil
+                result
+
+              case file: InMemory =>
+                val contentLines = file.code.split("\n").toSeq
```
Gedochao: seems like the only difference between how `InMemory` and `OnDisk` elements are handled is how the `contentLines` are extracted... the rest seems like duplicated code, should be abstracted

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+                val result = file.directivesPositions match
+                  case Some(positions) =>
+                    val (_, directives) = splitCodeAndDirectives(contentLines, positions)
```
Gedochao: btw. some logger message could be useful for `InMemory` elements, to inform the user that those directives can't be deleted, but they're still being migrated.

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+                    Seq(directives)
+                  case _ => Nil
+                result
+              case _ => Nil
+          }
+        val allDirectives = directivesSeq.fold(Nil)(_ ++ _).mkString("\n")
+        logger.message(s"Moving all directives to $projectFilePath...")
+        if (os.exists(projectFilePath)) {
+          val newProjectFileContent = allDirectives + "\n" + os.read(projectFilePath)
+          os.write.over(projectFilePath, newProjectFileContent.getBytes(StandardCharsets.UTF_8))
```
Gedochao: should we write over the existing `project.scala`?
what if there was existing logic in there?
I think the migrated directives should be appended after any existing directives there, but before any further code that already exists

Gedochao: also, it'd be useful to log whether the `project.scala` file was found, if it was created or if directives were added to it

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+        }
+        else
+          os.write(
+            projectFilePath,
+            allDirectives.getBytes(StandardCharsets.UTF_8),
+            createFolders = true
+          )
+        logger.message(s"Successfully moved all using directives to the project file!")
```
Gedochao:
```suggestion
        logger.message(s"Successfully moved all using directives to the project file: $projectFilePath")
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -120,17 +120,22 @@ object CrossSources {
     }
 
   /** @return
-    *   a CrossSources and Inputs which contains element processed from using directives
+    *   Inputs which contain elements processed from using directives and preprocessed sources
+    *   extracted from these Inputs
```
Gedochao: NIT
```suggestion
    *   extracted from given Inputs
```

modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
```diff
@@ -40,11 +40,13 @@ class ScalaCliCommands(
     doc.Doc,
     doctor.Doctor,
     export0.Export,
+    fix.Fix,
     fmt.Fmt,
     new HelpCmd(help),
     installcompletions.InstallCompletions,
     installhome.InstallHome,
     metabrowse.Metabrowse,
+    migratedirectives.MigrateDirectives,
```
lwronski: So for each option in the `fix` command we will create a separate command? I'm not convinced that `MigrateDirectives` should be a sub-commend, it might just be an option in the `fix`.

cc: @Gedochao

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+    ): (String, String) = {
+      val endLine   = positions.scope.endPos._1
+      val endColumn = positions.scope.endPos._2
+      val (indexedDirectives, indexedCode) =
+        contentLines.zipWithIndex.partition((_, idx) =>
+          idx < endLine || (idx == endLine && endColumn > 0)
+        )
+      val code       = indexedCode.map(_._1)
+      val directives = indexedDirectives.map(_._1)
+      (code.mkString("\n"), directives.mkString("\n"))
```
lwronski: It won't work on windows. You can do something similar as in `DependencyUpdate` [here](https://github.com/VirtusLab/scala-cli/blob/5639c4b84ae652e1f46120ef2e3eb1bda0b07931/modules/cli/src/main/scala/scala/cli/commands/dependencyupdate/DependencyUpdate.scala#L117).

modules/cli/src/main/scala/scala/cli/commands/migratedirectives/MigrateDirectives.scala
```diff
@@ -0,0 +1,104 @@
...
+                    val (_, directives) = splitCodeAndDirectives(contentLines, positions)
+                    Seq(directives)
+                  case _ => Nil
+                result
+              case _ => Nil
+          }
+        val allDirectives = directivesSeq.fold(Nil)(_ ++ _).mkString("\n")
+        logger.message(s"Moving all directives to $projectFilePath...")
+        if (os.exists(projectFilePath)) {
+          val newProjectFileContent = allDirectives + "\n" + os.read(projectFilePath)
```
lwronski: It looks like `allDirectives` may contain duplicates, for example different versions of scala.  Maybe some interactive mode will be helpful - @Gedochao what do you think?

lwronski: @wleczny And how does the fix command work if there are directives pointing to different versions of Scala in different files?

Gedochao: IMO it should be consistent with how it works with other commands, like `run`/`compile`/etc.
The way it currently works is that the Scala version from the first input will take priority.
A warning definitely would be in place, but IMO the same one should be migrated to `project.scala`, while the other ones just removed after running `fix`.
and yeah, I agree that an `--interactive` mode scenario would be useful here, but that could be extracted to a separate ticket/PR.

modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
Gedochao: We also have `dependencies-update`, which IMO should probably be a part of `fix` as well...
Dunno, I don't really mind having aliases for those features, although I agree it'd be easier to maintain if we just put all that under the `fix` sub-command...

lwronski: `dependencies-update` should be part of the fix, but this functionality deserves for own command. I am not convinced that is the same case with `MigrateDirectives`.

wleczny: I would also opt for not using alias for `migrate-directives` and just put it all under `fix` sub-command.

modules/cli/src/main/scala/scala/cli/commands/fix/Fix.scala
```diff
@@ -0,0 +1,117 @@
...
+import scala.build.internal.{Constants, CustomCodeWrapper}
+import scala.build.preprocessing.PreprocessedSource._
+import scala.build.preprocessing.{DirectivesPositions, PreprocessedSource, ScopePath}
+import scala.build.{CrossSources, Logger, Position, Sources}
+import scala.cli.commands.shared.SharedOptions
+import scala.cli.commands.{ScalaCommand, SpecificationLevel}
+
+object Fix extends ScalaCommand[FixOptions] {
+  override def group                   = "Main"
+  override def scalaSpecificationLevel = SpecificationLevel.RESTRICTED
```
Gedochao: Might be safer for a while, we can move it to `RESTRICTED` after it proves to be stable.
```suggestion
  override def scalaSpecificationLevel = SpecificationLevel.EXPERIMENTAL
```

modules/cli/src/main/scala/scala/cli/commands/fix/FixOptions.scala
```diff
@@ -0,0 +1,22 @@
+package scala.cli.commands.fix
+
+import caseapp.*
+import caseapp.core.help.Help
+
+import scala.build.internal.Constants
+import scala.cli.commands.shared.{HasSharedOptions, SharedOptions}
+
+// format: off
+@HelpMessage("Run fixes for a Scala CLI project")
```
Gedochao: Might want to add a detailed help message as well here, refer to  other commands' help messages.
context: #1852, #1872

modules/cli/src/main/scala/scala/cli/commands/fix/FixOptions.scala
```diff
@@ -0,0 +1,22 @@
...
+final case class FixOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
+  @HelpMessage(s"Move all using directives to the ${Constants.projectFileName} file")
+    migrateDirectives: Option[Boolean] = None
```
Gedochao:
```suggestion
  @HelpMessage(s"Move all using directives to the ${Constants.projectFileName} file")
  @Tag(tags.inShortHelp)
  @Tag(tags.experimental)
    migrateDirectives: Option[Boolean] = None
```

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Migrating directives
+It is recommended to keep all `using` directives centralised in one file. Fix command along with the `--migrate-directives` option will migrate all directives of given inputs to the `project.scala` configuration file (if the file does not exist, Scala CLI will create it in the project root directory).
+
+The following command will move all directives inside the current working directory to the `project.scala` file:
+```
+scala-cli fix --migrate-directives .
+```
+
+Please note that this is equivalent to:
```
Gedochao:
```suggestion
Note that this is equivalent to:
```

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
...
+sidebar_position: 20
+---
+
+The fix command can be used to perform fixes on a Scala CLI project.
+
+## Migrating directives
+It is recommended to keep all `using` directives centralised in one file. Fix command along with the `--migrate-directives` option will migrate all directives of given inputs to the `project.scala` configuration file (if the file does not exist, Scala CLI will create it in the project root directory).
+
+The following command will move all directives inside the current working directory to the `project.scala` file:
+```
```
Gedochao, line 21:
````suggestion
```bash ignore
````

File: modules/integration/src/test/scala/scala/cli/integration/FixTests.scala [hunk 20]
```diff
@@ -0,0 +1,124 @@
...
+      val barFileContent     = noCrLf(os.read(root / "files" / "Bar.scala"))
+      val expectedProjectFileContent =
+        """//> using scala "3.2.0"
+          |//> using options "-Xasync"
+          |
+          |
+          |//> using dep "com.lihaoyi::os-lib::0.8.1"
+          |
+          |
+          |""".stripMargin
```
Gedochao: why are there so many extra empty lines in the output?

website/docs/commands/fix.md
Gedochao, line 1: Some example with directives spread between multiple files would be useful.
An example output `project.scala` could then be included, plus there'd be no reason to ignore the `'''bash` snippets in `docs-tests`

website/docs/commands/fix.md
```diff
@@ -0,0 +1,19 @@
...
+scala-cli fix --migrate-directives .
+```
+
+Please note that this is equivalent to:
+```bash ignore
+scala-cli migrate-directives .
```
Gedochao: is this example still relevant?
since we're not adding the `migrate-directives` alias

modules/integration/src/test/scala/scala/cli/integration/FixTests.scala
```diff
@@ -0,0 +1,124 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+class FixTests extends ScalaCliSuite {
```
lwronski, line 5: What happens if I now have directives in different scopes? For example main directives in `Hello.scala` and test directives in `Hello.test.scala`?

We should make sure that we don't accidentally broke someone's Scala CLI configuration.

lwronski: It should be also tested with Scala.Js and Scala native scopes.

Gedochao:
Closing this, we'll make a new attempt soon.
to be tracked under #2104

PR #1630, opened 2022-12-01
Add repositories from build options when validating scala versions

lwronski:
Co-authored-by: Piotr Chabelski <ged.subfan@gmail.com>

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/RunTestsDefault.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/RunTestsDefault.scala
@@ -47,4 +47,21 @@ class RunTestsDefault extends RunTestDefinitions(scalaVersionOpt = None) {
       archLinuxTest()
     }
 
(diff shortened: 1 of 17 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunTestsDefault.scala
```diff
@@ -47,4 +47,21 @@ class RunTestsDefault extends RunTestDefinitions(scalaVersionOpt = None) {
       archLinuxTest()
     }
 
+  test("3.nightly") { // should run code using scala 3 nightly version
+    TestInputs(os.rel / "sample.sc" -> """println("Hello World")""").fromRoot {
```
Gedochao, line 51: NIT it'd be great to verify the used version here as well (similar to `2.12.nightly` and `2.13.nightly` tests), but I figure we don't have an option to depend on a nightly version of the compiler dynamically. We can fix it some other time, I guess.

Gedochao (approved):
LGTM

PR #1643, opened 2022-12-05
Fix - merge extra repos with resolve.repositories

lwronski:
Fix fetching external binary in M1 - [fail](https://github.com/VirtusLab/scala-cli/actions/runs/3619015290/jobs/6099621848#step:6:1675).

`withRepositories` only sets repositories from `params.extraRepos` but we should merge the default repositories from courier with `extraRepos`, so I'll use `addRepositories`.


tgodzik (approved):
LGTM!

PR #1645, opened 2022-12-05
Force to re-compiling native code when user select different main class in interactive

### LINKED ISSUE #1631: Native target ignores users choice of main after the first run
**Version(s)**
```
scala-cli --version
Scala CLI version: 0.1.18
```

**Describe the bug**
When the source has multiple main classes and the project is ran with scala-native, the first main class the user picks is always ran no matter if the user chooses another the next time.

**To Reproduce**

Setup
```bash
touch someProject.scala
echo -e "@main def bar=\n   println(\"bar\")" >> someProject.scala
echo -e "@main def foo=\n   println(\"foo\")" >> someProject.scala
```

Native
```bash
scala-cli . --interactive --native
Compiling project (Scala 3.2.1, Scala Native)
Compiled project (Scala 3.2.1, Scala Native)
Found several main classes. Which would you like to run?
[0] bar
[1] foo
0
[info] Linking (1257 ms)
[info] Discovered 668 classes and 3700 methods
[info] Optimizing (debug mode) (755 ms)
[info] Generating intermediate code (860 ms)
[info] Produced 16 files
[info] Compiling to native code (3999 ms)
[info] Total (7076 ms)
bar
scala-cli . --interactive --native
Found several main classes. Which would you like to run?
[0] bar
[1] foo
1
bar
```

⬆️ As we see here, the second time we ran it and chose another main, it still ran the main from the first invocation.

For reference both JS and JVM work correctly

JVM
```bash
scala-cli . --interactive
Compiling project (Scala 3.2.1, JVM)
Compiled project (Scala 3.2.1, JVM)
Found several main classes. Which would you like to run?
[0] bar
[1] foo
0
bar
scala-cli . --interactive
Found several main classes. Which would you like to run?
[0] bar
[1] foo
1
foo
```

JS
```bash
scala-cli . --interactive --js    
Compiling project (Scala 3.2.1, Scala.js)
Compiled project (Scala 3.2.1, Scala.js)
Found several main classes. Which would you like to run?
[0] bar
[1] foo
0
bar
scala-cli . --interactive --js
Found several main classes. Which would you like to run?
[0] bar
[1] foo
1
foo
```

**Expected behaviour**
The second time we run and pick `foo`, `foo` is printed as that is the main function the user picked.

lwronski:
Fixes #1631

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -982,7 +982,7 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
         Nil
     val pythonCliOptions = pythonLdFlags.flatMap(f => Seq("--linking-option", f)).toList
 
-    val allCliOptions = pythonCliOptions ++ cliOptions
+    val allCliOptions = pythonCliOptions ++ cliOptions ++ Seq(mainClass)
 
     val nativeWorkDir = build.inputs.nativeWorkDir
     os.makeDir.all(nativeWorkDir)


alexarchambault:
I'm trying a slightly different approach in https://github.com/VirtusLab/scala-cli/pull/1701. It should force future uses of interactive options to take into account CLI options equivalent to the interactively picked option.

PR #1649, opened 2022-12-06
Bumped up com.lihaoyi::os-lib version to 0.9.0

pingu1m:
I tried to run the example but it did not work using the latest install and default configuration.  The dependency version was increased to run the example successfully.


lwronski (approved):
@pingu1m Thanks for your contribution, LGTM

PR #1660, opened 2022-12-07
Don't clear compilation output dir

### LINKED ISSUE #1484: `scala-cli compile -d` always clears the directory
**Version(s)**
0.1.16

**Describe the bug**
for backwards compatibility, using the `-d` flag should not clear the directory before adding classes.

Instead, each compilation to the directory clears it, this makes it impossible to incrementally add classes to a directory.

**To Reproduce**

In this example, we simulate three compilation runs of "artefacts" that output to the same directory, the idea is that at runtime the `lib` classes should still exist, but they do not in this case:

```scala
// lib.scala
object lib:
  def foo = 23
```
```scala
// utils.scala
object utils:
  def bar = lib.foo
```
```scala
// app.scala
@main def app = println(utils.bar)
```

```
> mkdir out
> scala compile -d out lib.scala
> scala compile -d out -cp out utils.scala
> scala -cp out app.scala
Compiling project (Scala 3.2.0, JVM)
Compiled project (Scala 3.2.0, JVM)
Exception in thread "main" java.lang.NoClassDefFoundError: lib$
	at utils$.bar(Macros.scala:5)
	at App$package$.app(App.scala:3)
	at app.main(App.scala:3)
Caused by: java.lang.ClassNotFoundException: lib$
	at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 3 more
```

**Expected behaviour**
no crash - program prints 23

clutroth:


Gedochao (approved):

wleczny (approved):
LGTM

bishabosha:
Awesome !

PR #1661, opened 2022-12-07
fix if else in mill.bat

MFujarewicz:
When running mill.bat under windows warning shows up:
'else' is not recognized as an internal or external command, operable program or batch file.

This is because else is in new line, not on the same as ')' ending if
https://github.com/VirtusLab/scala-cli/blob/main/mill.bat#L42
https://stackoverflow.com/questions/12365374/else-is-not-recognized-as-an-internal-or-external-command-operable-program-or

Because of that else is ignored end next if is entered

This causes to in .config/mill-version to take precedence over mill-version in outermost directory, opposite to the intent of the code


lwronski (approved):
@MFujarewicz Thanks for your first contribution. LGTM

PR #1679, opened 2022-12-13
Add warnings for `-save` & `-nosave` legacy `scala` runner options instead of failing

### LINKED ISSUE #1670: Option `-save` is not supported

The current `scala` runner (based on Scala 2.13) does support a [`-save`](https://github.com/scala/scala/blob/7ab7c953d173fb82673badf091b48222874b0bc0/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala#L98) option.

The experimental `scala` runner (based on Scala CLI) does not recognize that option:

~~~ text
$ scala -save test.scala 
Unrecognized argument: -save

  /home/julien/.local/share/coursier/bin/.scala.aux --help
~~~

For the sake of compatibility with the current runner, it would be great to recognize the `-save` option. As far as I understand, Scala CLI now always behaves like the current `scala -save`, so the option is not necessary anymore. Therefore, it could simply print a warning like “Unnecessary option: -save. This option is always enabled.”

Gedochao:
Fixes #1670 

for `scala-cli`:
```bash
▶ scala-cli -e 'println("Hello")' -save -nosave
Deprecated option '-save' is ignored.
The compiled project files will be saved in the '.scala-build' directory in the project root folder.
If an actual jar file is necessary to be saved, run the 'package' sub-command after this one:
  scala-cli package --library -e println("Hello")
Deprecated option '-nosave' is ignored.
A jar file is not saved unless the 'package' sub-command is called.
Compiling project (Scala 3.2.1, JVM)
Compiled project (Scala 3.2.1, JVM)
Hello
```
for `scala` (SIP mode):
```bash
▶ scala -e 'println("Hello")' -save -nosave
Deprecated option '-save' is ignored.
The compiled project files will be saved in the '.scala-build' directory in the project root folder.
If an actual jar file is necessary to be saved, run the '--power package' sub-command after this one:
  scala --power package --library -e println("Hello")
Deprecated option '-nosave' is ignored.
A jar file is not saved unless the '--power package' sub-command is called.
Compiling project (Scala 3.2.1, JVM)
Compiled project (Scala 3.2.1, JVM)
Hello
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
@@ -52,7 +52,7 @@ class Default(
     sys.exit(0)
   }
 
-  override def runCommand(options: DefaultOptions, args: RemainingArgs, logger: Logger): Unit = {
+  override def runCommand(options: DefaultOptions, args: RemainingArgs, logger: Logger): Unit =
     if options.version then println(Version.versionInfo)
     else
       {
@@ -62,9 +62,8 @@ class Default(
           options.shared.snippet.executeMarkdown.nonEmpty ||
           (options.shared.extraJarsAndClassPath.nonEmpty && options.sharedRun.mainClass.mainClass.nonEmpty)
         if shouldDefaultToRun then RunOptions.parser else ReplOptions.parser
-      }.parse(rawArgs) match
+      }.parse(options.legacyScala.filterNonDeprecatedArgs(rawArgs, progName, logger)) match
         case Left(e)                              => error(e)
         case Right((replOptions: ReplOptions, _)) => Repl.runCommand(replOptions, args, logger)
         case Right((runOptions: RunOptions, _))   => Run.runCommand(runOptions, args, logger)
-  }
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
@@ -1,0 +1,67 @@
+package scala.cli.commands.default
+
+import caseapp.*
+import caseapp.core.Indexed
+
+import scala.build.Logger
+import scala.cli.ScalaCli
+import scala.cli.commands.default.LegacyScalaOptions.*
+import scala.cli.commands.package0.Package
+import scala.cli.commands.tags
+
+/** Options covering backwards compatibility with the old scala runner.
+  */
+// format: off
+case class LegacyScalaOptions(
+  @Group("Scala")
+  @HelpMessage(s"Ignored legacy option. Deprecated equivalent of running a subsequent `$PowerString${Package.name}` command.")
+  @Tag(tags.must)
+  @Name("-save")
+    save: Option[Indexed[Boolean]] = None,
+  @Group("Scala")
+  @HelpMessage("Ignored legacy option. Deprecated override canceling the `-nosave` option.")
+  @Tag(tags.must)
+  @Name("-nosave")
+    nosave: Option[Indexed[Boolean]] = None,
+) {
+// format: on
+
+  extension [T](indexedOption: Option[Indexed[T]]) {
+    private def findArg(args: Array[String]): Option[String] =
+      indexedOption.flatMap(io => args.lift(io.index))
+  }
+
+  def filterNonDeprecatedArgs(
+    args: Array[String],
+    progName: String,
+    logger: Logger
+  ): Array[String] = {
+    val saveOptionString   = save.findArg(args)
+    val noSaveOptionString = nosave.findArg(args)
+    val deprecatedArgs     = Seq(saveOptionString, noSaveOptionString).flatten
+    val filteredArgs       = args.filterNot(deprecatedArgs.contains)
+    val filteredArgsString = filteredArgs.mkString(" ")
+    saveOptionString.foreach { s =>
+      logger.message(
+        s"""Deprecated option '$s' is ignored.
+           |The compiled project files will be saved in the '.scala-build' directory in the project root folder.
+           |If you need to produce an actual jar file, run the '$PowerString${Package.name}' sub-command as follows:
+           |  ${Console.BOLD}$progName $PowerString${Package.name} --library $filteredArgsString${Console.RESET}""".stripMargin
+      )
+    }
+    noSaveOptionString.foreach { ns =>
+      logger.message(
+        s"""Deprecated option '$ns' is ignored.
+           |A jar file is not saved unless the '$PowerString${Package.name}' sub-command is called.""".stripMargin
+      )
+    }
+    filteredArgs
+  }
+}
+object LegacyScalaOptions {
+  implicit lazy val parser: Parser[LegacyScalaOptions] = Parser.derive
+  implicit lazy val help: Help[LegacyScalaOptions]     = Help.derive
+
+  private[default] lazy val PowerString =
+    if ScalaCli.allowRestrictedFeatures then "" else "--power "
+}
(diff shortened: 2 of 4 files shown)


modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.build.Logger
+import scala.cli.commands.package0.Package
+import scala.cli.commands.tags
+
+/** Options covering backwards compatibility with the old scala runner.
+  */
+// format: off
+case class LegacyScalaOptions(
+  @Group("Scala")
+  @HelpMessage("Ignored legacy option. Deprecated equivalent of running a subsequent `package` command.")
```
tgodzik: Is package by default in the scala runner actually?

lwronski: By default package command create a lightweight launcher JAR, like the "bootstrap" JAR files [generated by coursier](https://get-coursier.io/docs/cli-bootstrap#bootstraps). So this is different than what the save option generates.

I think scala runner generates Library JARs or fat JARs, @Gedochao could you check that?

modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
```diff
@@ -62,9 +62,8 @@ class Default(
           options.shared.snippet.executeMarkdown.nonEmpty ||
           (options.shared.extraJarsAndClassPath.nonEmpty && options.sharedRun.mainClass.mainClass.nonEmpty)
         if shouldDefaultToRun then RunOptions.parser else ReplOptions.parser
-      }.parse(rawArgs) match
+      }.parse(options.legacyScala.filterNonDeprecatedArgs(rawArgs, progName, logger).toSeq) match
```
lwronski:
```suggestion
      }.parse(options.legacyScala.filterNonDeprecatedArgs(rawArgs, progName, logger)) match
```

modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
julienrf: Good catch @tgodzik, `package` is not part of the `scala` runner. For now, I suggest changing the help message to not mention it.

modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
```diff
@@ -0,0 +1,62 @@
...
+    logger: Logger
+  ): Array[String] = {
+    val saveOptionString   = save.findArg(args)
+    val noSaveOptionString = nosave.findArg(args)
+    val deprecatedArgs     = Seq(saveOptionString, noSaveOptionString).flatten
+    val filteredArgs       = args.filterNot(deprecatedArgs.contains)
+    val filteredArgsString = filteredArgs.mkString(" ")
+    saveOptionString.foreach { s =>
+      logger.message(
+        s"""Deprecated option '$s' ignored.
```
julienrf:
```suggestion
        s"""Deprecated option '$s' is ignored.
```

modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
```diff
@@ -0,0 +1,62 @@
...
+           |The compiled project files will be saved in the '.scala-build' directory in the project root folder.
+           |If an actual jar file is necessary to be saved, run the 'package' command after this one:
+           |  ${Console.BOLD}$progName ${Package.name} $filteredArgsString${Console.RESET}""".stripMargin
+      )
+    }
+    noSaveOptionString.foreach { ns =>
+      logger.message(
+        s"""Deprecated option '$ns' ignored.
```
julienrf:
```suggestion
        s"""Deprecated option '$ns' is ignored.
```

modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
Gedochao: @tgodzik ugh, y're right, `package` isn't included in the `scala` runner... are we ok with mentioning it with the `--power` command here?
I don't think we can deliver creating the jar here without it.

@lwronski ha, this is tricky. It seems we'd have to fix the `package` sub-command a bit to be compatible, as we don't fill all the required things in the manifest. I wasn't able to run a jar coming from `scala-cli package` as things are. And yes, we'd need to pass `--library` or `--assembly` flags anyway, as the bootstrap wouldn't do.

And even then, this functionality is actually slightly borked in the old `scala` runner already, as even `dotty`'s `scala` runner `refers to the wrong compiler version in the manifest...  So I have rising doubts if it's still useful to anyone

Gedochao: @julienrf thoughts on mentioning `package` with the `--power` flag in the warning? Would it be acceptable when run as `scala`? That's the only sub-command capable of generating jars, even though it's not been accepted in the SIP

julienrf: Yes, I think that’s a good idea. But `--power` is not recognized (0.1.18). Is it still under implementation?

Gedochao: it should be recognised in `0.1.18`.
Did you pass it as a launcher option?
It has to be passed before the sub-command to work.
```bash
▶ scala --power package .  
```

Gedochao: anyhow, this will require some fixes in the `package` command, as we don't set the `Scala-Compiler-Version` and `Class-Path` fields in the jar manifest... (plus the `Main-Class` field is only available to be added for fat jars).
I'll create a separate issue for those and do it in a separate PR, I guess

julienrf: Ah, right, I didn’t pass `--power` _before_ `package`. Now it is recognized. But the option `--power` is not mentioned when I run `scala --help-full`. Is this intended?

Gedochao: ...yeah, I think it should be mentioned...
All of the launcher options (like `--cli-version` and `--power`) seem to be skipped from help, since they're passed before a subcommand is invoked (including the default sub-command).
That's another bug, it seems 😅

Gedochao: the side-issues were extracted to #1698 & #1699.
The rest should already be addressed.

modules/cli/src/main/scala/scala/cli/commands/default/LegacyScalaOptions.scala
```diff
@@ -0,0 +1,67 @@
...
+    val saveOptionString   = save.findArg(args)
+    val noSaveOptionString = nosave.findArg(args)
+    val deprecatedArgs     = Seq(saveOptionString, noSaveOptionString).flatten
+    val filteredArgs       = args.filterNot(deprecatedArgs.contains)
+    val filteredArgsString = filteredArgs.mkString(" ")
+    saveOptionString.foreach { s =>
+      logger.message(
+        s"""Deprecated option '$s' is ignored.
+           |The compiled project files will be saved in the '.scala-build' directory in the project root folder.
+           |If an actual jar file is necessary to be saved, run the '${PowerString}package' sub-command after this one:
```
julienrf:
```suggestion
           |If you need to produce an actual jar file, run the '${PowerString}${Package.name}' sub-command as follows:
```

julienrf (approved):

lwronski (approved):
LGTM

Gedochao:
relevant docs will be added as part of #1721

PR #1681, opened 2022-12-13
Fix BSP `workspace/reload` to always pick up dependencies from `using` directives

### LINKED ISSUE #1673: Fix `workspace/reload` not importing dependencies without re-compilation
It actually doesn't do nothing, we should be supporting `workspace/reload`.

https://github.com/VirtusLab/scala-cli/blob/adca9c2555d15640b6effdcf6b6ec204fada63ce/modules/build/src/main/scala/scala/build/bsp/BuildServerProxy.scala#L80
https://github.com/VirtusLab/scala-cli/blob/01960bec2cd54f2cd7664a3ebb8a0190481e4747/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala#L557

After testing it a bit just now, it technically does work, but in this use case it's a bit wonky.

https://user-images.githubusercontent.com/18601388/206666017-93d5e42c-479c-4598-8cd5-e24bc7b533a6.mov

after adding a new directive/dependency in the code, `workspace/reload` doesn't seem to pick it up until the first re-compilation is triggered. It doesn't matter if it's from the command line, or if it's a re-run from IntelliJ and the re-run should go through BSP).

Not sure where's the bug, but it's worth investigating on our side. I'll create a ticket.

_Originally posted by @Gedochao in https://github.com/VirtusLab/scala-cli/discussions/1659#discussioncomment-4350316_

Gedochao:
Fixes #1673 

https://user-images.githubusercontent.com/18601388/207319736-534f2d8a-862d-4c0a-8c8a-e52d95ac03e6.mov

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -390,12 +390,7 @@ final class BspImpl(
       build(bloopSession0, actualLocalClient, notifyChanges = true, reloadableOptions),
       ()
     )
-    lazy val bloopSession0: BloopSession = new BloopSession(
-      inputs,
-      remoteServer,
-      bspServer,
-      watcher
-    )
+    lazy val bloopSession0: BloopSession = BloopSession(inputs, remoteServer, bspServer, watcher)
 
     bloopSession0.registerWatchInputs()
     bspServer.newInputs(inputs)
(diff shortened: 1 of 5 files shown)


modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -578,8 +573,11 @@ final class BspImpl(
           }
         }
         val newInputs      = value(argsToInputs(ideInputs.args))
+        val newHash        = newInputs.sourceHash()
         val previousInputs = currentBloopSession.inputs
-        if (newInputs == previousInputs) CompletableFuture.completedFuture(new Object)
+        val previousHash   = currentBloopSession.inputsHash
+        if newInputs == previousInputs && newHash == previousHash then
```
lwronski, line 579:
```suggestion
        if newHash == previousHash then
```

It should be sufficient, as the sourceHash method calculates a hash that includes [the file path](https://github.com/VirtusLab/scala-cli/blob/5264acdea6de5509fc47b291456607c704eb7792/modules/build/src/main/scala/scala/build/input/Inputs.scala#L104).

Gedochao: I think it may not be sufficient in some edge cases, as Inputs has some fields beyond the actual elements (like i.e. if Markdown sources are to be included). I'd leave it as it is, just to be safe, the extra comparison is a cheap enough operation

lwronski (approved):
LGTM

julienrf:
Amazing, thank you @Gedochao and @lwronski!

PR #1693, opened 2022-12-15
Use snapshot repo to download stubs

lwronski:


tgodzik (approved):
Why was there `.root` used actually. What does the method do on the repo? Any diea?

lwronski:
> Why was there `.root` used actually. What does the method do on the repo? Any diea?

Some time ago I changed the argument type in the `artifact` methods from:
```
extraRepositories: Seq[String] -> extraRepositories: Seq[Repository]
```

so now we don't need to run `root` method to get plain repository name.

Gedochao (approved):

alexarchambault (approved):

PR #1694, opened 2022-12-15
Ignore `jsoniter` updates for JDK 8

lwronski:


Gedochao:
Hey, there's actually a separate sequence of upgrades for JDK8-compatible `jsoniter`.
The most recent one is [2.13.5.2](https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.13.5.2)
Not sure if we can force `scala-steward` to only look at those though

tgodzik (approved):

PR #1695, opened 2022-12-15
Adjust SIP help output & ensure `ScalaSipTests` are run on Windows

### LINKED ISSUE #1664: --help incorrectly says that the tool lets you package your Scala code
**Version(s)**

0.1.18 installed via `cs install scala-experimental`

**Describe the bug**

`scala --help` says:

> Scala code runner is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code.

However, the “packaging” features are not part of the `scala` tool (only `scala-cli`).

**Expected behaviour**

The message should be adapted to the features supported by `scala`:

> Scala code runner is a command-line tool to interact with the Scala language. It lets you compile, run, and test your Scala code.

Gedochao:
Fixes #1664

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
@@ -6,32 +6,40 @@ import scala.util.Properties
 
 class SipScalaTests extends ScalaCliSuite {
 
-  def noDirectoriesCommandTest(binaryName: String): Unit =
+  implicit class BinaryNameOps(binaryName: String) {
+    def prepareBinary(root: os.Path): os.Path = {
+      val cliPath    = os.Path(TestUtil.cliPath, os.pwd)
+      val ext        = if (Properties.isWin) ".exe" else ""
+      val newCliPath = root / s"$binaryName$ext"
+      os.copy(cliPath, newCliPath)
+      if (!Properties.isWin) os.perms.set(newCliPath, "rwxr-xr-x")
+      newCliPath
+    }
+
+    def isSip: Boolean = binaryName == "scala" || binaryName.endsWith("sip")
+  }
+
+  def testDirectoriesCommand(binaryName: String): Unit =
     TestInputs.empty.fromRoot { root =>
-      val cliPath = os.Path(TestUtil.cliPath, os.pwd)
+      val binary = binaryName.prepareBinary(root)
 
-      os.proc(cliPath, "directories").call(cwd = root)
-      os.proc(cliPath, "compile", "--help").call(cwd = root)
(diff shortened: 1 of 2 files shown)


modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -6,32 +6,40 @@ import scala.util.Properties
...
-  def noDirectoriesCommandTest(binaryName: String): Unit =
-    TestInputs.empty.fromRoot { root =>
-      val cliPath = os.Path(TestUtil.cliPath, os.pwd)
+  implicit class BinaryNameOps(binaryName: String) {
+    def prepareBinary(root: os.Path): os.Path = {
+      val cliPath    = os.Path(TestUtil.cliPath, os.pwd)
+      val ext        = if (Properties.isWin) ".exe" else ""
+      val newCliPath = root / s"$binaryName$ext"
+      os.copy(cliPath, newCliPath)
+      os.perms.set(newCliPath, "rwxr-xr-x")
```
lwronski: Should we set `perms` on windows?

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -122,6 +122,20 @@ class SipScalaTests extends ScalaCliSuite {
       }
     }
 
+  def checkHelp(binaryName: String): Unit = TestInputs.empty.fromRoot { root =>
+    val cliPath = os.Path(TestUtil.cliPath, os.pwd)
+    val ext = if (Properties.isWin) ".exe" else ""
+    val newCliPath = root / s"$binaryName$ext"
+    os.copy(cliPath, newCliPath)
+
+    for { helpOption <- Seq("help", "-help", "--help") } {
```
lwronski, line 112: Should be sufficient if this test will test only `help` command. For `-help` and `--help` it will be the same effect.

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: Yeah, I guess there's no need to do that, although it doesn't seem to have been affecting the tests, really...
Anyhow, should be fixed now.

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: I mean, we do already test for `version` along with `-version` and `--version` in `SipScalaTests`...
And `HelpTests` similarly already test for `help`, `-help` and `--help`...
So I see no reason to skip over that here, either (even if it's a little paranoid)

lwronski (approved):

PR #1703, opened 2022-12-17
Build native launcher for linux aarch64

lwronski:


Gedochao (approved):

.github/workflows/ci.yml
```diff
@@ -187,6 +187,58 @@ jobs:
         SCALA_CLI_IT_GROUP: 3
         SCALA_CLI_SODIUM_JNI_ALLOW: false
 
+  generate-linux-arm64-native-launcher:
```
alexarchambault, line 191: Feels this was [copied from coursier](https://github.com/coursier/coursier/blob/44e1a024897fb1046303f705e0e4143d808e1adc/.github/workflows/ci.yml#L263-L326), so not code you authored… Could you mention it?

alexarchambault (commented):
Did you try manually running the integration tests on a Linux/ARM64 machine? Just so that we have an idea if it works ok in most cases, or if it's just broken there?

lwronski:
> Did you try manually running the integration tests on a Linux/ARM64 machine? Just so that we have an idea if it works ok in most cases, or if it's just broken there?

I used generated launcher and run some tests such as(`RunTestsDefault`, `BspTestsDefault`, `CompileTestsDefault`) in Ubuntu with arm64 architecture and it works.

alexarchambault (approved):

mkurz:
Doesn't work for me, please see #1714

mkurz:
This binary does not work for all aarch64 kernels out there. Most of the aarch64 Linux kernel will have a page size of 4k, which this binary is compatible with, but if a kernel runs another page size like 64k or, like in my case, 16k, this binary will not work.
See https://github.com/coursier/coursier/issues/2636#issuecomment-1363491391

mkurz:
So this QEMU thing is quite slow, seems like James Ward came up with a faster approach: https://twitter.com/_JamesWard/status/1613201502469750785 (I didn't read it, just posting so you are aware and might like it)

lwronski:
> So this QEMU thing is quite slow, seems like James Ward came up with a faster approach: https://twitter.com/_JamesWard/status/1613201502469750785 (I didn't read it, just posting so you are aware and might like it)

Thanks for the suggestions. I did a slightly different approach, used an actual runner that scala-cli `builds` macos-m1 launchers and there I run build the `aarch64` linux launcher using docker. I reduced the build time to 20 minutes.

PR #1718, opened 2022-12-21
Add test-only cookbook

lwronski:


website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+- start with `test` - `test*`
+- end with `test` - `*test`
+- contains `test` -  `*test*`
+
+:::note
+The `test-only` is supported for every test frameworks running in Scala CLI.
+:::
+
+For example, passing `tests.only*` to the `--test-only` runs only test suite which start with `tests.only`:
```
tgodzik:
```suggestion
For example, passing `tests.only*` to the `--test-only` runs only test suites which start with `tests.only`:
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+<!-- Expected:
+tests.only.BarTests:
++ bar
+-->
+
+## Running a single test case by name
+
+:::note
+It's important that the test frameworks may have their own specific API for specifying the test cases to run.
```
tgodzik:
```suggestion
Test frameworks may have their own specific API for specifying the test cases to run aside from the test-only input.
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+## Filter the test suites
+
+The `test-only` parameter in Scala CLI supports using glob patterns to filter test suites to run. A glob pattern is a
+string which contains asterisk `*` characters to match a set of test suites.
+
+There are three examples of glob patterns how to filter test suites with the `test-only`:
+
+- start with `test` - `test*`
+- end with `test` - `*test`
+- contains `test` -  `*test*`
```
tgodzik, line 20: would the forth not be `.` I think . replaces a single char instead of * replacing 0+ ?

tgodzik (changes requested):
Thanks for working on this cookbook!

website/docs/cookbooks/test-only.md
lwronski:
```
➜  scala-demo cat Test.scala 
//> using lib "org.scalameta::munit::0.7.29"

class Tests extends munit.FunSuite {
  test("foo") {
    assert(2 + 3 == 5)
  }
}
➜  scala-demo scala-cli test .  -- '*foo' 

Tests:
  + foo 0.039s
➜  scala-demo scala-cli test .  -- '*fo.' 

Tests:
```

In my tests `.` isn't matched to any character.

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Filter the test suites to run
+sidebar_position: 11
+---
+
+The `test-only` parameter it is a useful option for when you only want to run a specific subset of tests.
```
Gedochao:
```suggestion
The `test-only` parameter is a useful option for when you only want to run a specific subset of tests.
```

Gedochao: I'd even be tempted to always call `test-only` an option here, since we only use it from the command line (and not through a `using` directive).
In which case:
```suggestion
The `--test-only` option is useful for when you only want to run a specific subset of tests.
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+In this cookbook, we will walk through how to use the `test-only` parameter with the `test` sub-command in Scala CLI,
```
Gedochao: NIT as per the earlier comment
```suggestion
In this cookbook, we will walk through how to use the `--test-only` option with the `test` sub-command in Scala CLI,
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+specifically when using the `munit` and `utest` test frameworks.
+
+## Filter the test suites
+
+The `test-only` parameter in Scala CLI supports using glob patterns to filter test suites to run. A glob pattern is a
```
Gedochao: NIT as per the earlier comment
```suggestion
The `--test-only` option in Scala CLI supports using glob patterns to filter test suites to run. A glob pattern is a
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+string which contains asterisk `*` characters to match a set of test suites.
+
+There are three examples of glob patterns how to filter test suites with the `test-only`:
```
Gedochao:
```suggestion
Here are three examples of glob patterns for how to filter test suites with `--test-only`:
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+- start with `test` - `test*`
+- end with `test` - `*test`
+- contains `test` -  `*test*`
+
+:::note
+The `test-only` is supported for every test frameworks running in Scala CLI.
```
Gedochao:
```suggestion
The `--test-only` option is supported for every test framework running with Scala CLI.
```

website/docs/cookbooks/test-only.md
Gedochao:
```suggestion
For example, passing `tests.only*` to the `--test-only` option runs only the test suites which start with `tests.only`:
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Running a single test case by name
+
+:::note
+It's important that the test frameworks may have their own specific API for specifying the test cases to run.
+:::
+
+### Munit
+
+To run a specific test cases inside the test suite pass `*test-name*` as an argument to `scala-cli`:
```
Gedochao:
```suggestion
To run a specific test case inside a test suite pass `*test-name*` as an argument to `scala-cli`:
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+#   + foo-again 0.001s
+```
+
+<!-- Expected:
+tests.only.Tests:
++ foo
++ foo-again
+-->
+
+This will run only the test which contains the name `foo`, and any other tests in your test suite will be skipped.
```
Gedochao:
```suggestion
This will only run the test which contains the name `foo`; any other tests in your test suite will be skipped.
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Utest
+
+Unfortunately, the `utest` test framework doesn't support using glob pattern `*` to filter the test cases to run. In
+order to run specific test case you will need to specify the exact names of the test you want to run.
```
Gedochao:
```suggestion
order to run aspecific test case you will need to specify the exact name of the test you want to run.
```

website/docs/cookbooks/test-only.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+<!-- Expected:
+Running Tests MyTests.foo
+Hello from tests
++ MyTests.foo
+Tests: 1, Passed: 1, Failed: 0
+-->
+
+This will run only the test with the name of `MyTests.foo`.
```
Gedochao:
```suggestion
This will run only the test case with the name `MyTests.foo`.
```

website/docs/cookbooks/test-only.md
tgodzik: Huh, so `scala-cli test .  -- 'MyTests.foo'` it's the only option to use the dot ?

Gedochao (approved):

website/docs/cookbooks/test-only.md
lwronski: Yes, it seems that `.` isn't treated as a special character.

tgodzik (approved):
LGTM!

PR #1724, opened 2022-12-23
Downgrade `coursier` M1 CLI back to 2.1.0-RC3-1

Gedochao:


Gedochao:
https://github.com/VirtusLab/coursier-m1/releases/tag/v2.1.0-RC4
no need to do this after all, since `v2.1.0-RC4` is up now

PR #1725, opened 2022-12-23
Include launcher options in the help for the default and `help` sub-commands

### LINKED ISSUE #1698: Launcher options (`--cli-version` and `--power`) should be mentioned in help
All of the launcher options (like `--cli-version` and `--power`) seem to be skipped from help, since they're passed before a subcommand is invoked (including the default sub-command).

_Originally posted by @Gedochao in https://github.com/VirtusLab/scala-cli/pull/1679#discussion_r1050511768_

Gedochao:
Fixes #1698 

Note: this adds the launcher options help to the default and `help` sub-commands **only**.
They probably should also smh be included in other sub-commands' help as well, but that'll be addressed separately.

```bash
▶ scala-cli --help
Usage: /Users/pchabelski/IdeaProjects/scala-cli/out/cli/base-image/nativeImage.dest/scala-cli <COMMAND>
Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test and package your Scala code.

Main commands:
  clean                  Clean the workspace
  compile                Compile Scala code
  dependency-update      Update dependencies in project
  doc                    Generate Scaladoc documentation
  fmt, format, scalafmt  Format Scala code
  repl, console          Fire-up a Scala REPL
  package                Compile and package Scala code
  publish
  publish local
  publish setup
  run                    Compile and run Scala code.
  test                   Compile and test Scala code

Miscellaneous commands:
  version  Print version

Other commands:
  export                                        Export current project to sbt or Mill
  help                                          Print help message
  install completions, install-completions      Installs completions into your shell
  github secret create, gh secret create
  github secret list, gh secret list
  setup-ide                                     Generate a BSP file that you can import into your IDE
  shebang                                       Like `run`, but more handy from shebang scripts
  uninstall                                     Uninstall scala-cli - only works when installed by the installation script
  uninstall completions, uninstall-completions  Uninstalls completions from your shell
  update                                        Update scala-cli - only works when installed by the installation script

Doctor commands:
  doctor  Print details about this application

See 'scala-cli <command> --help' to read about a specific subcommand. To see full help run 'scala-cli <command> --help-full'.

To use launcher options, specify them before any other argument.
For example, to run another Scala CLI version, specify it with the '--cli-version' launcher option:
  scala-cli --cli-version <version> args

Launcher options:
  --cli-version nightly|version  Set the Scala CLI version
  --power                        When called as 'scala', allow to use power commands too

When no subcommand is passed explicitly, an implicit subcommand is used based on context:
  - if the '--version' option is passed, it prints the 'version' subcommand output, unmodified by any other options
  - if any inputs were passed, it defaults to the 'run' subcommand
  - additionally, when no inputs were passed, it defaults to the 'run' subcommand in the following scenarios:
    - if a snippet was passed with any of the '--execute*' options
    - if a main class was passed with the '--main-class' option alongside an extra '--classpath'
  - otherwise, if no inputs were passed, it defaults to the 'repl' subcommand
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/HelpUtils.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/HelpUtils.scala
@@ -1,0 +1,19 @@
+package scala.cli.commands.util
+
+import caseapp.core.help.{Help, HelpFormat}
+
+object HelpUtils {
+  extension (help: Help[_]) {
+    private def abstractHelp(
+      format: HelpFormat,
+      showHidden: Boolean
+    )(f: (StringBuilder, HelpFormat, Boolean) => Unit): String = {
+      val b = new StringBuilder
+      f(b, format, showHidden)
+      b.result()
+    }
+
+    def optionsHelp(format: HelpFormat, showHidden: Boolean = false): String =
+      abstractHelp(format, showHidden)(help.printOptions)
+  }
+}
--- a/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCliCommands.scala
@@ -79,7 +79,10 @@ class ScalaCliCommands(
 
   override def summaryDesc =
     s"""|See '$baseRunnerName <command> --help' to read about a specific subcommand. To see full help run '$baseRunnerName <command> --help-full'.
-        |To run another $fullRunnerName version, specify it with '--cli-version' before any other argument, like '$baseRunnerName --cli-version <version> args'.""".stripMargin
+        |
+        |To use launcher options, specify them before any other argument.
+        |For example, to run another $fullRunnerName version, specify it with the '--cli-version' launcher option:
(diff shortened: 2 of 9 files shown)


modules/cli/src/main/scala/scala/cli/commands/util/HelpUtils.scala
```diff
@@ -0,0 +1,21 @@
...
+    private def abstractHelp(
+      format: HelpFormat,
+      showHidden: Boolean
+    )(f: (StringBuilder, HelpFormat, Boolean) => Unit): String = {
+      val b = new StringBuilder
+      f(b, format, showHidden)
+      b.result()
+    }
+
+    def usageHelp(format: HelpFormat, showHidden: Boolean = false): String =
```
lwronski: This function doesn't seem to be use.

Gedochao: True, removed it for now

lwronski (approved):
LGTM

PR #1728, opened 2022-12-23
Upgrade scripts to latest coursier

mkurz:
~This is just a reminder for myself.~
~RC5 is not yet released, but if it is, I hope it contains https://github.com/coursier/coursier/pull/2641 so I am also able to build scala-ci on my Apple Silicon hardware running Linux (I can do that already using locally build cs that works, this just makes it official).~

### DIFF
--- a/.github/scripts/build-linux-aarch64.sh
+++ b/.github/scripts/build-linux-aarch64.sh
@@ -6,7 +6,7 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
 
 mkdir -p artifacts
 mkdir -p utils
-cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.0.16/cs-aarch64-pc-linux)" utils/cs
+cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.1.4/cs-aarch64-pc-linux)" utils/cs
 chmod +x utils/cs
 
 cp "$DIR/build-linux-aarch64-from-docker.sh" utils/
--- a/.github/scripts/build-linux-aarch64.sh
+++ b/.github/scripts/build-linux-aarch64.sh
@@ -6,7 +6,7 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
 
 mkdir -p artifacts
 mkdir -p utils
-cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.1.4/cs-aarch64-pc-linux)" utils/cs
+cp "$(cs get https://github.com/VirtusLab/coursier-m1/releases/download/v2.1.4/cs-aarch64-pc-linux)" utils/cs
 chmod +x utils/cs
 
 cp "$DIR/build-linux-aarch64-from-docker.sh" utils/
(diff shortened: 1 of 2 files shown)


mill
```diff
@@ -2,7 +2,7 @@
 
 # Adapted from
 
-coursier_version="2.1.0-M6-53-gb4f448130"
+coursier_version="2.1.0-RC5"
```
mkurz: Not sure about this script? Is that an offical mill / millw script actually (it does not look like it is) Or just some hacked together custom script for scala-ci?

lwronski: It's a kind of customer script specific for `scala-cli`. You should also update courier version in this script. It's fine.

lwronski:
Thanks for the update, it looks like this bug appears also in `scala-cli` - https://github.com/coursier/coursier/issues/2624.

lwronski:
> Thanks for the update, it looks like this bug appears also in `scala-cli` - [coursier/coursier#2624](https://github.com/coursier/coursier/issues/2624).

Let's check if it works with static launcher of `cs`.

.github/workflows/ci.yml
```diff
@@ -210,7 +210,7 @@ jobs:
           setup: |
             mkdir -p artifacts
             mkdir -p utils
-            cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.0.16/cs-aarch64-pc-linux)" utils/cs
+            cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.1.0-RC5/cs-aarch64-pc-linux)" utils/cs
```
mkurz: @lwronski The url here is wrong. See https://github.com/coursier/coursier/pull/2658, the aarch64 files are hosted here now: https://github.com/VirtusLab/coursier-m1/releases.
However that release page does not have RC5 binaries yet...

Also in this line here the binary gets downladed, but the releases do not contain plain binaries anymore, but gz files...

mkurz:
@lwronski I rebased this long open PR, can you take a look again?
Thanks!

.github/scripts/build-linux-aarch64.sh
```diff
@@ -6,7 +6,7 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
 
 mkdir -p artifacts
 mkdir -p utils
-cp "$(cs get https://github.com/coursier/coursier/releases/download/v2.0.16/cs-aarch64-pc-linux)" utils/cs
+cp "$(cs get https://github.com/VirtusLab/coursier-m1/releases/download/v2.1.4/cs-aarch64-pc-linux)" utils/cs
```
mkurz, line 9: I changed the url here
* from coursier/coursier (https://github.com/coursier/coursier/releases/)
* to VirtusLab/coursier-m1 (https://github.com/VirtusLab/coursier-m1/releases)

is that correct?

lwronski: @mkurz Thank you for updating it. Yes, that is correct, as Coursier for M1 is built in the `coursier-m1` repository. When the CI passes, I will merge this PR.

mkurz: Thanks, CI is green.

lwronski (approved):

PR #1730, opened 2022-12-23
Allow to circumvent Scala version checks

alexarchambault:
Just in case. Having issues with it with locally published Scala versions right now, so I'd rather disable this check if I can.

Needs https://github.com/scalacenter/bloop-config/pull/20, https://github.com/scala-cli/bloop-core/pull/129

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
@@ -258,6 +258,8 @@ final case class BuildOptions(
     val svOpt: Option[String] = scalaOptions.scalaVersion match {
       case Some(MaybeScalaVersion(None)) =>
         None
+      case Some(MaybeScalaVersion(Some(svInput))) if svInput.endsWith("!") =>
+        Some(svInput.stripSuffix("!"))
       case Some(MaybeScalaVersion(Some(svInput))) =>
         val sv = value {
           svInput match {


tgodzik:
> Just in case. Having issues with it with locally published Scala versions right now, so I'd rather disable this check if I can.
> 
> Needs [scalacenter/bloop-config#20](https://github.com/scalacenter/bloop-config/pull/20), [scala-cli/bloop-core#129](https://github.com/scala-cli/bloop-core/pull/129)

What is the issue you are getting? It should work correctly for locally published jars, last time I checked the correct bridge was used if available.

alexarchambault:
> What is the issue you are getting? It should work correctly for locally published jars, last time I checked the correct bridge was used if available.

Precisely, bridges are fetched by Bloop from the default coursier repositories, so `~/.ivy2/local` and Maven Central. For reproducibility, we use a custom repo in an arbitrary directory in the test added here. Having Bloop accept the bridge JAR alongside the compiler JARs allows to get that bridge JAR via the custom repo, that the build tool (Scala CLI) knows about, while Bloop doesn't.

Note also that for Scala 3, the bridge JAR originates from the same organization as the compiler, and has the same version too. In essence, it feels like it's part of the compiler - so why should it be handled differently than the compiler JARs?

alexarchambault:
Just moved the non-regression test, along with the Bloop-related changes it requires, to https://github.com/VirtusLab/scala-cli/pull/1889

lwronski (commented):
Can you add docs how to avoid Scala version checks? It might be useful for Nix environment.

Gedochao:
Closing this for now, as there's been no activity neither here nor in the prerequisite PRs for a long while.
Feel free to reopen if you continue working on this.
Progress is to be tracked under #2447.

PR #1731, opened 2022-12-23
Fixing minor typos and some wordsmithing.

medale:
Great project! Looking forward to learning more about this and using it. Thanks for the project and excellent documentation. This PR contains just some minor doc changes:

* Fixed minor typos (compiled vs. complied, been vs. ben).
* Fixed link.
* Instances where I think the word should be "than" instead of "then" ([https://www.merriam-webster.com/words-at-play/when-to-use-then-and-than](https://www.merriam-webster.com/words-at-play/when-to-use-then-and-than)).
* Slight wordsmithing.


lwronski (approved):
@medale Thanks for your contribution. LGTM!

PR #1736, opened 2022-12-28
Remove suffix `.aux` from progName when installed by cs

### LINKED ISSUE #1662: Confusing path shown in --usage output
**Version(s)**

0.1.18 installed via `cs install scala-experimental`

**Describe the bug**

When `scala` shows the “usage” instructions it shows a confusing path for the `scala` executable itself:

~~~ text
$ scala test --usage
Usage: /home/julien/.local/share/coursier/bin/.scala.aux test [options]
~~~

**To Reproduce**

- Install `scala` via `cs install scala-experimental`
- Run `scala test --usage`

**Expected behaviour**

~~~ text
$ scala test --usage
Usage: scala test [options]
~~~

lwronski:
Fixes #1662.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -4,6 +4,7 @@ import sun.misc.{Signal, SignalHandler}
 
 import java.io.{ByteArrayOutputStream, File, PrintStream}
 import java.nio.charset.StandardCharsets
+import java.nio.file.Paths
 import java.util.Locale
 
 import scala.build.blooprifle.FailedToStartServerException
@@ -21,7 +22,15 @@ object ScalaCli {
     // the Scala CLI native image, no need to manually load it.
     coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
 
-  val progName = (new Argv0).get("scala-cli")
+  val progName = {
+    val argv0 = (new Argv0).get("scala-cli")
+    val last  = Paths.get(argv0).getFileName.toString
+    last match {
+      case s".${name}.aux" =>
+        name // cs installs binaries under .app-name.aux and adds them to the PATH
+      case _ => argv0
+    }
+  }
 
   private def checkName(name: String) = {
     val baseProgName = if (Properties.isWin) progName.stripSuffix(".exe") else progName
--- a/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
@@ -107,6 +107,19 @@ class SipScalaTests extends ScalaCliSuite {
       }
     }
 
+  if (TestUtil.isNativeCli)
+    test(s"usage instruction should point to scala when installing by cs") { // https://github.com/VirtusLab/scala-cli/issues/1662
+      TestInputs.empty.fromRoot {
+        root => // cs installs binaries under .app-name.aux and scala-cli should drop .aux from progName
+          val binary       = "scala".prepareBinary(root)
+          val csBinaryName = root / ".scala.aux"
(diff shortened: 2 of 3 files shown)


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -21,7 +21,15 @@ object ScalaCli {
     // the Scala CLI native image, no need to manually load it.
     coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
 
-  val progName = (new Argv0).get("scala-cli")
+  val progName = {
+    val argv0 = (new Argv0).get("scala-cli")
+    val idx   = argv0.lastIndexOf(File.separator)
+    val last  = if (idx < 0) argv0 else argv0.drop(idx + 1)
```
tgodzik:
```suggestion
    val last  = Paths.get(argv0).getFileName().toString()
```
?

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -107,6 +107,19 @@ class SipScalaTests extends ScalaCliSuite {
...
+  if (TestUtil.isNativeCli)
+    test(s"usage instruction should point to scala when installing by cs") { // https://github.com/VirtusLab/scala-cli/issues/1662
+      TestInputs.empty.fromRoot {
+        root => // cs install binaries under .app-name.aux and scala-cli should drop .aux from progName
+          val binary       = "scala".prepareBinary(root)
+          val csBinaryName = root / ".scala.aux"
+          os.move(binary, csBinaryName)
+          val output   = os.proc(csBinaryName, "test", "--usage").call(check = false).out.text()
+          val usageMsg = TestUtil.removeAnsiColors(output)
+          expect(usageMsg.contains(" scala "))
```
julienrf: Maybe this assertion is too weak? Should it be something like `expect(!(usageMsg.contains(".scala.aux")))`?

lwronski: Right, Thanks for review!

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -107,6 +107,20 @@ class SipScalaTests extends ScalaCliSuite {
...
+          expect(!usageMsg.contains(".scala.aux"))
```
julienrf: What is the entire message? I think it should be exactly:

> Usage: scala test [options]

Is this possible?

lwronski: If `case-app` doesn't change the message for usage options, it should be exactly `Usage: scala test [options]`.

But for now, it should be exactly `Usage: scala test [options]`, so I modify integration test.

julienrf (approved):
Thank you Łukas!

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -21,7 +22,14 @@ object ScalaCli {
     // the Scala CLI native image, no need to manually load it.
     coursier.jniutils.LoadWindowsLibrary.assumeInitialized()
 
-  val progName = (new Argv0).get("scala-cli")
+  val progName = {
+    val argv0 = (new Argv0).get("scala-cli")
+    val last  = Paths.get(argv0).getFileName.toString
+    last match {
+      case s".${name}.aux" => name // cs install binaries under .app-name.aux and add to the PATH
```
Gedochao: NIT
```suggestion
      case s".${name}.aux" => name // cs installs binaries under .app-name.aux and adds them to the PATH
```

Gedochao (approved):
LGTM

PR #1738, opened 2022-12-29
Add support for running jar as an input

### LINKED ISSUE #1641: Support jars as inputs
The current syntax for handling jars with Scala CLI is slightly verbose.
```bash
▶ scala-cli --classpath sample.jar --main-class Main
Hello
```
It'd be convenient to be able to skip the `--classpath` option and pass a jar straight up as an input.

Add support like:
```bash
▶ scala-cli sample.jar
Hello
```

lwronski:
Fixes #1641.

### DIFF
--- a/website/docs/commands/run.md
+++ b/website/docs/commands/run.md
@@ -106,6 +106,19 @@ You can also add java options with the using directive `//> using javaOpt`:
 //> using javaOpt "-Xmx1g", "-Dfoo=bar"
 ```
 
+### JAR
+
+`scala-cli` lets you run JAR files just like any other input.
+
+```bash ignore
+scala-cli Hello.jar
+```
+
+```text
+Hello World
+```
+
+When you provide a JAR file as input to `scala-cli`, it will be added to the `classPath`.
 
 ## Define source files in using directives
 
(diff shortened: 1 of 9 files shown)


modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
```diff
@@ -109,6 +109,21 @@ final case class JavaPreprocessor(
         }
         Some(res)
 
+      case jar: JarFile => Some(either {
```
Gedochao: NIT I wonder if it wouldn't be easier to follow this code path if Jars had their own preprocessor.
We probably should support virtual jars in the future, too, so there'd be more dedicated logic, really.

website/docs/commands/run.md
```diff
@@ -106,6 +106,19 @@ You can also add java options with the using directive `//> using javaOpt`:
 //> using javaOpt "-Xmx1g", "-Dfoo=bar"
 ```
 
+### JAR
```
Gedochao, line 109: this should also be mentioned in `website/docs/commands/basics.md`

Gedochao (approved):

PR #1740, opened 2022-12-30
Explain the differences in using shebang vs scala-cli directly in script

### LINKED ISSUE #1309: Documentation should explain the differences in using `shebang` command vs. using scala-cli directly in the shebangs
This is especially important for our SIP

lwronski:
Fixes #1309.

### DIFF
--- a/website/docs/commands/shebang.md
+++ b/website/docs/commands/shebang.md
@@ -3,10 +3,11 @@ title: Shebang
 sidebar_position: 30
 ---
 
-This command is equivalent to `run`, but it changes the way Scala CLI parses arguments in order to be
-compatible with shebang scripts.
+This command is equivalent to `run`, but it changes the way Scala CLI parses options (used to configure the tool) and
+inputs (the sources of your project) in order to be compatible with `shebang` scripts.
 
-Normally, inputs and `scala-cli` options can be mixed. Program arguments have to be specified after `--`
+Normally, inputs and `scala-cli` options can be mixed. Program arguments (to be passed to your app) have to be specified
+after `--` (double dash) separator.
 
 ```bash ignore
 scala-cli [command] [scala_cli_options | input]... -- [program_arguments]...
--- a/website/docs/commands/shebang.md
+++ b/website/docs/commands/shebang.md
@@ -1,0 +1,23 @@
+---
+title: Shebang
+sidebar_position: 30
+---
+
+This command is equivalent to `run`, but it changes the way Scala CLI parses arguments in order to be
+compatible with shebang scripts.
+
+Normally, inputs and `scala-cli` options can be mixed. Program arguments have to be specified after `--`
+
+```bash ignore
+scala-cli [command] [scala_cli_options | input]... -- [program_arguments]...
+```
+
+For the `shebang` command, only a single input can be set. All Scala CLI options must be set before
+the input, while everything after the input is considered a program argument.
+
+```bash ignore
+scala-cli shebang [scala_cli_options]... input [program_arguments]...
+```
+
+More details can be found in [Shebang guide](../guides/shebang).
+
--- a/website/docs/guides/shebang.md
+++ b/website/docs/guides/shebang.md
@@ -1,0 +1,176 @@
+---
+title: Shebang
+sidebar_position: 30
+---
+
+
+This guide explains the differences between the `run` and `shebang` sub-commands, mainly covering how each of them
+parses its arguments.
+
+### `shebang` script headers
+
+Before proceeding, let's discuss how `scala-cli` works in a script without the `shebang` command.
+Here is a simple `hello.sc` script with a `shebang` header:
+
+```scala title=hello.sc
+#!/usr/bin/env -S scala-cli -S 3
+
+println(args.size)
+println(args.headOption)
+```
+
+And it works almost correctly.
+
+<ChainedSnippets>
+
+```bash 
+chmod +x hello.sc
+./hello.sc    
+```
+
+```text
+0
+None
+```
+
+<!-- Expected:
+0
+None
+-->
+
+</ChainedSnippets>
+
+And it also works.
+
+<ChainedSnippets>
+
+```bash
+./hello.sc -- Hello World
+```
+
+```text
+2
+Some(Hello)
+```
+
+<!-- Expected:
+2
+Some(Hello)
+-->
+
+</ChainedSnippets>
+
+Note that the extra `--` must be added to make it work. If it is not supplied, the result is:
+
+<ChainedSnippets>
+
+```bash run-fail
+./hello.sc Hello World
+```
+
+```text
+[error] Hello: not found
+World: not found
+```
+
+<!-- Expected:
+Hello: not found
+World: not found
+-->
+
+</ChainedSnippets>
+
+If we modify our script slightly and use the `shebang` sub-command in the header, we will get the following:
+
+```scala title=hello.sc
+#!/usr/bin/env -S scala-cli shebang -S 3
+
+println(args.size)
+println(args.headOption)
+```
+
+<ChainedSnippets>
+
+```bash
+./hello.sc Hello World
+```
+
+```text
+2
+Some(Hello)
+```
+<!-- Expected:
+2
+Some(Hello)
+-->
+
+</ChainedSnippets>
+
+
+### `shebang` and the command line
+
+Let's now see how the `shebang` command works straight from the command line.
+
+```scala title=Main.scala 
+object Main {
+  def main(args: Array[String]): Unit = println(args.mkString(" "))
+}  
+```
+
+<ChainedSnippets>
+
+```bash                                                                                                                                                                                                                                                                
+scala-cli shebang Main.scala Hello world
+```
+
+```text
+Hello world
+```
+
+<!-- Expected:
+Hello world
+-->
+
+</ChainedSnippets>
+
+
+:::note
+Please note that `shebang` changing how arguments are parsed means that every option after the first input will be treated as
+an argument to the app.
+
+<ChainedSnippets>
+
+```bash
+scala-cli shebang Main.scala -S 2.13 #-S 2.13 is not recognised as an option, but as app arguments
+```
+
+```text
+-S 2.13
+```
+
+<!-- Expected:
+-S 2.13
+-->
+
+</ChainedSnippets>
+:::
+
+If we try to do the same with the `run` sub-command, we get the following error:
+
+<ChainedSnippets>
+
+```bash run-fail
+scala-cli run Main.scala Hello world
+```
+
+```text
+[error]  Hello: not found
+world: not found
+```
+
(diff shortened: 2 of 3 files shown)


website/docs/commands/shebang.md
```diff
@@ -0,0 +1,23 @@
...
+This command is equivalent to `run`, but it changes the way Scala CLI parses arguments in order to be
+compatible with shebang scripts.
+
+Normally, inputs and `scala-cli` options can be mixed. Program arguments have to be specified after `--`
+
+```bash ignore
+scala-cli [command] [scala_cli_options | input]... -- [program_arguments]...
+```
+
+For the shebang command, only a single input file can be set. All Scala CLI options must be set before
```
Gedochao:
```suggestion
For the `shebang` command, only a single input can be set. All Scala CLI options must be set before
```

website/docs/commands/shebang.md
```diff
@@ -0,0 +1,23 @@
...
+the input file, and program arguments must be set after the input file.
```
Gedochao:
```suggestion
the input, while everything after the input is considered a program argument.
```

website/docs/commands/shebang.md
```diff
@@ -0,0 +1,23 @@
...
+
+```bash ignore
+scala-cli shebang [scala_cli_options]... input [program_arguments]...
+```
+
+More details in the [Shebang guide](../guides/shebang).
```
Gedochao:
```suggestion
More details can be found in [the Shebang guide](../guides/shebang).
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Shebang
+sidebar_position: 30
+---
+
+This command is equivalent to `run`, but it changes the way Scala CLI parses arguments in order to be
+compatible with shebang scripts.
```
Gedochao:
```suggestion
This guide explains the differences between the `run` and `shebang` sub-commands, mainly covering how each of them parses its arguments.
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+### In shebang script
```
Gedochao:
```suggestion
### `shebang` script headers
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+Before proceeding, let's discuss how `scala-cli` works in a script without the `shebang` command.
+Here is a simple hello script with a shebang header:
```
Gedochao:
```suggestion
Here is a simple `hello.sc` script with a `shebang` header:
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+```scala title=hello.sc
+#!/usr/bin/env scala-cli -S 3
+
+println(args.size)
+println(args.headOption)
+```
+
+And it almost works correctly.
```
Gedochao:
```suggestion
And it works almost correctly.
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+And it also works.
+
+```bash
+./hello.sc -- Hello World
+2
+Some(Hello)
+```
+
+But note that the extra `--` must be added to make it work. If it is not supplied, the result is:
```
Gedochao:
```suggestion
Note that the extra `--` must be added to make it work. If it is not supplied, the result is:
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+```bash
+./hello.sc Hello World
+Hello: not found
+World: not found
+```
+
+If we modify our script slightly and use the `shebang` command, we will get this effect.
```
Gedochao:
```suggestion
If we modify our script slightly and use the `shebang` sub-command in the header, we will get the following:```

File: website/docs/guides/shebang.md [hunk 10]
```diff
@@ -0,0 +1,87 @@
...
+And it almost works correctly.
+
+```bash
+./hello.sc    
+```
+
+```text
+0
+None
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash
./hello.sc    
```

```text
0
None
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 11]
```diff
@@ -0,0 +1,87 @@
...
+
+And it also works.
+
+```bash
+./hello.sc -- Hello World
+2
+Some(Hello)
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash
./hello.sc -- Hello World
```

```text
2
Some(Hello)
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 12]
```diff
@@ -0,0 +1,87 @@
...
+
+But note that the extra `--` must be added to make it work. If it is not supplied, the result is:
+
+```bash
+./hello.sc Hello World
+Hello: not found
+World: not found
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash
./hello.sc Hello World
```

```text
Hello: not found
World: not found
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 13]
```diff
@@ -0,0 +1,87 @@
...
+
+println(args.size)
+println(args.headOption)
+```
+
+```bash
+./hello.sc Hello World
+2
+Some(Hello)
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash
./hello.sc Hello World
```

```text
2
Some(Hello)
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 14]
```diff
@@ -0,0 +1,87 @@
...
+
+### In command line
```
Gedochao:
```suggestion
### `shebang` and the command line
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+Let's now see how the shebang command works in the command line.
```
Gedochao:
```suggestion
Let's now see how the `shebang` command works straight from the command line.
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+
+```scala title=Main.scala 
+object Main {
+  def main(args: Array[String]): Unit = println(args.mkString)
+}  
+```
+```bash                                                                                                                                                                                                                                                                
+scala-cli shebang Main.scala Hello world
+# Hello world
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash                                                                                                                                                                                                                                                                
scala-cli shebang Main.scala Hello world
```

```text
Hello world
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 17]
```diff
@@ -0,0 +1,87 @@
...
+
+Please note that `shebang` change in how arguments are parsed means that every options after the first input will be treated as
+an argument to the app.
```
Gedochao:
````suggestion
:::note
Please note that `shebang` changing how arguments are parsed means that every option after the first input will be treated as
an argument to the app.

<ChainedSnippets>

```bash
scala-cli shebang Main.scala -S 2.13 #-S 2.13 is not recognised as an option, but as app arguments
```

```text
-S2.13
```
</ChainedSnippets>
:::
````

File: website/docs/guides/shebang.md [hunk 18]
```diff
@@ -0,0 +1,87 @@
...
+
+If we try to do the same with the run command, we get the following error:
```
Gedochao:
```suggestion
If we try to do the same with the `run` sub-command, we get the following error:
```

website/docs/guides/shebang.md
```diff
@@ -0,0 +1,87 @@
...
+```bash ignore
+scala-cli run Main.scala Hello world
+# [error]  Hello: not found
+#  : not found
+# world: not found
+```
```
Gedochao:
````suggestion
<ChainedSnippets>

```bash ignore
scala-cli run Main.scala Hello world
```

```text
[error]  Hello: not found
 : not found
world: not found
```

</ChainedSnippets>
````

File: website/docs/guides/shebang.md [hunk 20]
```diff
@@ -0,0 +1,87 @@
+---
+title: Shebang
+sidebar_position: 30
+---
+
```
Gedochao, line 5:
```suggestion

import {ChainedSnippets} from "../../src/components/MarkdownComponents.js";

```

Gedochao (approved):
LGTM

website/docs/commands/shebang.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Shebang
+sidebar_position: 30
+---
+
+This command is equivalent to `run`, but it changes the way Scala CLI parses arguments in order to be
+compatible with shebang scripts.
+
+Normally, inputs and `scala-cli` options can be mixed. Program arguments have to be specified after `--`
```
julienrf: It is not completely obvious what the differences between inputs, options and arguments are. I think it would be worth defining them better.

Also, since the `shebang` command is valid in the `scala` runner, it would be better to use “Scala runner” instead of `scala-cli` I think.

SethTisue:
Can we do better than `shebang` on the name? It seems like obscure jargon to me

website/docs/commands/shebang.md
lwronski: I added a short explanation of the differences between inputs, options, and arguments.

I think we should do a migration from `scala-cli` to `scala` in another PR. It could be confusing if only one doc page points to `scala`.

Gedochao:
> Can we do better than `shebang` on the name? It seems like obscure jargon to me

@SethTisue any suggestions for an alias?

I actually don't think it's that obscure of a jargon...
The current `shebang` sub-command has 2 main use cases:
- to be used in the `shebang` header, as per [the Unix standard](https://en.wikipedia.org/wiki/Shebang_(Unix)), for which the name seems appropriate;
- to be used from the command line with legacy syntax backwards compatible with the old `scala` runner.

Although I believe we'll be keeping the `shebang` name, as the former usecase is the main one, I think we could potentially have an alias to make it easier to guess the second one... provided that we can think of a good name for it, of course.
Would something like `run-legacy` do the trick? feels verbose to me...

Also, I wonder if an alias is even necessary... if a user cares about backwards compatibility in this context when migrating from the old `scala` runner to Scala CLI, then it's probably fair to assume he's a rare specimen of a power user. Which means he should probably be okay with obscurely named sub-commands...

We don't want to promote using `shebang` from the command line for regular users, they should just learn the `run` syntax.

Gedochao:
@SethTisue I created a dedicated discussion for the `shebang` naming in #1755, let's continue that topic there

julienrf (approved):

Issue #1240, opened 2022-08-09
Show help document if for the invalid subcommand


tanishiking:
**Is your feature request related to a problem? Please describe.**
If I provide an invalid subcommand to `scala-cli`, it shows a stacktrace.

```sh
$ ❯ scala-cli invalid
scala.build.errors.InputsException: invalid: not found
  scala.build.Inputs$.forNonEmptyArgs(Inputs.scala:436)
  scala.build.Inputs$.apply(Inputs.scala:471)
  scala.cli.commands.util.SharedOptionsUtil$SharedOptionsOps.inputs(SharedOptionsUtil.scala:276)
  scala.cli.commands.util.SharedOptionsUtil$SharedOptionsOps.inputsOrExit(SharedOptionsUtil.scala:231)
  scala.cli.commands.Run$.run(Run.scala:64)
  scala.cli.commands.Run$.run(Run.scala:34)
  scala.cli.commands.Default.run(Default.scala:37)
  scala.cli.commands.Default.run(Default.scala:30)
  caseapp.core.app.CaseApp.main(CaseApp.scala:149)
  caseapp.core.app.CommandsEntryPoint.main(CommandsEntryPoint.scala:120)
  scala.cli.ScalaCliCommands.main(ScalaCliCommands.scala:119)
  scala.cli.ScalaCli$.main0(ScalaCli.scala:184)
  scala.cli.ScalaCli$.main(ScalaCli.scala:83)
  scala.cli.ScalaCli.main(ScalaCli.scala)
```

**Describe the solution you'd like**
It would be helpful if `scala-cli` shows a help document (that is shown when we type `scala-cli help`) when we provide an invalid subcommand

```sh
$ scala-cli invalid
'invalid' is not a scala-cli command <- something like that

Usage: scala-cli <COMMAND>
Scala CLI is a command-line tool to interact with the Scala language. It lets you compile, run, test, and package your Scala code.

Other commands:
  export                                        Export current project to sbt or Mill
  help                                          Print help message
  install completions, install-completions      Installs completions into your shell
  ...
```

**Describe alternatives you've considered**
It might be better to show a small error message that recommends users to run `scala-cli help`.

For example, git command shows the following message for invalid command.

```
$ git invalid
git: 'invalid' is not a git command. See 'git --help'.
```

**Additional context**
none

Gedochao:
Hey, good point. ~~And it should be simple enough to do that, too.~~

EDIT:
On second thought, I wonder.
[...]

PR #1743, opened 2023-01-02
Make the default sub-command default to `shebang` instead of `run` when inputs are provided

### LINKED ISSUE #603: Natural command-line arguments handling
### Motivation ###

For in scripting or prototyping use-cases for which *scala-cli* was designed, it is very important to allow of use command-line arguments as conveniently as possible.

Early versions of *scala-cli* are likely the only opportunity to simplify behavior for such fundamental use-cases.

### Current state ###

When *scala-cli* is invoked without a specific command, application command-line arguments can only follow after a special '--' argument, for example:

```
scala-cli hello.sc -- world
```

This seems to be a consequence of *scala-cli* defaulting to a *run* command behavior when invoked without a specific command. This behavior seems to require the '--' argument separator to accommodate possible *scala-cli* options and multiple source inputs/files.

#### Disadvantages ####

Although this is an effective solution it breaks the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) in the following ways:
* Runs contrary to the quite well established standard used by other 'script-friendly' languages (e.g. Python, Go) where by default no extra '--' argument is needed
* Invoking a 'Scala script' via *scala-cli* provides less than optional user experience - one must remember the need for '--'
* Creating a self-contained 'Scala script' requires using a specific command in the shebang - one must remember the *shebang* command instead of just using `#/usr/bin/env scala-cli`

#### Overview ####

Trade the convenience of using multiple source inputs for greater convenience of supplying application command-line arguments.

#### Examples ####

**Default behavior**
```
scala-cli [scala-cli options] <source input> [application arguments]
```

**Enable multiple source inputs option**
```
scala-cli [scala-cli options] --sources <source input> [source input]... -- [application arguments]
```

#### Details ####

This would define the default structure of *scala-cli* command-line arguments when invoked without a command as follows:
  * Scala CLI options come first
  * Single source input in the middle (unless specific CLI option enables multiple sources mode)
  * Application command-line arguments follow after

This would require two changes when compared to the current behavior:
  * *scala-cli* options must be specified before anything else. This allows distinguishing them from subsequent application arguments and it is a solution commonly used by many other programming language interpreters.
  * Default number of expected plain source input (source file) arguments must be 1. A specific command-line option (-s, --sources) can be provided to lift this limitation. Such an option would enable access to the current 'multiple sources' behavior leveraging the '--' argument separator to separate input sources arguments from the application arguments the way it works now.

#### Implementation alternatives ####

**Alternative 1**
* Change the current behavior of the *run* command as outlined above
* Make the *run* command a default 'no-command' behavior in case it already isn't
* Make the *shebang* command to be a mere alias for a *run* command for backwards compatibility and possibly depreciate it

**Alternative 2**
* Change the current behavior of the *shebang* command as outlined above
* Make the *shebang* command a default 'no-command' behavior instead of the current one
* Leave the *run* command unaffected for backwards compatibility and convenience

Comments and additional ideas are very welcome.

Gedochao:
Fixes #603 

This switches out the default sub-command (when no sub-command is specified explicitly) from `run` to `shebang`.
What it effectively does is that the command-line arguments handling is switched to the natural style of other scripts` engines.

```bash
▶ cat Main.scala 
object Main {
  def main(args: Array[String]): Unit = println(args.mkString)
}                                                                                                                                                                                                                                                                   
▶ scala-cli Main.scala Hello ' ' world
Hello world
```

Please note that the `run` sub-command's behaviour remains unchanged.
```bash
▶ scala-cli run Main.scala Hello ' ' world
[error]  Hello: not found
 : not found
world: not found
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
@@ -8,9 +8,9 @@ import scala.build.internal.Constants
 import scala.build.options.BuildOptions
 import scala.cli.CurrentParams
 import scala.cli.commands.repl.{Repl, ReplOptions}
-import scala.cli.commands.run.{Run, RunOptions}
 import scala.cli.commands.shared.ScalaCliHelp.helpFormat
 import scala.cli.commands.shared.SharedOptions
+import scala.cli.commands.shebang.{Shebang, ShebangOptions}
 import scala.cli.commands.version.Version
 import scala.cli.commands.{ScalaCommand, ScalaCommandWithCustomHelp}
 import scala.cli.launcher.LauncherOptions
@@ -25,8 +25,8 @@ class Default(
        |
        |When no subcommand is passed explicitly, an implicit subcommand is used based on context:
        |  - if the '--version' option is passed, it prints the 'version' subcommand output, unmodified by any other options
-       |  - if any inputs were passed, it defaults to the 'run' subcommand
-       |  - additionally, when no inputs were passed, it defaults to the 'run' subcommand in the following scenarios:
+       |  - if any inputs were passed, it defaults to the 'shebang' subcommand
+       |  - additionally, when no inputs were passed, it defaults to the 'shebang' subcommand in the following scenarios:
        |    - if a snippet was passed with any of the '--execute*' options
        |    - if a main class was passed with the '--main-class' option alongside an extra '--classpath'
        |  - otherwise, if no inputs were passed, it defaults to the 'repl' subcommand""".stripMargin
@@ -46,14 +46,14 @@ class Default(
     if options.version then println(Version.versionInfo)
     else
       {
-        val shouldDefaultToRun =
+        val shouldDefaultToShebang =
           args.remaining.nonEmpty || options.shared.snippet.executeScript.nonEmpty ||
           options.shared.snippet.executeScala.nonEmpty || options.shared.snippet.executeJava.nonEmpty ||
           options.shared.snippet.executeMarkdown.nonEmpty ||
           (options.shared.extraJarsAndClassPath.nonEmpty && options.sharedRun.mainClass.mainClass.nonEmpty)
-        if shouldDefaultToRun then RunOptions.parser else ReplOptions.parser
+        if shouldDefaultToShebang then ShebangOptions.parser else ReplOptions.parser
       }.parse(options.legacyScala.filterNonDeprecatedArgs(rawArgs, progName, logger)) match
-        case Left(e)                              => error(e)
-        case Right((replOptions: ReplOptions, _)) => Repl.runCommand(replOptions, args, logger)
-        case Right((runOptions: RunOptions, _))   => Run.runCommand(runOptions, args, logger)
+        case Left(e)                               => error(e)
+        case Right((replOptions: ReplOptions, _))  => Repl.runCommand(replOptions, args, logger)
+        case Right((sbOptions: ShebangOptions, _)) => Shebang.runCommand(sbOptions, args, logger)
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/shebang/Shebang.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shebang/Shebang.scala
@@ -14,8 +14,7 @@ object Shebang extends ScalaCommand[ShebangOptions] {
 
   override def scalaSpecificationLevel = SpecificationLevel.MUST
 
-  override def sharedOptions(options: ShebangOptions): Option[SharedOptions] =
-    Run.sharedOptions(options.runOptions)
+  override def sharedOptions(options: ShebangOptions): Option[SharedOptions] = Some(options.shared)
 
   override def runCommand(options: ShebangOptions, args: RemainingArgs, logger: Logger): Unit =
     Run.runCommand(
--- a/modules/cli/src/main/scala/scala/cli/commands/shebang/ShebangOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shebang/ShebangOptions.scala
@@ -2,7 +2,7 @@ package scala.cli.commands.shebang
 
 import caseapp.*
 
-import scala.cli.commands.run.RunOptions
+import scala.cli.commands.run.{RunOptions, SharedRunOptions}
 import scala.cli.commands.shared.{HasSharedOptions, SharedOptions}
 
 @HelpMessage(
@@ -34,9 +34,11 @@ import scala.cli.commands.shared.{HasSharedOptions, SharedOptions}
 )
 final case class ShebangOptions(
   @Recurse
-  runOptions: RunOptions = RunOptions()
+  shared: SharedOptions = SharedOptions(),
+  @Recurse
+  sharedRun: SharedRunOptions = SharedRunOptions()
 ) extends HasSharedOptions {
-  override def shared: SharedOptions = runOptions.shared
+  def runOptions: RunOptions = RunOptions(shared, sharedRun)
 }
 
 object ShebangOptions {
--- a/modules/integration/src/test/scala/scala/cli/integration/DefaultTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/DefaultTests.scala
@@ -116,7 +116,7 @@ class DefaultTests extends WithWarmUpScalaCliSuite {
     }
   }
 
-  test("default to the run sub-command if -classpath and --main-class are passed") {
+  test("default to the shebang sub-command if -classpath and --main-class are passed") {
     val expectedOutput = "Hello"
     val mainClassName  = "Main"
     TestInputs(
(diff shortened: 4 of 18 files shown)


romanowski:
I understand why we want to do that however, I feel it may be actually confusing in the long run that `scala ...` and `scala run ...` has different semantics. 

Maybe we can just provide a meaningful warning in such case and actionable diagnostic?  Namely that you need to separate inputs from arguments with `--`?

Gedochao:
> I understand why we want to do that however, I feel it may be actually confusing in the long run that `scala ...` and `scala run ...` has different semantics.

@romanowski Yep, the problem is that no matter how we fix it, it will be confusing to some extent. So this is an attempt at a compromise.

Based on @julienrf's [post after the last SIP meeting](https://contributors.scala-lang.org/t/sip-46-scala-cli-as-default-scala-command/5996/14), we should try to make Scala CLI compatible with the legacy `scala` runner in how it handles app arguments. The old `scala` command doesn't require separating inputs from arguments with `--`.

Now, way I see it, this can't be done without a price, we have to change semantics to achieve it.
If we want non-confusing, consistent behaviour, then we'd have to drop the `run` command's  current semantics altogether in favor of the `shebang` command (effectively merging them).
Which is something I'd rather avoid, as the option to pass multiple inputs the way it's possible right now is very convenient.

> Maybe we can just provide a meaningful warning in such case and actionable diagnostic? Namely that you need to separate inputs from arguments with `--`?

The error message should definitely be improved, yeah, but that will be done in a separate issue - #1240 

BTW this PR was meant to be discussed before potentially being merged from the start, I'm aware this is a controversial change.

julienrf:
Thank you for starting this discussion.

> If we want non-confusing, consistent behaviour, then we'd have to drop the `run` command's current semantics altogether in favor of the `shebang` command (effectively merging them).
> Which is something I'd rather avoid, as the option to pass multiple inputs the way it's possible right now is very convenient.

Could you please elaborate on why it is very convenient?

Gedochao:
> Could you please elaborate on why it is very convenient?

With the `shebang` sub-command semantics (which are in sync with the old `scala` runner in this regard) only the first argument is treated as an input, while all the following arguments are treated as app arguments. 
This effectively means that if more inputs are to be passed, they have to be declared with using directives (like the [custom sources `using file` directive](https://scala-cli.virtuslab.org/docs/reference/directives#custom-sources)) and not straight from the command line.

Meanwhile, the `run` sub-command's semantics use the double dash `--` as a separator between inputs and app arguments. 
```bash
▶ scala-cli sources tmp/A.scala tmp2/B.scala -- Hello ' ' World
```
IMO this makes it especially convenient when one wants to mix on-disk with virtual sources.
```bash
▶ cat Messages.scala
# case class Messages(args: Array[String]) {
#   override def toString: String = args.mkString(" ")
# }
▶ scala-cli Messages.scala https://gist.github.com/Gedochao/5ed4f0701b5d891e5919807d24f52437 -- hello world
# Compiling project (Scala 3.2.1, JVM)
# Compiled project (Scala 3.2.1, JVM)
# hello world
```
And in particular, mixing piped sources with other inputs in any other manner is tricky.
```bash
▶ scala-cli _.scala https://gist.github.com/Gedochao/5ed4f0701b5d891e5919807d24f52437 -- hello world < Messages.scala
# Compiling project (Scala 3.2.1, JVM)
# Compiled project (Scala 3.2.1, JVM)
# hello world
```

julienrf:
I see, thank you for providing concrete examples. I wonder if mixing source types is common, though… Also, I wonder if using multiple local files is common too (could we simply use `scala run .` instead of `scala run A.scala B.scala`)?

Gedochao:
> I see, thank you for providing concrete examples. I wonder if mixing source types is common, though… Also, I wonder if using multiple local files is common too (could we simply use `scala run .` instead of `scala run A.scala B.scala`)?

Yes, we could. `scala-cli run .` is the default way of using Scala CLI, really.

However, if there's more source files in the working directory beyond `A.scala` and `B.scala`, a user may want to pass them explicitly to not include the whole directory. 
Passing multiple directories is an option, too, which may come in useful when 2 projects have a shared subset of sources.
```bash
▶ tree
# .
# ├── project1
# ├── project2
# └── shared
▶ scala-cli project1 shared -- arg1 arg2
# project1 output
▶ scala-cli project2 shared -- arg1 arg2
# project2 output
▶ scala-cli . -- arg1 arg2
## won't compile
```

Additionally, I'd imagine passing multiple inputs is convenient when working with repositories running on different platforms. For example, a project may use a different set of sources when running on Scala Native and different when running on the JVM...

julienrf:
> However, if there's more source files in the working directory beyond `A.scala` and `B.scala`, a user may want to pass them explicitly to not include the whole directory.

Is that really in the scope of the Scala runner? It seems to be a complicated setup and requires being careful when invoking the runner to pass the correct arguments. It could also be achieved with `//> using file ../shared`, no?

Gedochao:
> Is that really in the scope of the Scala runner? 

When looking at Scala CLI as "just" the new Scala runner, then maybe not. However, Scala CLI has been out for a while... And AFAIK this has been possible almost from the very beginning. Being able to pass inputs like this is a part of the Scala CLI UX by now... I feel one should always be careful when taking features away, someone's bound to be unhappy with it.

> It seems to be a complicated setup and requires being careful when invoking the runner to pass the correct arguments. It could also be achieved with `//> using file ../shared`, no?

Yes, using directives should cover this. Although I don't think it's possible straight from the command line when using the `shebang` sub-command. We could, theoretically, allow for passing extra inputs via an option... but I have mixed feelings about going that direction. I mean syntax like this:

```bash
scala-cli shebang --extra-input shared project1 arg1 arg2
```
as opposite to the earlier
```bash
scala-cli run project1 shared -- arg1 arg2
```

Also, do keep in mind that the `shebang` subcommand (similarly to the legacy `scala` runner) requires all options to be passed before the input source, as anything after the first input is considered an app argument. So when setting the Scala version:
```bash
# run allows to pass the options wherever before the app args section
scala-cli run . -S 2.13 -- arg1 arg2
scala-cli run -S 2.13 . -- arg1 arg2

# shebang requires options to be passed before the input
scala-cli shebang -S 2.13 . arg1 arg2

# this is similar to the legacy scala runner
scala -save Main.scala arg1 arg2
```

lwronski:
> Is that really in the scope of the Scala runner? It seems to be a complicated setup and requires being careful when invoking the runner to pass the correct arguments. It could also be achieved with `//> using file ../shared`, no?

It would be too limiting for me if we decided to limit inputs to only one file or directory. Sometimes, it is useful to add a directory such as a `utils` directory in this way:
```
scala-cli run Hello.scala utils -S 2.13
``` 

Sometimes it might be useful to specify different sources in CI, and in these cases, it can be difficult to modify the source to add a `using` file directive.

julienrf:
> Sometimes it might be useful to specify different sources in CI, and in these cases, it can be difficult to modify the source to add a `using` file directive.

Don’t we have a systematic way to supply configuration either via command-line arguments or using directives? Couldn’t we do something like:

~~~
scala run --source utils Hello.scala
~~~

Gedochao:
> Don’t we have a systematic way to supply configuration either via command-line arguments or using directives? Couldn’t we do something like:
> 
> ```
> scala run --source utils Hello.scala
> ```

We could, yes... But I'm not convinced we want to. 
Typing `--source` for every extra input beyond the first is a very verbose way to handle it.
A `--sources` flag accepting a list of inputs wouldn't be ideal, either.
Not to mention that and any other options would always have to be specified before the first input, as anything after that is considered an app arg.
TBH I don't think this loss of UX is a good idea for the sake of backwards compatibility and easier passing of app arguments... 

If a user wants the vanilla old runner experience on that matter, he can always specify the `shebang` sub-command explicitly... is that necessarily a deal breaker?

cc @tgodzik @romanowski

tgodzik:
> If a user wants the vanilla old runner experience on that matter, he can always specify the shebang sub-command explicitly... is that necessarily a deal breaker?

I don't love that we need to add the additional `--`, but I can't think of a solution that would actually be better. I don't think we should keep accepting just a single input for the sake of backward compatibility. Scala runner itself never had a lot of users and in reality most of it's behaviours are undocumented or even different between Scala 2/3, so the backwards compatibility argument doesn't have too much weight. Using `--` is already the case with coursier, so we do have an established precedence. Also Rust's Cargo is using exactly the same solution `cargo run [options] [-- args]`

I was also thinking that maybe we could treat the first argument that doesn't exist as a file as an actual program argument, but that would be prone to errors if users actually wanted to use a file as an argument.

What we should do is when users get an error about a file not existing, we tell them clearly that they need to add `--`.

julienrf:
Thank you @tgodzik for your comment.

> Scala runner itself never had a lot of users and in reality most of it's behaviours are undocumented or even different between Scala 2/3, so the backwards compatibility argument doesn't have too much weight.

I agree with this point. I believe we should try to preserve the backwards compatibility as much as possible, but if doing so would really reduce the overall usability of the tool, then it’s fine to introduce breaking changes. Please make sure to thoroughly document the differences between `shebang` and `run` (as tracked by #1309).

Also, it would be great to already create a list of incompatibilities and recipes to migrate from the old runner to the new one.

Gedochao:
Alright, seems like we won't be going ahead with this solution, so I'm closing the PR.

PR #1744, opened 2023-01-03
Cleanup `version`, `about` & `doctor` sub-commands

### LINKED ISSUE #698: Add doctor command to diagnose common problems and print information for further debug
Having more than an instance of scala-cli on PATH creates problems with updates, since an old version might appear sooner on PATH and get executed instead of the newer version!

How should it work:

The user runs 
`scala-cli doctor`

Then, if there are any duplicates on PATH it gets reported to the user!

One implementation strategy is 

- first seeing if `type -ap scala-cli` shows duplicates; 
- then if so, we run `ls -l` on the result to make sure they are not all symlinks to the same location. 
- Then if they were not symlinks, we `cat` the contents of them to see if they run the same script. (this is tricky)
- Now if they weren’t scripts executing the same thing, we produce a PATH duplicates error

Alternatively, this information can be provided by running the update, about, or version commands.

The team decided that the PATH duplicate part of issue #543  should be dealt with as an independent separate issue.
It is also related to issue #670

### LINKED ISSUE #1665: No sensible difference between commands doctor, about, and version

Currently (`scala` 0.1.18 installed via `cs install scala-experimental`), the documentation says the following:

~~~ text
Doctor commands:
  doctor  Print details about this application

Miscellaneous commands:
  version  Print version
~~~

It is not clear what are the differences between `doctor`, `about`, and `version`.

I tried all of them and I got the following output:

~~~ text
$ scala doctor
Your scala version (0.1.18) is current.
scala can be updated by your package manager since it is correctly installed in only one location: .
Your scala is a native application.

$ scala about
Scala code runner version: 0.1.18
$ scala version
Scala code runner version: 0.1.18
~~~

I suggest keeping the behavior of `version` as it is, but improving its documentation to the following:

~~~ text
Print the version of the scala runner and the default Scala version (unless specified in the project).
~~~

I suggest removing the `about` command, or renaming `doctor` into `about` (so, removing `doctor`) because “about” is more self-explanatory or intuitive to me (but that might be purely subjective).

### LINKED ISSUE #1667: Output of command doctor could be more helpful

When using `scala` installed via `cs install scala-experimental`, running `scala doctor` produces the following output:

~~~ text
$ scala doctor
Your scala version (0.1.18) is current.
scala can be updated by your package manager since it is correctly installed in only one location: .
Your scala is a native application.
~~~

Overall, I think the content could be improved, but I also see that the “location” of the installation is empty.

I am not sure, because it is still unclear what I should expect from this command based on its current documentation and output.

Here are some suggestions:

- ~~~ diff
  - Your scala version (0.1.18) is current.
  + Your scala runner version (0.1.18) is the latest.
  ~~~
- About “scala can be updated by your package manager …”, I am not sure which package manager you are talking about. Is `cs` the package manager? If yes, it would be good to clarify that because many people could assume you talk about `apt`, which is the package manager of my Linux distribution.
- Fix the installation location
- What does “is a native application” mean? It does not need a JVM?

I guess the documentation of this command could be improved as well. Currently, it says that it “prints details about this application”. It would be useful to understand the exact purpose of the command, and when I should be interested in using it? (maybe it’s primarily a command that I should run when I want to report a bug, to help the maintainers figure out details about my setup?)

I also see that there is a hidden option `--gh-token`. When is it used? Why is it hidden?

Gedochao:
Fixes #1665 
Fixes #1667 
Closes #698 

in accordance with https://github.com/VirtusLab/scala-cli/issues/1665#issuecomment-1362860955, this does the following:
- the `doctor` sub-command is being removed, its present and previously planned functionalities should be delivered in a more interactive manner
- the `about` sub-command is being removed, its functionalities merged into `version`
  - `version` will now check if Scala CLI is up-to-date (similarly how it was previously done in `about`)
  - the `--offline` option was added to `version` to allow to skip checking if Scala CLI is up-to-date
  - the `--scala-version` and `--cli-version` options similarly skip checking if Scala CLI is up-to-date, outputting just the raw version requested


modules/cli/src/main/scala/scala/cli/commands/version/VersionOptions.scala
```diff
@@ -20,14 +23,21 @@ final case class VersionOptions(
...
+  @HelpMessage("Show plain Scala version only")
   @Tag(tags.implementation)
   @Name("scala")
-    scalaVersion: Boolean = false
+    scalaVersion: Boolean = false,
+  @Hidden
+  @HelpMessage(HelpMessages.passwordOption)
+  @Tag(tags.implementation)
+    ghToken: Option[PasswordOption] = None,
+  @Group("Version")
```
lwronski, line 38: Maybe keep this options in `VerbosityOptions`. It should be useful globally since we wanted to add `offline` mode for scala-cli.

lwronski (approved):

modules/cli/src/main/scala/scala/cli/commands/version/VersionOptions.scala
Gedochao: It's not a verbosity option either, though... and I don't want to introduce a new category just for this, so for now I'd keep it in `Version` options.
That assignment would probably change once we implement a global `--offline` implementation

PR #1748, opened 2023-01-04
Generate OS package for aarch64 and let sdkman know

mkurz:
sdkman doesn't know about the aarch64/arm64 binary yet :cry: 
This is what I get on my aarch64 Linux:
```sh
$ sdk install scalacli

Stop! scalacli 0.1.19 is not available. Possible causes:
 * 0.1.19 is an invalid version
 * scalacli binaries are incompatible with your platform
 * scalacli has not been released yet

Tip: see all available versions for your platform:

  $ sdk list scalacli
```
````sh
$ sdk list scalacli
================================================================================
Available Scalacli Versions
================================================================================
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                

================================================================================
+ - local version
* - installed
> - currently in use
================================================================================

### DIFF
--- a/.github/scripts/generate-os-packages.sh
+++ b/.github/scripts/generate-os-packages.sh
@@ -1,6 +1,16 @@
 #!/usr/bin/env bash
 set -eu
 
+ARCHITECTURE="x86_64" # Set the default architecture
+if [[ "$OSTYPE" == "linux-gnu"* ]] || [[ "$OSTYPE" == "darwin"* ]]; then
+  # When running on macOS and Linux lets determine the architecture
+  ARCHITECTURE=$(uname -m)
+fi
+if [[ $# -eq 1 ]]; then
+  # architecture gets overridden by command line param
+  ARCHITECTURE=$1
+fi
+
 ARTIFACTS_DIR="artifacts/"
 mkdir -p "$ARTIFACTS_DIR"
 
@@ -103,10 +113,15 @@ generate_sdk() {
   local binName
 
   if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-    sdkDirectory="scala-cli-x86_64-pc-linux-static-sdk"
+    if [[ "$ARCHITECTURE" == "aarch64" ]] || [[ "$ARCHITECTURE" == "x86_64" ]]; then
+      sdkDirectory="scala-cli-$ARCHITECTURE-pc-linux-static-sdk"
+    else
+      echo "scala-cli is not supported on $ARCHITECTURE"
+      exit 2
+    fi
     binName="scala-cli"
   elif [[ "$OSTYPE" == "darwin"* ]]; then
-    if [[ "$(uname -m)" == "arm64" ]]; then
+    if [[ "$ARCHITECTURE" == "arm64" ]]; then
       sdkDirectory="scala-cli-aarch64-apple-darwin-sdk"
     else
       sdkDirectory="scala-cli-x86_64-apple-darwin-sdk"
@@ -133,11 +148,13 @@ generate_sdk() {
 }
 
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-  generate_deb
-  generate_rpm
+  if [ "$ARCHITECTURE" == "x86_64" ]; then
+    generate_deb
+    generate_rpm
+  fi
   generate_sdk
 elif [[ "$OSTYPE" == "darwin"* ]]; then
-  if [ "$(uname -m)" == "arm64" ]; then
+  if [ "$ARCHITECTURE" == "arm64" ]; then
     generate_pkg "aarch64"
   else
     generate_pkg "x86_64"
--- a/.github/scripts/publish-sdkman.sh
+++ b/.github/scripts/publish-sdkman.sh
@@ -15,13 +15,13 @@ version() {
 }
 
 SCALA_CLI_VERSION="$(version)"
-ARCHS=("x86_64" "x86_64" "aarch64" "x86_64")
-UNAMES=("pc-linux-static-sdk" "apple-darwin-sdk" "apple-darwin-sdk" "pc-win32-sdk")
-PLATFORMS=("LINUX_64" "MAC_OSX" "MAC_ARM64" "WINDOWS_64")
+ARCHS=("x86_64" "aarch64" "x86_64" "aarch64" "x86_64")
+UNAMES=("pc-linux-static-sdk" "pc-linux-static-sdk" "apple-darwin-sdk" "apple-darwin-sdk" "pc-win32-sdk")
+PLATFORMS=("LINUX_64" "LINUX_ARM64" "MAC_OSX" "MAC_ARM64" "WINDOWS_64")
 
 for i in "${!PLATFORMS[@]}"; do
 
-    SCALA_CLI_URL="https://github.com/VirtuslabRnD/scala-cli/releases/download/v$SCALA_CLI_VERSION/scala-cli-${ARCHS[i]}-${UNAMES[i]}.zip"
+    SCALA_CLI_URL="https://github.com/VirtusLab/scala-cli/releases/download/v$SCALA_CLI_VERSION/scala-cli-${ARCHS[i]}-${UNAMES[i]}.zip"
 
     # Release a new Candidate Version
     curl --silent --show-error --fail \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,7 @@ jobs:
           env: |
             JAVA_HOME: /jvm
             JAVA_OPTS: "-Djdk.lang.Process.launchMechanism=vfork"
+            GITHUB_ENV: "/tmp/github_env"
 
           # The shell to run commands with in the container
           shell: /bin/bash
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: .github/scripts/publish-sdkman.sh [hunk 1]
```diff
@@ -15,13 +15,13 @@ version() {
...
-UNAMES=("pc-linux-static-sdk" "apple-darwin-sdk" "apple-darwin-sdk" "pc-win32-sdk")
-PLATFORMS=("LINUX_64" "MAC_OSX" "MAC_ARM64" "WINDOWS_64")
+ARCHS=("x86_64" "aarch64" "x86_64" "aarch64" "x86_64")
+UNAMES=("pc-linux-static-sdk" "pc-linux-static-sdk" "apple-darwin-sdk" "apple-darwin-sdk" "pc-win32-sdk")
+PLATFORMS=("LINUX_64" "LINUX_ARM64" "MAC_OSX" "MAC_ARM64" "WINDOWS_64")
 
 for i in "${!PLATFORMS[@]}"; do
 
-    SCALA_CLI_URL="https://github.com/VirtuslabRnD/scala-cli/releases/download/v$SCALA_CLI_VERSION/scala-cli-${ARCHS[i]}-${UNAMES[i]}.zip"
+    SCALA_CLI_URL="https://github.com/VirtusLab/scala-cli/releases/download/v$SCALA_CLI_VERSION/scala-cli-${ARCHS[i]}-${UNAMES[i]}.zip"
```
mkurz, line 24: So you changed the GitHub org?

.github/workflows/ci.yml
```diff
@@ -228,11 +228,15 @@ jobs:
...
-            apt-get install -q -y build-essential libz-dev git python3-pip gnupg2 curl
+            apt-get install -q -y zip unzip build-essential libz-dev git python3-pip gnupg2 curl
           run: |
             set -e
             export PATH="/utils:$PATH"
             git config --global --add safe.directory "$(pwd)"
+            ./mill -i ci.setShouldPublish
+            if [[ $SHOULD_PUBLISH == "true" ]]; then
+              .github/scripts/generate-os-packages.sh
+            fi
```
mkurz, line 242: I am pretty sure this env variable thing here will not work, but I am waiting which error CI gives me.

.github/scripts/publish-sdkman.sh
lwronski: Yes, the organization was changed, but this script still works because it is correctly redirected. 

But let's change to `VirtusLab`

.github/workflows/ci.yml
lwronski: So why not add new job work before this [step](https://github.com/VirtusLab/scala-cli/blob/e818e9d4593f631ca059d87dfe4f44cef8a54f78/.github/workflows/ci.yml#L237)?  We don't have to generate sdk man package inside the docker with aarch64.

lwronski: Sorry for my confusing message. I didn't expect that `./mill -I ci.setShouldPublish` fail in this [way](https://github.com/VirtusLab/scala-cli/pull/1748/checks#step:5:721).

I think, your solution is better and it should work.

mkurz: Seems like it fails because the QEMU virtual machine runs under some different user (root or whatever) and so the GitHub actions user can't access them. I see if chown'ing all files works.

lwronski: @mkurz I think, we should include this PR into the next release. Can you able to return removed lines in `ci.yml` and then I think, we can merge it PR.

mkurz: @lwronski The PR does not work correctly. When running `.github/scripts/generate-os-packages.sh` it overrides the generated aarch64 binary with the "wrong" one meant for x86 systems (I tested this by downloading the binary, I could not run it).
The problem is: In the script the function `generate_sdk` calls the `launcher` function [here](https://github.com/VirtusLab/scala-cli/blob/d790aff03c7ebc69e383e2aa8bd2f6647b76d90e/.github/scripts/generate-os-packages.sh#L124) which runs `mill -i copyTo "cli.nativeImage" "scala" 1>&2` [here](https://github.com/VirtusLab/scala-cli/blob/d790aff03c7ebc69e383e2aa8bd2f6647b76d90e/.github/scripts/generate-os-packages.sh#L27) which again builds the binary and overrides the one generated by the virtual machine (at least this is what I think is happening).

What do you think is the best way to avoid this?

lwronski: Maybe we should back to your first idea and run `.github/scripts/generate-os-packages.sh` inside the Docker container. In this case, it should properly detect the architecture and take the `aarch64` launcher in the `.github/scripts/generate-os-packages.sh` script.

mkurz: Yeah, I will try that again.

mkurz:
@lwronski Finished, it should work now. Here you can see the artifacts generated also with the sdk packages for aarch64: https://github.com/VirtusLab/scala-cli/actions/runs/3897030762 This was the last run, I just squashed the commit now and removed added back the `if: github.event_name == 'push'`, so the `generate-linux-arm64-native-launcher` doesn't run for PRs, because it takes ages.
I tested the artifacts of the sdk aarch64 package and it works now.

Should be ready to merge.

.github/scripts/generate-os-packages.sh
```diff
@@ -1,6 +1,16 @@
...
 
+ARCHITECTURE="x86_64" # Set the default architecture
+if [[ "$OSTYPE" == "linux-gnu"* ]] || [[ "$OSTYPE" == "darwin"* ]]; then
+  # When running on macOS and Linux lets determine the architecture
+  ARCHITECTURE=$(uname -m)
+fi
+if [[ $# -eq 1 ]]; then
+  # architecture gets overridden by command line param
+  ARCHITECTURE=$1
+fi
```
mkurz, line 12: This is not really necessary anymore, but I still kept it, so in future, if possible with Graal, you can pass the architecture as first argument for cross building. Currently it has no use, but shouldn't hurt.

.github/scripts/generate-os-packages.sh
```diff
@@ -133,11 +148,13 @@ generate_sdk() {
 }
 
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-  generate_deb
-  generate_rpm
+  if [ "$ARCHITECTURE" == "x86_64" ]; then
+    generate_deb
+    generate_rpm
+  fi
```
mkurz, line 154: We don't need to generate deb and rpm files when running inside the aarch64 QEMU vm.

.github/workflows/ci.yml
```diff
@@ -218,6 +218,7 @@ jobs:
           env: |
             JAVA_HOME: /jvm
             JAVA_OPTS: "-Djdk.lang.Process.launchMechanism=vfork"
+            GITHUB_ENV: "/tmp/github_env"
```
mkurz, line 221: So, when starting the QEMU vm, all GitHub actions env variables are passed on to it, see here https://github.com/uraimo/run-on-arch-action/blob/a8003307a739516fdd80ee6d3da8924db811b8da/src/run-on-arch.sh#L101-L126
However the `GITHUB_ENV` var points to a file, which is not mapped inside the QEMU vm, so when trying to write to it it will fail, e.g. see [this comment](https://github.com/uraimo/run-on-arch-action/pull/53#issuecomment-1375690706).
So I override it with a dummy file `/tmp/github_env`, which kind of simulates the env file for mill when it calls `setShouldPublish` and [writes to that file](https://github.com/VirtusLab/scala-cli/blob/8ba8022069b2f35ff6374a8aeeb8f252cbc190cd/project/publish.sc#L213).

.github/workflows/ci.yml
```diff
@@ -228,12 +229,18 @@ jobs:
           # gnupg2 is for signing the resulting binary.
           install: |
             apt-get update -q -y
-            apt-get install -q -y build-essential libz-dev git python3-pip gnupg2 curl
+            apt-get install -q -y build-essential libz-dev git python3-pip gnupg2 curl zip
+            touch /tmp/github_env
```
mkurz, line 233: Just to make sure I create the dummy file first.
Also zip [is needed](https://github.com/VirtusLab/scala-cli/blob/8ba8022069b2f35ff6374a8aeeb8f252cbc190cd/.github/scripts/generate-os-packages.sh#L129) inside the vm

.github/workflows/ci.yml
mkurz: The `grep -q...` returns 0 if it finds the searched string and enters the if branch.
Thats part of the github_env var simulation...

.github/workflows/ci.yml
```diff
@@ -228,12 +229,18 @@ jobs:
...
           run: |
             set -e
             export PATH="/utils:$PATH"
             git config --global --add safe.directory "$(pwd)"
             ./mill -i copyDefaultLauncher artifacts/
+            ./mill -i ci.setShouldPublish
+            if grep -q "SHOULD_PUBLISH=true" /tmp/github_env; then
+              .github/scripts/generate-os-packages.sh
+            fi
+      - run: sudo chown -R $USER .
```
mkurz, line 243: Just to make sure I chown all files after the vm ran, so we don't have to worry about access denied errors etc.

.github/workflows/ci.yml
```diff
@@ -997,6 +1004,10 @@ jobs:
         with:
           name: linux-launchers
           path: artifacts/
+      - uses: actions/download-artifact@v3
+        with:
+          name: linux-aarch64-launchers
+          path: artifacts/
```
mkurz, line 1010: This was missing in the `update-packages` job. I hope this works now and the sdk -man script has all files available to upload them.

mkurz:
CI is green now as well.

.github/scripts/generate-os-packages.sh
lwronski: It's fine for me. We can still keep it here.

.github/workflows/ci.yml
lwronski: Good catch. Yeah, it should work.

lwronski (approved):
Thanks for your works! In the next week wa plan the next release of scala-cli, so we can verify your PR in release.

Gedochao (approved):

PR #1754, opened 2023-01-09
Add CLI option for suppressing the warning about directives in multiple files

### LINKED ISSUE #1747: Enable globally turning off warnings about `using` directives in multiple files
We should add a global configuration enabling users to silence the warning about migrating directives to `project.scala`.  It should be settable with the `config` sub-command.

So something like:
```bash
scala-cli config ignore-directives-in-multiple-files true
```

Ticket motivated by a comment by @lbialy in https://github.com/VirtusLab/scala-cli/issues/1632#issuecomment-1369764141

MaciejG604:


modules/cli/src/main/scala/scala/cli/commands/shared/SuppressWarningOptions.scala
```diff
@@ -0,0 +1,19 @@
...
+import caseapp.*
+
+import scala.cli.commands.tags
+
+// format: off
+final case class SuppressWarningOptions(
+  @Tag(tags.implementation)
+  @HelpMessage("Suppress warnings about using directives in multiple files")
+  @Name("suppress-warning-directives-in-multi-files")
+    multiFileDirectivesWarning: Boolean = false
```
Gedochao:
```suggestion
  @Hidden
  @Name("suppressWarningDirectivesInMultipleFiles")
    suppressDirectivesInMultipleFilesWarning: Boolean = false
```
You can use camel case in `@Name` aliases.
Also, I'm thinking if we could find a shorter alias that would bring this flag's name down to a sensible length... can't think of a better one atm 😐

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -300,6 +302,7 @@ object Build {
             buildClient,
             actualCompiler,
             buildTests,
+            suppressDirectivesInMultipleFilesWarning,
```
Gedochao: I'm thinking that instead of passing it separately in each of the sub-commands, you could add it as a field in `scala.build.options.BuildOptions` and pass it in `SharedOptions.buildOptions(...)`

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -126,6 +126,7 @@ object CrossSources {
     inputs: Inputs,
     preprocessors: Seq[Preprocessor],
     logger: Logger,
+    suppressDirectivesInMultipleFilesWarning: Option[Boolean],
```
Gedochao, line 129: I wonder if it's necessary to pass the flag all the way here like this, since it's already in `buildOptions`... 
`buildOptions` are initialised later in this method (line 218 as of this comment), I think it should be retrievable from there...
the warning would have to be printed slightly later though. which I think should be fine.

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -168,7 +169,11 @@ object CrossSources {
       (preprocessedInputFromArgs ++ preprocessedSourcesFromDirectives).distinct
 
     val preprocessedWithUsingDirs = preprocessedSources.filter(_.directivesPositions.isDefined)
-    if (preprocessedWithUsingDirs.length > 1) {
+    if (
+      preprocessedWithUsingDirs.length > 1 && !suppressDirectivesInMultipleFilesWarning.getOrElse(
```
Gedochao, line 173: As per the previous comment, the whole `if` clause producing the warning could potentially be moved after the `buildOptions` are initialised.
and then you should be able to:
```suggestion
      preprocessedWithUsingDirs.length > 1 && !buildOptions.suppressWarningOptions.suppressDirectivesInMultipleFilesWarning.getOrElse(
```

MaciejG604: I think this won't work here. This is what I found out the variable `buildOptions` looks like during the warning suppression test:
image

Looks like it has no values for our suppress warning flag

Gedochao: right, as those `buildOptions` are coming from processed sources... so they'd only include things from preprocessors, like `using` directives... sorry, I got confused there.
Nevermind my suggestion then.

Gedochao (approved):
LGTM

lwronski (approved):

carlosedp:
This apparently fixes #1828

PR #1760, opened 2023-01-10
Don't fail in case of connection errors in the version sub-command

Gedochao:
This should make our integration tests less shaky, especially `VersionTests` and `SipTests`.
Since `version` now checks for the upstream newest version by default, a lot of our tests on CI were failing randomly.

This will debug log any connection errors encountered by the `version` sub-command, instead of failing as it did before.

```bash
scala-cli version -v -v
# Scala CLI version: 0.1.20-SNAPSHOT
# Scala version (default): 3.2.1
# Failed to check for the newest Scala CLI version upstream: Error downloading https://api.github.com/repos/VirtusLab/scala-cli/releases: java.net.UnknownHostException: api.github.com
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
@@ -6,6 +6,7 @@ import com.github.plokhotnyuk.jsoniter_scala.macros.*
 import coursier.core
 
 import scala.build.Logger
+import scala.build.errors.CheckScalaCliVersionError
 import scala.build.internal.Constants.{ghName, ghOrg, version as scalaCliVersion}
 import scala.cli.CurrentParams
 import scala.cli.commands.{CommandUtils, ScalaCommand}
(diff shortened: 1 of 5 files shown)


modules/cli/src/main/scala/scala/cli/commands/default/Default.scala
```diff
@@ -44,7 +44,8 @@ class Default(
 
   override def runCommand(options: DefaultOptions, args: RemainingArgs, logger: Logger): Unit =
     // can't fully re-parse and redirect to Version because of --cli-version and --scala-version clashing
-    if options.version then Version.runCommand(VersionOptions(options.shared.logging), args, logger)
+    if options.version then
+      Version.runCommand(VersionOptions(options.shared.logging, offline = true), args, logger)
```
lwronski: Instead of setting `offline`, we can ignore errors when something is wrong while downloading the latest tag of `scala-cli`.

Gedochao: should be fixed.
i adjusted the PR description as well.

modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
```diff
@@ -31,29 +32,28 @@ object Update extends ScalaCommand[UpdateOptions] {
...
+    try {
+      val resp = ProcUtil.download(url, headers: _*)
+      readFromArray(resp)(releaseListCodec).filter(_.actualRelease)
+        .maxByOption(_.version)
+        .map(_.version.repr)
+        .toRight(CheckScalaCliVersionError(s"No $fullRunnerName versions found in $url"))
+    }
+    catch {
+      case e: JsonReaderException => Left(CheckScalaCliVersionError(s"Error reading $url", e))
+      case e: Exception => Left(CheckScalaCliVersionError(
```
lwronski:
```suggestion
      case e: Throwable => Left(CheckScalaCliVersionError(
```

modules/cli/src/main/scala/scala/cli/commands/version/Version.scala
```diff
@@ -26,19 +26,28 @@ object Version extends ScalaCommand[VersionOptions] {
   )
 
   override def runCommand(options: VersionOptions, args: RemainingArgs, logger: Logger): Unit = {
-    lazy val newestScalaCliVersion = Update.newestScalaCliVersion(options.ghToken.map(_.get()))
-    lazy val isVersionOutOfDate: Boolean =
-      CommandUtils.isOutOfDateVersion(newestScalaCliVersion, Constants.version)
+    lazy val maybeNewerScalaCliVersion: Option[String] =
+      Update.newestScalaCliVersion(options.ghToken.map(_.get())) match {
+        case Left(e) =>
+          logger.debug(e.message)
```
lwronski, line 32:
```suggestion
          logger.debug(e)
```

modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
```diff
@@ -107,7 +107,10 @@ object Update extends ScalaCommand[UpdateOptions] {
 
   private def update(options: UpdateOptions, currentVersion: String, logger: Logger): Unit = {
 
-    val newestScalaCliVersion0 = newestScalaCliVersion(options.ghToken.map(_.get()))
+    val newestScalaCliVersion0 = newestScalaCliVersion(options.ghToken.map(_.get())) match {
+      case Left(e)  => sys.error(e.message)
```
lwronski: Before exit print exception stack traces using [Logger.log](https://github.com/VirtusLab/scala-cli/blob/518d63de7592b665c3134b5f2b0905e7a8a4abdc/modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala#L126)

modules/cli/src/main/scala/scala/cli/commands/version/Version.scala
Gedochao: uh, that I can't do, since `CheckScalaCliVersionError` isn't a `BuildException`...

lwronski (approved):

modules/cli/src/main/scala/scala/cli/commands/version/Version.scala
Gedochao: to be addressed in #1766

PR #1768, opened 2023-01-11
Add toolkit using directive and cli options

### LINKED ISSUE #1750: Scala Toolkit support
Please add a new `using directive`:
```scala
//> using toolkit latest
//> using toolkit "x.y.z"
```
The first directive should be equivalent to adding a dependency on the library `org.virtuslab::toolkit:latest.release`. This library will be released for jvm, js and native. The second version of the directive, instead of `latest.release`, should use the given version. There is high probability that, with time, the toolkit will be moved to another org on maven central (`org.scala-lang`)

lwronski:
Fixes #1750

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -16,6 +16,7 @@ import scala.build.EitherCps.{either, value}
 import scala.build.*
 import scala.build.blooprifle.BloopRifleConfig
 import scala.build.compiler.{BloopCompilerMaker, ScalaCompilerMaker, SimpleScalaCompilerMaker}
+import scala.build.directives.DirectiveDescription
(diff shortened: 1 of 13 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -180,6 +182,9 @@ final case class SharedOptions(
   @ValueDescription("/example/path")
   @Tag(tags.must)
     compilationOutput: Option[String] = None,
+  @HelpMessage("Add toolkit to classPath")
+  @ValueDescription("version|latest")
+    toolkit: Option[String] = None
```
Gedochao: NIT to be consistent with `--with-compiler`
```suggestion
  @Name("toolkit")
    withToolkit: Option[String] = None
```

Gedochao (approved):
LGTM

PR #1771, opened 2023-01-11
Set workspace dir to `os.tmp.dir` for virtual sources

### LINKED ISSUE #1459: When using process substitution or -e, Scala CLI shouldn't create .scala-build in the current directory
**Version(s)**
`0.1.16`

**Describe the bug**
Running commands such as
```bash
$ scala-cli <(echo 'println("Hello")')
$ scala-cli -e 'println("Hello")'
```
creates a `.scala-build` directory, and uses it to compile the provided input. Scala CLI should rather use a directory under the home directory for that purpose.

lwronski:
Fixes #1459

### DIFF
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -157,6 +157,11 @@ object Inputs {
               )
             (f.path / os.up, true, WorkspaceOrigin.SourcePaths)
         }
+      }.orElse {
+        validElems.collectFirst {
+          case _: Virtual =>
+            (os.temp.dir(), true, WorkspaceOrigin.VirtualForced)
+        }
       }.getOrElse((os.pwd, true, WorkspaceOrigin.Forced))
     }
 
--- a/modules/build/src/main/scala/scala/build/input/WorkspaceOrigin.scala
+++ b/modules/build/src/main/scala/scala/build/input/WorkspaceOrigin.scala
@@ -9,5 +9,6 @@ object WorkspaceOrigin {
 
   case object ResourcePaths extends WorkspaceOrigin
 
-  case object HomeDir extends WorkspaceOrigin
+  case object HomeDir       extends WorkspaceOrigin
+  case object VirtualForced extends WorkspaceOrigin
 }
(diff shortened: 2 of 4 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunPipedSourcesTestDefinitions.scala
```diff
@@ -202,4 +202,14 @@ trait RunPipedSourcesTestDefinitions { _: RunTestDefinitions =>
       expect(output == expectedOutput)
     }
   }
+
+  test("running script snipped should not create workspace dir in cwd") {
```
Gedochao:
```suggestion
  test("running a script snippet should not create the workspace dir in cwd") {
```

modules/integration/src/test/scala/scala/cli/integration/RunPipedSourcesTestDefinitions.scala
```diff
@@ -202,4 +202,14 @@ trait RunPipedSourcesTestDefinitions { _: RunTestDefinitions =>
...
+    emptyInputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "-e", """println("Hello")""", extraOptions)
+        .call(cwd = root)
+        .out.trim()
```
Gedochao: using the right quotation marks is necessary when passing snippets on Windows
```suggestion
      val quotation = TestUtil.argQuotationMark
      val msg = "Hello"
      val res = os.proc(TestUtil.cli, "-e", s"""println($quotation$msg$quotation")""", extraOptions)        
        .call(cwd = root)
        .out.trim()
      expect(res == msg)
```

modules/integration/src/test/scala/scala/cli/integration/RunPipedSourcesTestDefinitions.scala
Gedochao: btw. this test should go to either `RunSnippetTestDefinitions` or `DefaultTests`

modules/integration/src/test/scala/scala/cli/integration/RunPipedSourcesTestDefinitions.scala
```diff
@@ -202,4 +202,14 @@ trait RunPipedSourcesTestDefinitions { _: RunTestDefinitions =>
...
+      expect(!os.exists(root / Constants.workspaceDirName))
+    }
+  }
```
Gedochao: Also, we could have another test confirming the workspace directory **does** get created in the project folder for mixed virtual and on-disk sources, just to be extra sure.

Gedochao (approved):

Issue #1749, opened 2023-01-05
Document how to migrate from the old scala runner to the new one


julienrf:
It would be great to already create a list of incompatibilities and recipes to migrate from the old runner to the new one, in the documentation.

_Originally posted by @julienrf in https://github.com/VirtusLab/scala-cli/issues/1743#issuecomment-1372199016_

Issue #1721, opened 2022-12-22
Support or silently warn about legacy-`scala`-runner-specific options without failing


Gedochao:
**Describe what needs to be done and why**
We need to be backwards compatible with all the options available from the old Scala runner.
This doesn't necessarily mean we have to implement any missing functionality (unless it makes sense), but we shouldn't fail builds when they're passed and a meaningful warning explaining what works and why should be given (similarly to how it's done in #1679).
The ones we still have left to handle:
- Scala 2 options:
  - `-howtorun`
  - `-i`
  - `-I`
  - `-Dname=prop`
  - `-J<arg>`
  - `-nobootcp`
- Scala 3 options:
  - `-run`
  - `-with-compiler`

We should also add a thorough doc on our docs website explaining which legacy options are supported and how.

**Additional context**
- Relevant issues:
  - #1670 
  - #1150 
  - #1310 
  - #1526 
- Relevant PRs:
  - #1679 
  - #1313 
  - #1340
  - #1166 
- Old Scala 2 runner:
  - [options reference](https://github.com/scala/scala/blob/8403e7047458cb0f55dc2da2a49e13bf0bea0974/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala#L93)
  - [options implementation](https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/GenericRunnerSettings.scala)
- Old Scala 3 runner: 
  - [options implementation](https://github.com/lampepfl/dotty/blob/0397de1bba206e4b41e9ba2be4bcd0d418ca05d4/compiler/src/dotty/tools/MainGenericRunner.scala#L124)

Gedochao:
Regarding Scala 2's exclusive [`-howtorun` (or `--how-to-run`)](https://github.com/scala/scala/blob/8403e7047458cb0f55dc2da2a49e13bf0bea0974/src/compiler/scala/tools/nsc/GenericRunnerCommand.scala#L102) option, I wonder if (and if at all, to what degree) should we support it.

As the legacy option is barely documented, let me first explain how it seems to be meant to work.
What it is basically an override switch.
There's 5 available modes (6, if counting the invalid arg): `script`, `object`, `jar`, `repl` and `guess`.
  - `guess` is the assumed default and it means the user trusts the runner knows how to run the inputs which have been passed.
  - `repl` means the runner should go straight into the repl, loading the input on classpath.
[...]

PR #1774, opened 2023-01-13
Add support for deprecated Scala `2.13.x`-specific `scala` runner options

Gedochao:
Context: #1721 

Added the following deprecated options from the Scala 2 runner, along with appropriate warnings:
- `-howtorun`
- `-I`
- `-nc`

All of them are ignored, with deprecation warnings explaining how Scala CLI does things.

I deliberately skipped over `-i`, which will stay as an abbrevation for Scala CLI's `--interactive` mode. It will be documented in #1749.

Did some relevant fixes as well, like adding the missing `-verbose` alias for `-v`.
The legacy `scala` runner options will now also be included in the `--full-help` (they're all `@Hidden`, so not in regular `-help`).

```
Legacy Scala runner options:
  -save, --save                                  (hidden) Ignored legacy option. Deprecated equivalent of running a subsequent `package` command.
  -nosave, --nosave                              (hidden) Ignored legacy option. Deprecated override canceling the `-nosave` option.
  -howtorun, --how-to-run object|script|jar|repl|guess  (hidden) Ignored legacy option. Deprecated override defining how the runner should treat the input. Use the appropriate sub-command instead.
  -I file                                        (hidden) Ignored legacy option. Deprecated option allowing to preload inputs for the repl or command execution.
  -nc, -nocompdaemon, --no-compilation-daemon    (hidden) Ignored legacy option. Deprecated option allowing to prevent the use of the legacy fsc compilation daemon.
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/ScalaCommandWithCustomHelp.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/ScalaCommandWithCustomHelp.scala
@@ -4,7 +4,7 @@ import caseapp.core.Error
 import caseapp.core.help.{Help, HelpCompanion, RuntimeCommandsHelp}
 import caseapp.core.parser.Parser
 
-import scala.cli.commands.default.DefaultOptions
+import scala.cli.commands.default.{DefaultOptions, LegacyScalaOptions}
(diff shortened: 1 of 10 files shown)


lwronski (approved):
LGTM

PR #1775, opened 2023-01-13
Fix reading passwords from commands

alexarchambault:
It seems using `sys.process` adds superfluous new line characters when reading the command output. Better use `ProcessBuilder` / `Process` directly.


Gedochao (commented):
Can we have some sort of anti-regression test for this?
EDIT: and if not, maybe just explain more in some inline comments in the code on why we have to use `ProcessBuilder` rather than `sys.process`...

alexarchambault:
> EDIT: and if not, maybe just explain more in some inline comments in the code on why we have to use `ProcessBuilder` rather than `sys.process`...

Right, I thought I had done that already, but didn't… I just fixed it.

Gedochao (approved):

PR #1777, opened 2023-01-13
Fix a broken link

xerial:


lwronski:
@xerial, Thank you for your contribution. Good catch! I have modified a little  your changes and use an absolute path in the docs link.

MaciejG604 (approved):

Gedochao (approved):

PR #1779, opened 2023-01-16
Suppress warning using flag set in global config

### LINKED ISSUE #1747: Enable globally turning off warnings about `using` directives in multiple files

MaciejG604:


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -733,6 +733,43 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+    )
+    inputs.fromRoot { root =>
+      val warningMessage = "Using directives detected in"
+
+      val outputBeforeConfig =
+        os.proc(TestUtil.cli, ".").call(
+          cwd = root,
+          stderr = os.Pipe
+        ).err.trim()
+      expect(outputBeforeConfig.contains(warningMessage))
```
Gedochao: NIT this part feels unecessary, as we already are testing this part elsewhere

Gedochao (approved):
LGTM

Gedochao:
context: this PR is a spin-off of #1754

PR #1780, opened 2023-01-16
Add support for the `-with-compiler` runner option

Gedochao:
Context: #1721

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
@@ -540,4 +540,32 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
       expect(classFiles.contains(os.rel / "Hello.class"))
     }
   }
+
+  private def compilerArtifactName: String =
+    if (actualScalaVersion.startsWith("3")) "scala3-compiler" else "scala-compiler"
+
+  test(s"ensure the -with-compiler option adds $compilerArtifactName to the classpath") {
+    TestInputs(os.rel / "s.sc" -> """println("Hello")""")
+      .fromRoot { root =>
+        val compileRes = os.proc(
+          TestUtil.cli,
+          "compile",
+          "s.sc",
+          "--print-classpath",
+          extraOptions
+        )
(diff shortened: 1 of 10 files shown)


modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -540,4 +540,38 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
       expect(classFiles.contains(os.rel / "Hello.class"))
     }
   }
+
+  private def compilerArtifactName: String =
+    if (actualScalaVersion.startsWith("3")) "scala3-compiler" else "scala-compiler"
+
+  test(s"ensure the -with-compiler option $compilerArtifactName to the classpath") {
```
MaciejG604: Is there some word missing in the test name? I don't quite get the meaning.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -540,4 +540,38 @@ abstract class CompileTestDefinitions(val scalaVersionOpt: Option[String])
...
+        val compileRes = os.proc(
+          TestUtil.cli,
+          "compile",
+          "s.sc",
+          "-with-compiler",
+          "--print-classpath",
+          extraOptions
+        )
+          .call(cwd = root)
+        expect(compileRes.out.trim().contains(compilerArtifactName))
```
lwronski: condition from the next test could be added to this test, when the `-with-compiler` option is not passed. It should be much faster, because we have already compiled sources.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
Gedochao: yup, typo, one word missing. should be fixed now.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
Gedochao: should be fixed

lwronski (approved):
LGTM

PR #1784, opened 2023-01-16
Allow to pass supported scala version url from command line

lwronski:


modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -288,7 +293,8 @@ final case class SharedOptions(
             dependencies.compilerPlugin.map(Positioned.none),
             ignoreErrors
           ),
-        platform = platformOpt.map(o => Positioned(List(Position.CommandLine()), o))
+        platform = platformOpt.map(o => Positioned(List(Position.CommandLine()), o)),
+        supportedScalaVersionsUrl = supportedScalaVersionsUrl
```
kubukoz, line 297: **question:** is the output cached anywhere? I don't suppose we call this URL on every CLI command

Gedochao:
Can we add some tests?

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
lwronski: @kubukoz, I have another question about the root cause of your problem. Can you attach the error message that you encounter? Is it related to a proxy issue?

kubukoz: It's related to a issue of building inside Nix's sandbox - I need a reproducible way to download everything that needs to be downloaded. I'm planning to post a larger issue with things that would help Nix usage.

Unless other people need this solution, I would suggest not spending too much time until I describe the usecase in more context - maybe there's a better way to solve my problem :)

lwronski: Yes, I agree that providing more context would be much better. I will close this pull request as it isn't very useful for user.

PR #1785, opened 2023-01-17
Fix - suggest actionable update for toolkit

lwronski:


modules/integration/src/test/scala/scala/cli/integration/DependencyUpdateTests.scala
```diff
@@ -34,4 +35,31 @@ class DependencyUpdateTests extends ScalaCliSuite {
       expect(out == message)
     }
   }
+
+  test("update toolkit dependence") {
```
Gedochao, line 39:
```suggestion
  test("update toolkit dependency") {
```

Gedochao (approved):
LGTM

bishabosha:
I think this should special case "latest" when used as the version: #1809

PR #1792, opened 2023-01-18
Quit flag not suppresses compilation errors

### LINKED ISSUE #1800: Quiet flag even suppresses compilation errors
The following seems undesirable. Suppose I have a shebang script that doesn't compile:

`foo.sc`:

```scala
#!/usr/local/bin/env -S scala-cli shebang --quiet
gjkhfhjkgfkjgfhjkghfgjk
```

because `--quiet` is present, when I run it I just get:

```
% ./foo.sc
Compilation failed
```

(Perhaps the fix for #701 was over-broad?)

lwronski:
Fixes #1800


Gedochao:
Some simple test could be added, I think

Gedochao (approved):

PR #1796, opened 2023-01-19
Fix warning about using directives in multiple files when two java files are present

### LINKED ISSUE #1786: False warning about directives in multiple java files
**Version(s)**
0.1.19

**Describe the bug**
If the compiled sources contain more than one java file, a warning about using directives in multiple files  is reported, even if there are no using directives in the source files.

**To Reproduce**
```sh
mkdir /tmp/bug-repro
cd /tmp/bug-repro
echo 'public class Foo {}' > Foo.java
echo 'public class Bar {}' > Bar.java
scala-cli compile .
```

This outputs:
```scala
[warn] ./Bar.java:1:1: Using directives detected in multiple files. It is recommended to keep them centralized in the /private/tmp/bug-repro/project.scala file.
[warn] public class Bar {}
[warn] ^
[warn] ./Foo.java:1:1: Using directives detected in multiple files. It is recommended to keep them centralized in the /private/tmp/bug-repro/project.scala file.
[warn] public class Foo {}
[warn] ^
```

**Expected behaviour**
Compilation should report no warnings

MaciejG604:
Resolves #1786

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
@@ -14,7 +14,7 @@ import scala.build.internal.JavaParserProxyMaker
 import scala.build.options.BuildRequirements
 import scala.build.preprocessing.ExtractedDirectives.from
 import scala.build.preprocessing.PreprocessingUtil.optionsAndPositionsFromDirectives
-import scala.build.preprocessing.ScalaPreprocessor._
+import scala.build.preprocessing.ScalaPreprocessor.*
 
 /** Java source preprocessor.
   *
--- a/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
@@ -89,11 +89,19 @@ object ExtractedDirectives {
       val specialCommentDirectives = byKind(UsingDirectiveKind.SpecialComment)
       val plainCommentDirectives   = byKind(UsingDirectiveKind.PlainComment)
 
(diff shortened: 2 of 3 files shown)


modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
```diff
@@ -53,7 +53,11 @@ final case class JavaPreprocessor(
               allowRestrictedFeatures
             )
           }
-          Seq(PreprocessedSource.OnDisk(
+          if (
+            updatedOptions.global == BuildOptions.monoid.zero && updatedOptions.scoped.isEmpty && updatedOptions.unused.isEmpty
```
lwronski: Move this condition to the class `DirectivesProcessorOutput`.

MaciejG604: Moved checking for no directives to the `ExtractedDirectives` class, I think it makes much more sense there.

lwronski: Good catch!

modules/build/src/main/scala/scala/build/preprocessing/JavaPreprocessor.scala
```diff
@@ -11,10 +11,10 @@ import scala.build.Logger
 import scala.build.errors.BuildException
 import scala.build.input.{Inputs, JavaFile, SingleElement, VirtualJavaFile}
 import scala.build.internal.JavaParserProxyMaker
-import scala.build.options.BuildRequirements
+import scala.build.options.{BuildOptions, BuildRequirements}
```
lwronski: Remove unused import.

lwronski (approved):
LGTM

PR #1799, opened 2023-01-19
Ignore Bloop server early exit if it signals an already running server

alexarchambault:


modules/bloop-rifle/src/main/scala/scala/build/blooprifle/internal/Operations.scala
```diff
@@ -236,7 +236,7 @@ object Operations {
       () =>
         try {
           val completionOpt =
-            if (!p.isAlive())
+            if (!p.isAlive() && p.exitValue() != 222)
```
Gedochao: maybe leave a comment here, indicating what the magic number `222` stands for?
smth like:
```suggestion
            // 222 is a special code indicating a server is already running, refer to https://github.com/scala-cli/bloop-core/pull/128
            if (!p.isAlive() && p.exitValue() != 222) 
```

alexarchambault: Should be fixed

Gedochao (approved):
LGTM

PR #1802, opened 2023-01-23
Add support for running script files with no extension using shebang subcommand

### LINKED ISSUE #466: Allow executable scripts to not have any suffix (.sc or .scala) at all?
When you're using Scala as a scripting language, with `#!/usr/bin/env scala-cli shebang`, it would be nice if I could just have an executable script called `foo` and not have to name it `foo.sc` or `foo.scala`.

For me as the author of the script having the suffix is not so bad, but for those who are _using_ my script, it might be weird and annoying to have to say `foo.sc`; they would prefer to just run the script as `foo` and not even care whether it's written in C or Shell or Scala or what have you. To the user it's just some executable thing. Perhaps it's even on their `$PATH`.

I've been writing scripts for decades now in Perl, Bash, Scala, and assorted other languages and none of those languages have disallowed me from using a bare filename. (I have Scala on that list because until recently I was using sbt's `ScriptMain` support, which allowed this.)

MaciejG604:
Resolves #466

### DIFF
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -13,6 +13,7 @@ import scala.build.internal.Constants
 import scala.build.internal.zip.WrappedZipInputStream
 import scala.build.options.Scope
 import scala.build.preprocessing.ScopePath
+import scala.build.preprocessing.SheBang.isShebangScript
 import scala.util.Properties
 import scala.util.matching.Regex
 
@@ -261,7 +262,8 @@ object Inputs {
     download: String => Either[String, Array[Byte]],
     stdinOpt: => Option[Array[Byte]],
     acceptFds: Boolean,
-    enableMarkdown: Boolean
+    enableMarkdown: Boolean,
+    isRunWithShebang: Boolean
   ): Seq[Either[String, Seq[Element]]] = args.zipWithIndex.map {
     case (arg, idx) =>
       lazy val path      = os.Path(arg, cwd)
@@ -297,10 +299,21 @@ object Inputs {
       else if os.isDir(path) then Right(Seq(Directory(path)))
       else if acceptFds && arg.startsWith("/dev/fd/") then
         Right(Seq(VirtualScript(content, arg, os.sub / s"input-${idx + 1}.sc")))
+      else if isRunWithShebang && os.exists(path) then
+        if isShebangScript(String(content)) then Right(Seq(Script(dir, subPath)))
+        else
+          Left(s"""$arg does not contain shebang header
+                  |possible fixes:
+                  |  Add '#!/usr/bin/env scala-cli shebang' to the top of the file
+                  |  Add extension to the file's name e.q. '.sc'
+                  |""".stripMargin)
       else {
         val msg =
-          if (os.exists(path))
-            s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
+          if os.exists(path) then
+            if isShebangScript(String(content)) then
+              s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
+            else
+              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
           else s"$arg: not found"
         Left(msg)
       }
@@ -320,10 +333,11 @@ object Inputs {
     forcedWorkspace: Option[os.Path],
     enableMarkdown: Boolean,
     allowRestrictedFeatures: Boolean,
-    extraClasspathWasPassed: Boolean
+    extraClasspathWasPassed: Boolean,
+    isRunWithShebang: Boolean
   ): Either[BuildException, Inputs] = {
     val validatedArgs: Seq[Either[String, Seq[Element]]] =
-      validateArgs(args, cwd, download, stdinOpt, acceptFds, enableMarkdown)
+      validateArgs(args, cwd, download, stdinOpt, acceptFds, enableMarkdown, isRunWithShebang)
     val validatedSnippets: Seq[Either[String, Seq[Element]]] =
       validateSnippets(scriptSnippetList, scalaSnippetList, javaSnippetList, markdownSnippetList)
     val validatedArgsAndSnippets = validatedArgs ++ validatedSnippets
@@ -364,7 +378,8 @@ object Inputs {
     forcedWorkspace: Option[os.Path] = None,
     enableMarkdown: Boolean = false,
     allowRestrictedFeatures: Boolean,
-    extraClasspathWasPassed: Boolean
+    extraClasspathWasPassed: Boolean,
+    isRunWithShebang: Boolean
   ): Either[BuildException, Inputs] =
     if (
       args.isEmpty && scriptSnippetList.isEmpty && scalaSnippetList.isEmpty && javaSnippetList.isEmpty &&
@@ -388,7 +403,8 @@ object Inputs {
         forcedWorkspace,
         enableMarkdown,
         allowRestrictedFeatures,
-        extraClasspathWasPassed
+        extraClasspathWasPassed,
+        isRunWithShebang
       )
 
   def default(): Option[Inputs] = None
--- a/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
@@ -5,6 +5,8 @@ import scala.util.matching.Regex
 object SheBang {
   private val sheBangRegex: Regex = s"""(^(#!.*(\\r\\n?|\\n)?)+(\\s*!#.*)?)""".r
 
+  def isShebangScript(content: String): Boolean = sheBangRegex.unanchored.matches(content)
+
   def ignoreSheBangLines(content: String): (String, Boolean) =
     if (content.startsWith("#!")) {
       val regexMatch = sheBangRegex.findFirstMatchIn(content)
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -177,7 +177,7 @@ final case class SharedOptions(
   @Recurse
     input: SharedInputOptions = SharedInputOptions(),
   @Recurse
-  helpGroups: HelpGroupOptions = HelpGroupOptions(),
+    helpGroups: HelpGroupOptions = HelpGroupOptions(),
 
   @Hidden
     strictBloopJsonCheck: Option[Boolean] = None,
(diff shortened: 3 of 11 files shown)


modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -297,6 +298,7 @@ object Inputs {
       else if os.isDir(path) then Right(Seq(Directory(path)))
       else if acceptFds && arg.startsWith("/dev/fd/") then
         Right(Seq(VirtualScript(content, arg, os.sub / s"input-${idx + 1}.sc")))
+      else if isRunWithShebang then Right(Seq(Script(dir, subPath)))
```
lwronski: Check if file has no extension.

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -177,11 +177,16 @@ final case class SharedOptions(
...
-  helpGroups: HelpGroupOptions = HelpGroupOptions(),
+    helpGroups: HelpGroupOptions = HelpGroupOptions(),
 
   @Hidden
     strictBloopJsonCheck: Option[Boolean] = None,
 
+  @Hidden
+  @Tag(tags.implementation)
+  @HelpMessage("Has the run been initiated using the shebang subcommand")
+    runWithShebang: Option[Boolean] = None,
```
lwronski: Do we need this option from command line? cc: @Gedochao

lwronski (commented):
[Documentation](https://scala-cli.virtuslab.org/docs/commands/shebang) and [guide](https://scala-cli.virtuslab.org/docs/guides/shebang) for shebang should be also updated.

modules/build/src/main/scala/scala/build/input/Inputs.scala
Gedochao: technically if it has some extension that isn't supported by Scala CLI, but it does have the `shebang` header, I think we should be able to run it...

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
Gedochao: I don't think we do, no.
it's only `run with shebang` if it's, well, run with the `shebang` sub-command, really. 😅 
a flag in the `sharedOptions` would enable to pass it from almost anywhere (even from `fmt` or `repl`...), which is something we definitely don't want.

Gedochao: I guess that maaaaybe we could have a hidden flag like this in the `run` sub-command only... but I dunno

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -326,4 +326,18 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
         expect(p.out.trim() == "List(1, 2, 3, -v)")
       }
     }
+
+  test("script file with shebang line and no extension run with scala-cli shebang") {
```
Gedochao: we might want an extra sanity check verifying that an input without the shebang header can't be run if it has no supported extension.
so another test beyond this one, maybe?

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
MaciejG604: Don't we need something to verify that the `run` is actually originating from calling `shebang` to accept files with no extensions?

Gedochao: We do, but it shouldn't be through an option available from most sub-commands.
You can add it as a hidden option for the `run` sub-command only (`SharedRunOptions`)... That'd be fine, I think.
Or else bypass the `RunOptions` altogether and pass it straight to the run logic, up to you

MaciejG604: Sure, run-only options make much more sense

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -297,6 +298,8 @@ object Inputs {
       else if os.isDir(path) then Right(Seq(Directory(path)))
       else if acceptFds && arg.startsWith("/dev/fd/") then
         Right(Seq(VirtualScript(content, arg, os.sub / s"input-${idx + 1}.sc")))
+      else if isRunWithShebang && String(content).startsWith("#!") then
```
Gedochao: we have a regex for shebang headers defined at `scala.build.preprocessing.SheBang.sheBangRegex`, we could check against it as well

modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
```diff
@@ -5,6 +5,11 @@ import scala.util.matching.Regex
 object SheBang {
   private val sheBangRegex: Regex = s"""(^(#!.*(\\r\\n?|\\n)?)+(\\s*!#.*)?)""".r
 
+  def isShebangScript(content: String): Boolean = sheBangRegex.findFirstMatchIn(content) match {
+    case Some(_) => true
+    case None    => false
+  }
```
Gedochao:
```suggestion
  def isShebangScript(content: String): Boolean = sheBangRegex.matches(content)
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -326,4 +326,29 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+    val inputs = TestInputs(
+      os.rel / "script-with-shebang" ->
+        s"""|#!/usr/bin/env -S ${TestUtil.cli.mkString(" ")} shebang -S 2.13
+            |//> using scala "$actualScalaVersion"
+            |println(args.toList)""".stripMargin,
+      os.rel / "script-no-shebang" ->
```
Gedochao, line 351: NIT better to do the negative test (with no shebang) in a separate test clause, the test result logs are more readable then

website/docs/commands/shebang.md
```diff
@@ -5,6 +5,7 @@ sidebar_position: 30
 
 This command is equivalent to `run`, but it changes the way Scala CLI parses options (used to configure the tool) and
 inputs (the sources of your project) in order to be compatible with `shebang` scripts.
+The command `shebang` also allows for executing script files that don't have any file extensions.
```
Gedochao:
```suggestion
The command `shebang` also allows script files to be executed even if they have no file extension, provided they start with the [`shebang` header](../guides/shebang.md#shebang-script-headers).
```

modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
MaciejG604: This tries to match whole `content`, that's why I changed it to a longer version. But after looking more into the documentation `sheBangRegex.unanchored.matches(content)` should work.

Gedochao: Or just
```suggestion
  def isShebangScript(content: String): Boolean = sheBangRegex.findFirstMatchIn(content).isDefined
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -326,4 +326,35 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
         expect(p.out.trim() == "List(1, 2, 3, -v)")
       }
     }
+
+  test("script file with shebang header and no extension run with scala-cli shebang") {
```
Gedochao, line 330: Both tests seem to fail on Windows, as it doesn't have shebang support by default (unless with `cygwin`, but I don't think it's available on our CI). I think we can simply skip them on Windows.

```suggestion
  if (!Properties.isWin) {
    test("script file with shebang header and no extension run with scala-cli shebang") {
```

Gedochao (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -326,4 +326,35 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+  }
+
+  test("script file with NO shebang header and no extension run with scala-cli shebang") {
+    val inputs = TestInputs(
+      os.rel / "script-no-shebang" ->
+        s"""//> using scala "$actualScalaVersion"
+           |println(args.toList)""".stripMargin
+    )
+    inputs.fromRoot { root =>
+      os.perms.set(root / "script-no-shebang", os.PermSet.fromString("rwx------"))
```
Gedochao: actually, this test should work on Windows once this line is removed, I think
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -326,4 +326,35 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+  test("script file with shebang header and no extension run with scala-cli shebang") {
+    val inputs = TestInputs(
+      os.rel / "script-with-shebang" ->
+        s"""|#!/usr/bin/env -S ${TestUtil.cli.mkString(" ")} shebang -S 2.13
+            |//> using scala "$actualScalaVersion"
+            |println(args.toList)""".stripMargin
+    )
+    inputs.fromRoot { root =>
+      os.perms.set(root / "script-with-shebang", os.PermSet.fromString("rwx------"))
+      val output = os.proc("./script-with-shebang", "1", "2", "3", "-v").call(cwd = root).out.trim()
```
Gedochao: Alternatively, a version that should work for Windows:
```suggestion
      val output = 
        (if (!Properties.isWin) {
          os.perms.set(root / "script-with-shebang", os.PermSet.fromString("rwx------"))
          os.proc("./script-with-shebang", "1", "2", "3", "-v")
        } else 
          os.proc(TestUtil.cli, "shebang", "script-no-shebang", "1", "2", "3", "-v")
          ).call(cwd = root).out.trim()
```

julienrf (commented):
Thank you for working on this issue!

Regarding the documentation, I wonder if the “shebang” guide and the “scripts” guide should be merged? At least, I would expect the information that files with no extension are supported to be mentioned in the scripts guide.

I think it is important also to highlight better that the semantics of files with no extension is always a “script” semantics, not a “program” semantics.

PR #1803, opened 2023-01-23
Dont warn about target directives

### LINKED ISSUE #1632: frustrating warning for using target.platform "scala-js" directive
I'm frustrated when I would like to have some files compile only when Scala.js is picked as a target, however when I put using directives (`//> using target.platform "scala-js"`) in that file, I get a warning to put the using directive in `project.scala`, but I don't want to do that because I don't want the whole project to be scala-js.

Concretely for directories structured like:

```
- src
  - file1.scala (compiles any platform)
  - file2.scala (scala.js only)
- project.scala
```

not see these warnings for files with a different target, or suggest a way to scope files for target platforms in `project.scala`

perhaps having directories that are scoped to specific platforms?

**Additional context**
N/A

MaciejG604:
Resolves #1632

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
@@ -11,6 +11,7 @@ import com.virtuslab.using_directives.{Context, UsingDirectivesProcessor}
 
 import scala.annotation.targetName
 import scala.build.errors.*
+import scala.build.preprocessing.UsingDirectivesOps.*
 import scala.build.preprocessing.directives.{DirectiveUtil, ScopedDirective, StrictDirective}
 import scala.build.{Logger, Position}
 import scala.collection.mutable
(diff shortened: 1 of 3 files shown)


modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -90,10 +90,16 @@ object ExtractedDirectives {
...
-          plainCommentDirectives.getFlattenedMap.isEmpty
+        if ( // check if there's no directives or all of them are target directives
+          codeDirectives.getFlattenedMap.keySet().asScala.forall(
+            _.toString.startsWith("target.")
+          ) &&
+          specialCommentDirectives.getFlattenedMap.keySet().asScala.forall(
+            _.toString.startsWith("target.")
+          ) &&
+          plainCommentDirectives.getFlattenedMap.keySet().asScala.forall(
+            _.toString.startsWith("target.")
```
Gedochao: NIT extracting this to a util would greatly improve the readability of this `if` clause IMO
so basically, we could have a util object somewhere with:
```scala
extension (ud: UsingDirective) {
  def keySet: Set[String] = ud.getFlattenedMap.keySet().asScala
  def containsTargetDirectivesOnly: Boolean = ud.keySet.forall(_.toString.startsWith("target."))
}
```

and then we could

```suggestion
          codeDirectives.containsTargetDirectivesOnly &&
          specialCommentDirectives.containsTargetDirectivesOnly &&
          plainCommentDirectives.containsTargetDirectivesOnly
```

Gedochao (approved):

PR #1804, opened 2023-01-23
Add warnings for the deprecated `-Yscriptrunner` legacy `scala` runner option instead of passing it to `scalac`

### LINKED ISSUE #1721: Support or silently warn about legacy-`scala`-runner-specific options without failing

Gedochao:
Fixes #1721 
Depends on #1801 (relies on some of the refactors from there)
Despite the `-Y` prefix, `-Yscriptrunner` is actually a `scala` runner option (and not a `scalac` option) and thus will be covered as part of #1721.
Up until now, Scala CLI has been assuming it's a `scalac` option and has been passing it to the compiler.
However, the compiler doesn't know what to do with it.
```
▶ scalac -Yscriptrunner resident
bad option '-Yscriptrunner' was ignored
source file not found: resident
▶ scala-cli -Yscriptrunner -O resident
bad option '-Yscriptrunner' was ignored
Ignoring spurious arguments: resident
```

With these changes, we'll be logging a reasonable warning about the option being deprecated and no longer supported in Scala CLI (it being the new `scala` runner). We'll no longer pass the option to the compiler, either.
```
▶ scala-cli . -Yscriptrunner resident 
Deprecated option '-Yscriptrunner' is ignored.
The script runner can no longer be picked as before.
scala.tools.nsc.fsc.ResidentScriptRunner (the resident script runner) is no longer available.
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
@@ -35,12 +35,13 @@ object ScalacOptions {
     origin = Some("ScalacOptions")
   )
   // .withIsFlag(true) // The scalac options we handle accept no value after the -… argument
+  val YScriptRunnerOption = "-Yscriptrunner"
   private val scalacOptionsPurePrefixes =
     Set("-V", "-W", "-X", "-Y")
   private val scalacOptionsPrefixes =
     Set("-g", "-language", "-opt", "-P", "-target", "-source") ++ scalacOptionsPurePrefixes
   private val scalacAliasedOptions = // these options don't require being passed after -O and accept an arg
-    Set("-encoding", "-release", "-color")
+    Set("-encoding", "-release", "-color", YScriptRunnerOption)
   private val scalacNoArgAliasedOptions = // these options don't require being passed after -O and don't accept an arg
     Set(
       "-nowarn",
@@ -64,6 +65,9 @@ object ScalacOptions {
     "-classpath", // redirected to --extra-jars
     "-d"          // redirected to --compilation-output
   )
+  val ScalacDeprecatedOptions: Set[String] = Set(
+    YScriptRunnerOption // old 'scala' runner specific, no longer supported
+  )
 
   private val scalacOptionsArgument: Argument[List[String]] =
     new Argument[List[String]] {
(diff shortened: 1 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
```diff
@@ -81,7 +85,7 @@ object ScalacOptions {
         formatter: Formatter[Name]
       ): Either[(Error, List[String]), Option[(Option[List[String]], List[String])]] =
         args match {
-          case h :: t if scalacOptionsPrefixes.exists(h.startsWith) =>
+          case h :: t if scalacOptionsPrefixes.exists(h.startsWith) && h != YScriptRunnerOption =>
```
lwronski: You can if `ScalacDeprecatedOptions` set doesn't contain `h`,

lwronski (approved):
LGTM.

PR #1810, opened 2023-01-26
Add cross-platform toolkit dependency

### LINKED ISSUE #1808: using toolkit with native does not select the native dependencies
**Version(s)**
0.1.20

**Describe the bug**

`using toolkit "0.1.4"` does not pick native dependencies for Scala Native.

**To Reproduce**
The following fails to link with `scala-cli --native ls.scala`
```scala
//> using toolkit "0.1.4"

@main def ls =
  os.list(os.pwd).foreach(println)
```

**Expected behaviour**
behave as if I put the explicit `lib` directive

bishabosha:
this should select the toolkit dependency that matches the current platform.


lwronski (approved):
@bishabosha Thanks! LGTM!

PR #1811, opened 2023-01-27
Invalid subcommand arg

### LINKED ISSUE #1240: Show help document if for the invalid subcommand

MaciejG604:


modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -313,8 +314,17 @@ object Inputs {
...
             else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+              s"""$arg: unrecognized source type (expected .scala or .sc extension, or a directory),
+                 |if it's meant to be a script add a '!#' pointing to 'scala-cli shebang' in the top line
+                 |and run the source with 'scala-cli shebang $arg' or ${
+                  if path.toString == arg then arg else s"./$arg"
+                }
+                 |""".stripMargin
+          else if isRunWithDefault && idx == 0 && !arg.contains('.') then
```
Gedochao:
```suggestion
          else if isRunWithDefault && idx == 0 && arg.forall(_.isLetterOrDigit) then
```

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -313,8 +314,17 @@ object Inputs {
             if isShebangScript(String(content)) then
               s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
             else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+              s"""$arg: unrecognized source type (expected .scala or .sc extension, or a directory),
```
Gedochao, line 317: shouldn't explicitly use `scala-cli` here, we should switch to the actual command name that's been used... we probably need to pass `progName` from `cli`.
It's probably also safer to use `Shebang.names.headOption.map(_.mkString(" ")).getOrElse(Shebang.name)` instead of hardcoding `shebang` in the string 

So something like:
```suggestion
                 |and run the source with '$progName ${Shebang.name} $arg' or ${
```

Doesn't have to be fixed in this PR, as this applies to other strings in `Inputs.scala` as well, just noting

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,7 +263,8 @@ object Inputs {
     stdinOpt: => Option[Array[Byte]],
     acceptFds: Boolean,
     enableMarkdown: Boolean,
-    isRunWithShebang: Boolean
+    isRunWithShebang: Boolean,
+    isRunWithDefault: Boolean
```
Gedochao: it might be a better idea to handle this by an `enum` rather than a myriad of boolean flags
```suggestion
    invokingCommand: ScalaCliCommandRef
```

Gedochao:
`scala.cli.integration.SipScalaTests.test directories command when run as*` needs to be adjusted, as it relies on the old error message

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -313,8 +314,17 @@ object Inputs {
             if isShebangScript(String(content)) then
               s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
             else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+              s"""$arg: unrecognized source type (expected .scala or .sc extension, or a directory),
+                 |if it's meant to be a script add a '!#' pointing to 'scala-cli shebang' in the top line
+                 |and run the source with 'scala-cli shebang $arg' or ${
```
Gedochao: shouldn't explicitly use `scala-cli` here, we should switch to the actual command name that's been used... we probably need to pass `progName` from `cli`.
It's probably also safer to use `Shebang.names.headOption.map(_.mkString(" ")).getOrElse(Shebang.name)` instead of hardcoding `shebang` in the string 

So something like:
```suggestion
                 |and run the source with '$progName ${Shebang.name} $arg' or ${
```

Doesn't have to be fixed in this PR, as this applies to other strings in `Inputs.scala` as well, just noting

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -313,8 +314,17 @@ object Inputs {
             if isShebangScript(String(content)) then
               s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
             else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+              s"""$arg: unrecognized source type (expected .scala or .sc extension, or a directory),
+                 |if it's meant to be a script add a '!#' pointing to 'scala-cli shebang' in the top line
```
Gedochao: also, I wonder if we should print the same error on Windows... 🤔 
`shebang` headers don't normally work there, unless used with `cygwin`...

modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
```diff
@@ -40,7 +40,7 @@ object Bsp extends ScalaCommand[BspOptions] {
       argsSeq =>
         either {
           val sharedOptions = getSharedOptions()
-          val initialInputs = value(sharedOptions.inputs(argsSeq, () => Inputs.default()))
+          val initialInputs = value(sharedOptions.inputs(argsSeq, ScalaCliInvokeData(progName, actualCommandName, SubCommand.Other), () => Inputs.default()))
```
lwronski: Creation of this object `ScalaCliInvokeData ` move to `ScalaCommand`.

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -304,17 +311,27 @@ object Inputs {
         else
           Left(s"""$arg does not contain shebang header
                   |possible fixes:
-                  |  Add '#!/usr/bin/env scala-cli shebang' to the top of the file
+                  |  Add '#!/usr/bin/env $fullProgramCall' to the top of the file
```
Gedochao: we should always advise to use the `shebang` sub-command in the `shebang` header, I think
```suggestion
                  |  Add '#!/usr/bin/env ${programInvokeData.progName} shebang' to the top of the file
```

MaciejG604: In this If branch it's known that shebang command has been called first, that's why it's full call name here.

Gedochao: Ah yeah... got confused, y're right

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -304,17 +312,27 @@ object Inputs {
...
```
lwronski: This error shouldn't be displayed on Windows.  

`#!/usr/bin/env $fullProgramCall` it doesn't work on Windows.

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,14 +263,22 @@ object Inputs {
...
       lazy val subPath   = path.subRelativeTo(dir)
       lazy val stdinOpt0 = stdinOpt
       lazy val content   = os.read.bytes(path)
+      val (isRunWithShebang, isRunWithDefault) = programInvokeData.subCommand match
+        case SubCommand.Shebang => (true, false)
+        case SubCommand.Default => (false, true)
+        case _                  => (false, false)
+      lazy val fullProgramCall = programInvokeData.progName +
+        s"${if isRunWithDefault then "" else s" ${programInvokeData.subCommandName}"}"
+      lazy val progName = os.Path(programInvokeData.progName, cwd).baseName
```
lwronski: What is the value of `progName` if someone is running `scala-cli` from PATH? For example:
```
scala-cli shebang script
```

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -304,17 +312,27 @@ object Inputs {
...
         val msg =
           if os.exists(path) then
             if isShebangScript(String(content)) then
-              s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
+              s"$arg scripts with no file extension should be run with '$progName shebang'"
             else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+              s"""$arg: unrecognized source type (expected .scala or .sc extension, or a directory),
+                 |if it's meant to be a script add a '!#' pointing to '$progName shebang' in the top line
```
lwronski: The same as above, `!#` doesn't work with Windows.

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,14 +263,22 @@ object Inputs {
...
     case (arg, idx) =>
       lazy val path      = os.Path(arg, cwd)
       lazy val dir       = path / os.up
       lazy val subPath   = path.subRelativeTo(dir)
       lazy val stdinOpt0 = stdinOpt
       lazy val content   = os.read.bytes(path)
+      val (isRunWithShebang, isRunWithDefault) = programInvokeData.subCommand match
+        case SubCommand.Shebang => (true, false)
+        case SubCommand.Default => (false, true)
+        case _                  => (false, false)
```
lwronski:
```suggestion
      val (isRunWithShebang, isRunWithDefault) =  
          (programInvokeData.subCommand.is(SubCommand.Shebang),
           programInvokeData.subCommand.is(SubCommand.Default))
```
Maybe this approach will be more readable.

modules/build/src/main/scala/scala/build/input/Inputs.scala
Gedochao: I mean, technically `shebang` headers can be made to work on Windows with `cygwin`... but I agree, we shouldn't suggest it in our terminal outputs (although maybe we could try doing a cookbook on it sometime...)

modules/build/src/main/scala/scala/build/input/Inputs.scala
MaciejG604: It's `scala-cli` when running from PATH. But actually I think it would be better to use `baseRunnerName` instead of `progName`, as `progName` may be a path so the output of errors will be less readable. What do you think?

modules/build/src/main/scala/scala/build/input/Inputs.scala
MaciejG604: But still, even on Windows we will check whether the first line is a shebang header when `scala-cli shebang` is used. Should we suggest that the Windows users put just "#!" in the first line?

modules/build/src/main/scala/scala/build/input/Inputs.scala
lwronski: Will be sufficient if we use `progName`.

modules/build/src/main/scala/scala/build/input/ScalaCliInvokeData.scala
```diff
@@ -0,0 +1,37 @@
...
+case class ScalaCliInvokeData(progName: String, subCommandName: String, subCommand: SubCommand)
+
+enum SubCommand:
+  case Default extends SubCommand
+  case Shebang extends SubCommand
+  case Other   extends SubCommand
+
+object ScalaCliInvokeData {
+
+  // Copied from https://github.com/scalacenter/bloop/blob/main/shared/src/main/scala/bloop/io/Environment.scala
```
lwronski: Better to use permalink - https://github.com/scalacenter/bloop/blob/a249e0a710ce169ca05d0606778f96f44a398680/shared/src/main/scala/bloop/io/Environment.scala

modules/build/src/main/scala/scala/build/input/ScalaCliInvokeData.scala
```diff
@@ -0,0 +1,37 @@
...
+    "/bin/pdksh",
+    "/bin/posh",
+    "/bin/tcsh",
+    "/bin/zsh",
+    "/bin/fish"
+  )
+
+  def isShebangCapableShell = Option(System.getenv("SHELL")) match
+    case Some(currentShell) if shebangCapableShells.exists(sh => currentShell.contains(sh)) => true
+    case _                                                                                  => false
```
lwronski: Move this functions to some utils object, for example to `scala.cli.internal.ProcUtil`.

lwronski (approved):

modules/build/src/main/scala/scala/build/input/ScalaCliInvokeData.scala
```diff
@@ -0,0 +1,21 @@
...
+  *   the actual Scala CLI program name which was run
+  * @param subCommand
+  *   the sub-command that was invoked by user
+  */
+
+case class ScalaCliInvokeData(
+  progName: String,
+  subCommandName: String,
+  subCommand: SubCommand,
+  isShebangCabaleShell: Boolean
```
Gedochao:
```suggestion
  isShebangCapableShell: Boolean
```

modules/build/src/main/scala/scala/build/input/ScalaCliInvokeData.scala
```diff
@@ -0,0 +1,21 @@
+package scala.build.input
+
+/** Stores information about how the program has been evoked
+  *
+  * @param progName
+  *   the actual Scala CLI program name which was run
+  * @param subCommand
+  *   the sub-command that was invoked by user
+  */
```
Gedochao, line 13: 2 params missing from the scaladoc

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,14 +263,25 @@ object Inputs {
...
+      lazy val subPath          = path.subRelativeTo(dir)
+      lazy val stdinOpt0        = stdinOpt
+      lazy val content          = os.read.bytes(path)
+      val isRunWithShebang      = programInvokeData.subCommand == SubCommand.Shebang
+      val isRunWithDefault      = programInvokeData.subCommand == SubCommand.Default
+      val isShebangCapableShell = programInvokeData.isShebangCabaleShell
+
+      lazy val fullProgramCall = programInvokeData.progName +
+        s"${if isRunWithDefault then "" else s" ${programInvokeData.subCommandName}"}"
+      lazy val progName = programInvokeData.progName
```
Gedochao: NIT as above, seems to be unnecessary

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,14 +263,25 @@ object Inputs {
...
-      lazy val subPath   = path.subRelativeTo(dir)
-      lazy val stdinOpt0 = stdinOpt
-      lazy val content   = os.read.bytes(path)
+      lazy val path             = os.Path(arg, cwd)
+      lazy val dir              = path / os.up
+      lazy val subPath          = path.subRelativeTo(dir)
+      lazy val stdinOpt0        = stdinOpt
+      lazy val content          = os.read.bytes(path)
+      val isRunWithShebang      = programInvokeData.subCommand == SubCommand.Shebang
+      val isRunWithDefault      = programInvokeData.subCommand == SubCommand.Default
```
Gedochao: having 2 separate vals for this seems verbose, if it's run with shebang, it's definitely not run with default and vice versa...
if you really want the descriptive names to be used here, you can make `isRunWithShebang` and `isRunWithDefault` methods of `ScalaCliInvokeData`.
or just do the comparison on the fly, it's self-descriptive

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -263,14 +263,25 @@ object Inputs {
...
+      val isShebangCapableShell = programInvokeData.isShebangCabaleShell
```
Gedochao: NIT seems unnecessary, just use `programInvokeData.isShebangCapableShell` further down

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -302,19 +313,30 @@ object Inputs {
...
       else {
         val msg =
           if os.exists(path) then
-            if isShebangScript(String(content)) then
-              s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
-            else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+            if isShebangCapableShell && isShebangScript(String(content)) then
+              s"$arg scripts with no file extension should be run with '$progName shebang'"
```
Gedochao:
```suggestion
              s"$arg scripts with no file extension should be run with '$progName ${Shebang.name}'"
```

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -302,19 +313,30 @@ object Inputs {
...
+          else unrecognizedSourceError)
       else {
         val msg =
           if os.exists(path) then
-            if isShebangScript(String(content)) then
-              s"$arg scripts with no file extension should be run with 'scala-cli shebang'"
-            else
-              s"$arg: unrecognized source type (expected .scala or .sc extension, or a directory)"
-          else s"$arg: not found"
+            if isShebangCapableShell && isShebangScript(String(content)) then
```
Gedochao: NIT I feel a match clause on `ScalaCliInvokeData` could improve readability of this overly complex nest of if clauses...

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -511,9 +511,8 @@ final case class SharedOptions(
 
   def inputs(
     args: Seq[String],
-    defaultInputs: () => Option[Inputs] = () => Inputs.default(),
-    isRunWithShebang: Boolean = false
-  ): Either[BuildException, Inputs] =
+    defaultInputs: () => Option[Inputs] = () => Inputs.default()
+  )(implicit programInvokeData: ScalaCliInvokeData): Either[BuildException, Inputs] =
```
Gedochao: NIT might be better to use [Scala 3's `using`/`given` syntax](https://docs.scala-lang.org/scala3/book/ca-given-using-clauses.html), rather than the old implicits for passing `ScalaCliInvokeData` around
```suggestion
  )(using programInvokeData: ScalaCliInvokeData): Either[BuildException, Inputs] =
```

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
```diff
@@ -55,7 +55,23 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
       args.remaining,
       args.unparsed,
       () => Inputs.default(),
-      logger
+      logger,
+      invokeData
+    )
+
+  def runCommand(
```
Gedochao: NIT adding a third layer of `runCommand` seems overkill here...

website/docs/guides/shebang.md
```diff
@@ -218,18 +218,12 @@ scala-cli shebang hello-no-shebang Hello World
...
 <!-- Expected:
-[error]  hello-no-shebang does not contain shebang header
-possible fixes:
-  Add '#!/usr/bin/env scala-cli shebang' to the top of the file
-  Add extension to the file's name e.q. '.sc'
+hello-no-shebang: unrecognized source type (expected .scala or .sc extension, or a directory)
 -->
 
-Note that files with no extensions are always run as scripts even though they may contain e.g. valid `.scala` program.
+Please note that files with no extensions are always run as scripts even though they may contain e.g. valid `.scala` program.
```
Gedochao:
```suggestion
:::note
Files with no extensions are always run as scripts, even though they may contain e.g. valid `.scala` program.
:::
```

modules/build/src/main/scala/scala/build/input/Inputs.scala
MaciejG604: This will introduce a circular dependency between modules `cli` and `build`. Let's address this in a different PR.

PR #1818, opened 2023-01-30
Add support for entrypoint annotation

alexarchambault:


modules/directives/src/main/scala/scala/build/preprocessing/directives/Entrypoints.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.build.directives.*
+import scala.build.errors.BuildException
+import scala.build.options.{BuildOptions, PostBuildOptions}
+import scala.cli.commands.SpecificationLevel
+
+@DirectiveExamples("//> using entrypoints")
+@DirectiveUsage("//> using entrypoints", "`//> using entrypoints`")
+@DirectiveDescription("Enable entrypoint annotation support")
+@DirectiveLevel(SpecificationLevel.EXPERIMENTAL)
+final case class Entrypoints(
```
tgodzik, line 12: Sorry for the drive-by comment, but couldn't this be added just by using `//> using lib "com.github.alexarchambault::case-app-entrypoint-annotation:0.1.0-SNAPSHOT"` ? Is a separate directive needed?

modules/integration/src/test/scala/scala/cli/integration/RunEntrypointsTests.scala
```diff
@@ -0,0 +1,123 @@
...
+  val extraArgs = Seq(
+    "--scala=3.3.0-RC1-bin-20230120-d6cc101-NIGHTLY",
+    "--server=false",
+    "-r",
+    "sonatype:snapshots"
+  )
+
+  test("simple") {
+    val inputs = TestInputs(
+      os.rel / "Hello.scala" ->
```
Gedochao, line 16: would `@entrypoint` also work with `.sc` files?

modules/directives/src/main/scala/scala/build/preprocessing/directives/Entrypoints.scala
alexarchambault: Having the using directive should make the completion stuff try to actually compile the project, and provide completions for the entrypoints' options. Ideally, it should also import the annotation itself.

modules/integration/src/test/scala/scala/cli/integration/RunEntrypointsTests.scala
alexarchambault: It ought to, but it currently runs into TASTy version issues (that I need to look into)

Gedochao:
Closing this for now.
It seems we wouldn't be able to merge it until macro annotations become stable in Scala 3, which doesn't have a date just yet.

Gedochao:
Closing this for now.
It seems we wouldn't be able to merge it until macro annotations become stable in Scala 3, which doesn't have a date just yet.

PR #1821, opened 2023-02-01
Add build/taskStart and taskFinish to the exception reporting BSP mechanism

MaciejG604:
This resolves the issue with only one $ivy or $dep import error being visible at a time. The behaviour has been checked with metals.

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspClient.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspClient.scala
@@ -2,7 +2,7 @@ package scala.build.bsp
 
 import ch.epfl.scala.{bsp4j => b}
 
-import java.lang.{Boolean => JBoolean}
+import java.lang.Boolean as JBoolean
 import java.net.URI
 import java.nio.file.Paths
 import java.util.concurrent.{ConcurrentHashMap, ExecutorService}
@@ -155,12 +155,11 @@ class BspClient(
       case None =>
         logger.debug(s"Not reporting $ex to users (no build target id)")
       case Some(targetId) =>
-        val allExceptions = ex match {
-          case c: CompositeBuildException => c.exceptions
-          case _                          => Seq(ex)
-        }
-        val touchedFiles =
-          allExceptions.flatMap(reportDiagnosticForFiles(targetId, reset = reset)).toSet
+        val touchedFiles = (ex match {
+          case c: CompositeBuildException =>
+            reportDiagnosticsForFiles(targetId, c.exceptions, reset = reset)
+          case _ => reportDiagnosticsForFiles(targetId, Seq(ex), reset = reset)
+        }).toSet
 
         // Small chance of us wiping some Bloop diagnostics, if these happen
         // between the call to remove and the call to actualBuildPublishDiagnostics.
@@ -182,7 +181,34 @@ class BspClient(
         }
     }
 
-  def reportDiagnosticForFiles(
+  def reportDiagnosticsForFiles(
+    targetId: b.BuildTargetIdentifier,
+    diags: Seq[Diagnostic],
+    reset: Boolean = true
+  ): Seq[os.Path] =
+    if reset then // send diagnostic with reset only once for every file path
+      diags.flatMap { diag =>
+        diag.positions.map { position =>
+          Diagnostic(diag.message, diag.severity, Seq(position), diag.textEdit)
+        }
+      }
+        .groupBy(_.positions.headOption match
+          case Some(File(Right(path), _, _)) => Some(path)
+          case _                             => None
+        )
+        .filter(_._1.isDefined)
+        .values
+        .toSeq
+        .flatMap {
+          case head :: tail =>
+            reportDiagnosticForFiles(targetId, reset = reset)(head)
+              ++ tail.flatMap(reportDiagnosticForFiles(targetId))
+          case _ => Nil
+        }
+    else
+      diags.flatMap(reportDiagnosticForFiles(targetId))
+
+  private def reportDiagnosticForFiles(
     targetId: b.BuildTargetIdentifier,
     reset: Boolean = false
   )(diag: Diagnostic): Seq[os.Path] =
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -7,6 +7,7 @@ import org.eclipse.lsp4j.jsonrpc
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError
 
 import java.io.{InputStream, OutputStream}
+import java.util.UUID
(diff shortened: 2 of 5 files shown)


modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -287,10 +293,37 @@ final class BspImpl(
 
     preBuild.thenCompose {
       case Left((ex, scope)) =>
+        val targetId  = currentBloopSession.bspServer.targetIds.head
+        val targetUri = targetId.getUri
+        val target    = targetUri.substring(targetUri.lastIndexOf('=') + 1, targetUri.length)
+
+        val taskStartParams = new TaskStartParams(new TaskId("1"))
```
lwronski: Maybe just use:
```
val taskId = new TaskId(UUID.randomUUID().toString)
```

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -287,10 +293,37 @@ final class BspImpl(
 
     preBuild.thenCompose {
       case Left((ex, scope)) =>
+        val targetId  = currentBloopSession.bspServer.targetIds.head
+        val targetUri = targetId.getUri
+        val target    = targetUri.substring(targetUri.lastIndexOf('=') + 1, targetUri.length)
```
lwronski: This won't work if targetUri contains `=` twice.

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -287,10 +293,37 @@ final class BspImpl(
 
     preBuild.thenCompose {
       case Left((ex, scope)) =>
+        val targetId  = currentBloopSession.bspServer.targetIds.head
```
lwronski: We need to handle the case when `buildTargetId` doesn't exist and when there are several `targetIds`:
```suggestion
        for (targetId <- currentBloopSession.bspServer.targetIds) {
          ...
        }
```

modules/build/src/main/scala/scala/build/bsp/BspClient.scala
```diff
@@ -155,12 +155,24 @@ class BspClient(
...
-        val touchedFiles =
-          allExceptions.flatMap(reportDiagnosticForFiles(targetId, reset = reset)).toSet
+        val touchedFiles = (ex match {
+          case c: CompositeBuildException =>
+            if reset then // send one build exception with reset == true for every file path
+              c.exceptions.groupBy {
+                _.positions.headOption match
+                  case Some(File(Right(path), _, _)) => Some(path)
+                  case _                             => None
+              }
```
lwronski: I think you should process all `positions` from list instead only the first.

lwronski (commented):
Add some test to `BspTestDefinitions` to check your changes.

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
Gedochao: +1, `UUID`s should do the trick here

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -1,5 +1,6 @@
 package scala.build.bsp
 
+import ch.epfl.scala.bsp4j._
```
Gedochao: don't import `bsp4j._` here.
instead, use the types from there with the `b` prefix imported on the line below.
For example, `b.TaskId`.

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -287,10 +293,43 @@ final class BspImpl(
...
-          ex
-        )
+        val taskId = new b.TaskId(UUID.randomUUID().toString)
+
+        for targetId <- currentBloopSession.bspServer.targetScopeIdOpt(scope) do {
+          val idPrefix = "?id="
+          val target = targetId.getUri.substring(
+            targetId.getUri.lastIndexOfSlice(idPrefix) + idPrefix.length,
+            targetId.getUri.length
+          )
```
lwronski:
```suggestion
          val target = targetId.getUri match {
            case s"$_?id=$targetId" => targetId
            case targetIdUri => targetIdUri
          }
```

modules/build/src/main/scala/scala/build/bsp/BspClient.scala
```diff
@@ -182,7 +181,33 @@ class BspClient(
...
+      diags.filter(_.positions.nonEmpty)
+        .flatMap { diag =>
+          diag.positions.map { position =>
+            Diagnostic(diag.message, diag.severity, Seq(position), diag.textEdit)
+          }
+        }
+        .groupBy(_.positions.head match
+          case File(Right(path), _, _) => Some(path)
+          case _                       => None
+        )
```
lwronski:
```suggestion
      diags.
        .flatMap { diag =>
          diag.positions.map { position =>
            Diagnostic(diag.message, diag.severity, Seq(position), diag.textEdit)
          }
        }
        .groupBy(_.positions.headOption match
          case Some(File(Right(path), _, _)) => Some(path)
          case _                       => None
        )
```

modules/build/src/main/scala/scala/build/bsp/BspClient.scala
```diff
@@ -182,7 +181,33 @@ class BspClient(
...
+        .filter(_._1.isDefined)
+        .values
+        .toSeq
+        .flatMap { diagsForPath =>
+          reportDiagnosticForFiles(targetId, reset = reset)(diagsForPath.head) ++
+            diagsForPath.tail.flatMap(reportDiagnosticForFiles(targetId))
```
lwronski:
```suggestion
            diagsForPath match {
               case head :: tail =>
                  reportDiagnosticForFiles(targetId, reset = reset)(head)
                    ++ tail.flatMap(reportDiagnosticForFiles(targetId))
               case _ => Nil
```

modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
```diff
@@ -287,10 +293,43 @@ final class BspImpl(
...
+            Some(targetId),
+            ex
+          )
+
+          val taskFinishParams = new b.TaskFinishParams(taskId, b.StatusCode.ERROR)
+          taskFinishParams.setEventTime(System.currentTimeMillis())
+          taskFinishParams.setMessage(s"Preprocessed '$target'")
+          taskFinishParams.setDataKind(b.TaskDataKind.COMPILE_REPORT)
+
+          val errorNumber = ex match {
```
lwronski:
```suggestion
          val errorSize = ex match {
```

lwronski (approved):
LGTM

Gedochao (approved):

PR #1826, opened 2023-02-02
Pass user arguments to JS runner

alexarchambault:


Gedochao (approved):
LGTM
it could be cool to add some docs for this, but can be done separately

PR #1827, opened 2023-02-02
Rename using lib directive to using dep

lwronski:
To be more consistent with command line options where dependency is passed by `--dep`.

I still keep the alias of this using directive (`lib`, `libs`) for backwards compatibility

### DIFF
--- a/examples/junit/MyTests.scala
+++ b/examples/junit/MyTests.scala
@@ -1,4 +1,4 @@
-//> using lib "com.github.sbt:junit-interface:0.13.2"
+//> using dep "com.github.sbt:junit-interface:0.13.2"
 import org.junit.Test
 
 class MyTests {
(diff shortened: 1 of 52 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunTestsDefault.scala
```diff
@@ -82,7 +82,7 @@ class RunTestsDefault extends RunTestDefinitions(scalaVersionOpt = None) {
     val inputs = TestInputs(
       libSourcePath -> libSource("Hello"),
       os.rel / "app" / "TestApp.scala" ->
-        """//> using lib "test-org::messages:0.1.0"
```
Gedochao: NIT we might want to keep some random test with the `lib` alias

Gedochao (approved):
LGTM

PR #1840, opened 2023-02-08
Add export --json option

### LINKED ISSUE #1722: `project` subcommand
For larger projects with multiple dependencies it may be hard to grasp all of the relevant configurations, dependencies declared in multiple files, etc. 
It'd be useful to have a sub-command for printing the details of a given Scala CLI project.

Something like this:
```bash
$ scala-cli project .
# Scala version used: 3.2.1
# dependencies:
#   com.lihaoyi::os-lib:0.8.1
#   com.lihaoyi::pprint:0.8.1
# Scala compiler options
#    -release 11
#   -new-syntax
# ...
```
We could give the summary of all of the dependencies, compiler options, platform, Scala & JVM versions and everything else configured by using directives in the project pass as inputs.

**Additional context**
Created from a comment posted by @prolativ in https://github.com/VirtusLab/scala-cli/issues/1665#issuecomment-1342679214

### LINKED ISSUE #1791: Export information about the project
This has been raised in numerous places:
- https://github.com/VirtusLab/scala-cli/discussions/1005
- https://github.com/VirtusLab/scala-cli/issues/1722#issuecomment-1384182072
- Scala Steward could use information about the dependencies

Have an export subcommand that would dump information in the json format for other tools to consume.

```
scala-cli export --json
```

Not sure about the exact format and what data we would need but at the minimum we should support a list of dependencies.

We could potentially copy over the json used for Bloop and analyse it.

MaciejG604:


modules/integration/src/test/scala/scala/cli/integration/ExportJsonTestDefinitions.scala
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    inputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "export", "--json", ".")
+        .call(cwd = root)
+
+      val fileContents = os.read(root / "dest" / "export.json")
+        .replaceAll("\\s", "")
+
+      expect(fileContents.contains("\"scalaVersion\":\"3.2.2\""))
```
tgodzik: How large would be the full output? Maybe we could just compare the whole thing?

MaciejG604: Not very large, especially since there's one dependency in the file. We can do it, I'll change it.

modules/cli/src/main/scala/scala/cli/commands/export/Export.scala
```diff
@@ -100,8 +115,12 @@ object Export extends ScalaCommand[ExportOptions] {
...
     }
 
-    val buildToolName = if (shouldExportToMill) "mill" else "sbt"
-    System.out.println(s"Exporting to a $buildToolName project...")
+    if (!shouldExportToJson) {
+      val buildToolName = if (shouldExportToMill) "mill" else "sbt"
+      println(s"Exporting to a $buildToolName project...")
+    }
+    else
+      println(s"Exporting to JSON...")
```
lwronski: Instead of `println` use `logger`.

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
```diff
@@ -0,0 +1,95 @@
...
+}
+
+case class ExportDependencyFormat(groupId: String, artifactId: ArtifactId, version: String)
+
+case class ArtifactId(name: String, fullName: String)
+
+object ExportDependencyFormat {
+  def apply(dep: Dependency): ExportDependencyFormat = {
+    val scalaVersionStartIndex = dep.module.name.value.lastIndexOf('_')
+    val shortDepName = dep.module.name.value.substring(
```
lwronski: I suggest not to use `substring` since it can throw exception.

modules/cli/src/main/scala/scala/cli/exportCmd/Json.scala
```diff
@@ -0,0 +1,186 @@
...
+import java.nio.file.Path
+
+import scala.build.internal.Constants
+import scala.build.internal.Runner.frameworkName
+import scala.build.options.{BuildOptions, Platform, ScalaJsOptions, ScalaNativeOptions, Scope}
+import scala.build.testrunner.AsmTestRunner
+import scala.build.{Logger, Sources}
+import scala.cli.util.SeqHelpers.*
+
+final case class Json(
```
Gedochao: NIT I feel a more descriptive name would be better, `Json` is too general and doesn't really indicate what this class is for...
Maybe smth like:
```suggestion
final case class JsonProjectDescriptor(
```

Gedochao: same applies to the `Mill` and `Sbt` classes, so maybe they could be renamed consistently as well...

Gedochao (approved):
NIT it'd be good to extract the refactor changes to a separate NIT commit
aside from the nitpicks, LGTM

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.build.options.ConfigMonoid
+import scala.cli.util.SeqHelpers.*
+import scala.reflect.NameTransformer
+import scala.util.{Properties, Using}
+
+final case class JsonProject(
+  projectName: Option[String] = None,
+  scalaVersion: Option[String] = None,
+  platform: Option[String] = None,
+  scalacOptions: Seq[String] = Nil,
```
prolativ: Regarding my comment here https://github.com/VirtusLab/scala-cli/issues/1722#issuecomment-1362960092:
Why do some properties like `*Deps` or `*Sources` have a a prefix for `main` or `test` but others don't? What if e.g. a user needed an additional resolver for a dependency that is used only in tests or if they wanted to add some resources only for tests? I mean in general its possible to add using directives in a file which belongs to the test scope and they should apply to all other sources in the test scope, right? Maybe the structure here should be something like
```scala
case class JsonProject(
   projectName: Option[String],
   scalaVersion: Option[String],
   main: JsonMainProject,
   test: JsonTestProject,
   ...
)

case class JsonMainProject(
  scalacOptions: : Seq[String],
  resolvers: Seq[ExportResolverFormat],
  ...
)

case class JsonTestProject(
  scalacOptions: : Seq[String],
  resolvers: Seq[ExportResolverFormat],
  ...
)
```
???

MaciejG604: Good point, I'll split project information into scopes.

prolativ: I would be also good to figure out how this should behave if the project can be cross-compiled to different platforms or scala versions. Should the output json actually be a list of objects with such shape and each object would correspond to a combination of scala version and platform? Or should we return one object for the default values of these unless a user specifies them explicitly? In other words consider these two situations:
```shell
scala-cli export --json .
```

```shell
scala-cli export --json . -S 2.13 --native
```

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
```diff
@@ -0,0 +1,92 @@
...
+      .withTransientEmpty(false))
+  implicit lazy val StringSeqCodec: JsonValueCodec[Seq[String]] = JsonCodecMaker.make
+}
+
+object JsonProject {
+  implicit val monoid: ConfigMonoid[JsonProject]           = ConfigMonoid.derive
+  implicit lazy val jsonCodec: JsonValueCodec[JsonProject] = JsonCodecMaker.make
+}
+
+case class ExportDependencyFormat(groupId: String, artifactId: ArtifactId, version: String)
```
alexarchambault, line 88: Does that imply that dependencies will end up like this in the output JSON?
```json
{
  "groupId": "org.something",
  "artifactId": {
    "name": "foo",
    "fullName": "foo_3"
  },
  "version": "X.Y.Z"
}
```

That's way too heavyweight IMO. Dependencies have a concise string representation, meant to be used in that kind of case: `"org.something::foo:X.Y.Z"`, that can be parsed back to a non-ambiguous dependency.

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+case class ExportResolverFormat(
+  name: String,
+  location: Option[String],
+  user: Option[String] = None,
+  password: Option[String] = None,
+  headers: Seq[(String, String)] = Nil
+)
```
alexarchambault: Just like for dependencies, that looks too heavyweight. A simple string should be enough to describe a repository non-ambiguously. For `MavenRepository` instances, just the URL location is enough, and will be parsed back as a `MavenRepository`. For `IvyRepository`, they have a `ivy:…` representation too, see how these are parsed [here](https://github.com/coursier/coursier/blob/b8b48dc1809726b52e0ccaa054a87211e9c98344/modules/coursier/shared/src/main/scala/coursier/internal/SharedRepositoryParser.scala#L37-L47).

About authentication, I wouldn't put it in that file by default, as this could unexpectedly leak passwords, and users using these info and fetching things with coursier down-the-line will recover the globally-configured authentication anyway, via the global coursier config files. I'd add authentication only if users explicitly ask for it.

So overall, I'd represent repositories as a simple string by default. If users ask for authentication, an object should do, with the former string in a field such as `"repo"`, and authentication fields alongside it.

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.cli.util.SeqHelpers.*
+import scala.reflect.NameTransformer
+import scala.util.{Properties, Using}
+
+final case class JsonProject(
+  projectName: Option[String] = None,
+  scalaVersion: Option[String] = None,
+  platform: Option[String] = None,
+  scalacOptions: Seq[String] = Nil,
+  scalaCompilerPlugins: Seq[String] = Nil,
```
alexarchambault: Aren't these supposed to be dependencies too? (so something like `Seq[ExportDependencyFormat]`)

modules/cli/src/main/scala/scala/cli/exportCmd/JsonProject.scala
alexarchambault: We should have dependencies as `dependency.DependencyLike` instances (or one of its aliases). It has a [`render`](https://github.com/coursier/dependency/blob/3a77788ad257ddf81a9d578b032e9c420c512bd4/dependency/src/main/scala/dependency/DependencyLike.scala#L36) method meant to provide that concise string.

prolativ: I would say the format we should use here depends on the particular use cases that we want to support. On the one hand encoding a dependency as a single string is more space-efficient but if we expect other tools to be built on the base of scala-cli, we would force them to reimplement the logic of parsing this dependency format, which might get slightly more complicated in the future (e.g. there have been some suggestions about implementing some heuristics to allow users to always write just `:` instead of either `:` or `::`). Also if this JSON grows too large it might get hard to read for humans and one of the motivations of this feature was also to let users clearly see the entire configuration of their project in case using directives are split between multiple files (which might be a case e.g. if you need different settings for different targets)

MaciejG604:
After consideration I think this PR should focus on exporting basic information about the project and mostly on what has been requested, so the list of dependencies and resolvers. The discussion on the cross-compilation for platforms and scala versions should be moved to a follow-up issue. 

For now the `export --json` result contains the following information about the project:
-   **projectName**
- **scalaVersion** that's visible in the Main scope
- **platform** only one that's visible in the Main scope
- **scalacOptions**
- **scalaCompilerPlugins** presented as `String`s with ':' as delimiters
- **scalaJsVersion** that's visible in the Main scope, only if the platform is `JS`
- **scalaNativeVersion** that's visible in the Main scope, only if the platform is `Native`
- **mainClass**
- **extraDecls** contains [`jsModuleKind`](https://scala-cli.virtuslab.org/docs/reference/directives#scalajs-options) information
- scope specific infromation:
  - **scopeName**, e.g. "main", "test"
  - **sources** list of subpaths to the source files
  - **dependencies** presented as `String`s with ':' as delimiters
  - **resolvers** presented as URL for Maven and `ivy:pattern` for Ivy
  - **resourceDirs**
  - **extraDecls** contains custom `jar`s added to the project


Follow-ups to consider:
-
- cross-compilation
- resolver authentication

MaciejG604:
@fthomas, @ckipp01
Is this format okay?

ckipp01:
> Is this format okay?

Just took a look. Really happy you're working on this! One thing I did notice though, which @prolativ already mentioned is that I like the more structured output than the concise string. By doing the concise string you're now forcing every consumer of this output to have to both know how and to manually parse that output to their needs instead of having it already structured. I'm not really sure what the benefit of the concise string is. Having it structured like you originally had it is much more flexible imo.

prolativ:
1) Showing only the configuration for the current/default platform seems reasonable. We could later (as an independent improvement) add `--cross` flag to return a json list/object with configurations for different platforms and scala versions.
2) One can have different `scalacOptions` for main and test so these should be split. I assume this is true also for `scalaCompilerPlugins`
3) is `extraDecls` supposed to be dependent on the platform (e.g. can it hold some options for scala-native?) or is it scala-js only? In the latter case why isn't it called explicitly `jsModuleKind`? And why is this property both top-level and separately for main/test?
4) If we specify `scalaJsVersion` and `scalaNativeVersion`, shouldn't we also include e.g. the version of js itself or JVM, if it's specified explicitly?
5) Maybe `scopes` should be a Map rather than a list? Then `scopeName` wouldn't be necessary and extracting a desired scope from the json structure would be easier

MaciejG604:
@prolativ 
> * is `extraDecls` supposed to be dependent on the platform (e.g. can it hold some options for scala-native?) or is it scala-js only? In the latter case why isn't it called explicitly `jsModuleKind`? And why is this property both top-level and separately for main/test?

I would delete this field altogether and leave scala-j/scala-native options for follow-up PRs if needed. WDYT?

> * If we specify `scalaJsVersion` and `scalaNativeVersion`, shouldn't we also include e.g. the version of js itself or JVM, if it's specified explicitly?

By `"the version of js itself"` do you mean Scala.js ECMA Script version?

MaciejG604:
[UPDATE]
Added structured dependency json objects, moved scalacOptions and scalaCompilerPlugins to scopes.
Added jvmVersion and jsEsVersion field.

For now the `export --json` result contains the following information about the project:
-   **projectName**
- **scalaVersion** that's visible in the Main scope
- **platform** only one that's visible in the Main scope
- **jvmVersion** only if declared explicitly
- **scalaJsVersion** that's visible in the Main scope, only if the platform is `JS`
- **jsEsVersion** only if declared explicitly
- **scalaNativeVersion** that's visible in the Main scope, only if the platform is `Native`
- **mainClass**
- scope specific infromation as a map of (scope name -> following information):
  - **sources** list of subpaths to the source files
  - **scalacOptions**
  - **scalaCompilerPlugins** as structured json
  - **dependencies** as structured json
  - **resolvers** presented as URL for Maven and `ivy:pattern` for Ivy
  - **resourceDirs**
  - **customJarsDecls** contains custom `jar`s added to the project


Follow-ups to consider:
- cross-compilation
- resolver authentication
- scala-js/scala-native options

prolativ (approved):
LGTM in general.
Personally I'm not a big fan of `jsEsVersion`, because:
* this name is not very intuitive IMO
* ECMAScript seems to be the standard of the language rather than its implementation but for JVM we can specify that we want to use e.g. `--jvm adopt:1.8.0-232` specifically rather than just `--jvm 8`.
But at least it seems consistent with the name of the command line option

lwronski (approved):

kubukoz:
Just curious, what was the motivation for this change? I like it, might be very useful for packaging in Nix, just wanted to know what other usecases you had :)

tgodzik:
The main use case we were thinking about is Scala Steward support, since there was a suggestion that export to sbt might be used as intermediary, which scared me a bit :sweat_smile:

fthomas:
That suggestion turned quickly into reality: https://github.com/scala-steward-org/scala-steward/pull/2933 :see_no_evil: :rofl:

PR #1842, opened 2023-02-08
Read home directory from env variable instead of option from command line

lwronski:


modules/cli/src/main/scala/scala/cli/commands/directories/Directories.scala
```diff
@@ -21,7 +21,7 @@ object Directories extends ScalaCommand[DirectoriesOptions] {
       sys.exit(1)
     }
 
-    val directories = options.directories.directories
+    val directories = scala.build.Directories.directories
```
Gedochao, line 24: NIT just import it here
```suggestion
    val directories = Directories.directories
```

Gedochao (approved):

modules/cli/src/main/scala/scala/cli/commands/directories/Directories.scala
lwronski: It doesn't work, but this class is named `Directories`.

Issue #1843, opened 2023-02-09
Failed to start bloop server on armle using Debian 11


Flowdalic:
Using scala-cli 0.1.20 to execute a script on an armle machine running Debian 11 (Bullseye) with openjdk-jre-17-headless installed fails with blooprifle throwing a `FailedToStartServerException`.

```
flo@raspi:~$ scala-cli ./myscript.sc 
Starting compilation server                                          
Error: scala.build.blooprifle.FailedToStartServerException: Server didn't start
For more details, please see '/home/flo/.scala-build/stacktraces/1675934079-9520059747794623223.log'
Running                                                              
  scala-cli bloop output                                             
might give more details.
```

```
flo@raspi:~$  cat /home/flo/.scala-build/stacktraces/1675949061-6542722033418601453.log
scala.build.blooprifle.FailedToStartServerExitCodeException: Server failed with code 1
  scala.build.blooprifle.internal.Operations$.check0$1$$anonfun$1(Operations.scala:250)
  scala.build.blooprifle.BloopRifleLogger.runnable$$anonfun$1(BloopRifleLogger.scala:13)
  java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
  java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
  java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
  java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
  java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
  java.base/java.lang.Thread.run(Thread.java:833)
```
running `scala-cli bloop output` shows

```
flo@raspi:~$ scala-cli bloop output
Unrecognized VM option 'MaxInlineLevel=20'
Did you mean 'MaxForceInlineLevel=<value>'? Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

[...]

PR #1844, opened 2023-02-09
blooprifle: report exit code in exception

Flowdalic:
To help debugging, the failure code is now part of the exception in case the bloop server failed to start.

I've worked on this during debugging #1843. Turns out, knowing the exit code was not really helpful in this case. However, I believe that it is never wrong to load the exception with it.

### DIFF
--- a/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/FailedToStartServerException.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/FailedToStartServerException.scala
@@ -2,5 +2,11 @@ package scala.build.blooprifle
 
 import scala.concurrent.duration.Duration
 
-final class FailedToStartServerException(timeoutOpt: Option[Duration] = None)
-  extends Exception("Server didn't start" + timeoutOpt.fold("")(t => s" after $t"))
+abstract class FailedToStartServerException(message: String)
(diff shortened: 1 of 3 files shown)


lwronski (approved):
@Flowdalic Thanks for your contribution! LGTM

The error in CI is unrelated to your changes, and the Scala-CLI team will fix it soon.

Gedochao:
@Flowdalic, the CI should pass if you rebase your branch with the newest changes from `main`.
The same applies to #1845

Gedochao (approved):

Gedochao:
@Flowdalic it seems the following tests are failing and will need to be adjusted.
```text
integration.test.test 3 tests failed: 
  scala.cli.integration.BloopTests.invalid bloop options passed via cli cause bloop start failure 
  scala.cli.integration.BloopTests.invalid bloop options passed via global bloop config json file cause bloop start failure 
  scala.cli.integration.UpdateTests.updating dummy scala-cli using update command 
Error: Process completed with exit code 1.
```

I suspect `scala.cli.integration.UpdateTests.updating dummy scala-cli using update command` is an intermittent connection failure, but the other 2 will need code adjustments.

BTW. let me know if you'll have trouble to find the time to fix this, in which case either me or someone else from the maintenance team will have a look.

PR #1845, opened 2023-02-09
blooprifle: add -XX:+IgnoreUnrecognizedVMOptions to hardCodedDefaultJavaOpts

### LINKED ISSUE #1843: Failed to start bloop server on armle using Debian 11

Flowdalic:
This was (is?) an attempt to fix #1843. However, adding this line had no effect. Just like removing the `XX:MaxInlineLevel` line. I assume that the error in #1843 is caused by a different code path. I suspect https://github.com/scalacenter/bloop/blob/a249e0a710ce169ca05d0606778f96f44a398680/bloopgun/src/main/scala/bloop/bloopgun/util/Environment.scala#L87, where, according to the code comment, the code in scala-cli originates from.

I wonder if it is wise to hard code non-standard JVM options like it is done here and in scalacenter/bloop. As apparently #1843 demonstrates, some (most?) JVM will bail out if they are invoked with options they do not recognize (which is IMHO sensible behavior).  I would be great of all TCK compatible JVMs would be supported by scala-cli (and bloop). So either do not use non-standard hard coded options, or add `-XX:+IgnoreUnrecognizedVMOptions` (which, according to [JDK-6788376](https://bugs.openjdk.org/browse/JDK-6788376), is available since Java 6)<sup>1</sup>.

1: A third option would be to only issue the option if it was detected the receiving JVM supports it. But that is probably not trivial to implement.

### DIFF
--- a/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifleConfig.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifleConfig.scala
@@ -77,6 +77,7 @@ object BloopRifleConfig {
       "-XX:+UseZGC", // ZGC returns unused memory back to the OS, so Bloop does not occupy so much memory if unused
       "-XX:ZUncommitDelay=30",
       "-XX:ZCollectionInterval=5",
+      "-XX:+IgnoreUnrecognizedVMOptions", // Do not fail if an non-standard (-X, -XX) VM option is not recognized.
       "-Dbloop.ignore-sig-int=true"
     )
 
--- a/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
@@ -56,20 +56,6 @@ class BloopTests extends ScalaCliSuite {
     testScalaTermination(Constants.bloopVersion, shouldRestart = false)
   }
 
-  test("invalid bloop options passed via cli cause bloop start failure") {
-    TestInputs.empty.fromRoot { root =>
-      runScalaCli("bloop", "exit").call(cwd = root)
-      val res = runScalaCli("bloop", "start", "--bloop-java-opt", "-zzefhjzl").call(
-        cwd = root,
-        stderr = os.Pipe,
-        check = false,
-        mergeErrIntoOut = true
-      )
-      expect(res.exitCode == 1)
-      expect(res.out.text().contains("Server failed with exit code 1"))
-    }
-  }
-
   test("invalid bloop options passed via global bloop config json file cause bloop start failure") {
     val inputs = TestInputs(
       os.rel / "bloop.json" ->


lwronski:
The comment [here](https://github.com/scalacenter/bloop/blob/a249e0a710ce169ca05d0606778f96f44a398680/bloopgun/src/main/scala/bloop/bloopgun/util/Environment.scala#L87) indicates that this JVM option is ignored by GraalVM, but it isn't.

I found [here](https://github.com/oracle/graal/blob/59740919ef06bd6a0787880cfe1ab9212d8ed1aa/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/ReservedStackAccessTest.java#L78) an interesting flag in the Graal project that may address your issue.

Can you add `"-Dgraal.Inline=false"` around [here](https://github.com/VirtusLab/scala-cli/blob/f0e6e9b268fe356774adb255c5cbb4633a050069/project/settings.sc#L169) and check if it fix your problem?

Flowdalic:
> Can you add `"-Dgraal.Inline=false"` around [here](https://github.com/VirtusLab/scala-cli/blob/f0e6e9b268fe356774adb255c5cbb4633a050069/project/settings.sc#L169) and check if it fix your problem?

Note that I see this error with the standalone JVM launcher and not with a native image created by Graal<sup>1</sup>. So I assume your suggestion does not apply?

1: I execute scala-cli on an armle machine and Graal has no support for 32-bit arm.

lwronski:
@Flowdalic Thank you for working on this issue. 

~~I would like to suggest an alternative solution. Would it be possible for you to make the option `-XX:+IgnoreUnrecognizedVMOptions` configurable through an environment variable instead of hardcoding it in the default JVM options?~~

lwronski:
> @Flowdalic Thank you for working on this issue. I would like to suggest an alternative solution. Would it be possible for you to make the option `-XX:+IgnoreUnrecognizedVMOptions` configurable through an environment variable instead of hardcoding it in the default JVM options?

We discussed this internally and we agree to add `-XX:+IgnoreUnrecognizedVMOptions`, sorry for the confusion.

 It seems the following tests are failing and will need to be adjusted:
```
scala.cli.integration.BloopTests.invalid bloop options passed via cli cause bloop start failure
```

I suspect this test can be removed as invalid JVM options should not fail bloop.

BTW. let me know if you'll  don't have time to fix this, I can have a look.

Flowdalic:
Thanks. I'll look into the test.

Your suggestion to create an environment variable made me wonder if it wouldn't be sensible to be able to override the default hardcoded JVM options via an environment variable e.g., `OVERRIDE_BLOOP_JVM_OPTIONS`. This would allow to experiment with different JVM options without having to modify the code.

Gedochao (approved):

Gedochao:

PR #1846, opened 2023-02-09
Add clang to scala-cli docker image

### LINKED ISSUE #1756: Include `clang` to docker image with scala-cli

Include `clang` which would be required to produce scala native binaries using `scala-cli` docker image.

It should be updated [ScalaCliDockerFile](https://github.com/VirtusLab/scala-cli/blob/main/.github/scripts/docker/ScalaCliDockerFile).

**Additional context**
See discussion on Discord around @lolgab message [here](https://discord.com/channels/632150470000902164/901153021638082660/1059940928917356695).

lwronski:
Fixes #1756.

### DIFF
--- a/modules/integration/docker/src/test/scala/scala/cli/integration/RunDockerTests.scala
+++ b/modules/integration/docker/src/test/scala/scala/cli/integration/RunDockerTests.scala
@@ -10,6 +10,9 @@ class RunDockerTests extends munit.FunSuite {
   lazy val imageName = Option(System.getenv("SCALA_CLI_IMAGE")).getOrElse {
     sys.error("SCALA_CLI_IMAGE not set")
   }
+  lazy val termOpt = if (System.console() == null) Nil else Seq("-t")
+  lazy val ciOpt   = Option(System.getenv("CI")).map(v => Seq("-e", s"CI=$v")).getOrElse(Nil)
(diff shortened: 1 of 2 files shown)


modules/integration/docker/src/test/scala/scala/cli/integration/RunDockerTests.scala
```diff
@@ -42,4 +42,21 @@ class RunDockerTests extends munit.FunSuite {
...
+    )
+    inputs.fromRoot { root =>
+      val cmdPackage = Seq[os.Shellable](
+        // format: off
+        "docker", "run", "--rm", termOpt, "-v", s"$root:/data", "-w", "/data", ciOpt,
+        imageName, "package", "--native", fileName, "-o", "Hello"
+        // format: on
+      )
+      val procPackage = os.proc(cmdPackage).call(cwd = root)
+      expect(procPackage.exitCode == 0)
```
Gedochao: is this line necessary? if the `exitCode` wasn't 0, the test would fail, since there's no `check = false`, no?

lwronski: To highlight in this test that we are only checking if it was packaged correctly, I added this `assert`. 

Right, I missed `check = false`, so I added it.

Gedochao (approved):

PR #1848, opened 2023-02-10
Suppress lib update warning

### LINKED ISSUE #1825: Allow users to disable dependency updates
Currently if a user wants to keep a certain version of the dependency they will get a warning that it's outdated, but it might not always be possible to update it.

Let's at a very minimum add and option to disable these warnings, we can follow up with something more concrete

Disable all the actionable diagnostics with `scala-cli config actions false`

**Additional context**
https://github.com/VirtusLab/scala-cli/issues/1822

MaciejG604:
Resolves #1825

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -38,6 +38,7 @@ import scala.cli.commands.shared.{
 }
 import scala.cli.commands.tags
 import scala.cli.commands.util.ScalacOptionsUtil.*
+import scala.cli.config.Key.BooleanEntry
 import scala.cli.config.{ConfigDb, ConfigDbException, Keys}
 import scala.concurrent.ExecutionContextExecutorService
 import scala.concurrent.duration.*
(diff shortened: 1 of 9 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -432,8 +442,8 @@ final case class SharedOptions(
     }
   }
 
-  def suppressDirectivesInMultipleFilesWarning: Option[Boolean] =
-    suppress.suppressDirectivesInMultipleFilesWarning.filter(identity).orElse(
+  def getOptionOrFromConfig(cliOption: Option[Boolean], configDbKey: BooleanEntry) =
+    cliOption.orElse(
       configDb.map(_.get(Keys.suppressDirectivesInMultipleFilesWarning))
```
lwronski:
```suggestion
      configDb.map(_.get(configDbKey))
```

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -25,30 +25,35 @@ case object ActionableDependencyHandler
...
-      if (dependency.userParams.contains("toolkit"))
-        Some(ActionableDependencyUpdateDiagnostic(
-          setting.positions,
-          currentVersion,
-          latestVersion,
-          dependencyModuleName = "toolkit",
-          suggestion = latestVersion
-        ))
+  ): Either[BuildException, Option[ActionableDependencyUpdateDiagnostic]] =
+    if (buildOptions.suppressWarningOptions.suppressOutdatedDependencyWarning.getOrElse(false))
```
lwronski: Can you move this condition higher? to the `extractSettings` function.

lwronski (approved):

Gedochao (approved):

SethTisue:
Is there a way to suppress these warnings by adding something to the `.scala` file itself, either per-dependency or project-wide? I'd rather not suppress them globally, but I also don't want to have to enter the flag on the command line every time.

tgodzik:
> Is there a way to suppress these warnings by adding something to the `.scala` file itself, either per-dependency or project-wide? I'd rather not suppress them globally, but I also don't want to have to enter the flag on the command line every time.

Unfortunately, those are the only two options currently. It might be hard to figure out an API to do it per case basis.

PR #1849, opened 2023-02-12
Don't modify features based on binary name

### LINKED ISSUE #1824: Don't modify features based on binary name
Currently, if you use ScalaCLI as `scala` you will need to use --power to use advanced feature. This is not the case when changing the name of the binary. This basically means we would need two kinds of documentation and makes it necessary to have two different releases.

We discussed it with both @julienrf and @romanowski that in the end we would only want to have a single `scala` runner with the ability to use --power options or using https://github.com/VirtusLab/scala-cli/issues/1823 to transform scala into a full scala runner.

This also allows us to have a single documentation page with the advanced features clearly marked via https://github.com/VirtusLab/scala-cli/issues/1700

Let's make the behaviour with `scala` as the default.

Keep it as it is or do the update later on.

lwronski:
I'd like to highlight the implications of this change implying the usage of scala-cli:

1. The non sip commands (`package` and `export`) have stoped working with the scala-cli launcher. Running these commands will now result in an error.
```
package is not a scala-cli sub-command and it is not a valid path to an input file or directory
                Try 'scala-cli--help' to see the list of available sub-commands and options
```
so this warning should be extended to add info - that to avoiding this error, users should either use the `--power` option or set power to true globally using the `scala-cli config power true` command.

2.  https://github.com/VirtusLab/scala-cli-setup - should be added a configuration that enables the power functionalities in `scala-cli`, something like this:
``` 
    - uses: VirtusLab/scala-cli-setup@main
      with:
        power: true
```
3. Running docker scala-cli image will require passing `--power`with non sip commands when executing this image:
```
docker run virtuslab/scala-cli:latest --power package .
```
or we can publish docker image with enabling power by default.

Fixes #1824


modules/build/src/main/scala/scala/build/preprocessing/DirectivesProcessor.scala
```diff
@@ -35,8 +35,8 @@ object DirectivesProcessor {
     ) =
       if (!allowRestrictedFeatures && handler.isRestricted)
         val msg =
-          """This directive is not supported with 'scala' command.
-            |Please run it with `scala-cli` command or with `--power` flag or turn on this flag globally running command `config power true`.""".stripMargin
+          """This directive is not supported.
+            |Please run it with `--power` flag or turn on this flag globally running command `config power true`.""".stripMargin
```
Gedochao:
```suggestion
            |Please run it with the `--power` flag or turn this flag on globally by running `config power true`.""".stripMargin
```

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -34,26 +34,18 @@ object ScalaCli {
...
     } yield !power
 
-    isSipConfigDb.getOrElse(checkName("scala") || checkName("scala-cli-sip"))
+    isSipScala.getOrElse(true)
   }
   def allowRestrictedFeatures = !isSipScala
-  def fullRunnerName          = if (isSipScala) "Scala code runner" else "Scala CLI"
-  def baseRunnerName          = if (isSipScala) "scala" else "scala-cli"
+  def fullRunnerName = if (progName.contains("scala-cli")) "Scala CLI" else "Scala code runner"
+  def baseRunnerName = if (progName.contains("scala-cli")) "scala-cli" else "scala"
```
Gedochao: maybe the `scala-cli` string should be a val?

modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
```diff
@@ -36,8 +36,8 @@ object RestrictedCommandsParser {
         case Right(Some(_, arg, _)) if !arg.isSupported =>
           Left((
             Error.UnrecognizedArgument(
-              s"""`${args(index)}` option is not supported in `scala` command.
-                 |Please run it with `scala-cli` command or with `--power` flag or turn on this flag globally running command `config power true`.""".stripMargin
+              s"""`${args(index)}` option is not supported.
+                 |Please run it with `--power` flag or turn on this flag globally running command `config power true`.""".stripMargin
```
Gedochao:
```suggestion
            |Please run it with the `--power` flag or turn this flag on globally by running `config power true`.""".stripMargin
```

modules/cli/src/main/scala/scala/cli/commands/compile/CompileOptions.scala
```diff
@@ -33,7 +33,6 @@ final case class CompileOptions(
 
   @Name("p")
   @Name("printClasspath")
-  @Tag(tags.restricted)
```
Gedochao, line 36: shouldn't we still have a tag?
```scala
@Tag(tags.should)
```

website/docs/release_notes.md
```diff
@@ -84,7 +84,7 @@ Added by [@lwronski](https://github.com/lwronski) in [#1739](https://github.com/
 
 ### Docker packaging with `using` directives
 It is now possible to configure packaging into a docker image via `using` directives.
-```scala compile
+```bash ignore
```
Gedochao:
````suggestion
```scala compile power
````

File: website/docs/release_notes.md [hunk 6]
```diff
@@ -94,7 +94,7 @@ Added by [@lwronski](https://github.com/lwronski) in [#1753](https://github.com/
 
 ### Pass GraalVM args via a `using` directive
 It is now possible to pass args to GraalVM via the following `using` directive:
-```scala compile
+```bash ignore
```
Gedochao:
````suggestion
```scala compile power
````

Gedochao (approved):
LGTM

Issue #1669, opened 2022-12-08
Output of `scala run --help` is unclear


julienrf:
**Is your feature request related to a problem? Please describe.**

Running `scala run --help` (with `scala` installed via `cs install scala-experimental`) does not explain how the program is executed. Furthermore, it shows a lot of options that should probably stay in incubation in Scala CLI for some time.

~~~ text
$ scala run --help 
Usage: /home/julien/.local/share/coursier/bin/.scala.aux run [options]
Compile and run Scala code.

To pass arguments to the application, just add them after `--`, like:

```sh
scala-cli MyApp.scala -- first-arg second-arg
```

Help options:
  --usage                                        Print usage and exit
  -h, -help, --help                              Print help message and exit
  -help-full, -full-help, --help-full, --full-help  Print help message, including hidden options, and exit

Scala options:
  --js                                           Enable Scala.js. To show more options for Scala.js pass `--help-js`
  --native                                       Enable Scala Native. To show more options for Scala Native pass `--help-native`
  -P, --plugin, --compiler-plugin string*        Add compiler plugin dependencies
  -O, --scala-opt, --scala-option, --scalac-option option  Add a `scalac` option. Note that options starting with `-g`, `-language`, `-opt`, `-P`, `-target`, `-V`, `-W`, `-X`, and `-Y` are assumed to be Scala compiler options and don't require to be passed after `-O` or `--scalac-option`.
  -S, --scala, --scala-version version           Set the Scala version (3.2.1 by default)
  --scalac-help, --help-scalac                   Show help for scalac. This is an alias for --scalac-option -help
  --scalac-verbose, --verbose-scalac             Turn verbosity on for scalac. This is an alias for --scalac-option -verbose
  --script-snippet string*                       Allows to execute a passed string as a Scala script
  --scala-snippet string*                        Allows to execute a passed string as Scala code

Java options:
  --java-home path                               Set the Java home directory
  -j, --jvm jvm-name                             Use a specific JVM, such as `14`, `adopt:11`, or `graalvm:21`, or `system`
  --java-snippet string*                         Allows to execute a passed string as Java code
  --execute-java string*                         A synonym to --scala-snippet, which defaults the sub-command to `run` when no sub-command is passed explicitly
  -cp, --jar, --jars, --class, --classes, -classpath, --classpath, --extra-jar, --extra-jars, --class-path, --extra-class, --extra-classes, --extra-class-path paths  Add extra JARs and compiled classes to the class path
[...]

PR #1852, opened 2023-02-13
Add more detail for sub-commands' help messages

Gedochao:
Context: #1669 

This is the first batch of help tweaks aiming at improving their usefulness.
It focuses on individual sub-commands' help messages (and not their options' help messages, which will be addressed separately).

Example for the `run` sub-command:

````
▶ scala-cli run --help
Usage: scala-cli run [options]
Compile and run Scala code.

Specific run configurations can be specified with both command line options and using directives defined in sources.
Command line options always take priority over using directives when a clash occurs, allowing to override configurations defined in sources.
Using directives can be defined in all supported input source file types.

For a run to be successful, a main method must be present on the classpath.
.sc scripts are an exception, as a main class is provided in their wrapper.

Multiple inputs can be passed at once.
Paths to directories, URLs and supported file types are accepted as inputs.
Accepted file extensions: .scala, .sc, .java, .jar, .md, .jar, .c, .h, .zip
For piped inputs use the corresponding alias: _.scala, _.java, _.sc, _.md
All supported types of inputs can be mixed with each other.

To pass arguments to the actual application, just add them after `--`, like:

```sh
/Users/pchabelski/IdeaProjects/scala-cli/out/cli/base-image/nativeImage.dest/scala-cli run Main.scala AnotherSource.scala -- first-arg second-arg
```

For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/run

Help options:
  --usage                                        Print usage and exit
  (...)
````

### CODE REVIEW

File: modules/cli/src/main/scala/scala/cli/commands/run/RunOptions.scala [hunk 1]
```diff
@@ -3,16 +3,26 @@ package scala.cli.commands.run
...
+                 |
+                 |For a run to be successful, a main method must be present on the classpath.
+                 |.sc scripts are an exception, as a main class is provided in their wrapper.
+                 |
+                 |${HelpMessages.acceptedInputs}
+                 |
+                 |To pass arguments to the actual application, just add them after `--`, like:
+                 |
+                 |```sh
+                 |${ScalaCli.progName} run Main.scala AnotherSource.scala -- first-arg second-arg
```
lwronski:
```suggestion
                |${ScalaCli.progName}  Main.scala AnotherSource.scala -- first-arg second-arg
```
Do we want to suggest here running `run` command? as it is default command

modules/cli/src/main/scala/scala/cli/commands/version/VersionOptions.scala
```diff
@@ -10,13 +10,14 @@ import scala.cli.signing.util.ArgParsers.*
...
-     |The version of the scala runner is the version of the command-line tool that runs Scala programs, which
-     |is distinct from the Scala version of a program. We recommend you specify the version of Scala of a
-     |program in the program itself (via a configuration directive). Otherwise, the runner falls back to the default
-     |Scala version defined by the runner.
-     |""".stripMargin
+  s"""|Prints the version of the $fullRunnerName and the default version of Scala (which can be overridden in the project).
+      |If network connection is available, this sub-command also checks if the installed $fullRunnerName is up-to-date.
+      |
+      |The version of the $fullRunnerName is the version of the command-line tool that runs Scala programs, which
+      |is distinct from the Scala version of a program. We recommend to specify the version of Scala for a
```
lwronski: >  Scala version of a program.

Do you mean `Scala version of compiler`?

modules/cli/src/main/scala/scala/cli/commands/test/TestOptions.scala
```diff
@@ -6,14 +6,27 @@ import caseapp.core.help.Help
...
 
 // format: off
-@HelpMessage("Compile and test Scala code")
+@HelpMessage(s"""Compile and test Scala code.
+                |
+                |Test sources are compiled separately (after the 'main' sources), and may use different dependencies, compiler options, and other configurations.
+                |A source file is treated as a test source if:
+                |  - it contains the `//> using target.scope "test"` directive
+                |  - the file name ends with `.test.scala`
+                |  - the file comes from a directory that is provided as input, and the relative path from that file to its original directory contains a `test` directory
```
lwronski:
```suggestion
                |  - the file comes from a directory that is provided as input, and the relative path from that file to its workspace directory contains a `test` directory
```

modules/cli/src/main/scala/scala/cli/commands/setupide/SetupIdeOptions.scala
```diff
@@ -2,11 +2,28 @@ package scala.cli.commands.setupide
...
 // format: off
-@HelpMessage("Generate a BSP file that you can import into your IDE")
+@HelpMessage(
+  s"""Generates a BSP file that you can import into your IDE.
+     |
+     |The `setup-ide` sub-command allows to pre-configure a $fullRunnerName project to import to an IDE with BSP support.
+     |It is also ran implicitly when `compile`, `run`, `shebang` or `test` sub-commands are called.
+     |
+     |The pre-configuration should be saved in a BSP json connection file under the path:
+     |    {project-root}/.bsp/$baseRunnerName.json
```
lwronski, line 22: From the user's point of view, this information is irrelevant. I would suggest to  remove it

modules/cli/src/main/scala/scala/cli/commands/publish/PublishLocalOptions.scala
```diff
@@ -3,16 +3,20 @@ package scala.cli.commands.publish
...
-  SharedWatchOptions
-}
+import scala.cli.commands.shared._
 
 // format: off
+@HelpMessage(
+  s"""Publishes build artifacts to the local Ivy2 repository.
+     |
+     |The local Ivy2 repository usually lives under `~/.ivy2/local`.
+     |It is taken into account most of the time by most Scala tools when fetching artifacts.
```
lwronski, line 13:
```suggestion
     |It is taken into account by Scala tools when fetching artifacts.
```
Are there tools that don't fetch dependencies from local ivy?

modules/cli/src/main/scala/scala/cli/commands/bsp/BspOptions.scala
```diff
@@ -2,10 +2,21 @@ package scala.cli.commands.bsp
...
 
 // format: off
-@HelpMessage("Start BSP server")
+@HelpMessage(
+  s"""Start BSP server.
+     |
+     |BSP stands for Build Server Protocol.
+     |For more information refer to https://build-server-protocol.github.io/
+     |
+     |This sub-command is not designed to be used by a human.
```
lwronski, line 16:
```suggestion
     |This sub-command is not designed to be used by a user.
```

modules/cli/src/main/scala/scala/cli/commands/package0/PackageOptions.scala
```diff
@@ -13,15 +13,17 @@ import scala.build.{BuildThreads, Positioned}
...
 import scala.cli.commands.shared.*
 
-// format: off
-@HelpMessage(
+@HelpMessage({
+  val cmdName = "package"
   s"""Compile and package Scala code.
      |
-     |${HelpMessages.commandConfigurations("package")}
+     |${HelpMessages.commandConfigurations(cmdName)}
```
lwronski, line 20: Why not use `Package.name`? The same suggestion for similar cases

modules/cli/src/main/scala/scala/cli/commands/run/RunOptions.scala
Gedochao: it is the help for the `run` sub-command though... so I'd rather not point to the default sub-command here, if only to avoid confusion

modules/cli/src/main/scala/scala/cli/commands/test/TestOptions.scala
Gedochao: I'd be careful about calling it the `workspace`directory, as we call the `.scala-build` directory that elsewhere: https://scala-cli.virtuslab.org/docs/reference/cli-options/#workspace-options

modules/cli/src/main/scala/scala/cli/commands/publish/PublishLocalOptions.scala
Gedochao: I mean... Maven doesn't, I suppose.

Anyhow, it's a direct quote from our [publish local doc](https://scala-cli.virtuslab.org/docs/commands/publishing/publish-local)

modules/cli/src/main/scala/scala/cli/commands/bsp/BspOptions.scala
Gedochao: I mean... technically, automation or an IDE is also a user.

however, the BSP command isn't designed to communicate in a human-comprehensible way, but it is meant to be machine-processable instead. Thus - it's not designed to be used by a human

modules/cli/src/main/scala/scala/cli/commands/package0/PackageOptions.scala
Gedochao: `Package` is parametrised with `PackageOptions`, so `Package.name` isn't available before `PackageOptions` is loaded.
in other words, that wouldn't compile

modules/cli/src/main/scala/scala/cli/commands/setupide/SetupIdeOptions.scala
Gedochao: We do mention it in our [IDE support docs](https://scala-cli.virtuslab.org/docs/guides/ide)... why not mention it here?
by verifying if the `.bsp` directory has been created, it's easier for the user to check if `setup-ide` has been run in a directory. it also marks the project root to be opened in his IDE.
IMO won't hurt to mention it in the help as well...

modules/cli/src/main/scala/scala/cli/commands/version/VersionOptions.scala
Gedochao: Yes. 
Fixed, should be clearer now

lwronski (approved):

PR #1865, opened 2023-02-17
bloop-rifle: increase timeout values

Flowdalic:
To allow scala-cli being run on low-end devcies, increase the timeouts to be more generous. Yes, it does not make much fun (yet) to use scala-cli on those devices, waiting 6 minutes for bloop to get exec'ed, but having larger timeouts at least enables user to run Scala scripts. Whereas with the smaller timeouts, scala-cli aborts for no good reason.

The increased timeouts may also help to prevent sporadic CI failures due to false negatives, i.e., because scala-cli run into such a relative low timeout -- some timeouts where as low as 30 seconds -- on busy CI nodes.

### DIFF
--- a/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifleConfig.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifleConfig.scala
@@ -119,6 +119,10 @@ object BloopRifleConfig {
       .orElse(fromProps)
       .getOrElse(hardCodedDefaultScalaVersion)
   }
+  lazy val extraTimeout: FiniteDuration = Option(System.getenv("SCALA_CLI_EXTRA_TIMEOUT"))
+    .map(Duration(_))
+    .collect { case d: FiniteDuration => d }
+    .getOrElse(Duration.Zero)
 
   def default(
     address: Address,
@@ -136,10 +140,10 @@ object BloopRifleConfig {
       bspStdout = None,
       bspStderr = None,
       period = 100.milliseconds,
-      timeout = 10.seconds,
+      timeout = 10.seconds + extraTimeout,
       startCheckPeriod = 100.millis,
-      startCheckTimeout = 1.minute,
-      initTimeout = 30.seconds,
+      startCheckTimeout = 1.minute + extraTimeout,
+      initTimeout = 30.seconds + extraTimeout,
       minimumBloopJvm = 8,
       retainedBloopVersion = AtLeast(BloopVersion(Constants.bloopVersion))
     )
--- a/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/internal/Operations.scala
+++ b/modules/bloop-rifle/src/main/scala/scala/build/blooprifle/internal/Operations.scala
@@ -473,7 +473,7 @@ object Operations {
     val nailgunClient0 = nailgunClient(address)
     val streams        = Streams(None, out, err)
 
-    timeout(30.seconds, scheduler, logger) {
+    timeout(30.seconds + BloopRifleConfig.extraTimeout, scheduler, logger) {
       nailgunClient0.run(
         "about",
         Array.empty,


lwronski:
> To allow scala-cli being run on low-end devcies, increase the timeouts to be more generous. Yes, it does not make much fun (yet) to use scala-cli on those devices, waiting 6 minutes for bloop to get exec'ed, but having larger timeouts at least enables user to run Scala scripts. Whereas with the smaller timeouts, scala-cli aborts for no good reason.
> 
> The increased timeouts may also help to prevent sporadic CI failures due to false negatives, i.e., because scala-cli run into such a relative low timeout -- some timeouts where as low as 30 seconds -- on busy CI nodes.

@Flowdalic Hi, thanks for your contribution, What do you think about making it configurable via an environment variable?

Flowdalic:
> @Flowdalic Hi, thanks for your contribution, What do you think about making it configurable via an environment variable?

Ideally timeouts like probably most of those in question would be not necessary. For example `startCheckTimeout` should not strictly be required, because either the server is still starting, or it failed with an exit code. Of course, we do not life in an ideal world and the starting server could be trapped in the "server starting" state for eternity because, for example, it entered an endless loop due to a bug. However, I believe one minute, that is the current value, to be to low for that kind of timeouts, because on heavily-loaded machines processes may only get a few milliseconds in a one minute window. I also wouldn't be surprised if this low timeout is the cause of sporadic CI failures.

Therefore, I think the current values are simply to low. If the new values I chose are perfectly fine is a good bikeshedding topic. I tried to make them as low as possible so that they still would allow me to run scala-cli on my low-end system.

In any case, I think we can agree that hard-coding such values, without a mechanism to override them, is not desirable. Having an environment variable to override those is better than nothing and we can certainly start with that. But ideally scala-cli would also grow support for user and system configuration files (the former overriding values from the latter). As I actually would want to configure those values system-wide in e.g., `/etc/scala-cli/config.yaml` on every low-end machines, if we can not agree on increasing the current default timeouts.

lwronski:
@Flowdalic Thank you for initiating this discussion. I apologize for the delayed response, as I have just returned from vacation.

> I also wouldn't be surprised if this low timeout is the cause of sporadic CI failures.

What errors on the CI are you referring to? If you are referring to the errors in the Scala CLI repository, then most of them are caused by internet connection problems, `503` errors often occur. 

> In any case, I think we can agree that hard-coding such values, without a mechanism to override them, is not desirable. Having an environment variable to override those is better than nothing and we can certainly start with that. But ideally scala-cli would also grow support for user and system configuration files (the former overriding values from the latter). As I actually would want to configure those values system-wide in e.g., /etc/scala-cli/config.yaml on every low-end machines, if we can not agree on increasing the current default timeouts. 


Making configurable timeouts via `/etc/scala-cli/config.yaml` would be a desirable solution. Alternatively, we could use the `configDb` mechanism in Scala CLI, which allows users to define global configurations (per user). This way, users can set global configurations that apply to all their Scala CLI projects.

```
scala-cli config bloop-rifle-timeout 10.minutes
```

This `configDb` is stored in the home directory of the user. It could be handled something similar as [here](https://github.com/VirtusLab/scala-cli/blob/7fa41b99b2c983918c0301da58a71df5eebf0772/modules/cli/src/main/scala/scala/cli/ScalaCli.scala#L40) and new key `bloop-rifle-timeout` should be added [here](https://github.com/VirtusLab/scala-cli/blob/7fa41b99b2c983918c0301da58a71df5eebf0772/modules/config/src/main/scala/scala/cli/config/Keys.scala#L40).
 

I'm not convinced that changing the previous default to high values. Sometimes, users may have to wait a long time to get any response from `scala-cli`. However, it should be possible to change the default values, and for me, it is sufficient to make it configurable via environment variables at the beginning. If you are interested in implementing it through the `configDb` mechanism in `scala-cli`, let me know, and I can explain the mechanism further.

Flowdalic:
> @Flowdalic Thank you for initiating this discussion. I apologize for the delayed response, as I have just returned from vacation.

No worries. I hope you enjoyed your vacation. :)

> Alternatively, we could use the `configDb` mechanism in Scala CLI, which allows users to define global configurations (per user). This way, users can set global configurations that apply to all their Scala CLI projects.

A per-user mechanism is not really an alternative. Consider fleet of low end devices that run scala-cli to execute Scala scripts, probably under various different users for isolation purposes. Then what you really want is a system-wide setting that is deployed (and managed) via a configuration management system. Ideally a file under `/etc/scala-cli`, following the pattern of other applications like ssh and git.

> I'm not convinced that changing the previous default to high values. Sometimes, users may have to wait a long time to get any response from `scala-cli`.

Yes, but getting a positive response after a relatively long time is better then getting a negative response after a short time. What makes matters worse, the negative response is not caused by an actual failure, just by waiting not long enough. The only reason you do not want to wait infinitely is to protect users from cases where the child process ran into an endless loop due to an bug.

But how often do users actually experience such a bug? And compare this to how often users are unable to run scala-cli just because the timeouts where to low? I would assume the latter is far more often than the former. Which means increasing the timeouts is sensible, even if it means that in rare cases, i.e., when the child process entered an endless loop, users have to wait longer till they get a negative response.

In any case, I am going to implement the environment variable approach. As it allows me to keep experimenting with scala-cli on low-end devices. But I really think that those timeouts that simply be increased. At least, if scala-cli is run on a low-end device.

lwronski:
@Flowdalic Thanks that you continue working on this PR.

> Yes, but getting a positive response after a relatively long time is better then getting a negative response after a short time. What makes matters worse, the negative response is not caused by an actual failure, just by waiting not long enough. The only reason you do not want to wait infinitely is to protect users from cases where the child process ran into an endless loop due to an bug.

It can be difficult to find the right balance for the timeout value. Sometimes, when you get stuck, you may have to wait a long time to receive a response from Scala CLI. However, from another side sometimes it is better to receive a positive response after a long wait than to experience a failure.


> But how often do users actually experience such a bug? And compare this to how often users are unable to run scala-cli just because the timeouts where to low? I would assume the latter is far more often than the former. Which means increasing the timeouts is sensible, even if it means that in rare cases, i.e., when the child process entered an endless loop, users have to wait longer till they get a negative response.

Could you provide more information about the errors being referred to? maybe some link to an issue or thread? Unfortunately, there are some bloop errors that occur, although these are typically caused by other issues rather than the timeout problem which we're trying to address on a regular basis.

Flowdalic:
I just made the case that low timeouts are more likely to cause unnecessary aborts as they prevent infinite wait. And that it is therefore sensible to increase them so that they cause less unnecessary aborts. That is really all I wrote. :)

lwronski:
> I just made the case that low timeouts are more likely to cause unnecessary aborts as they prevent infinite wait. And that it is therefore sensible to increase them so that they cause less unnecessary aborts. That is really all I wrote. :)

I suggest merging your changes now, and if other users also report that the timeout is too low, we can increase the default timeout. Does this sound good to you?

modules/bloop-rifle/src/main/scala/scala/build/blooprifle/BloopRifleConfig.scala
```diff
@@ -119,6 +119,10 @@ object BloopRifleConfig {
       .orElse(fromProps)
       .getOrElse(hardCodedDefaultScalaVersion)
   }
+  lazy val extraTimeout: FiniteDuration = Option(System.getenv("BLOOP_EXTRA_TIMEOUT"))
```
lwronski:
```suggestion
  lazy val extraTimeout: FiniteDuration = Option(System.getenv("SCALA_CLI_EXTRA_TIMEOUT"))
```

We can make this variable more name generic so that we can use this value in other places in the future.

lwronski (approved):
LGTM @Flowdalic Thanks!

PR #1867, opened 2023-02-17
Drop metabrowse command

lwronski:


tgodzik (approved):
LGTM

A comment from me, this is being done in order to limit the amount of features that we need to support, especially that the metabrowse project is no being very actively worked on. I appreciate the work that was done here and we can return the command in the future if needed. Currently this can just be replaced by running Intellij or Metals

CC @alexarchambault I hope it's not too radical, but I feel like having too many features is something raised against scala-cli a lot and we can reduce the ones that can be work around.

Gedochao (approved):

PR #1872, opened 2023-02-20
Shorter options help

Gedochao:
Fixes #1669 

- bump `scala-cli-signing` and `case-app`
- only show `tags.important` or `tags.must` help options in the short help (`sub-command --help`)
- example output:
```bash
$ scala-cli fmt -h
Usage: scala-cli fmt [options]
Formats Scala code.

You are currently viewing the basic help for the fmt sub-command. You can view the full help by running: 
   scala-cli fmt --help-full
For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/fmt

Format options:
  --check                                     Check if sources are well formatted
  --save-scalafmt-conf                        Saves .scalafmt.conf file if it was created or overwritten
  -F, --scalafmt-arg string*                  Pass an argument to scalafmt.
  --scalafmt-conf, --scalafmt-config string?  Custom path to the scalafmt configuration file.
  --dialect, --scalafmt-dialect string?       Pass a global dialect for scalafmt. This overrides whatever value is configured in the .scalafmt.conf file or inferred based on Scala version used.
  --fmt-version, --scalafmt-version string?   Pass scalafmt version before running it (3.6.1 by default). If passed, this overrides whatever value is configured in the .scalafmt.conf file.

Help options:
  --help-js                                      Show options for ScalaJS
  --help-native                                  Show options for ScalaNative
  --help-doc, --doc-help, --help-scaladoc, --scaladoc-help  Show options for Scaladoc
  --help-repl, --repl-help                       Show options for Scala REPL
  --help-fmt, --fmt-help, --help-scalafmt, --scalafmt-help  Show options for Scalafmt
```

PR #1880, opened 2023-02-22
NIT Use enums for help groups and help command groups

Gedochao:
Follow-up for https://github.com/VirtusLab/scala-cli/pull/1872#discussion_r1112959410

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroups.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroups.scala
@@ -1,0 +1,36 @@
+package scala.cli.commands.shared
+
+enum HelpGroup:
+  case Benchmarking, BSP, BuildToolExport,
+    Config, Compilation, CompilationServer,
+    Debian, Debug, Default, Dependency, Doc, Docker,
+    Entrypoint,
+    Format,
+    Help,
+    Install,
+    Java,
+    Launcher, LegacyScalaRunner, Logging,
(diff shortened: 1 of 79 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroups.scala
```diff
@@ -0,0 +1,13 @@
+package scala.cli.commands.shared
+
+enum HelpGroup:
+  case Help, Logging, Dependency, Scala, Java, Entrypoint, BSP, Markdown, Benchmarking,
+    `Scala Native`, `Build Tool export`, `Scala.js`, Debug, Uninstall, Watch, Run, Config, Update,
```
tgodzik:
```suggestion
    ScalaNative, BuildToolExport, ScalaJS, Debug, Uninstall, Watch, Run, Config, Update,
```
maybe and have a custom toString? I just prefer to avoid values enclosed in backticks, there're less clear and a lot of the places don't highlight them well (for example here)

tgodzik (changes requested):
Looks good, but could we avoid backticked cases?

modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroups.scala
Gedochao: Should be better now

tgodzik (approved):
LGTM! Thanks for indulging my OCD

PR #1897, opened 2023-03-01
Add extra class path to generated bootstrap launcher

### LINKED ISSUE #1891: sbt-assembly fat jars run fine using "run" command but not when packaged or turned into docker image
Scala CLI version: 0.2.0

I noticed a bug where a fatjar created with assembly-sbt works fine in programs consuming it from sbt. However, the fatjar will fail in scala-cli when compiled as a package or Docker image. It does work if you simply "run" it. Ultimately, getting this running Docker means that I can create many transformation jobs much simpler compared to using sbt. These would eventually run on Kubernetes orchestrated by Airflow. 

The same fatjar is used in all consuming code.

Step 1. Create a fat jar - I removed all the jars but one to keep this simple. The same issue persists when creating a fatjar of one jar.
```scala
name := "testjar"
version := "0.1.0"
scalaVersion := "3.2.2"

val hadoopVersion = "3.3.4"
val hadoopClient = "org.apache.hadoop" % "hadoop-client" % hadoopVersion

enablePlugins(AssemblyPlugin)

libraryDependencies ++= Seq(
  hadoopClient
)

assemblyMergeStrategy := {  
	case PathList("META-INF", xs @ _*) => MergeStrategy.discard
	case _ => MergeStrategy.first
}
autoScalaLibrary := false
assemblyJarName := s"${name.value}_${scalaBinaryVersion.value}_${version.value}.jar"
```
Step 2. Build a program using sbt - this code works
`build.sbt`
```scala
name := "assTest"
version := "1.0"
scalaVersion := "3.2.2"

enablePlugins(AssemblyPlugin)

unmanagedJars in Compile += file(
  "/src/testjar/target/scala-3.2.2/testjar_3_0.1.0.jar"
)

assemblyJarName := s"${name.value}_${scalaBinaryVersion.value}_${version.value}.jar"
```
`main.scala`
```scala
import org.apache.hadoop.conf.Configuration

object Main extends App:
  val hadoopConf = new Configuration()
  hadoopConf.set("fs.s3a.path.style.access", "true")

  println("Hello, world")
  println(hadoopConf.get("fs.s3a.path.style.access", "true"))
```
3. Run the same fat jar using this code:
`test.scala`
This Works: `scala-cli run test.scala --main-class Main `
```scala
//> using jar "/src/testjar/target/scala-3.2.2/testjar_3_0.1.0.jar"

import org.apache.hadoop.conf.Configuration
object Main extends App:
  val hadoopConf = new Configuration()
  hadoopConf.set("fs.s3a.path.style.access", "true")

  println("Hello, world")
  println(hadoopConf.get("fs.s3a.path.style.access", "true"))
``` 
4. Now build as package
```
scala-cli --power package  -f test.scala --main-class Main -o test 
Wrote /Users/kf/src/json2parquet-cli/test, run it with
  ./test
```
Builds fine
5. `./test` fails with 
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration
        at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3402)
        at java.base/java.lang.Class.getMethodsRecursive(Class.java:3543)
        at java.base/java.lang.Class.getMethod0(Class.java:3529)
        at java.base/java.lang.Class.getMethod(Class.java:2225)
        at coursier.bootstrap.launcher.a.a(Unknown Source)
        at coursier.bootstrap.launcher.ResourcesLauncher.main(Unknown Source)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.conf.Configuration
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:587)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
        ... 7 more
```
6. Now build as assembly `scala-cli --power package  -f test.scala --main-class Main -o test --assembly`
builds fine
7. `./test` fails with 
```
Error: Unable to initialize main class Main
Caused by: java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration
```

When it works, it should print
```
Hello, world
true
```

lwronski:
Fixes #1891

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -729,8 +729,11 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
             ClassPathEntry.Url(url)
       }
     val byteCodeEntry = ClassPathEntry.Resource(s"${destPath.last}-content.jar", 0L, tmpJarContent)
+    val extraClassPath = build.options.classPathOptions.extraClassPath.map { classPath =>
+      ClassPathEntry.Resource(classPath.last, os.mtime(classPath), os.read.bytes(classPath))
+    }
 
-    val allEntries    = Seq(byteCodeEntry) ++ dependencyEntries
+    val allEntries    = Seq(byteCodeEntry) ++ dependencyEntries ++ extraClassPath
     val loaderContent = coursier.launcher.ClassLoaderContent(allEntries)
     val preamble = Preamble()
       .withOsKind(Properties.isWin)
@@ -850,7 +853,8 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
       }
 
     val provided = build.options.notForBloopOptions.packageOptions.provided ++ extraProvided
-    val allFiles = build.artifacts.artifacts.map(_._2)
+    val allFiles =
+      build.artifacts.artifacts.map(_._2) ++ build.options.classPathOptions.extraClassPath
     val files =
       if (provided.isEmpty) allFiles
       else {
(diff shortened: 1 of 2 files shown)


Gedochao (commented):
can we add an anti-regression test?

Gedochao (approved):
LGTM

PR #1898, opened 2023-03-02
Prevent null exception, throw exception on secret key / password missing

### LINKED ISSUE #1899: Misleading output of publish setup when pgp keys stored in config
**Version(s)**
Scala CLI version: 0.2.0

**Describe the bug**
When using the `publish setup` subcommand all the parameters get written to `publish-conf.scala` except `publish.secretKey` and `publish.secretKey` if they reside in config (which is sensible since using directives don't support "config:..."). But the output of the subcommand is as follows:

(the subcommand is run for the second time so the default values are almost all there except pgp values)

```
1 option needs to be set
  pgp-secret-key


Wrote ./publish-conf.scala

Project is ready for publishing!
To publish your project, run
  scala-cli publish .
``` 
But in fact nothing is written to ./publish-conf.scala and running `scala-cli publish .` results in artifacts not getting signed which is mandatory for publishing to Sonatype repo (`"central-s01"`).
So correct command in this case is 
```
scala-cli publish  \
  --secret-key config:pgp.secret-key \
  --secret-key-password config:pgp.secret-key-password .
```
as in [documentation](https://scala-cli.virtuslab.org/docs/commands/publishing/publish#signing).

**To Reproduce**
clone a github project and use commands:
```
scala-cli config --create-pgp-key
scala-cli config publish.credentials s01.oss.sonatype.org env:SONATYPE_USER env:SONATYPE_PASSWORD
scala-cli publish setup .
```

**Expected behaviour**
The `publish setup` should take care of everything so using directives should support `"config:..."` or a proper output shudl be provided given user stores pgp keys in config:

```
1 option needs to be set
  pgp-secret-key


Values for pgp-secret-key and pgp-secret-password found in config.

Project is ready for publishing!
To publish your project, run
  scala-cli publish  \
    --secret-key config:pgp.secret-key \
    --secret-key-password config:pgp.secret-key-password .
```

MaciejG604:
When using `scala-cli publish` with `--secret-key` but no value for   `--secret-key-password` an ambiguous error is presented to the user due to `null` being referenced (line 888 in old version). This should fix the exception and also warn users if one of the values is missing.

Also fixes #1899 
Follow-up to #1432

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
@@ -942,7 +942,7 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
               "secret key",
               "--secret-key",
               directiveName = "",
-              configKeys = Seq("pgp.secret-key"),
+              configKeys = Seq(Keys.pgpSecretKey.fullName),
               extraMessage = shouldSignMsg
             )
           )
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
@@ -865,6 +865,9 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
         }
       }
 
+      if (secretKeyPasswordOpt.isEmpty)
+        logger.diagnostic("PGP signing with no password is not recommended since it's not stable")
+
       if (forceSigningBinary)
         (new scala.cli.internal.BouncycastleSignerMakerSubst).get(
           secretKeyPasswordOpt.fold(null)(_.toCliSigning),
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
@@ -948,7 +948,10 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
           )
         } yield getBouncyCastleSigner(secretKey, getSecretKeyPasswordOpt)
       case _ =>
-        logger.message("Artifacts not signed as it's not required nor has it been specified")
+        if (!publishOptions.contextual(isCi).signer.contains(PSigner.Nop))
+          logger.message(
+            "\ud83d\udd13 Artifacts NOT signed as it's not required nor has it been specified"
+          )
         Right(NopSigner)
     }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
@@ -948,7 +948,7 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
           )
         } yield getBouncyCastleSigner(secretKey, getSecretKeyPasswordOpt)
       case _ =>
-        logger.diagnostic("Not signing files as it's not needed nor has it been specified")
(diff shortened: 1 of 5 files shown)


modules/integration/src/test/scala/scala/cli/integration/PublishTestDefinitions.scala
```diff
@@ -271,4 +271,68 @@ abstract class PublishTestDefinitions(val scalaVersionOpt: Option[String])
       expect(mainClasses == Set(scalaFile1, scalaFile2, s"$scriptsDir.${scriptName}_sc"))
     }
   }
+
+  test("missing secret key pasword") {
+    val baseExpectedArtifacts = Seq(
```
Gedochao: NIT this could be a `Set` from the get-go, then no `.toSet` would be necessary below
```suggestion
    val baseExpectedArtifacts = Set(
```

modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
```diff
@@ -835,77 +836,93 @@ object Publish extends ScalaCommand[PublishOptions] with BuildCommandHelpers {
         }
     }
 
-    val signerOpt = publishOptions.contextual(isCi).signer.orElse {
+    val signerKind = publishOptions.contextual(isCi).signer.getOrElse {
```
Gedochao: NIT I get the feeling that using a `match` clause on `publishOptions.contextual(isCi)` rather than the nested if clauses could significantly improve the readability of this part

MaciejG604: I think that will just replace the getOrElse, so there's no gain from that. Also I've flattened the if clause so it's a little more readable now.

modules/integration/src/test/scala/scala/cli/integration/PublishTestDefinitions.scala
```diff
@@ -271,4 +273,212 @@ abstract class PublishTestDefinitions(val scalaVersionOpt: Option[String])
...
+      "--secret-key", s"file:key.skr",
+      "--signer", "bc"
+    )
+    // format: on
+
+    TestCase.testInputs.fromRoot { root =>
+      os.proc(
+        TestUtil.cli,
+        "--power",
+        "config",
```
alexarchambault: Doesn't this change the global user config? Better set `SCALA_CLI_CONFIG` when running commands in this test (look for `SCALA_CLI_CONFIG` in `RunScalaNativeTestDefinitions` or `ConfigTests` for example), so that the config changes apply only to the tests, and the global config is ignored.

lwronski: Alternatively, we could consider changing the default config directory for integration tests. It could be handled around [here](https://github.com/VirtusLab/scala-cli/blob/0d4c5f762010c13172b5c413f662f63cfa66c803/build.sc#L870).

Even if we forget to set the `SCALA_CLI_CONFIG` in tests, any changes in `config` would not affect the global user config.

MaciejG604:
The mechanism of PGP singing files with BouncyCastle using no secret key password is unstable with native launcher. I've excluded the test for it from native launcher tests, but it should be fixed.

I've created a follow-up issue: #1913

modules/integration/src/test/scala/scala/cli/integration/PublishTestDefinitions.scala
lwronski: > Even if we forget to set the SCALA_CLI_CONFIG in tests, any changes in config would not affect the global user config.

it should be handled in another PR.

lwronski (approved):
LGTM

PR #1903, opened 2023-03-03
Add 'dependency' and 'dependencies' alias for using directive

MaciejG604:


modules/directives/src/main/scala/scala/build/preprocessing/directives/Dependency.scala
```diff
@@ -24,8 +24,10 @@ import scala.cli.commands.SpecificationLevel
 @DirectiveLevel(SpecificationLevel.MUST)
 final case class Dependency(
   @DirectiveName("lib")
+  @DirectiveName("dependency")
   @DirectiveName("libs")
   @DirectiveName("deps")
+  @DirectiveName("dependencies")
   dep: List[Positioned[String]] = Nil
```
lwronski:
```suggestion
  @DirectiveName("lib")
  @DirectiveName("libs")
  @DirectiveName("dep")
  @DirectiveName("deps")
  @DirectiveName("dependencies")
  dependency: List[Positioned[String]] = Nil
```

lwronski (approved):
LGTM.

Gedochao (approved):

PR #1909, opened 2023-03-06
Fix pgp creation option name

MaciejG604:
Changed `config --create-key`, which is incorrect, to `config --create-pgp-key`.

### DIFF
--- a/website/docs/commands/config.md
+++ b/website/docs/commands/config.md
@@ -85,11 +85,12 @@ scala-cli --power config --password github.token
 
 </ChainedSnippets>
 
-Use `--create-key` to create a PGP key pair, protected by a randomly-generated password, to
+Use `--create-pgp-key` to create a PGP key pair, protected by a randomly-generated password, to
 be used by the `publish setup` sub-command:
 ```sh
-scala-cli config --create-key
+scala-cli --power config --create-pgp-key --email "some_email"
 ```
+The `--email` option or `publish.user.email` has to be specified for this subcommand to work properly.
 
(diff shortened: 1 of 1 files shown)


website/docs/commands/misc/config.md
```diff
@@ -79,8 +79,8 @@ scala-cli --power config --password github.token
 
 Use `--create-key` to create a PGP key pair, protected by a randomly-generated password, to
 be used by the `publish setup` sub-command:
-```sh
-scala-cli config --create-key
+```bash
+scala-cli config --create-pgp-key
```
Gedochao:
```suggestion
scala-cli --power config --create-pgp-key
```

website/docs/commands/misc/config.md
```diff
@@ -77,10 +77,10 @@ scala-cli --power config --password github.token
 
 </ChainedSnippets>
 
-Use `--create-key` to create a PGP key pair, protected by a randomly-generated password, to
+Use `--create-pgp-key` to create a PGP key pair, protected by a randomly-generated password, to
 be used by the `publish setup` sub-command:
-```sh
-scala-cli config --create-key
+```bash
+scala-cli --power config --create-pgp-key
```
lwronski: The same situation as in #1898, maybe we can change config directory by default for `scalaclicheck` module. It could be handled around [here](https://github.com/VirtusLab/scala-cli/blob/0d4c5f762010c13172b5c413f662f63cfa66c803/build.sc#L117).

lwronski (approved):

PR #1910, opened 2023-03-07
Add a reference for available config keys in help & docs

### LINKED ISSUE #1623: Add a table of references in the docs for the `config` sub-command
The current docs for the `config` sub-command only mention it in the context of publishing, but we actually also use it for other things (actionable diagnostics, interactive mode, etc).
We could use a reference table for all the keys we currently support with a short description of what they're for.
Optimally, that table would be generated dynamically based on changes in the code, but that could be done separately.

Gedochao:
Fixes #1623 

These changes add the available config keys to the `config` sub-command's `--help` and `--full-help` output (and thus, to the reference docs).
Only non-hidden (important for the average user) keys are listed in the short `--help`.

I also tweaked the `config` sub-command doc:
- it will now be in the main `commands` website level
- it's no longer a power command, so I fixed it not to be tagged as such

Short `config` help output:
```bash
scala-cli config -h
# Usage: scala-cli config [options]
# Configure global settings for Scala CLI.
# 
# Available keys:
#   actions                                        Globally enables actionable diagnostics. Enabled by default.
#   interactive                                    Globally enables interactive mode (the '--interactive' flag).
#   power                                          Globally enables power mode (the '--power' launcher flag).
#   suppress-warning.directives-in-multiple-files  Globally suppresses warnings about directives declared in multiple source files.
#   suppress-warning.outdated-dependencies-files   Globally suppresses warnings about outdated dependencies.
# 
# You are currently viewing the basic help for the config sub-command. You can view the full help by running: 
#    scala-cli config --help-full
# For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/misc/config
# 
# Config options:
#   --unset, --remove  Remove an entry from config
```

Full `config` help output:
```bash
scala-cli config --full-help
# Usage: scala-cli config [options]
# Configure global settings for Scala CLI.
# 
# Syntax:
#   scala-cli config key value
# For example, to globally set the interactive mode:
#   scala-cli config interactive true
#   
# Available keys:
#   actions                                        Globally enables actionable diagnostics. Enabled by default.
#   github.token                                   (hidden) GitHub token.
#   httpProxy.address                              (hidden) HTTP proxy address.
#   httpProxy.password                             (hidden) HTTP proxy password (used for authentication).
#   httpProxy.user                                 (hidden) HTTP proxy user (used for authentication).
#   interactive                                    Globally enables interactive mode (the '--interactive' flag).
#   interactive-was-suggested                      (hidden) Setting indicating if the global interactive mode was already suggested.
#   pgp.public-key                                 (hidden) The PGP public key, used for signing.
#   pgp.secret-key                                 (hidden) The PGP secret key, used for signing.
#   pgp.secret-key-password                        (hidden) The PGP secret key password, used for signing.
#   power                                          Globally enables power mode (the '--power' launcher flag).
#   publish.credentials                            (hidden) Publishing credentials, syntax: s1.oss.sonatype.org value:user value:password
#   publish.user.email                             (hidden) The 'email' user detail, used for publishing.
#   publish.user.name                              (hidden) The 'name' user detail, used for publishing.
#   publish.user.url                               (hidden) The 'url' user detail, used for publishing.
#   repositories.credentials                       (hidden) Repository credentials, syntax: value:user value:password
#   repositories.default                           (hidden) Default repository, syntax: https://first-repo.company.com https://second-repo.company.com
#   repositories.mirrors                           (hidden) Repository mirrors, syntax: repositories.mirrors maven:*=https://repository.company.com/maven
#   suppress-warning.directives-in-multiple-files  Globally suppresses warnings about directives declared in multiple source files.
#   suppress-warning.outdated-dependencies-files   Globally suppresses warnings about outdated dependencies.
# 
# For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/misc/config
# 
# Config options:
#   --dump             (hidden) Dump config DB as JSON
#   --unset, --remove  Remove an entry from config
#  (...)
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
@@ -93,11 +93,11 @@ object ConfigOptions {
     s"""$helpHeader
        |
        |Available keys:
-       |  ${configKeysBulletPoints(includeHidden = false).mkString(s"${System.lineSeparator}  ")}
+       |  ${configKeyMessages(includeHidden = false).mkString(s"${System.lineSeparator}  ")}
        |
        |${HelpMessages.commandFullHelpReference(cmdName)}
        |${HelpMessages.commandDocWebsiteReference(websiteSuffix)}""".stripMargin
-  private def configKeysBulletPoints(includeHidden: Boolean): Seq[String] = {
+  private def configKeyMessages(includeHidden: Boolean): Seq[String] = {
     val allKeys: Seq[Key[_]] = Keys.map.values.toSeq
     val keys: Seq[Key[_]]    = if includeHidden then allKeys else allKeys.filterNot(_.hidden)
     val maxFullNameLength    = keys.map(_.fullName.length).max
@@ -108,7 +108,7 @@ object ConfigOptions {
           if currentKeyFullNameLength > 0 then " " * currentKeyFullNameLength else ""
         val hiddenString =
           if key.hidden then s"${ScalaCliConsole.GRAY}(hidden)${Console.RESET} " else ""
-        s"- ${Console.YELLOW}${key.fullName}${Console.RESET}$extraSpaces  $hiddenString${key.description}"
+        s"${Console.YELLOW}${key.fullName}${Console.RESET}$extraSpaces  $hiddenString${key.description}"
       }
   }
   val detailedHelpMessage: String =
@@ -120,7 +120,7 @@ object ConfigOptions {
        |  ${Console.BOLD}$progName $cmdName interactive true${Console.RESET}
        |  
        |Available keys:
-       |  ${configKeysBulletPoints(includeHidden = true).mkString(s"${System.lineSeparator}  ")}
+       |  ${configKeyMessages(includeHidden = true).mkString(s"${System.lineSeparator}  ")}
        |
        |${HelpMessages.commandDocWebsiteReference(websiteSuffix)}""".stripMargin
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
@@ -88,7 +88,6 @@ object ConfigOptions {
   implicit lazy val help: Help[ConfigOptions]     = Help.derive
   private val helpHeader: String = s"Configure global settings for $fullRunnerName."
   private val cmdName            = "config"
-  private val websiteSuffix      = s"misc/$cmdName"
   val helpMessage: String =
     s"""$helpHeader
        |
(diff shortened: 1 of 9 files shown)


modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
```diff
@@ -93,7 +93,17 @@ object ConfigOptions {
...
        |${HelpMessages.commandDocWebsiteReference(websiteSuffix)}""".stripMargin
-  private val configKeysBulletPoints: Seq[String] = Keys.map.keys.toSeq.sorted.map("- " + _)
+  private val configKeysBulletPoints: Seq[String] = {
+    val keys: Seq[Key[_]] = Keys.map.values.toSeq
+    val maxFullNameLength = keys.map(_.fullName.length).max
+    keys.sortBy(_.fullName)
+      .map { key =>
+        val currentKeyFullNameLength = maxFullNameLength - key.fullName.length
+        val extraSpaces =
+          if currentKeyFullNameLength > 0 then " " * currentKeyFullNameLength else ""
```
lwronski, line 107: ~~You can use the built-in `padTo` function. Something like this should work:~~
```
key.fullName.padTo(maxFullNameLength, " ")
```

modules/config/src/main/scala/scala/cli/config/Key.scala
```diff
@@ -40,6 +40,8 @@ abstract class Key[T] {
 
   def description: String
 
+  def hidden: Boolean = false
```
lwronski, line 45: Add a short description/comment of what is `hidden` in the configuration key entries.

modules/config/src/main/scala/scala/cli/config/Key.scala
```diff
@@ -79,7 +81,8 @@ object Key {
   final class StringEntry(
     val prefix: Seq[String],
     val name: String,
-    val description: String = ""
+    val description: String = "",
+    override val hidden: Boolean = false
```
lwronski, line 87: Is this necessary since the default value for `hidden` is false? The same example below.

modules/config/src/main/scala/scala/cli/config/Key.scala
Gedochao: Good catch, I added some scaladocs for new `Key` fields

modules/config/src/main/scala/scala/cli/config/Key.scala
Gedochao: Maybe not strictly necessary, but IMO it's useful.
It allows to not have to strictly pass the value to the constructor, if a key isn't to be hidden.
And at the same time, the default in `Key` is defined so that it's  inherited for keys initialized via anonymous classes (and has to be overridden there only if a key is hidden, like i.e. `Keys.repositoryCredentials`).

lwronski (approved):
LGTM

PR #1915, opened 2023-03-09
Fix handling for `experimental` features

Gedochao:
- fix mistagged features to be treated as `experimental`:
  - everything tied to publishing
  - the `export` sub-command
  - misc others
- fix `experimental` sub-commands  to only be available in power mode


lwronski (approved):
LGTM

PR #1930, opened 2023-03-16
Move toolkit to scalalang org

szymon-rd:
@lwronski please take a look
To be coordinated with https://github.com/scala/toolkit/pull/10


project/deps.sc
```diff
@@ -181,7 +181,7 @@ object Deps {
   def svm             = ivy"org.graalvm.nativeimage:svm:$graalVmVersion"
   def swoval          = ivy"com.swoval:file-tree-views:2.1.9"
   def testInterface   = ivy"org.scala-sbt:test-interface:1.0"
-  def toolkit         = ivy"org.virtuslab:toolkit:0.1.0"
+  def toolkit         = ivy"org.scala-lang:toolkit:0.1.0"
```
adpi2:
```suggestion
  def toolkit         = ivy"org.scala-lang:toolkit:0.2.0"
```

I am going to publish `0.2.0` because we already have a tag for `0.1.0` on the toolkit repo.

szymon-rd: I just run the CI to deploy 0.1.6, so I am changing to this version. 0.2.0 will be the first public beta (that will be released soon too)

bishabosha:
there are a couple of mentions of `0.1.4` explicitly in the tests, and that fails because that version is not published under `org.scala-lang` organisation

bishabosha:
it seems that https://github.com/VirtusLab/scala-cli/blob/main/website/docs/release_notes.md has an explicit code fence that is being compiled and failing the `docs-tests`:
~~~
```scala compile
//> using toolkit "0.1.4"
```
~~~
should there be conditional processing of the organisation based on the version for backwards compatibility ? otherwise I guess we can stop compiling this snippet

lwronski (approved):
@szymon-rd Thanks! LGTM

PR #1932, opened 2023-03-16
chore: Remove all deprecated using directive styles

tgodzik:
It seems that we still support multiple ways of declaring using directives, which should be removed before 1.0.x.

It's also needed for SIP so that it's clear what can be accepted by the SIP commitee.

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
@@ -18,13 +18,12 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
 case class ExtractedDirectives(
-  offset: Int,
   directives: Seq[StrictDirective],
   positions: Option[DirectivesPositions]
 ) {
   @targetName("append")
   def ++(other: ExtractedDirectives): ExtractedDirectives =
-    ExtractedDirectives(offset, directives ++ other.directives, positions)
+    ExtractedDirectives(directives ++ other.directives, positions)
 }
 
 case class DirectivesPositions(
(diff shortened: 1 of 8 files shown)


lwronski (commented):
Can you also remove invalid example from [documentation](https://scala-cli.virtuslab.org/docs/guides/using-directives#semantics)?

tgodzik:
> Can you also remove invalid example from [documentation](https://scala-cli.virtuslab.org/docs/guides/using-directives#semantics)?

I realized we have some more invalid syntax, so trying to update it altogether. Not sure which tests will fail, so will need to check on CI :sweat_smile:

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -136,28 +128,11 @@ object ExtractedDirectives {
             StrictDirective(k.getPath.asScala.mkString("."), l.asScala.toSeq)
         }
 
-      val offset =
-        if (usedDirectives.getKind != UsingDirectiveKind.Code) 0
-        else usedDirectives.getCodeOffset
```
lwronski, line 141: I think you are using the wrong offset below, for the `UsingDirectiveKind.SpecialComment` directive you should pass offset `0` into `ExtractedDirectives`, but you pass `specialCommentDirectives.getCodeOffset`.

I think you can remove the `offset` field from `ExtractedDirectives`, now it is no longer needed.

lwronski (commented):
If you fix my comment, the unit tests should pass.

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
tgodzik: Removed it!

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -245,27 +245,6 @@ class SourcesTests extends munit.FunSuite {
     }
   }
 
-  test("should fail dependencies in .java  with using keyword") {
```
tgodzik, line 248: I removed the test since code `using` can no longer be used

lwronski (approved):
LGTM, @tgodzik thanks!

tgodzik:
@lwronski the test failure looks unrelated or should I take a look?

PR #1920, opened 2023-03-13
Print warnings when experimental features are used

Gedochao:
Fixes #1281 
Using experimental features will now cause Scala CLI to print an appropriate warning.
The warning can be suppressed with the `--suppress-experimental-warning` option or alternatively with the `suppress-warning.experimental-features` global config key.
```
▶ scala-cli config suppress-warning.experimental-features true
```

Directives:
```bash
▶ scala-cli --power -e '//> using publish.name "my-library"'
# The '//> publish.name "my-library"' directive is an experimental feature.
# Please bear in mind that non-ideal user experience should be expected.
# If you encounter any bugs or have feedback to share, make sure to reach out to the maintenance team at https://github.com/VirtusLab/scala-cli
# Compiling project (Scala 3.2.2, JVM)
# Compiled project (Scala 3.2.2, JVM)
```

Options:
```bash
▶ scala-cli --power -e 'println("Hello")' --markdown
# The '--markdown' option is an experimental feature.
# Please bear in mind that non-ideal user experience should be expected.
# If you encounter any bugs or have feedback to share, make sure to reach out to the maintenance team at https://github.com/VirtusLab/scala-cli
# Compiling project (Scala 3.2.2, JVM)
# Compiled project (Scala 3.2.2, JVM)
# Hello
```
Sub-commands:
```bash
▶ scala-cli --power export 
# The 'export' sub-command is an experimental feature.
# Please bear in mind that non-ideal user experience should be expected.
# If you encounter any bugs or have feedback to share, make sure to reach out to the maintenance team at # https://github.com/VirtusLab/scala-cli
# Exporting to a sbt project...
# [error]  No inputs provided (expected files with .scala, .sc, .java or .md extensions, and / or directories).
```
Sub-commands' help:
```bash
▶ scala-cli --power publish -h
# Usage: scala-cli publish [options]
# Publishes build artifacts to Maven repositories.
# 
# You are currently viewing the basic help for the publish sub-command. You can view the full help by running: 
#    scala-cli --power publish --help-full
# For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/publishing/publish
# 
# The 'publish' sub-command is an experimental feature.
# Please bear in mind that non-ideal user experience should be expected.
# If you encounter any bugs or have feedback to share, make sure to reach out to the maintenance team at https://github.com/VirtusLab/scala-cli
# 
# Publishing options:
#   --organization string?                         Organization to publish artifacts under
# (...)
```

PR #1936, opened 2023-03-17
Add a verbosity guide

Gedochao:
Context: https://github.com/VirtusLab/scala-cli/pull/1920#discussion_r1139381785
~Depends on #1920~

### DIFF
--- a/website/docs/guides/verbosity.md
+++ b/website/docs/guides/verbosity.md
@@ -1,0 +1,161 @@
+---
+title: Verbosity
+sidebar_position: 44
+---
+
+import {ChainedSnippets} from "../../src/components/MarkdownComponents.js";
+
+Logging in Scala CLI can be controlled in a number of ways.
+
+```scala title=Hello.sc
+println("Hello")
+```
+
+Logs, warnings and errors will be printed by default.
+
+<ChainedSnippets>
+
+```bash
+scala-cli Hello.sc
+```
+
+```text  
+Compiling project (Scala 3.2.2, JVM)
+Compiled project (Scala 3.2.2, JVM)
+Hello
+```
+
+</ChainedSnippets>
+
+## Silencing logs with `-q`
+
+All logs except for errors can be silenced by passing the `-q` option.
+
+<ChainedSnippets>
+
+```bash
+scala-cli Hello.sc -q
+```
+
+```text
+Hello
+```
+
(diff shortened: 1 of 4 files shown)


website/docs/guides/verbosity.md
```diff
@@ -0,0 +1,161 @@
...
+
+import {ChainedSnippets} from "../../src/components/MarkdownComponents.js";
+
+Logging in Scala CLI can be controlled in a number of ways.
+
+```scala title=Hello.sc
+println("Hello")
+```
+
+Some things will be printed by default.
```
lwronski: I think, we can add what things we print, maybe something like this?

```suggestion
Logs, warning and error will be printed by default.
```

website/docs/guides/verbosity.md
```diff
@@ -0,0 +1,161 @@
...
+
+```text
+Hello
+```
+
+</ChainedSnippets>
+
+## Increasing verbosity with `-v`
+
+You can increase verbosity by passing the `-v` option.
```
lwronski:
```suggestion
You can increase verbosity by passing the `-v` option to print debugging logs.
```

website/docs/guides/verbosity.md
```diff
@@ -0,0 +1,161 @@
...
+(...)
+Hello
+
+```
+
+</ChainedSnippets>
+
+## Warnings suppression
+
+Some specific warning logs can be suppressed individually.
```
lwronski, line 98: For now, it should be good enough. However, in the future, we may want to consider adding a table that contains options for suppressing them.

website/docs/guides/verbosity.md
Gedochao: fixed

website/docs/guides/verbosity.md
Gedochao: 👍

website/docs/guides/verbosity.md
Gedochao: Agreed.
If we end up having more of them, an automatically generated table would be useful...
We potentially already have a larger one with all the available keys in the `config` sub-command reference doc, but those don't include the corresponding command line options. We'd have to pair them somehow...
But regardless, as you say, it's probably good enough for now.

lwronski (approved):
LGTM

PR #1939, opened 2023-03-20
Performance tweaks

lwronski:


modules/cli/src/main/scala/scala/cli/util/ConfigDbUtils.scala
```diff
@@ -1,9 +1,9 @@
...
 
-import scala.build.errors.BuildException
+import scala.build.errors.{BuildException, ConfigDbException}
 import scala.build.{Directories, Logger}
 import scala.cli.commands.publish.ConfigUtil.wrapConfigException
-import scala.cli.config.{ConfigDb, ConfigDbException, Key}
+import scala.cli.config.{ConfigDb, Key}
 
 object ConfigDbUtils {
   lazy val configDb: Either[ConfigDbException, ConfigDb] =
```
Gedochao, line 9: the `configDb` initialised here could be the same one as everywhere else, I think

Gedochao (approved):

PR #1949, opened 2023-03-21
Make credential entries respect the --password-value option

MaciejG604:
Should work like this - mentioned [here](https://scala-cli.virtuslab.org/docs/commands/publishing/publish-setup#sonatype-credentials)
Also documentation for repositories.credentials was wrong - it accepts a format with repository address at the beginning and an optional realm at the end.

### DIFF
--- a/modules/config/src/main/scala/scala/cli/config/Keys.scala
+++ b/modules/config/src/main/scala/scala/cli/config/Keys.scala
@@ -216,8 +216,9 @@ object Keys {
 
   val repositoryCredentials: Key[List[RepositoryCredentials]] =
     new Key[List[RepositoryCredentials]] {
-      override val description: String = "Repository credentials, syntax: value:user value:password"
-      override val hidden: Boolean     = true
+      override val description: String =
+        "Repository credentials, syntax: repositoryAddress value:user value:password [realm]"
+      override val hidden: Boolean = true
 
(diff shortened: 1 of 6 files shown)


modules/config/src/main/scala/scala/cli/config/Keys.scala
```diff
@@ -341,7 +342,7 @@ object Keys {
 
   val publishCredentials: Key[List[PublishCredentials]] = new Key[List[PublishCredentials]] {
     override val description: String =
-      "Publishing credentials, syntax: s1.oss.sonatype.org value:user value:password"
```
MaciejG604, line 344: Should this be kept as an example address as it was? WDYT?

Gedochao: it's nice to have an example in a guide or general command doc, but IMO reference docs & help outputs (like here) should provide general syntax instead
tl;dr don't keep it here, but it may be worth having in a guide doc or smth

Gedochao (approved):
LGTM

modules/config/src/main/scala/scala/cli/config/Keys.scala
MaciejG604: There are examples of the publish credentials already in `publish` and `publish setup` docs.

PR #1963, opened 2023-03-26
Vcs.parse fix

KuceraMartin:
The `split` method in this form expects a regex. Because the pipe is not escaped, we will get url = first character, conn = second character, and devConn = the rest.

I think this solution is better than escaping the pipe because a string with more than 2 pipes should be considered as malformed input.


lwronski (commented):
@KuceraMartin Thank you for your contribution. Can you add some unit test? You can add a new test suite in this [directory](https://github.com/VirtusLab/scala-cli/tree/main/modules/options/src/test/scala/scala/build/options).

To execute the `parse` method from `Vcs`, you can use the `Positioned.none` constructor:
```
Vcs.parse(Positioned.none("???"))
```

KuceraMartin:
@lwronski It seems that tests in the directory you suggested are not being executed (I created a separate issue: https://github.com/VirtusLab/scala-cli/issues/1976)

For now moving it to a different directory.

lwronski:
> @lwronski It seems that tests in the directory you suggested are not being executed (I created a separate issue: #1976)
> 
> For now moving it to a different directory.

I think, it is good enough if unit tests were moved to a different directory. Can you fix the failing test? and then I would be happy to merge your changes

lwronski (approved):
@KuceraMartin Thanks for your contribution. LGTM

PR #1964, opened 2023-03-27
Add support for nativeLto options in command line and using directives

### LINKED ISSUE #1951: Scala Native should allow nativeLTO option
I am trying to create a reproducer for XCode LLVM tools that only happens when using LTO "thin" option. Scala-cli seems to be the only reasonable way outside of the Scala world to share Scala Native code.

In Scala Native you can have the Link Time Optimization (LTO) set to "thin", "full", or "none" (default)

I would have them install sbt instead and have a special repo on github for the reproducer.m Perhaps use a template and then add sbt options.

**Additional context**
Selecting a garbage collector would be nice as well although default "immix" is good but single threaded.

lwronski:
Fixes #1951.

I added support for passing the `--native-lto` option via the command line:
```
scala-cli Hello.scala --native --native-lto thin
```

or via using directives:

```
$ cat Hello.scala
//> using nativeLto "thin"

object Main extends App {
  println("Hello2")
}
```

### DIFF
--- a/website/docs/reference/cli-options.md
+++ b/website/docs/reference/cli-options.md
@@ -1291,6 +1291,10 @@ Set the Scala Native version (0.4.12 by default).
 
 Set Scala Native compilation mode
 
+### `--native-lto`
+
+Link-time optimisation mode
+
 ### `--native-gc`
 
 Set the Scala Native garbage collector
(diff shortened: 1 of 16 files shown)


website/docs/reference/cli-options.md
```diff
@@ -1285,12 +1285,16 @@ Enable Scala Native. To show more options for Scala Native pass `--help-native`
 
 ### `--native-version`
 
-Set the Scala Native version (0.4.9 by default).
```
Gedochao, line 1288: NIT the reference docs should be regenerated in an earlier commit, where Scala Native is actually bumped

Gedochao (approved):
LGTM

PR #1969, opened 2023-03-28
refactor: change data in bsp.Diagnostic to contain a workspaceEdit

### LINKED ISSUE #1957: Change TextEdit to WorkspaceEdit for actionable diagnostics
There is an ongoing work around the compiler to introduce actionable diagnostics and the consensus is that WorkspaceEdit from LSP spec would be much more powerful.

Change TextEdit to WorkspaceEdit and also make a fallback mechanism in Metals.

ckipp01:
This change is part of the discussion that we had at the Scala Tooling
Summit where we decided that using a `WorkspaceEdit` rather than a
`TextEdit` provided a more future-proof way of actionable diagnostics
since it's richer. This PR makes that changes, but in a limited way.

- It only actually changes this when we go from the `Diagnostic` to a
  BSP Diagnostic. It doesn't for example change the `TextEdit` that the
  internal representation of Diagnostic has. That's left as is to
  minimize changes internally.
- Right now really only `changes` are supported in the `WorkspaceEdit`
  mainly because that's all we really need at this time. In the future
  this could be expanded to use the richer `documentChanges` if the BSP
  client supports it.
- This also changes the structure of `data` slightly to not have the
  actual edit be at the top level and instead nests it under an `edit`
  key. There is no gaurantee that we won't have other keys in `data` in
  the future, so this sort of allows for that and ensures that we don't
  get stuck with only being able to have the actual `WorkspaceEdit` at
  the top level in the `data` object.

Fixes #1957


modules/build/src/main/scala/scala/build/bsp/protocol/DiagnosticData.scala
```diff
@@ -0,0 +1,51 @@
...
+  *   A Map of the URI that contains the diagnostic to the TextEdits that it contains.
+  * @param documentChanges
+  *   NOT CURRENTLY USED
+  * @param changeAnnotations
+  *   NOT CURRENTLY USED
+  */
+case class WorkspaceEdit(
+  changes: Map[String, Array[TextEdit]],
+  documentChanges: Array[Object] = Array.empty,
+  changeAnnotations: Map[String, Object] = Map.empty
```
ckipp01: I wasn't 100% sure what to do with these as I could create an `enum` for the different type of `documentChanges`, but we don't use any of them at the moment. The same with the `changeAnnotations`. So I opted to just use an `Object` here and make them empty. Any thoughts on this or a better alternative? Should I make the entire structure right away even though we aren't using it? During `toJsonTree()` these also get added to the json object with an empty array. I'll need to check if that screws up clients since they might check if `documentChanges` _exists_ and if it does, they might not check `changes`.

modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala
```diff
@@ -85,7 +85,11 @@ class CliLogger(
 
         for (textEdit <- textEditOpt) {
           val bTextEdit = TextEdit(range, textEdit.newText)
-          diag.setData(bTextEdit.toJsonTree())
+          val workspaceEdit = WorkspaceEdit(changes =
+            Map(f.path.getOrElse(os.Path("")).toNIO.toUri().toString() -> Array(bTextEdit))
```
ckipp01: I was also a bit unsure of what to do here. `f.path` is `Either[String, Path]`. I'm assuming this is due to diagnostics potentially not being tied to a file URI, but instead to a target? What's an example of this being left here? We can't leave this as is since `""` is invalid here, so it's just a placeholder until I better understand what the left is used for here. If it's truly meant to representation maybe a target, should we instead detect left and then just not even attempt to add `data` since the `WorkspaceEdit` will need a valid URI?

modules/build/src/main/scala/scala/build/bsp/protocol/DiagnosticData.scala
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/** A WorkspaceEdit meant to mimic the LSP WorkspaceEdit.
+  *
+  * https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit
+  *
+  * For now this only utilizes changes, hench the default empty for both documentChanges and
+  * changeAnnotations. In the future when we support more operations like, CreateFile, RenameFile,
+  * etc we can expland on this and change the type of documentChanges to actually be an
+  * Array[TextDocumentEdit] | Array[TextDocumentEditTypes].
```
Gedochao: NIT typos
```suggestion
  * For now this only utilizes changes, hence the default empty for both documentChanges and
  * changeAnnotations. In the future when we support more operations like, CreateFile, RenameFile,
  * etc we can expand on this and change the type of documentChanges to actually be an
  * Array[TextDocumentEdit] | Array[TextDocumentEditTypes].
```

modules/build/src/main/scala/scala/build/bsp/protocol/DiagnosticData.scala
Gedochao: if we're fairly sure what values we expect there in the future, then defining the entire structure (including enums) straight away sounds good, even if we wouldn't be using them just yet... especially as this will require corresponding changes in clients, if I understand correctly. 
So the earlier the model is there, the better. But I consider it nice-to-have.

modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala
Gedochao: If I'm not mistaken, the potential `Left` value is reserved for virtual inputs (command line snippets, piped sources, etc) which wouldn't have a conventional file path and instead have a String indicating the source here: https://github.com/VirtusLab/scala-cli/blob/c878f742d2c0f0b62579f7283c1b88c9ecb4b124/modules/build/src/main/scala/scala/build/input/Element.scala#L19 
example `Left` values would be `"<stdin>-scala-file"` or  `"<snippet>-java-snippet2"`.
https://github.com/VirtusLab/scala-cli/blob/c878f742d2c0f0b62579f7283c1b88c9ecb4b124/modules/build/src/main/scala/scala/build/input/Inputs.scala#L240

ckipp01: Ah so that should really never be the URI in a diagnostic since it wouldn't actually be a valid file system URI right? If so, I'll just add logic that for `Left` you can't have a valid URI, therefore no `data`. If right, then you can.

Gedochao: Yup, I believe you're correct

modules/build/src/main/scala/scala/build/bsp/protocol/DiagnosticData.scala
ckipp01: Alright, so I broke this out and basically made all the `ResourceOperation`s. I did make a few assumptions though about the spec that I'll comment inline.

modules/build/src/main/scala/scala/build/bsp/protocol/WorkspaceEdit.scala
```diff
@@ -0,0 +1,18 @@
...
+  *
+  * @param changes
+  *   A Map of the URI that contains the diagnostic to the TextEdits that it contains.
+  * @param documentChanges
+  *   An array of resource operations that can just be an array of TextDocumentEdit[] or the various
+  *   file resource operation.
+  */
+case class WorkspaceEdit(
+  changes: Map[String, Array[TextEdits.TextEdit]],
+  documentChanges: Array[ResourceOperation] = Array.empty
```
ckipp01: So _technically_ `documentChanges` here can actually be an `Either[Array[ResourceOperation.TextDocumentEdit], Array[ResourceOperation]]`. What I have here to me is more simplified yet accomplishing the same thing. Do you think this is fine?

modules/cli/src/main/scala/scala/cli/internal/CliLogger.scala
```diff
@@ -83,9 +83,15 @@ class CliLogger(
...
-          diag.setData(bTextEdit.toJsonTree())
+        (textEditOpt, f.path) match {
+          case (Some(textEdit), Right(path)) =>
+            val bTextEdit = TextEdits.TextEdit(range, textEdit.newText)
+            val workspaceEdit = WorkspaceEdit(changes =
+              Map(path.toNIO.toUri().toString() -> Array(bTextEdit))
+            )
+            val data = DiagnosticData(edits = Array(workspaceEdit))
+            diag.setData(data.toJsonTree())
+          case _ => ()
```
ckipp01: Alright, so I did what I mentioned earlier. if this ends up being a `Left` it's not a valid `URI`, so we don't process it for any `data`.

modules/build/src/main/scala/scala/build/bsp/protocol/ResourceOperation.scala
```diff
@@ -0,0 +1,121 @@
...
+  object DeleteFile {
+
+    /** Delete file options
+      *
+      * @param recursive
+      *   Delete the content recursively if a folder is denoted.
+      * @param ignoreIfNotExists
+      *   Ignore the operation if the file doesn't exist.
+      */
+    case class DeleteFileOptions(recursive: Boolean, ignoreIfNotExists: Boolean)
```
ckipp01: So pretty much all of these options don't make a ton of sense to me, but it's what the spec has. The tricky part is that they are all optional, or nullable rather. If I put them as `Option[Boolean]`, then they will render super odd with `recursive: { value: true }` or something like that, which we don't want. However, leaving them both just as it _should_ work, but sort of forces you to set both, which doesn't make a ton of sense. I'm not sure what's the best approach here. Any thoughts?

modules/build/src/main/scala/scala/build/bsp/protocol/ResourceOperation.scala
```diff
@@ -0,0 +1,121 @@
...
+    *   A delete
+    * @param uri
+    *   The file to delete
+    * @param options
+    *   Additional delete options
+    */
+  case class DeleteFile(
+    kind: "delete" = "delete",
+    uri: String,
+    options: DeleteFile.DeleteFileOptions
```
ckipp01: All of these `options` in the various operations are technically optional, nullable. So again, I wasn't sure how to model this to best achieve that to keep the json clean, but also not use nullable. Any thoughts?

ckipp01 (commented):
Still need to fix the tests, but I wanted to get the full structure up here and get some feedback.

modules/build/src/main/scala/scala/build/bsp/protocol/WorkspaceEdit.scala
lwronski: Why not to use `WorkspaceEdit` from `org.eclipse.lsp4j`? and I'm fine with adding new dependency `org.eclipse.lsp4j` to `scala-cli` project

cc: @tgodzik

ckipp01: Ha, I originally thought of that since it's a but silly to have to mimic all that, but also noticed that the specific needed lsp4j package wasn't in scope here. If we can add it, then yea this makes all of this way easier.

lwronski: Sure, we can add it. This won't increase the significant size of the scala-cli launcher.

ckipp01: Ok cool, I'll update this then to go that route... should have asked first 😬

lwronski:
I saw there is issue with generating the native image and therefore native tests fail, I think that reflection configuration is missing for graalvm. Let me check and I will get back to you

lwronski:
> I saw there is issue with generating the native image and therefore native tests fail, I think that reflection configuration is missing for graalvm. Let me check and I will get back to you

I think my commit should resolve the issues with the native tests. It was added the `lsp4j` dependency and then used `Gson` to serialize a class which uses reflection. However, for the GraalVM image, the class/method that uses reflection must be specified in reflect-config.json

ckipp01:
> I think my commit should resolve the issues with the native tests. It was added the lsp4j dependency and then used Gson to serialize a class which uses reflection. However, for the GraalVM image, the class/method that uses reflection must be specified in reflect-config.json

Ahh perfect, thanks! This should be ready to review now but I'll also work on a draft pr in Metals just to show that this still works as expected there as well.

lwronski:
> > I think my commit should resolve the issues with the native tests. It was added the lsp4j dependency and then used Gson to serialize a class which uses reflection. However, for the GraalVM image, the class/method that uses reflection must be specified in reflect-config.json
> 
> Ahh perfect, thanks! This should be ready to review now but I'll also work on a draft pr in Metals just to show that this still works as expected there as well.

Thanks for your contribution @ckipp01 , overall LGTM. So I wait until you finish your work in Metals.

ckipp01:
Alright, so this should be ready to go now. I have a PR in [Metals](https://github.com/scalameta/metals/pull/5173) that shows that it can still handle the old way of doing it, and we can add another test in there as well for the new way of doing it when we get a new release.

Gedochao (approved):
LGTM
edit: will merge this after https://github.com/scalameta/metals/pull/5173 is ready

ckipp01:
> will merge this after https://github.com/scalameta/metals/pull/5173 is ready

That pr is actually ready. The only reason it's marked as a draft is because it's waiting for this one 😆 .

tgodzik:
> That pr is actually ready. The only reason it's marked as a draft is because it's waiting for this one laughing .

och, but if we don't merge it in Metals it will be broken there :sweat_smile: We should merge the Metals one first IMHO

tgodzik:
The metals PR is merged :tada: Do you need help finishing this one @ckipp01 ?

ckipp01:
> The metals PR is merged 🎉 Do you need help finishing this one @ckipp01 ?

Hey! I'll actually jump back on this now. I'll get to it this week.

ckipp01:
Superseded by https://github.com/VirtusLab/scala-cli/pull/2284

PR #1978, opened 2023-03-29
Bump Scala to `3.3.0-RC3`

Gedochao:


Gedochao:
Seems like we won't be going ahead with this after all.

PR #1980, opened 2023-03-29
Enable to pass custom docker-cmd to execute application in docker

lwronski:


modules/cli/src/main/scala/scala/cli/commands/package0/PackagerOptions.scala
```diff
@@ -132,6 +132,12 @@ final case class PackagerOptions(
   )
   @Tag(tags.restricted)
   dockerImageTag: Option[String] = None,
+  @Group(HelpGroup.Docker.toString)
+  @HelpMessage(
+    "Executable that will run an application in docker, default sh"
```
Gedochao: technically, the platform influences what it defaults too, no?
```suggestion
    "Allows to override the executable used to run the application in docker, otherwise it defaults to sh for the JVM platform and node for the JS platform"
```

Gedochao (approved):

PR #1982, opened 2023-03-29
Comply with optional password in scala-cli-signing

### LINKED ISSUE #1079: Rely on a private key rather than secret key + password in publish / publish setup
For now, `publish` (and `publish setup` in https://github.com/VirtusLab/scala-cli/pull/926), when using Bouncycastle-based Scala CLI-managed keys rather than `gpg`, rely on a "secret key" plus a password to sign artifacts, like how sbt-ci-release deals with such keys. The Bouncycastle API offers a better way to handle those keys: using "private keys". In the Bouncycastle API (I guess that terminology must not be specific to it), "private keys" are "clear" keys (can be used without a password to sign or encrypt things), while "secret keys" are private keys protected by a password.

When we upload a key as GitHub release asset, or when we generate one and keep it in the Scala CLI config, there's no need to keep a secret key plus a password - just a private key would do. That way, in `publish setup`, we should only need to create a GitHub repository secret for the private key, rather than one for the secret key and one for its password.

MaciejG604:
Based on: https://github.com/VirtusLab/scala-cli-signing/pull/28
Closes #1079

### DIFF
--- a/project/deps.sc
+++ b/project/deps.sc
@@ -88,7 +88,7 @@ object Deps {
     def scalaMeta          = "4.7.1"
     def scalaNative        = "0.4.9"
     def scalaPackager      = "0.1.29"
-    def signingCli         = "0.1.16"
+    def signingCli         = "0.1.17"
   }
   // DO NOT hardcode a Scala version in this dependency string
   // This dependency is used to ensure that Ammonite is available for Scala versions
--- a/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
@@ -74,7 +74,7 @@ object Config extends ScalaCommand[ConfigOptions] {
             val (pgpPublic, pgpSecret0) =
               ThrowawayPgpSecret.pgpSecret(
                 mail,
-                password,
+                Some(password),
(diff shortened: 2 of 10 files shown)


project/deps.sc
```diff
@@ -88,7 +88,7 @@ object Deps {
     def scalaMeta          = "4.7.1"
     def scalaNative        = "0.4.9"
     def scalaPackager      = "0.1.29"
-    def signingCli         = "0.1.16"
```
Gedochao, line 91: NIT maybe do the bump in a separate commit 🤔

Gedochao:
NIT I'd suggest some commit structure tweaks:
- b8dbf04867e23d913fc662ec7507cc2032bd51f4 and 25ffee2673f88e0249020ba7c1fb1c4132c9061c could be squashed into a single commit, IMO
- c1bd87db432bc3490f93a14f06b85c20f58fcc70 should have the `NIT` prefix in the commit message

project/deps.sc
Gedochao: ah, but it's breaking changes, right? so you actually have to do it together with the other changes... 
disregard this comment then, nvm 😅

lwronski (approved):
LGTM

PR #1985, opened 2023-03-30
Print a better error when workspace path contains the `File.pathSeparator`

### LINKED ISSUE #1411: If a scala file has a colon anywhere in the path (including parent directories) the scala-cli can't run and does not give helpful warnings
**Version(s)**
Probably all but tested on 0.1.15

**Describe the bug**
scala-cli cannot run scala code that is in a path that contains a `:` anywhere in the path. Compiling is possible but running is not. There also is not a helpful warning to help the user, the error is just:

```
Exception in thread "main" java.lang.ClassNotFoundException: main
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        at scala.cli.runner.Runner$.main(Runner.scala:13)
        at scala.cli.runner.Runner.main(Runner.scala)
```

in my case the code was:
```scala
@main def main = println("test")
```
, but the bug is reproducible with any code

**To Reproduce**

1. Create a directory with `:` in the name, for example `scala:cli`
2. Create a scala file in the newly created directory and put in some code, any code, that you know would run perfectly otherwise.
3. Type `scala-cli run .` or `scala-cli run FileName.scala`.

Note: This bug is irrelevant on Windows, since you can't create directories with `:` in the name on Windows. However it is tested on both Mac and Ubuntu and the bug exists on both.

**Expected behaviour**
The code should run as normal or at least give a meaningful warning.

Gedochao:
Fixes #1411 

`:` is the `File.pathSeparator` on UNIX systems, used as the separator on classpaths by the compiler. 
Unfortunately, it seems that there is no way of escaping it, even though `:` is a valid character for directory names.
The recommended workaround is to either move your project to a sane path or at the very least override your workspace with one.

```bash
scala-cli run Hello.scala
# [error]  Invalid workspace path: ~/IdeaProjects/scala-cli-tests/weird:directory
# Workspace path cannot contain a : (colon character).
# Consider moving your project to a different path.
# Alternatively, you can force your workspace with the '--workspace' option:
#     scala-cli run --workspace <alternative-workspace-path> Hello.scala
```

### DIFF
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -1,13 +1,13 @@
 package scala.build.input
 
-import java.io.ByteArrayInputStream
+import java.io.{ByteArrayInputStream, File}
 import java.math.BigInteger
 import java.nio.charset.StandardCharsets
 import java.security.MessageDigest
 
 import scala.annotation.tailrec
 import scala.build.Directories
-import scala.build.errors.{BuildException, InputsException}
+import scala.build.errors.{BuildException, InputsException, WorkspaceError}
 import scala.build.input.ElementsUtils.*
 import scala.build.internal.Constants
 import scala.build.internal.zip.WrappedZipInputStream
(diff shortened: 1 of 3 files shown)


modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -386,15 +347,70 @@ object Inputs {
...
-        extraClasspathWasPassed
-      ))
+      val (workspace, needsHash, workspaceOrigin0) = forcedWorkspace match {
+        case None => (inferredWorkspace, inferredNeedsHash, workspaceOrigin)
+        case Some(forcedWorkspace0) =>
+          val needsHash0 = forcedWorkspace0 != inferredWorkspace || inferredNeedsHash
+          (forcedWorkspace0, needsHash0, WorkspaceOrigin.Forced)
+      }
+
+      if !Properties.isWin && workspace.toString.contains(":") then
```
tgodzik: Do we allow it on windows? I think in that case classpath separator is `;` , so maybe we should have a different one per system?

Gedochao: oh, interesting.
indeed, it seems that Windows allows `;` in filenames (which is a bad idea, as apparently a lot of apps don't handle this well)
I suppose we can generalize this to `File.pathSeparator` to keep it working on all OS...
lemme fix it

Gedochao: fixed, should account for the Windows class path separator now

lwronski (approved):
LGTM.

PR #1987, opened 2023-03-31
Properly handle pgp keychains generated by Scala CLI

MaciejG604:
It is illogical to encrypt the private key (it's actually a keychain) with a password just to keep them together in the same config file. We should require explicitly setting a password or explicitly opting for `none` or `random` (although `random` is the least useful if you think about it, maybe for CI).

The only possible problem with my solution is that this introduces some new option values (random and none), that the user would need to know how to use.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
@@ -142,7 +142,7 @@ object Config extends ScalaCommand[ConfigOptions] {
                   valueOpt match {
                     case Some(value) =>
                       for (v <- value)
-                        if (options.password && entry.isPasswordOption)
+                        if (options.passwordValue && entry.isPasswordOption)
                           PasswordOption.parse(v) match {
                             case Left(err) =>
                               System.err.println(err)
--- a/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
@@ -51,10 +51,15 @@ object Config extends ScalaCommand[ConfigOptions] {
       args.all match {
         case Seq() =>
           if (options.createPgpKey) {
-            val coursierCache       = options.coursier.coursierCache(logger.coursierLogger(""))
-            val secKeyEntry         = Keys.pgpSecretKey
-            val secKeyPasswordEntry = Keys.pgpSecretKeyPassword
-            val pubKeyEntry         = Keys.pgpPublicKey
+            if (options.pgpPassword.isEmpty) {
+              logger.error(
+                s"--pgp-password not specified, use 'none' to create an unprotected keychain or 'random' to generate a password"
+              )
+              sys.exit(1)
+            }
+            val coursierCache = options.coursier.coursierCache(logger.coursierLogger(""))
+            val secKeyEntry   = Keys.pgpSecretKey
+            val pubKeyEntry   = Keys.pgpPublicKey
 
             val mail = options.email
               .filter(_.trim.nonEmpty)
@@ -64,17 +69,23 @@ object Config extends ScalaCommand[ConfigOptions] {
                   .orExit(logger)
               }
               .getOrElse {
-                System.err.println(
-                  s"Error: --email ... not specified, and ${Keys.userEmail.fullName} not set (either is required to generate a PGP key)"
+                logger.error(
+                  s"--email ... not specified, and ${Keys.userEmail.fullName} not set (either is required to generate a PGP key)"
                 )
                 sys.exit(1)
               }
 
-            val password = ThrowawayPgpSecret.pgpPassPhrase()
+            val passwordOpt = if (options.pgpPassword.contains("none"))
+              None
+            else if (options.pgpPassword.contains("random"))
+              Some(ThrowawayPgpSecret.pgpPassPhrase())
+            else
+              options.pgpPassword.map(scala.cli.signing.shared.Secret.apply)
+
             val (pgpPublic, pgpSecret0) =
               ThrowawayPgpSecret.pgpSecret(
                 mail,
-                Some(password),
+                passwordOpt,
                 logger,
                 coursierCache,
                 () =>
@@ -88,11 +99,20 @@ object Config extends ScalaCommand[ConfigOptions] {
             val pgpSecretBase64 = pgpSecret0.map(Base64.getEncoder.encodeToString)
 
             db.set(secKeyEntry, PasswordOption.Value(pgpSecretBase64.toConfig))
-            db.set(secKeyPasswordEntry, PasswordOption.Value(password.toConfig))
             db.set(pubKeyEntry, PasswordOption.Value(pgpPublic.toConfig))
             db.save(directories.dbPath.toNIO)
               .wrapConfigException
               .orExit(logger)
+
+            logger.message("PGP keychains written to config")
+            if (options.pgpPassword.contains("random"))
+              passwordOpt.foreach { password =>
+                println(
+                  s"""Password: ${password.value}
+                     |Don't lose it!
+                     |""".stripMargin
+                )
+              }
           }
           else {
             System.err.println("No argument passed")
--- a/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
@@ -42,12 +42,11 @@ final case class ConfigOptions(
   @Tag(tags.experimental)
     email: Option[String] = None,
   @Group(HelpGroup.Config.toString)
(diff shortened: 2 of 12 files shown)


modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
```diff
@@ -88,11 +99,19 @@ object Config extends ScalaCommand[ConfigOptions] {
...
               .wrapConfigException
               .orExit(logger)
+
+            logger.message("PGP keychains written to config")
+            passwordOpt.foreach { password =>
+              logger.message(
+                s"""Password: ${password.value}
+                   |Don't lose it!
+                   |""".stripMargin
+              )
```
Gedochao: I think this should be printed only if `options.pgpPassword.contains("random")`...
Also, `logger.message` means that this might not get printed when verbosity is decreased, while it's actually crucial to print the random password, right?
so maybe `println` would be better 🤔

modules/cli/src/main/scala/scala/cli/commands/config/ConfigOptions.scala
```diff
@@ -27,10 +27,18 @@ final case class ConfigOptions(
...
   @Group(HelpGroup.Config.toString)
-  @HelpMessage("Create PGP key in config")
+  @HelpMessage("Create PGP keychain in config")
+  @Tag(tags.inShortHelp)
   @Tag(tags.experimental)
     createPgpKey: Boolean = false,
   @Group(HelpGroup.Config.toString)
+  @HelpMessage("A password used to encode the private PGP keychain")
+  @Tag(tags.inShortHelp)
+  @Tag(tags.restricted)
```
Gedochao: everything tied to `publish` should be considered experimental for now, I think
```suggestion
  @Tag(tags.experimental)
```

modules/integration/src/test/scala/scala/cli/integration/ConfigTests.scala
```diff
@@ -179,21 +181,58 @@ class ConfigTests extends ScalaCliSuite {
...
+        pgpPasswordOption
+      )
+        .call(cwd = root, env = extraEnv, mergeErrIntoOut = true)
+
+      val pgpPasswordOpt: Option[String] = pgpCreated.out.text()
+        .linesIterator
+        .toSeq
+        .filter(_.startsWith("Password"))
+        .collect(_.stripPrefix("Password:").trim())
+        .headOption
```
Gedochao:
```suggestion
      val pgpPasswordOpt: Option[String] = pgpCreated.out.text()
        .linesIterator
        .find(_.startsWith("Password"))
        .map(_.stripPrefix("Password:").trim())
```

website/docs/commands/publishing/publish-setup.md
```diff
@@ -75,9 +75,13 @@ scala-cli --power config publish.user.email ""
 
 Generate a PGP key pair for publishing with
 ```bash
-scala-cli --power config --create-pgp-key
+scala-cli --power config --create-pgp-key --pgp-password MY_CHOSEN_PASSWORD
 ```
 
+It's not mandatory, although recomended, to use a password to encrypt your keychains.
+To store the private keychain in an unencrypted form use `--pgp-password none`, use
+`--pgp-password random` for Scala CLI to randomly generate a password for you.
```
Gedochao: NIT
```suggestion
To store the private keychain in an unencrypted form use `--pgp-password none` 
To randomly generate a pasword, use `--pgp-password random` instead.
```

website/docs/commands/publishing/publish.md
```diff
@@ -133,23 +133,27 @@ it can also be overriden with `--signer` with one of the values:
 
 #### Bouncy Castle
 
-Bouncy Castle library is the recommended way of singing artifacts with Scala CLI. 
+Bouncy Castle library is the quickest way of singing artifacts with Scala CLI.
```
Gedochao:
```suggestion
Bouncy Castle library is the quickest way of signing artifacts with Scala CLI.
```

website/docs/commands/publishing/publish.md
```diff
@@ -133,23 +133,27 @@ it can also be overriden with `--signer` with one of the values:
...
 ```bash ignore
-scala-cli --power config --create-pgp-key
+scala-cli --power config --create-pgp-key --pgp-password MY_CHOSEN_PASSWORD
 ```
 
-This generates a public key and password protected private key, all values are kept in config
-and will be used by default unless specified otherwise:
+It's not mandatory, although recomended, to use a password to encrypt your keychains.
+To store the private keychain in an unencrypted form use `--pgp-password none`, use
+`--pgp-password random` for Scala CLI to randomly generate a password for you.
```
Gedochao:
```suggestion
To store the private keychain in an unencrypted form use `--pgp-password none` 
To randomly generate a pasword, use `--pgp-password random` instead.
```

website/docs/commands/publishing/publish.md
```diff
@@ -133,23 +133,27 @@ it can also be overriden with `--signer` with one of the values:
...
+
+The generated values are kept in config and will be used by default unless specified otherwise:
```
Gedochao:
```suggestion
The generated values are kept in the `config` and will be used by default unless specified otherwise:
```

website/docs/commands/publishing/publish.md
```diff
@@ -172,6 +176,9 @@ Using GPG to sign artifacts requires the `gpg` binary to be installed on your sy
 A benefit of using `gpg` to sign artifacts over Bouncy Castle is: you can use keys from
 your GPG key ring, or from external devices that GPG may support.
 
+To get started consult the [documentation on the library's website](https://gnupg.org/documentation/guides.html) and be sure to read about
```
Gedochao:
```suggestion
To get started, consult the [documentation on the library's website](https://gnupg.org/documentation/guides.html) and be sure to read about
```

modules/cli/src/main/scala/scala/cli/commands/publish/checks/PgpSecretKeyCheck.scala
```diff
@@ -392,34 +390,48 @@ final case class PgpSecretKeyCheck(
...
+        // Prevent potential leakage of a secret value
+        val passwordDirectives = getDirectiveValue(setupKeys.secretKeyPasswordOpt)
+          .flatMap {
+            case rawValueRegex(rawValue) =>
+              logger.diagnostic(
+                s"""The value of PGP password WON'T be written to a potentially public file!
+                   |Provide it as an option to the publish subcommand with:
+                   | --secret-key-password value:$rawValue
+                   |""".stripMargin
+              )
```
Gedochao:
```suggestion
              logger.diagnostic(
                s"""The value of the PGP password ${Console.BOLD}will not ${Console.RESET} be written to a potentially public file!
                   |Provide it as an option to the publish subcommand with:
                   | --secret-key-password value:$rawValue
                   |""".stripMargin
              )
```

modules/cli/src/main/scala/scala/cli/commands/publish/checks/PgpSecretKeyCheck.scala
```diff
@@ -392,34 +390,48 @@ final case class PgpSecretKeyCheck(
...
+
+        // Prevent potential leakage of a secret value
+        val secretKeyDirValue = getDirectiveValue(setupKeys.secretKeyOpt).flatMap {
+          case rawValueRegex(rawValue) =>
+            logger.diagnostic(
+              s"""The value of PGP secret key WON'T be written to a potentially public file!
+                 |Provide it as an option to the publish subcommand with:
+                 | --secret-key value:$rawValue
+                 |""".stripMargin
+            )
```
Gedochao:
```suggestion
            logger.diagnostic(
              s"""The value of the PGP secret key ${Console.BOLD}will not${Console.RESET} be written to a potentially public file!
                 |Provide it as an option to the publish subcommand with:
                 | --secret-key value:$rawValue
                 |""".stripMargin
            )
```

Gedochao: btw. this message could be abstracted and added to `WarningMessages` I think

modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
MaciejG604: I had in mind adding some assurance for the user that the password that they provided has been interpreted as intended and has not been misspelled.

True, it should be `println` here.

Gedochao: tbh I wouldn't print any secrets in any output, unless they're generated... but I may be paranoid here 🤔

lwronski: > I think this should be printed only if options.pgpPassword.contains("random")...
 
I agreed, we should print password only if it is generated by us.

lwronski (approved):

Issue #1729, opened 2022-12-23
Defining project.scala dependencies in test scope?


francisdb:
**Version(s)**
Scala CLI version: 0.1.19
Scala version (default): 3.2.1

I see this warning on every compile
```
[warn] ./files.test.scala:1:1: Using directives detected in multiple files. It is recommended to keep them centralized in the /xyz/project.scala file.
```

But was wondering how I define dependencies in test scope inside project.scala or is there no such thing as a test scope?
I have not found anything related to that in the docs?

Gedochao:
Hey, sorry for the late answer. There indeed should be a way to define test scope directives in the configuration file, but there currently isn't. It's definitely a bug, and it hasn't been covered in the docs properly, as you say.

All sources, including the `project.scala` configuration file, should respect being in the test scope, as described in our [test command doc](https://scala-cli.virtuslab.org/docs/commands/test#test-sources). The warning doesn't seem to take it into account.

I know it may be annoying, but it seems the best I can advise is to ignore the warning until we fix this.
As things are, the warning will only go away if you keep all your directives in a single source file, which probably isn't ideal, as  you probably wouldn't want your test dependencies in the main scope.

prolativ:
There is a similar problem when one needs a different set of directives when cross-compiling to different targets: JVM/native/js and different scala version

tgodzik:
We are thinking of having a short term solution to disable the warning, but long term we should have somethin like:
`//> using test.lib ...` which would mean we are adding the library only to the test scope. Similar with native/js/jvm

Gedochao:
> We are thinking of having a short term solution to disable the warning

That can be tracked under a separate issue: #1747

Gedochao:
> We are thinking of having a short term solution to disable the warning, but long term we should have somethin like: `//> using test.lib ...` which would mean we are adding the library only to the test scope. Similar with native/js/jvm

It's been decided we will proceed with this syntax, as in the prefix `test.*` will indicate a directive should only affect the `test` scope.
no separate project file will be used. (like `project.test.scala`)
example:
```scala
//> using test.dep "(...)"
```

prolativ:
This might solve the problem for test/main and potentially for jvm/js/native. But what if one e.g. needs different library dependencies for scala 2.12, 2.13 and 3? Would it be
```scala
//> using scala.2_12.dep "..."
//> using scala.2_13.dep "..."
//> using scala.3.dep "..."
```
?
[...]

PR #1988, opened 2023-03-31
Don't print the spread directives warning if there's only a single file per scope

Gedochao:
Context: #1729

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -170,31 +170,6 @@ object CrossSources {
     val preprocessedSources =
       (preprocessedInputFromArgs ++ preprocessedSourcesFromDirectives).distinct
 
-    val preprocessedWithUsingDirs = preprocessedSources.filter(_.directivesPositions.isDefined)
-    if preprocessedWithUsingDirs.length > 1 &&
-      !suppressWarningOptions.suppressDirectivesInMultipleFilesWarning.getOrElse(false)
-    then {
-      val projectFilePath = inputs.elements.projectSettingsFiles.headOption match
-        case Some(s) => s.path
-        case _       => inputs.workspace / Constants.projectFileName
-      preprocessedWithUsingDirs
-        .filter(_.scopePath != ScopePath.fromPath(projectFilePath))
-        .foreach { source =>
-          source.directivesPositions match
-            case Some(positions) =>
-              val pos = Seq(
-                positions.codeDirectives,
-                positions.specialCommentDirectives,
-                positions.plainCommentDirectives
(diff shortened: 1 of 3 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -258,6 +237,44 @@ object CrossSources {
...
+          val path: os.Path = withBuildRequirements.value match
+            case im: Sources.InMemory =>
+              im.originalPath match
+                case Right((_, p: os.Path)) => p
+                case _                      => inputs.workspace / im.generatedRelPath
+            case (p: os.Path, _) => p
+          directivesPositions.map((path, scope, _))
+        }
+        .groupBy((_, scope, _) => scope)
+        .map { (scope, entries) =>
```
MaciejG604:
```suggestion
        .mapValues { _.map((path, _, directivesPostions) => path -> directivesPositions}
```

MaciejG604 (approved):

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -258,6 +237,44 @@ object CrossSources {
...
+          scope -> entries.map((path, _, directivesPositions) => path -> directivesPositions)
+        }
+    val anyScopeHasMultipleSourcesWithDirectives =
+      Scope.all.exists(allPathsWithDirectivesByScope.get(_).map(_.length).getOrElse(0) > 1)
+    val shouldSuppressWarning =
+      suppressWarningOptions.suppressDirectivesInMultipleFilesWarning.getOrElse(false)
+    if anyScopeHasMultipleSourcesWithDirectives && !shouldSuppressWarning then {
```
lwronski:
```suggestion
    if !shouldSuppressWarning && anyScopeHasMultipleSourcesWithDirectives then {
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -258,6 +237,44 @@ object CrossSources {
...
+            case (p: os.Path, _) => p
+          directivesPositions.map((path, scope, _))
+        }
+        .groupBy((_, scope, _) => scope)
+        .map { (scope, entries) =>
+          scope -> entries.map((path, _, directivesPositions) => path -> directivesPositions)
+        }
+    val anyScopeHasMultipleSourcesWithDirectives =
+      Scope.all.exists(allPathsWithDirectivesByScope.get(_).map(_.length).getOrElse(0) > 1)
+    val shouldSuppressWarning =
```
lwronski, line 258: It should be `lazy val` to avoid initialising this variable when spread directive warning is disabled.

PR #1990, opened 2023-04-03
Expose the name of script via `scriptPath` variable in Scala Script

### LINKED ISSUE #1937: Provide way to get "the name of this script" (like $0 in Bash)

The basic problem is wanting to print a "usage message" whenever someone misuses a scala-cli script. I'm looking for the equivalent of `$0` in the following bash script:
```bash
$ cat t/script.sh 
#!/usr/bin/env bash

if [[ "$#" -ne 1 ]]; then
  echo "Requires exactly 1 argument!"
  echo "Usage: $0 <file>"
fi

$ ./t/script.sh 
Requires exactly 1 argument!
Usage: ./t/script.sh <file>
```

I don't really care what the API is, just something that gets the current script name. I think there is subtly in making sure that the path given is relative to the current working directory of the shell invoking the script, and also making it easy to just get the script name (without directory) if desired as well.

I could elide the name of the script in "Usage" messages but I think it would look more professional and be more useful to have it.

**Additional context**

Discussed a little with @lwronski on Scala Discord

lwronski:
Fixes #1937.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -288,7 +288,7 @@ object CrossSources {
       else if (sourcePath == os.sub / Constants.projectFileName)
         Right(Seq(ProjectScalaFile(dir, subPath)))
       else if (sourcePath.ext == "scala") Right(Seq(SourceScalaFile(dir, subPath)))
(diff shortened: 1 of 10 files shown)


Gedochao (approved):
LGTM

PR #1998, opened 2023-04-04
add newline to topWrapper

### LINKED ISSUE #1996: No metals completions in `foo.sc` files for first line
**Version(s)**
1.0.0-RC1

**Describe the bug**
if I run metals in a file `foo.sc` after `setup-ide`, then there is no completions on the first line, but there are on the second line.

**To Reproduce**
- Open VS Code (with Scala (Metals) extension 0.11.11)
- create new directory, open it as the workspace
- create `foo.sc` file
- in terminal `scala setup-ide foo.sc`
- try to type `args.mkString(" ")` on the first line, there will be no automatic completions
- then on second line try to type `args.mkString(" ")` and immediately there are automatic completions

**Expected behaviour**
the same behaviour for the first line

bishabosha:
fixes #1996

In my testing locally, completions are working for `.sc` files on the first line (and all others)

didn't know if there were tests for this, so I tried building it locally and starting metals in vscode with the local build of scala-cli

### DIFF
--- a/modules/core/src/main/scala/scala/build/internals/CodeWrapper.scala
+++ b/modules/core/src/main/scala/scala/build/internals/CodeWrapper.scala
@@ -17,7 +17,7 @@ abstract class CodeWrapper {
     val (topWrapper, bottomWrapper, userCodeNestingLevel) =
       apply(code, pkgName, indexedWrapperName, extraCode0)
     val (topWrapper0, bottomWrapper0) =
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks! I think this should work from the Metals perspective unless anyone knows if this could cause some issues?

Gedochao (approved):
I was worried it might break debugging scripts, but I just tested and it seems to recognise the correct line... so LGTM
cc @lwronski

lwronski (approved):

PR #2005, opened 2023-04-05
Publish to maven central a bootstrapped jar

romanowski:
Such a jar will open Scala CLI to people in restricted environments.


lwronski (approved):
Let's see how it works with the nightly release. If everything goes ok, we have to update the documentation.

PR #2008, opened 2023-04-05
Make 'export --json' print to stdout by default

### LINKED ISSUE #1887: Output `export --json` to `stdout`
The `export` sub-command currently exports `--json` outputs to a `JSON` file, consistent with the `--mill` and `--sbt` exports.
However, although that behaviour should still be available, it probably shouldn't be the default and instead be hidden behind an option, or perhaps only be used when an `--output` path is passed explicitly.
As `--json` outputs are likely to be piped, it'd be great to print the json to `stdout` as the default instead.

`--json` should print the output on `stdout` by default.
the old `--json` behaviour saving the JSON to a file should either be available behind an option or only be used whan an `--output` path is passed explicitly.

MaciejG604:
Closes #1887

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
@@ -104,9 +104,12 @@ object Export extends ScalaCommand[ExportOptions] {
   override def runCommand(options: ExportOptions, args: RemainingArgs, logger: Logger): Unit = {
     val initialBuildOptions = buildOptionsOrExit(options)
 
-    val output = options.output.getOrElse("dest")
-    val dest   = os.Path(output, os.pwd)
-    if (os.exists(dest)) {
+    val output                   = options.output.getOrElse("dest")
+    val dest                     = os.Path(output, os.pwd)
+    val shouldExportToJson       = options.json.getOrElse(false)
+    val shouldExportJsonToStdout = shouldExportToJson && options.output.isEmpty
+
+    if (!shouldExportJsonToStdout && os.exists(dest)) {
       logger.error(
         s"""Error: $dest already exists.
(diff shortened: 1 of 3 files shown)


modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -104,17 +104,18 @@ object Export extends ScalaCommand[ExportOptions] {
...
     val initialBuildOptions = buildOptionsOrExit(options)
 
-    val output = options.output.getOrElse("dest")
-    val dest   = os.Path(output, os.pwd)
-    if (os.exists(dest)) {
+    val output             = options.output.getOrElse("dest")
+    val dest               = os.Path(output, os.pwd)
+    val shouldExportToJson = options.json.getOrElse(false)
+
+    if (!shouldExportToJson && os.exists(dest)) {
```
lwronski: I think we should also enable users to save JSON to a file. Something like this should work:

```
scala-cli export . --json -o dest/scala-cli.json
scala-cli export . --json -o dest
```

modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -155,44 +154,52 @@ object Export extends ScalaCommand[ExportOptions] {
...
+
+      if (
+        optionsMain0.scalaOptions.scalaVersion.isEmpty && optionsTest0.scalaOptions.scalaVersion.nonEmpty
+      ) {
+        logger.error(
+          s"""Detected that the Scala version is only set in test scope.
+             |Please set the Scala version explicitly in the main and test scopes with using directives or pass -S, --scala-version as parameter""".stripMargin
+        )
+        sys.exit(1)
+      }
```
lwronski: These conditions should also be checked when users export JSON to stdout.

lwronski (approved):

PR #2014, opened 2023-04-06
Improve error messages for malformed `config` values

### LINKED ISSUE #2000: Confusing error message when passing malformed boolean value in scala-cli config command
Right now, when passing a malformed boolean to a `scala-cli config` command, we get an error like
```text
$ scala-cli config power true.
[error]  Config DB error: Malformed values true. for power, expected value
```

The `expected value` message is confusing. Something like `expected a boolean value "true" or "false"` would work better for example.

Gedochao:
Fixes #2000 
Fixed some of the error messages for the `config` subcommand + did a wider refactor.

Before:
```bash
scala-cli config power true.
# [error]  Config DB error: Malformed values true. for power, expected value
scala-cli config power true true false
# [error]  Config DB error: Malformed values true, true, false for power, expected value
```

After the changes:
```bash
scala-cli config power true.          
# [error]  Config DB error: Malformed value 'true.' for the 'power' entry, expected a single boolean value ('true' or 'false').
scala-cli config power true true false
# [error]  Config DB error: Malformed values 'true', 'true', 'false' for the 'power' entry, expected a single boolean value ('true' or 'false').
```

### DIFF
--- a/modules/config/src/main/scala/scala/cli/config/Key.scala
+++ b/modules/config/src/main/scala/scala/cli/config/Key.scala
@@ -73,8 +73,13 @@ object Key {
     messageOrExpectedShape: Either[String, String],
     cause: Option[Throwable] = None
   ) extends EntryError(
-        s"Malformed values ${input.mkString(", ")} for ${entry.fullName}, " +
-          messageOrExpectedShape.fold(shape => s"expected $shape", identity),
+        {
+          val valueWord    = if (input.length > 1) "values" else "value"
+          val valuesString = input.map(s => s"'$s'").mkString(", ")
+          val errorMessage = messageOrExpectedShape
+            .fold(shape => s", expected $shape", errorMessage => s". $errorMessage")
+          s"Malformed $valueWord $valuesString for the '${entry.fullName}' entry$errorMessage"
+        },
         cause
       )
 
@@ -109,7 +114,7 @@ object Key {
     def fromString(values: Seq[String]): Either[MalformedValue, String] =
       values match {
         case Seq(value) => Right(value)
-        case _          => Left(new MalformedValue(this, values, Left("value")))
+        case _          => Left(new MalformedValue(this, values, Left("a single string value.")))
       }
   }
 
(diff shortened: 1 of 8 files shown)


MaciejG604 (approved):

modules/config/src/main/scala/scala/cli/config/Key.scala
```diff
@@ -73,43 +74,51 @@ object Key {
...
         s"Malformed entry ${entry.fullName}, " +
           messages.mkString(", ")
       )
 
-  private val stringCodec: JsonValueCodec[String]   = JsonCodecMaker.make
-  private val booleanCodec: JsonValueCodec[Boolean] = JsonCodecMaker.make
+  abstract class KeyWithJsonCodec[T] extends Key[T] {
+    def jsonCodec: JsonValueCodec[T]
+
+    override def parse(json: Array[Byte]): Either[Key.EntryError, T] =
```
MaciejG604: NIT `override` keyword is optional here. To keep it consistent with the rest of the file I'd say use it only when it's needed.

modules/config/src/main/scala/scala/cli/config/Key.scala
```diff
@@ -119,21 +128,19 @@ object Key {
...
-    def parse(json: Array[Byte]): Either[EntryError, Boolean] =
-      try Right(readFromArray(json)(booleanCodec))
-      catch {
-        case e: JsonReaderException =>
-          Left(new JsonReaderError(e))
-      }
-    def write(value: Boolean): Array[Byte] =
-      writeToArray(value)(booleanCodec)
+  ) extends KeyWithJsonCodec[Boolean] {
+    val jsonCodec: JsonValueCodec[Boolean] = JsonCodecMaker.make
```
MaciejG604: As above. This val doesn't use extra `override` although it could.

modules/config/src/main/scala/scala/cli/config/Key.scala
```diff
@@ -83,32 +84,35 @@ object Key {
...
     messages: ::[String]
   ) extends EntryError(
         s"Malformed entry ${entry.fullName}, " +
           messages.mkString(", ")
       )
 
-  private val stringCodec: JsonValueCodec[String]   = JsonCodecMaker.make
-  private val booleanCodec: JsonValueCodec[Boolean] = JsonCodecMaker.make
+  abstract class KeyWithJsonCodec[T] extends Key[T] {
+    def jsonCodec: JsonValueCodec[T]
```
lwronski: I would prefer to avoid returning `JsonValueCodec[T]` from a function, since it can generate the same codec multiple times for only one type.

It might be better to initialize this codec in the companion object and define it as `implicit lazy val` to ensure that it is only initialized once.

Gedochao: should be fixed

lwronski (approved):
LGTM

PR #2025, opened 2023-04-08
Add first-class support for Typelevel and other toolkits

### LINKED ISSUE #2018: First-class support for "alternative" toolkits

Inspired by the [Scala Toolkit](https://github.com/scala/toolkit), we have created a [Typelevel Toolkit](https://typelevel.org/toolkit/). cc @toniogela

```
//> using toolkit "typelevel:latest"
```

Or something similarly simple e.g.:
```
//> using kit "typelevel:latest"
```

Currently we are doing:

```
//> using lib "org.typelevel::toolkit::0.0.3"
```

armanbilge:
Closes https://github.com/VirtusLab/scala-cli/issues/2018

### DIFF
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
@@ -39,9 +39,17 @@ final case class Toolkit(
 }
 
 object Toolkit {
-  def resolveDependency(toolkitVersion: Positioned[String]) = toolkitVersion.map(version =>
-    val v = if version == "latest" then "latest.release" else version
-    dep"${Constants.toolkitOrganization}::${Constants.toolkitName}::$v,toolkit"
+  def resolveDependency(toolkitCoord: Positioned[String]) = toolkitCoord.map(coord =>
+    val tokens  = coord.split(':')
+    val version = tokens.last
+    val v       = if version == "latest" then "latest.release" else version
+    val flavor  = tokens.dropRight(1).headOption
+    val org = flavor match {
+      case Some("typelevel") => Constants.typelevelOrganization
+      case Some(org)         => org
+      case None              => Constants.toolkitOrganization
+    }
+    dep"$org::${Constants.toolkitName}::$v,toolkit"
   )
   val handler: DirectiveHandler[Toolkit] = DirectiveHandler.derive
 }
--- a/build.sc
+++ b/build.sc
@@ -398,6 +398,7 @@ trait Core extends ScalaCliSbtModule with ScalaCliPublishModule with HasTests
          |
          |  def toolkitOrganization = "${Deps.toolkit.dep.module.organization.value}"
          |  def toolkitName = "${Deps.toolkit.dep.module.name.value}"
+         |  def typelevelOrganization = "org.typelevel"
          |
          |  def defaultScalaVersion = "${Scala.defaultUser}"
          |  def defaultScala212Version = "${Scala.scala212}"
--- a/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
@@ -107,4 +107,23 @@ class DirectiveTests extends munit.FunSuite {
     }
   }
 
+  test("resolve typelevel toolkit dependency") {
+    val testInputs = TestInputs(
+      os.rel / "simple.sc" ->
+        """//> using toolkit "typelevel:latest"
+          |""".stripMargin
+    )
+    testInputs.withBuild(baseOptions, buildThreads, bloopConfigOpt) {
+      (_, _, maybeBuild) =>
+        val build = maybeBuild.orThrow
+        val dep   = build.options.classPathOptions.extraDependencies.toSeq.headOption
+        assert(dep.nonEmpty)
+
+        val toolkitDep = dep.get.value
+        expect(toolkitDep.organization == "org.typelevel")
+        expect(toolkitDep.name == "toolkit")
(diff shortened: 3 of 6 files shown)


modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
```diff
@@ -39,9 +39,17 @@ final case class Toolkit(
...
+  def resolveDependency(toolkitCoord: Positioned[String]) = toolkitCoord.map(coord =>
+    val tokens  = coord.split(':')
+    val version = tokens.last
+    val v       = if version == "latest" then "latest.release" else version
+    val flavor  = tokens.dropRight(1).headOption
+    val org = flavor match {
+      case Some("typelevel") => Constants.typelevelOrganization
+      case Some(org)         => org
+      case None              => Constants.toolkitOrganization
+    }
```
armanbilge, line 51: Supports
```
//> using toolkit "typelevel:latest"
```

But there's a question what to do about:
```
//> using toolkit "foo:latest"
```
We could reject `foo` as an unknown toolkit. But instead what I did here was treat it as the group id. So then anyone can publish a toolkit artifact under their group id and it would Just Work.:tm: Thoughts?

TonioGela: An alternative can be having a system similar to how sbt treats the testing frameworks: the most notorious are manually added to sbt while there's the support to use them, adding a minimal configuration (while waiting for sbt to upstream the particular testing framework)

tgodzik: I think that should fine and allow us to not modify ScalaCLI every time someone wants to create or test out a new toolkit.

tgodzik (approved):
Looks good! I will wait a bit to gather some more opinions since toolkit is not officially out yet. You're too fast! :D

modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
armanbilge: Another interesting alternative is to somehow use GH org (e.g. `typelevel`, `scala`) instead of group id (`org.typelevel`, `org.scala-lang`). But this would be harder to implement :)

TonioGela: Also, not everybody might use GH to distribute their toolkit, right?

bishabosha:
I wonder if leaving this feature "open" might lead to a rush to make many "toolkits" just to have an easier import for a project, when perhaps the more general solution would be a flexible way to declare dependencies without versions (and possibly organisations?) and let Scala CLI figure out the conflicts.

I know dependencies can be declared with `latest.release` or `latest.stable` as the version, but I don't know if that handles conflicts at all

tgodzik:
> I wonder if leaving this feature "open" might lead to a rush to make many "toolkits" just to have an easier import for a project, when perhaps the more general solution would be a flexible way to declare dependencies without versions (and possibly organisations?) and let Scala CLI figure out the conflicts.
> 

I don't think there would be a lot of people willing to invest here, but I understand the concern. Anyone else feels this might become problematic?

> I know dependencies can be declared with `latest.release` or `latest.stable` as the version, but I don't know if that handles conflicts at all

I think it should handle the conflicts just as normal.

TonioGela:
> I don't think there would be a lot of people willing to invest here, but I understand the concern. Anyone else feels this might become problematic?

Also, after the development of the toolkit itself, the investment to make it automatically supported consists of just adding a [new case in a pattern match](https://github.com/VirtusLab/scala-cli/pull/2025/files#diff-d79b4ac6aedd989c84f67e7f04c9eccf816498fde9eee286a0cb7683952f0bd4R48). I don't think that's a lot to pay.

lefou:
I wonder whether support for "toolkit" is a special thing in general. Couldn't it be a simple  pom dependency, which transitively depends on all the artifacts that make up a "toolkit"? (If this was discussed before, can you please provide a link to it?)

bishabosha:
> I wonder whether support for "toolkit" is a special thing in general. Couldn't it be a simple pom dependency, which transitively depends on all the artifacts that make up a "toolkit"? (If this was discussed before, can you please provide a link to it?)

Yeah the current implementation is syntax sugar for adding a dependency to `<organisation>::toolkit::<version>`

TonioGela:
> Yeah the current implementation is syntax sugar for adding a dependency to `<organisation>::toolkit::<version>`

Plus, the implementation of the TL toolkit is exactly what you described: a metalib that depends on the libraries you want to include in it. See https://github.com/typelevel/toolkit/blob/main/build.sbt#L16.

modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
armanbilge: I was thinking the toolkit itself would still be on Maven, but somehow we use the GH orgs as the namespace to map to Maven coordinates.

e.g. `org.scala-lang` can just be `scala`, `org.typelevel` can just be `typelevel` (and we won't even need to hardcode it), etc.

Compare:

```scala
//> using toolkit "typelevel:latest"
```

With:

```scala
//> using toolkit "org.typelevel:latest"
```

And finally:
```scala
//> using dep "org.typelevel::toolkit::latest.release"
```

SethTisue:
> Anyone else feels this might become problematic?

in short, no. and even if somebody did overdo it, 🤷, hard to see any real harm in it

tgodzik (approved):
I don't see any objections to merging this, thanks @armanbilge !

PR #2026, opened 2023-04-08
only hash when boolean is true

bishabosha:
This means that vanilla `scala compile foo` with no directives and one source file will compile to `foo/.scala-build/project/main/classes`

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/HasHashData.scala
+++ b/modules/options/src/main/scala/scala/build/options/HasHashData.scala
@@ -49,6 +49,10 @@ object HasHashData:
   given option[T](using hasher: HashedType[T]): HasHashData[Option[T]] =
     (name, opt, update) => opt.foreach(t => update(s"$name=${hasher.hashedValue(t)}"))
 
+  given boolean: HasHashData[Boolean] =
+    (name, bool, update) =>
+      if bool then update(s"$name=true")
+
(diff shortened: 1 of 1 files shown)


bishabosha:
I don't really expect this to be much useful, or give much performance win, but I noticed when doing some debugging that the only reason a "vanilla" project was getting a hashed name was because of a single invocation of the hashing callback with `.scalaJsOptions.emitSourceMaps=false`. So, seeing as `None` does not hash, why should `false`?

lwronski:
Thanks for your contribution, @bishabosha. I have some doubts because there might be cases where `false` is not the default value for some option.

For example, if we have some option that default value is `true` and a user sets this option to `false` -  `.some.option = false`, it doesn't change the hash of the project after you changes. Am I Right?

bishabosha:
If the default was true, then if the user changes it to false that should change the hash when compared to the setting being true, because that setting would no longer be mixed with the hash.

MaciejG604:
I think this makes it possible to introduce hard-to-find bugs.
E.g., if there exists (or will ever exist) an option `defaultTrue: Option[Bool] = None`, that is used in code like this: `defaultTrue.getOrElse(true)`. It would not be hashed in both the case of being `None` and `Some(false)`, while its meaning for the project would differ.

bishabosha:
sure, ok

PR #2028, opened 2023-04-11
Add --as-jar option

### LINKED ISSUE #2016: Add --print-class-path-as-jars option to the compile command
The `--print-class-path` is handy when one wants to pass the class path of a Scala CLI project to another tool expecting a class path The directory containing the project byte code is added to this class path by Scala CLI. Yet, some tools require class path inputs to consist only in JARs - they don't accept directories. (The use case I'm having right now is passing this class path to the `spark.jars` config entry of Spark.)

`scala-cli compile` should have a `--print-class-path-as-jars` option, generating a library JAR like how `scala-cli package --library` would do, and putting it in the output class path, instead of the project byte code directory.

As a workaround, one can grab the output of `scala-cli package --library`, manually add it to the output of `scala-cli compile --print-class-path`, and remove the project byte code directory from it if it's not being ignored downstream. A `--print-class-path-as-jars` option would much more handy.

alexarchambault:
Fixes https://github.com/VirtusLab/scala-cli/issues/2016

This adds an `--as-jar` option to the `compile` / `run` / `test` / `repl` commands, that puts the project byte code as a JAR in the class path, rather than as a directory, which is useful in contexts where JARs are required.

### DIFF
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -117,6 +117,8 @@ final case class Inputs(
     workspace / Constants.workspaceDirName / projectName / "native"
   def nativeImageWorkDir: os.Path =
     workspace / Constants.workspaceDirName / projectName / "native-image"
+  def libraryJarWorkDir: os.Path =
+    workspace / Constants.workspaceDirName / projectName / "jar"
   def docJarWorkDir: os.Path =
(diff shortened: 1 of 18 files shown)


lwronski (commented):
Can you update the documentation?  I think it could be added  near the `--print-class-path` [section](https://scala-cli.virtuslab.org/docs/commands/compile/#printing-a-class-path). 
Other than that, LGTM.

lwronski (approved):
LGTM

PR #2033, opened 2023-04-12
Add class script wrapper

### LINKED ISSUE #532: Object encapsulation sometimes required in order for the script to execute without blocking/Timeout
As soon as a script have to deal with an execution context, or asynchronous processing  such as scala Future, ZIO, cats, ...) it is needed to encapsulate the code within an object in order for the script to execute properly, if not it blocks/timeouts without giving any results.

Ammonite has the same issue, see : https://github.com/com-lihaoyi/Ammonite/issues/802 and I was able in some cases to mitigate the issue by adding an ammonite startup option : `amm --class-based ...`.

I think this is an important issue, as it makes our scripts more verbose by systematically having to encapsulate part of the code within objects, and requiring us then to call a function within that object to start the execution.

See next examples, but if required I can provide more examples as the issue is quite easy to reproduce as soon as you use libraries such as AKKA, ZIO, CATS, DOODLE, ...

I was using scala-cli 0.0.9 in this test and scala3, but it is reproductible with any other scala releases.

**KO  example** : (https://gist.github.com/dacr/47ca242e6d672e0154c5816a0330eea7)
```scala3
// ---------------------
// using scala 3.1.0
// ---------------------

import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
import scala.concurrent.duration.*

implicit val ec: ExecutionContextExecutor = ExecutionContext.global

val future =
  for
    message <- Future("Hello world")
    _       <- Future(println(message))
  yield ()

Await.ready(future, 5.second)
```

**OK example** : (https://gist.github.com/dacr/13ff195adf279798ed5dcc16da2a33ba)
```scala3
// ---------------------
// using scala 3.1.0
// ---------------------

import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
import scala.concurrent.duration.*

object Encapsulated:
  implicit val ec: ExecutionContextExecutor = ExecutionContext.global

  def run()=
    val future =
    for
      message <- Future("Hello world")
      _ <- Future(println(message))
    yield ()
    Await.ready(future, 5.second)

Encapsulated.run()
```

**Executions results** : 
```bash
$ scala-cli https://gist.github.com/dacr/13ff195adf279798ed5dcc16da2a33ba
Hello world
$ scala-cli https://gist.github.com/dacr/47ca242e6d672e0154c5816a0330eea7
...
Caused by: java.util.concurrent.TimeoutException: Future timed out after [5 seconds]
	at scala.concurrent.impl.Promise$DefaultPromise.tryAwait0(Promise.scala:212)
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:104)
	at scala.concurrent.Await$.$anonfun$ready$1(package.scala:174)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:62)
	at scala.concurrent.Await$.ready(package.scala:124)
	at example$minusko$.<clinit>(47ca242e6d672e0154c5816a0330eea7-master/example-ko.sc:26)
```

### LINKED ISSUE #1933: zio-streams example does not terminate in scala-cli, but does when run in IntelliJ
**Version(s)**

```
$ scala-cli --version
Scala CLI version: 0.2.1
```

This is running on OS X.

**Describe the bug**
A simple  zio-streams example, included below, hangs when used in scala-cli, though it terminates normally when run in IntelliJ as a ZIO App:

```scala
object StreamExample extends ZIOAppDefault {
  override val run = {
    val points = Chunk(5, 10, 15, 20, 100, 200, 300)
​
    val natStream = ZStream.iterate(0) { i =>
      println(i)
      i + 1
    }
​
    ZIO.foreach(points)(point => natStream.take(point).runCollect.debug)
  }
}
```

**To Reproduce**

```scala
using lib "dev.zio::zio-streams:2.0.10"

import zio.*
import zio.stream.*

val points = Chunk(5, 10, 15, 20, 100, 200, 300)

val natStream = ZStream.iterate(0)( /* _ + 1*/ foo => { println(foo); foo + 1 })

val program =
  ZIO.foreach(points)(p => natStream.take(p).runCollect.map(_.sum))

Unsafe.unsafe { implicit unsafe =>
  Runtime.default.unsafe.run(program)
}
```

Then try running the script - for me it hangs at the end of executing the first stream (`.take(5)`):

```
$ scala-cli ZStreamIncrementalTake.sc 
Compiling project (Scala 3.2.2, JVM)
Compiled project (Scala 3.2.2, JVM)
0
1
2
3
4
```

**Expected behaviour**
The script should terminate normally.

MaciejG604:
This should avoid deadlocks when threads are run from the static initalizer of wrapper object.

Closes #532 and #1933 

The key to this issue is what's visible in the thread dump from [this comment](https://github.com/VirtusLab/scala-cli/issues/532#issuecomment-1236074174)
```
ZScheduler-Worker-1" #15 daemon prio=5 os_prio=0 cpu=109,44ms elapsed=32,71s tid=0x00007f62544c7290 nid=0x22344 in Object.wait()  [0x00007f6214706000]
   java.lang.Thread.State: RUNNABLE
	at issue$minus7301$minusexample$minusko$Photos$$anon$1$$Lambda$481/0x0000000800e2c798.apply(Unknown Source)
	- waiting on the Class initialization monitor for issue$minus7301$minusexample$minusko$
	at zio.ZIO.map$$anonfun$1(ZIO.scala:927)
	at zio.ZIO$$Lambda$87/0x0000000800d3f3e0.apply(Unknown Source)
```

The background thread worker is waiting for object class to be initialized, but the object class initialization process is waiting for the background thread to finish, this introduces a deadlock.

As stated by David Crosson in #532 encapsulating the `wait/join/Await.ready` method into a additonal object fixed this in some cases by making the background thread not dependent on the initialization of the outer wrapper object introduced by Scala CLI.

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
@@ -1,6 +1,11 @@
 package scala.build.internal
 
-case object CustomCodeWrapper extends CodeWrapper {
+/** Script code wrapper that solves problem of deadlocks when using threads. The code is placed in a
+  * class instance constructor, the created object is kept in 'mainObjectCode'.script to support
+  * running interconnected scripts using Scala CLI <br> <br> Incompatible with Scala 2 - it uses
+  * Scala 3 feature 'export'<br> Incompatible with native JS members - the wrapper is a class
+  */
+case object ClassCodeWrapper extends CodeWrapper {
   def mainClassObject(className: Name): Name =
     Name(className.raw ++ "_sc")
 
@@ -12,13 +17,8 @@ case object CustomCodeWrapper extends CodeWrapper {
     extraCode: String,
     scriptPath: String
   ) = {
-    val name               = mainClassObject(indexedWrapperName).backticked
-    val aliasedWrapperName = name + "$$alias"
-    val funHashCodeMethod =
-      if (name == "main_sc")
-        s"$aliasedWrapperName.alias.hashCode()" // https://github.com/VirtusLab/scala-cli/issues/314
-      else s"${indexedWrapperName.backticked}.hashCode()"
-    // We need to call hashCode (or any other method so compiler does not report a warning)
+    val name             = mainClassObject(indexedWrapperName).backticked
+    val wrapperClassName = Name(indexedWrapperName.raw ++ "$_").backticked
     val mainObjectCode =
       AmmUtil.normalizeNewlines(s"""|object $name {
                                     |  private var args$$opt0 = Option.empty[Array[String]]
@@ -29,22 +29,20 @@ case object CustomCodeWrapper extends CodeWrapper {
                                     |  def args$$: Array[String] = args$$opt.getOrElse {
                                     |    sys.error("No arguments passed to this script")
                                     |  }
+                                    |
+                                    |  lazy val script = new $wrapperClassName
+                                    |
                                     |  def main(args: Array[String]): Unit = {
                                     |    args$$set(args)
-                                    |    $funHashCodeMethod // hasCode to clear scalac warning about pure expression in statement position
+                                    |    script.hashCode() // hashCode to clear scalac warning about pure expression in statement position
                                     |  }
                                     |}
+                                    |
+                                    |export $name.script as ${indexedWrapperName.backticked}
                                     |""".stripMargin)
 
     val packageDirective =
       if (pkgName.isEmpty) "" else s"package ${AmmUtil.encodeScalaSourcePath(pkgName)}" + "\n"
-
-    val aliasObject =
-      if (name == "main_sc")
-        s"""object $aliasedWrapperName {
-           |  val alias = ${indexedWrapperName.backticked}
-           |}""".stripMargin
-      else ""
 
     // indentation is important in the generated code, so we don't want scalafmt to touch that
     // format: off
@@ -52,14 +50,14 @@ case object CustomCodeWrapper extends CodeWrapper {
 $packageDirective
 
 
-object ${indexedWrapperName.backticked} {
+final class $wrapperClassName {
 def args = $name.args$$
 def scriptPath = \"\"\"$scriptPath\"\"\"
 """)
     val bottom = AmmUtil.normalizeNewlines(s"""
 $extraCode
 }
-$aliasObject
+
 $mainObjectCode
 """)
     // format: on
--- a/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
@@ -6,9 +6,6 @@ package scala.build.internal
   * Scala 3 feature 'export'<br> Incompatible with native JS members - the wrapper is a class
   */
 case object ClassCodeWrapper extends CodeWrapper {
-  def mainClassObject(className: Name): Name =
-    Name(className.raw ++ "_sc")
-
   private val userCodeNestingLevel = 1
   def apply(
     code: String,
@@ -17,7 +14,7 @@ case object ClassCodeWrapper extends CodeWrapper {
     extraCode: String,
     scriptPath: String
   ) = {
-    val name             = mainClassObject(indexedWrapperName).backticked
+    val name             = CodeWrapper.mainClassObject(indexedWrapperName).backticked
     val wrapperClassName = Name(indexedWrapperName.raw ++ "$_").backticked
     val mainObjectCode =
       AmmUtil.normalizeNewlines(s"""|object $name {
--- a/modules/build/src/main/scala/scala/build/internal/util/WarningMessages.scala
+++ b/modules/build/src/main/scala/scala/build/internal/util/WarningMessages.scala
@@ -34,6 +34,15 @@ object WarningMessages {
        | $directiveName value:$rawValue
        |""".stripMargin
 
+  /** Using @main is impossible in new [[scala.build.internal.ClassCodeWrapper]] since none of the
+    * definitions can be accessed statically, so those errors are swapped with this text
+    * @param annotationIgnored
+    *   will annotation be ignored (or will compilation fail)
+    */
+  def mainAnnotationNotSupported(annotationIgnored: Boolean): String =
+    val consequencesString = if annotationIgnored then s", it will be ignored" else ""
+    s"Annotation @main in .sc scripts is not supported$consequencesString, use .scala format instead"
+
   private def powerFeatureUsedInSip(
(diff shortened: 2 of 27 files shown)


tgodzik:
Looks like some tests might need fixing up?

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
```diff
@@ -0,0 +1,62 @@
...
+                                    |
+                                    |  lazy val script = new $wrapperClassName
+                                    |
+                                    |  def main(args: Array[String]): Unit = {
+                                    |    args$$set(args)
+                                    |    script.hashCode() // hashCode to clear scalac warning about pure expression in statement position
+                                    |  }
+                                    |}
+                                    |
+                                    |export $name.script as ${indexedWrapperName.backticked}
```
tgodzik, line 38: Where do we check if it's Scala 3?

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -330,7 +330,11 @@ final case class SharedOptions(
         platform = platformOpt.map(o => Positioned(List(Position.CommandLine()), o))
       ),
       scriptOptions = bo.ScriptOptions(
-        codeWrapper = None
+        codeWrapper = platformOpt match {
+          case Some(_: Platform.JS.type)                    => Some(ObjectCodeWrapper)
+          case _ if scalaVersion.exists(_.startsWith("2.")) => Some(ObjectCodeWrapper)
```
MaciejG604: @tgodzik Here, when creating BuildOptions

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
```diff
@@ -0,0 +1,62 @@
+package scala.build.internal
+
+case object ClassCodeWrapper extends CodeWrapper {
```
Gedochao, line 8: we'll need scaladocs for `ClassCodeWrapper` and `ObjectCodeWrapper` explaining what and why, to avoid confusion in the future

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
tgodzik: thanks!

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
alexarchambault, line 38: The use of `export` for that is really neat!

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
alexarchambault: @MaciejG604 It seems this will only work if the Scala version is specified on the command-line, right? What if it's specified via a using directive in a script?

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val userCodeNestingLevel = 1
+  def apply(
+    code: String,
+    pkgName: Seq[Name],
+    indexedWrapperName: Name,
+    extraCode: String,
+    scriptPath: String
+  ) = {
+    val name             = mainClassObject(indexedWrapperName).backticked
+    val wrapperClassName = indexedWrapperName.backticked + "0"
```
alexarchambault: `$` would be a better suffix here IMO. Usually, the use of `$` in class names says that the class is compiler-generated, or at least, that it's not for end-users to use directly. Which is what we want here.

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
```diff
@@ -0,0 +1,62 @@
...
+    val packageDirective =
+      if (pkgName.isEmpty) "" else s"package ${AmmUtil.encodeScalaSourcePath(pkgName)}" + "\n"
+
+    // indentation is important in the generated code, so we don't want scalafmt to touch that
+    // format: off
+    val top = AmmUtil.normalizeNewlines(s"""
+$packageDirective
+
+
+class $wrapperClassName {
```
alexarchambault: You might want to mark the class as `final` here. I think it helps the JVM (maybe scalac too) optimizing things.

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
MaciejG604: Let's use `$_` since just `$` breaks `package --doc`

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
MaciejG604: Should work now for both using directives and cli options, wrapper inference is now postponed to the moment after preprocessing, when build options are extracted from both inputs and cli options.

lwronski (approved):
LGTM

modules/build/src/main/scala/scala/build/internal/util/WarningMessages.scala
```diff
@@ -29,4 +29,13 @@ object WarningMessages {
        |Provide it as an option to the publish subcommand with:
        | $directiveName value:$rawValue
        |""".stripMargin
+
+  /** Using @main is impossible in new [[scala.build.internal.ClassCodeWrapper]] since none of the
+    * definitions can be accessed statically, so those errors are swapped with this text
+    * @param consequences
+    *   \- "it will be ignored" or "" (if the compilation will fail)
+    */
+  def mainAnnotationNotSupported(consequences: String): String =
```
Gedochao: why can't `consequences` just be a `Boolean` then...?

website/docs/guides/old-runner-migration.md
```diff
@@ -358,7 +358,7 @@ println("Args: " + args.mkString(" "))
 
 This in turn is the Scala `3.x` equivalent for its own old `scala` runner.
 
-```scala compile title=old-scala-shebang-3.sc
+```scala ignore title=old-scala-shebang-3.sc
```
Gedochao: NIT
````suggestion
```scala title=old-scala-shebang-3.sc
````

Gedochao (approved):
LGTM, commented on some nitpicks

PR #2040, opened 2023-04-17
Explicit handling of paths in using directives

### LINKED ISSUE #1098: Explicit handling of paths in using directives

lwronski:
Fixes #1098

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
@@ -4,7 +4,7 @@ import com.eed3si9n.expecty.Expecty.expect
 
 import java.io.IOException
 import scala.build.{BuildThreads, Directories, LocalRepo, Position, Positioned}
-import scala.build.options.{BuildOptions, InternalOptions, MaybeScalaVersion}
+import scala.build.options.{BuildOptions, InternalOptions, MaybeScalaVersion, ScalacOpt}
 import scala.build.tests.util.BloopServer
 import build.Ops.EitherThrowOps
 import scala.build.Position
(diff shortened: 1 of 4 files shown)


modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
```diff
@@ -107,4 +107,41 @@ class DirectiveTests extends munit.FunSuite {
...
+    )
+    testInputs.withBuild(baseOptions, buildThreads, bloopConfigOpt) {
+      (_, _, maybeBuild) =>
+        val build = maybeBuild.orThrow
+        val scalacOptions: Option[Positioned[ScalacOpt]] =
+          build.options.scalaOptions.scalacOptions.toSeq.headOption
+        assert(scalacOptions.nonEmpty)
+
+        val scalacOpt = scalacOptions.get.value.value
+        expect(scalacOpt == """-coverage-out:$${.}""")
```
sjrd: The expected value should be `-coverage-out:${.}` here, with a single `$`.

modules/directives/src/main/scala/scala/build/directives/DirectiveSpecialSyntax.scala
```diff
@@ -0,0 +1,28 @@
...
+    *   the value of the directive, e.g., "-coverage-out:${.}" for example for the directive "//>
+    *   using options "-coverage-out:${.}""
+    * @param path
+    *   the file path from which the directive is read; replacement occurs only if the directive is
+    *   from a local file
+    * @return
+    *   the directive value with the `${.}` pattern replaced by the parent directory, if applicable
+    */
+  def handlingSpecialPathSyntax(directiveValue: String, path: Either[String, os.Path]): String = {
+    val pattern: Regex = """(?<!\$)\$\{\.}""".r
```
sjrd: That won't work with `$$${.}`, which should actually replace the trailing `${.}` (and prepend it with a *single* `$`).

lwronski: Thanks for review @sjrd!

website/docs/commands/compile.md
```diff
@@ -209,6 +209,22 @@ There's a separate directive for passing multiple options at one time:
 @main def hello = if (true) println("Hello")
 ```
 
+### Explicit handling of paths in using directives
```
Gedochao: I wonder if this section belongs in the `compile` command doc... wouldn't it make more sense in the using directives guide? since it applies to all of them, no?

Gedochao (approved):
LGTM

PR #2046, opened 2023-04-19
Support declaring some test directives from the main scope

### LINKED ISSUE #1729: Defining project.scala dependencies in test scope?

Gedochao:
Fixes #1729 

Some select directives (and not others) can now be declared for the test scope even in main scope sources. 
Those now support a separate key with the `test.*` prefix, indicating the test scope target.
The recommended location for declaring such directives would be the optional `project.scala` configuration file.

Directives supporting this feature:
```scala
//> using test.jar "<custom jar path>"
//> using test.dep "<dependency>"
//> using test.option "<scalac option>"
//> using test.javacOpt "<javac option>"
//> using test.javaOpt "<java option>"
//> using test.javaProp "<java properties>"
//> using test.toolkit "latest"
//> using test.resourceDir "<resources directory>"
```

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
@@ -7,7 +7,7 @@ import java.nio.charset.StandardCharsets
 import scala.build.EitherCps.{either, value}
 import scala.build.Logger
 import scala.build.errors.{BuildException, FileNotFoundException}
-import scala.build.options.{BuildOptions, SuppressWarningOptions}
+import scala.build.options.{BuildOptions, SuppressWarningOptions, WithBuildRequirements}
 import scala.build.preprocessing.DirectivesProcessor.DirectivesProcessorOutput
 import scala.build.preprocessing.ExtractedDirectives.from
 import scala.build.preprocessing.ScalaPreprocessor.*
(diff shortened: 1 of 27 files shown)


modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
```diff
@@ -49,6 +53,26 @@ object PreprocessingUtil {
       allowRestrictedFeatures,
       suppressWarningOptions
     ))
-    (updatedOptions, directivesPositions)
+
+    val directives1 = updatedOptions.unused
```
lwronski, line 57: Rename this variable.

lwronski (commented):
Add explanation in the directive guide about the `test.` prefix, and include a list of the available options.

modules/build/src/main/scala/scala/build/preprocessing/PreprocessingUtil.scala
Gedochao: I'll do a followup PR with refactorings, including this part.
For now, I abide the existing convention here (however unreadable that may be)

Gedochao:
> Add explanation in the directive guide about the test. prefix, and include a list of the available options.

@lwronski I'll do a followup PR with refactorings and docs, don't want to make this one bigger

lwronski (approved):

PR #2053, opened 2023-04-24
Add exclude using directive

lwronski:


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -305,4 +317,43 @@ object CrossSources {
...
+        case Seq(one) => Right(one)
+        case excludes => Left(new MultipleExcludeError(excludes.flatMap(_.flatMap(_.positions))))
+      }
+    val excludePatterns = value(exclude).map(_.value).flatMap { p =>
+      val maybeRelPath = Try(os.RelPath(p)).toOption
+      maybeRelPath match {
+        case Some(relPath) if os.isFile(workspaceDir / relPath) =>
+          Seq(p, (workspaceDir / relPath).toString) // exclude relative file paths
+        case Some(relPath) if os.isDir(workspaceDir / relPath) =>
+          Seq(p, s"${workspaceDir / relPath}${File.separator}*") // exclude relative directory paths
```
MaciejG604: A little more readable IMO
```suggestion
          Seq(p, (workspaceDir / relPath / "*").toString) // exclude relative directory paths
```

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -305,4 +317,43 @@ object CrossSources {
...
+    val exclude: Either[BuildException, Seq[Positioned[String]]] =
+      sources.flatMap(_.options).map(_.internal.exclude).filter(_.nonEmpty) match {
+        case Nil      => Right(Nil)
+        case Seq(one) => Right(one)
+        case excludes => Left(new MultipleExcludeError(excludes.flatMap(_.flatMap(_.positions))))
+      }
+    val excludePatterns = value(exclude).map(_.value).flatMap { p =>
+      val maybeRelPath = Try(os.RelPath(p)).toOption
+      maybeRelPath match {
+        case Some(relPath) if os.isFile(workspaceDir / relPath) =>
```
MaciejG604: There's no need to check if the relPath is a file. Additionally if we don't check it, exclude regex like `src/*.scala` should work since they will fall into this case.

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -501,4 +506,132 @@ class SourcesTests extends munit.FunSuite {
     }
   }
 
+  test("throw error when exclude found in multiple file") {
```
romanowski: I think we should also test error if exclude comes from outside `project.scala` and from `project.scala` that is not top-level

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -501,4 +506,132 @@ class SourcesTests extends munit.FunSuite {
...
+  }
+
+  test("exclude absolute file paths") {
+    val testInputs = TestInputs(
+      os.rel / "Hello.scala" -> "object Hello",
+      os.rel / "Main.scala" ->
+        """object Main {
+          |}""".stripMargin,
+      os.rel / "project.scala" ->
+        s"""//> using exclude "$${.}${File.separator}Main.scala" """
```
romanowski: hmm, is that an absolute path? TBH I do not understand how this path work

modules/core/src/main/scala/scala/build/errors/MultipleExcludeError.scala
```diff
@@ -0,0 +1,10 @@
+package scala.build.errors
+
+import scala.build.Position
+
+final class MultipleExcludeError(positions: Seq[Position], expectedProjectFilePath: os.Path)
```
romanowski: Nit: as it is used to mark exclude outside of `project.scala` I think it should be called `ExcludeDefinitionError`

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sources.scala
```diff
@@ -23,7 +23,8 @@ import scala.util.Try
 final case class Sources(
   @DirectiveName("file")
     files: DirectiveValueParser.WithScopePath[List[Positioned[String]]] =
-      DirectiveValueParser.WithScopePath.empty(Nil)
+      DirectiveValueParser.WithScopePath.empty(Nil),
+   exclude: List[Positioned[String]] = Nil
```
romanowski: It seems that the reference documentation is not updated - it seems that the mechanism to detect changes and/or generate the reference is broken.

romanowski (commented):
It generally looks good. There is an issue with documentation generation but not related to this PR.

Also, shouldn't we also provide a command line option to exclude things?

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
lwronski: It was added a new special syntax `${.}` in using directive for handling path. More info you can find [here](https://github.com/VirtusLab/scala-cli/blob/main/website/docs/guides/using-directives.md#explicit-handling-of-paths-in-using-directives).

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sources.scala
lwronski: Currently, reference documentation is only generated based on `@DirectiveGroupName`, `@DirectiveExamples`, `@DirectiveUsage`, and `@DirectiveDescription`. However, it seems that this mechanism should be updated to generate fields from the definition of using directives.

In this PR, I will fix this by adding a new definition of `Exclude` directive, and now the reference documentation is generated correctly.

modules/build/src/test/scala/scala/build/tests/ExcludeTests.scala
```diff
@@ -0,0 +1,183 @@
...
+        def file(artifact: Artifact) = sys.error("shouldn't be used")
+
+        def ec = sys.error("shouldn't be used")
+      }
+    ),
+    None,
+    () => sys.error("shouldn't be used")
+  )
+
+  test("throw error when exclude found in multiple file") {
```
MaciejG604:
```suggestion
  test("throw error when exclude found in multiple files") {
```

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -65,6 +65,8 @@ final case class Inputs(
 
   def add(extraElements: Seq[Element]): Inputs =
     if elements.isEmpty then this else copy(elements = (elements ++ extraElements).distinct)
+  def filter(filterElements: Seq[Element] => Seq[Element]): Inputs =
```
MaciejG604: The name 'filter' suggests an argument of type `T => Boolean`. Maybe let's change it to something like `withElementsTransformed` or just `withElements`?

MaciejG604 (approved):

Gedochao (approved):

PR #2069, opened 2023-05-01
Remove some dead code in build

alexarchambault:
Just stumbled upon that while trying to re-use that code in Almond


lwronski (approved):
Thanks!

PR #2076, opened 2023-05-04
improvement: Bump using directives to 1.0.0

### LINKED ISSUE #1954: Switch from Scala String + comma separated values syntax to whitespace delimited list
To improve the ability to copy-paste snippets, such as compiler options.

implement changes in the directives parser so that 
```scala
//> using options "-Xasync", "-Xfatal-warnings"
```
can be written as
```scala
//> using options -Xasync -Xfatal-warnings
```
Quotes can still be used for when a value needs to include whitespace, with the same escape characters for scala strings.

Here is a larger example:
```scala
//> using scala 3.2.2
//> using toolkit latest
//> using options -Xfatal-warnings -Yexplicit-nulls -Ysafe-init -source:future
//> using dep com.lihaoyi::cask:0.9.0
//> using some.key "I have spaces"
```

We do not anticipate issues in adapting [scala/vscode-scala-syntax](https://github.com/scala/vscode-scala-syntax) to support the new syntax. Additionally, values are still of `string` type.

no alternative considered.

**Additional context**
Since directives will not become scala syntax, there is no need to reuse the same rules as the scala parser.

_Discussed at the Scala Tooling Summit March 2023_

### LINKED ISSUE #2049: Multi-line comment syntax is still supported for directives
**Version(s)**

1.0.0-RC1

**Describe the bug**

It seems #1932 did not remove the support of multi-line directives:

~~~ scala
/*> using
  option "-Xfatal-warning"
*/
~~~

This code still compiles.

**Expected behaviour**

An error should be reported to the user, indicating what is the correct syntax to use.

Related: #2048 which also shows directives using multi-line comments.

tgodzik:
This includes the changes to simplify using directives:
- remove the possibility of nesting from API
- allow to omit commas in lists of values
- don't allow multiline comments
- remove multiline strings
- remove require and @require syntax support
- allow values without quotes
- remove @using
- remove multiline directives

Fixes https://github.com/VirtusLab/scala-cli/issues/2049
Fixes https://github.com/VirtusLab/scala-cli/issues/1954


tgodzik:
I updated some of the tests to use the new syntax, not sure if there is a new to do it for everything, but I would write new tests with the new syntax.

Gedochao (approved):

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -71,65 +51,21 @@ object ExtractedDirectives {
...
-      val specialCommentDirectives = byKind(UsingDirectiveKind.SpecialComment)
-      val plainCommentDirectives   = byKind(UsingDirectiveKind.PlainComment)
-
-      val directivesPositionsOpt =
-        if (
-          codeDirectives.containsTargetDirectivesOnly &&
-          specialCommentDirectives.containsTargetDirectivesOnly &&
-          plainCommentDirectives.containsTargetDirectivesOnly
-        )
+      val directives = allDirectives.head
```
lwronski: Are we sure that this list will always be non-empty?

tgodzik: I think that was the assumption before and we never get an empty one, but we can add a check to be on the safe side.

tgodzik: Ok changed to headOption, but now I realize that we could probably simplify the API and return a single `UsingDirective` instead of a list, but I don't want to release 2.0.0 of using directives just yet. There might some more things to fix.

In the meantime let's leave it as a list

lwronski (approved):

PR #2083, opened 2023-05-05
NIT Refactor test scope directives

Gedochao:
Cleaned up the test scope directive definitions & logic a bit, it should now be simpler to add new ones


MaciejG604 (approved):

lwronski (approved):
LGTM

PR #2101, opened 2023-05-08
Fix line conversion logic by simplifying topWrapperLen to line count of top wrapper

### LINKED ISSUE #2045: errors not reported in script files on last line (ensure "insert newline" is off)
**Version(s)**
1.0.0-RC2

**Describe the bug**
if you put erroneous code on the last line of a `.sc` file, (in an editor with BSP support) then no error will be reported.

**To Reproduce**
- Please ensure that your editor does not have any "insert newlines at the end of file" option turned on
- create a file `foo.sc` and run `scala setup-ide foo.sc`
- open the parent directory of `foo.sc` with an editor, e.g. VS Code + Metals, wait for metals to recognise the bsp project
- in the foo.sc file add code like:
```scala
val ok = "hello"
prntln(ok)
```
notice that if no newline is inserted, then no error is reported for the last line

**Expected behaviour**
error that "prntln" does not exist

I believe it is related to the wrapper code for scripts not inserting newlines after the last line.

MaciejG604:
Fixes #2045

### DIFF
--- a/modules/build/src/main/scala/scala/build/ConsoleBloopBuildClient.scala
+++ b/modules/build/src/main/scala/scala/build/ConsoleBloopBuildClient.scala
@@ -9,6 +9,7 @@ import java.nio.file.Paths
 import scala.build.errors.Severity
 import scala.build.internal.util.ConsoleUtils.ScalaCliConsole
 import scala.build.options.Scope
+import scala.build.postprocessing.LineConversion.scalaLineToScLine
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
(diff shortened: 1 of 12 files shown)


modules/build/src/main/scala/scala/build/ConsoleBloopBuildClient.scala
```diff
@@ -46,14 +47,14 @@ class ConsoleBloopBuildClient(
...
-          diag.getRange.getStart.getLine + lineOffset,
+          scalaLineToScLine(lineOffset)(diag.getRange.getStart.getLine).get,
           diag.getRange.getStart.getCharacter
         )
         val end = new bsp4j.Position(
-          diag.getRange.getEnd.getLine + lineOffset,
+          scalaLineToScLine(lineOffset)(diag.getRange.getEnd.getLine).get,
           diag.getRange.getEnd.getCharacter
         )
         val range = new bsp4j.Range(start, end)
```
Gedochao: ...or you could avoid the whole `.isDefined` => `.get` with something like:
```suggestion
    diagnosticMappings.get(path)
      .map { case (originalPath, lineOffset) =>
        (
          originalPath,
          scalaLineToScLine(lineOffset)(diag.getRange.getStart.getLine),
          scalaLineToScLine(lineOffset)(diag.getRange.getEnd.getLine)
        )
      }
      .collect {
        case (originalPath, Some(scLineStart), Some(scLineEnd)) =>
          val start       = new bsp4j.Position(scLineStart, diag.getRange.getStart.getCharacter)
          val end         = new bsp4j.Position(scLineEnd, diag.getRange.getEnd.getCharacter)
          val range       = new bsp4j.Range(start, end)
```

Gedochao (approved):
LGTM

lwronski (approved):

PR #2113, opened 2023-05-11
Fix test watch infinite loop

### LINKED ISSUE #2056: Fix `--watch` with `test`
### Discussed in https://github.com/VirtusLab/scala-cli/discussions/2043

<sup>Originally posted by **alvinj** April 19, 2023</sup>
Hi, I’m wondering if the `--watch` option should work with the `test` option? What I’m seeing is that if I save my file once, all is well, but if I change the file, it goes into a continuous/infinite loop.

Here’s my test file, which uses Munit:

```scala
//> using scala "3"
//> using dep "org.scalameta::munit::0.7.29"

def isTrue(a: Int | String | Boolean): Boolean = a match
    case 0 | "" | "0" | false => false
    case _ => true

class MyTests extends munit.FunSuite:
    test("isTrue tests") {
        assert(isTrue(1))
    }
```

And this is the command I’m running:

```sh
scala-cli test --watch Tests.scala
```

As mentioned, the first time I run that command, I see a few lines of output, and then it stops. But if I just add a blank space to the file and save it again, the command keeps running and running. The output looks like this:

```sh
> scala-cli test --watch Tests.scala
Compiling project (Scala 3.2.2, JVM)
Compiled project (Scala 3.2.2, JVM)
MyTests:
  + isTrue tests 0.019s
Watching sources, press Ctrl+C to exit, or press Enter to re-run.
Compiling project (Scala 3.2.2, JVM)
Compiled project (Scala 3.2.2, JVM)
MyTests:
  + isTrue tests 0.019s
Watching sources, press Ctrl+C to exit, or press Enter to re-run.
MyTests:
  + isTrue tests 0.018s
Watching sources, press Ctrl+C to exit, or press Enter to re-run.
MyTests:
  + isTrue tests 0.019s
Watching sources, press Ctrl+C to exit, or press Enter to re-run.
(this goes on forever ...)
```

This happens on a macOS system, regardless of the editor I use. Thanks!</div>

MaciejG604:
Fixes #2056
Bug introduced in #1727

File-change watcher gets registered for path `.../.scala-build/project_XXX/classes/main` which causes perpetual sheduling of operations since the files change during building.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
@@ -2,6 +2,7 @@ package scala.cli.integration
 
 import com.eed3si9n.expecty.Expecty.expect
 
+import scala.cli.integration.TestUtil.normalizeConsoleOutput
 import scala.util.Properties
 
 trait RunScriptTestDefinitions { _: RunTestDefinitions =>
@@ -21,13 +22,6 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
         expect(output == message)
     }
   }
-
-  def normalizeConsoleOutput(text: String) =
-    text.replace(Console.RED, "")
-      .replace(Console.YELLOW, "")
-      .replace(Console.CYAN, "")
-      .replace(Console.MAGENTA, "")
-      .replace(Console.RESET, "")
 
   test("simple script") {
     simpleScriptTest()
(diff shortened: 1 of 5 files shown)


modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -444,6 +437,8 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
     }
 
     test("user readable error when @main is used") {
+      import TestUtil.normalizeConsoleOutput
```
Gedochao: NIT import it in the imports section of this source instead, since it's used in a couple spots

modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala
```diff
@@ -291,4 +292,12 @@ object TestUtil {
     }
     Await.result(readLineF, timeout)
   }
+
+  def normalizeConsoleOutput(text: String) = {
```
Gedochao, line 296: NIT This is essentially a duplicate of the functionality from `scala.build.internal.util.ConsoleUtils`... I wonder if it'd be worth the hassle to extract it to some common module, so that we can reuse it in tests.
can also leave it as is though. 🤷‍♂️

Gedochao (approved):
LGTM, commented on some nitpicks

PR #893, opened 2022-04-14
Fix quotes in shebang help

crater2150:
Replaces single quotes with backticks in the help for shebang, consistent with other help.
The single quotes cause issues with zsh completion, as they are placed in single quoted strings.

Closes #695.

PR #2118, opened 2023-05-15
Fix flaky completions for `zsh`

Jasper-M:
Quotes break zsh completions.

Or more specifically, escaped quotes `\'` seem to break completions on zsh.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala
@@ -17,7 +17,7 @@ class CompleteTests extends ScalaCliSuite {
   test("zsh bug") {
     // guard against https://github.com/alexarchambault/case-app/issues/475
     TestInputs.empty.fromRoot { root =>
-      val res = os.proc(TestUtil.cli, "complete", "zsh-v1", "2", "scala-cli").call(cwd = root)
+      val res = os.proc(TestUtil.cli, "complete", shellFormat, "2", "scala-cli").call(cwd = root)
       expect(res.exitCode == 0)
       expect(!res.out.text().contains(raw"\'"))
     }
--- a/modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala
@@ -14,6 +14,15 @@ class CompleteTests extends ScalaCliSuite {
     }
   }
 
+  test("zsh bug") {
+    // guard against https://github.com/alexarchambault/case-app/issues/475
+    TestInputs.empty.fromRoot { root =>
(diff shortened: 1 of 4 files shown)


Gedochao:
I tried to replicate the problem with completions  and `zsh` and ran into something like this:
```
▶ scala-cli 
(eval):37: unmatched '
(eval):37: unmatched '
(eval):37: unmatched '
```
(`scala-cli ` with a space and then `TAB`, breaks)
```
▶ scala-cli c
clean    -- Clean the workspace.
compile  -- Compile Scala code.
config   -- Configure global settings for Scala CLI.
console  -- Fire-up a Scala REPL.
```
(adding a `c` brings the completions back)

Is this the behaviour you're trying to fix?

Gedochao:
We might need to add a relevant anti-regression test.
The completions integration tests are here: https://github.com/Gedochao/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala#L7

Do you have the time to add a test?
If not, I can try to look at this later.

Gedochao:
the `reference-doc` check is failing on the CI, btw.
This is because reference docs haven't been regenerated in sync with your changes.

You can fix it by running
```bash
```
(as per [CONTRIBUTING.md](https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md))

Jasper-M:
> (adding a c brings the completions back)

Indeed, once you manage to limit the output to completions that don't contain quotes in their description it works.

> Is this the behaviour you're trying to fix?

Yes that's exactly it. Turns out a similar change has already been attempted in the past https://github.com/VirtusLab/scala-cli/pull/893

> Do you have the time to add a test?

I will give it a go.

I just noticed that completions are implemented by case-app instead of natively in scala-cli. I should probably open an issue there.

Jasper-M:
> I just noticed that completions are implemented by case-app instead of natively in scala-cli. I should probably open an issue there.

alexarchambault/case-app#475

modules/integration/src/test/scala/scala/cli/integration/CompleteTests.scala
```diff
@@ -14,6 +14,15 @@ class CompleteTests extends ScalaCliSuite {
     }
   }
 
+  test("zsh bug") {
+    // guard against https://github.com/alexarchambault/case-app/issues/475
+    TestInputs.empty.fromRoot { root =>
+      val res = os.proc(TestUtil.cli, "complete", "zsh-v1", "2", "scala-cli").call(cwd = root)
```
Gedochao:
```suggestion
      val res = os.proc(TestUtil.cli, "complete", shellFormat, "2", "scala-cli").call(cwd = root)
```
won't hurt to enforce this to work the same on `bash`.

Gedochao (approved):

PR #2120, opened 2023-05-16
Fix the `--source-jar` option & add corresponding `using` directives

### LINKED ISSUE #2097: Missing `using` directive for handling source jars
**Version(s)**
`Scala CLI version: 1.0.0-RC1`

**Describe the bug**

When using `//> using jar` directives to add a local jar into a project it seems that those jars aren't included when you look at the BSP communication about the `buildTarget.data`. This can be problematic as Metals at times will look there for certain things.

**To Reproduce**
You'll need something like this:

```scala
//> using scala "3.3.0-RC5"
//> using jvm "17"
//> using jar "./lib/core.jar"

@main def run() = println("hi")
```

You'll also need to create a `lib/` dir and stick a jar in there or just make a `core.jar` dummy. If you then look at the `bsp.trace.json` file if you're using Metals or trace the BSP communication in some other way you'll see this:

```
[Trace - 05:12:47 pm] Received response 'workspace/buildTargets - (2)' in 8ms
Result: {
  "targets": [
    ... I removed the test target as it isn't relevant
    {
      "id": {
        "uri": "file:/Users/ckipp/Documents/scala-workspace/generative/.scala-build/?id\u003dproject_a67e43e310"
      },
      "displayName": "project_a67e43e310",
      "baseDirectory": "file:///Users/ckipp/Documents/scala-workspace/generative/.scala-build/",
      "tags": [...],
      "languageIds": [...],
      "dependencies": [],
      "capabilities": {...},
      "dataKind": "scala",
      "data": {
        "scalaOrganization": "org.scala-lang",
        "scalaVersion": "3.3.0-RC5",
        "scalaBinaryVersion": "3.3",
        "platform": 1,
        "jars": [
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/3.3.0-RC5/scala3-compiler_3-3.3.0-RC5.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-interfaces/3.3.0-RC5/scala3-interfaces-3.3.0-RC5.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_3/3.3.0-RC5/scala3-library_3-3.3.0-RC5.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/tasty-core_3/3.3.0-RC5/tasty-core_3-3.3.0-RC5.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-asm/9.4.0-scala-1/scala-asm-9.4.0-scala-1.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/compiler-interface/1.3.5/compiler-interface-1.3.5.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-reader/3.19.0/jline-reader-3.19.0.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal/3.19.0/jline-terminal-3.19.0.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal-jna/3.19.0/jline-terminal-jna-3.19.0.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.10/scala-library-2.13.10.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/com/google/protobuf/protobuf-java/3.7.0/protobuf-java-3.7.0.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/util-interface/1.3.0/util-interface-1.3.0.jar",
          "file:///Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/net/java/dev/jna/jna/5.3.1/jna-5.3.1.jar"
        ],
        "jvmBuildTarget": {
          "javaHome": "file:///Users/ckipp/.sdkman/candidates/java/17.0.5-tem/"
        }
      }
    }
  ]
}
Error: null
```

Notice that in `data.jars` you don't see the `core.jar` I'd expect to see.

**Expected behaviour**
I'd expect that jars added in this way are reported in `data.jars`.

Gedochao:
Fixes #2097 

It turns out the `--source-jar` command line option was implemented, but not at all passed to the build (and thus, non-functioning). 
I fixed it, added corresponding directives and added some tests & docs.

```scala
//> using sourceJar path/to/jar-sources
```

Additionally, the code now assumes any JARs with the `*-sources.jar` name suffix are assumed to be sources JARs, and are treated appropriately (along with a warning).

```bash
scala-cli . --extra-jar '/path/to/custom-sources.jar'
# [warn] Jars with the *-sources.jar name suffix are assumed to be source jars.
# The following jars were assumed to be source jars and will be treated as such: /path/to/custom-sources.jar
```

### DIFF
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/ClasspathUtils.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/ClasspathUtils.scala
@@ -1,0 +1,7 @@
+package scala.build.preprocessing.directives
+
+object ClasspathUtils {
+  extension (classpathItem: os.Path) {
+    def hasSourceJarSuffix: Boolean = classpathItem.last.endsWith("-sources.jar")
+  }
+}
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -1373,7 +1373,10 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
       }
     }
 
-  test("source jars handled correctly from the command line") {
+  def testSourceJars(
+    directives: String = "//> using jar Message.jar",
+    passSourceJarFromCommandLine: Boolean = false
+  ): Unit = {
     val jarSources    = os.rel / "jarStuff"
     val mainSources   = os.rel / "src"
     val jarPath       = mainSources / "Message.jar"
@@ -1381,7 +1384,7 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     val inputs = TestInputs(
       jarSources / "Message.scala" -> "case class Message(value: String)",
       mainSources / "Main.scala" ->
-        s"""//> using jar Message.jar
+        s"""$directives
            |object Main extends App {
            |  println(Message("Hello").value)
            |}
@@ -1412,11 +1415,14 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
         extraOptions
       )
         .call(cwd = root)
+      val bspOptions =
+        if (passSourceJarFromCommandLine) List("--source-jar", sourceJarPath.toString)
+        else List.empty
       withBsp(
         inputs,
         Seq(mainSources.toString),
         reuseRoot = Some(root),
-        bspOptions = List("--source-jar", sourceJarPath.toString)
+        bspOptions = bspOptions
       ) {
         (_, _, remoteServer) =>
           async {
@@ -1444,6 +1450,17 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
           }
       }
     }
+  }
+
+  test("source jars handled correctly from the command line") {
+    testSourceJars(passSourceJarFromCommandLine = true)
+  }
+
+  test("source jars handled correctly from a using directive") {
+    testSourceJars(directives =
+      """//> using jar Message.jar
+        |//> using source.jar Message-sources.jar""".stripMargin
+    )
   }
 
   private def checkIfBloopProjectIsInitialised(
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -1375,7 +1375,8 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
 
   def testSourceJars(
     directives: String = "//> using jar Message.jar",
-    passSourceJarFromCommandLine: Boolean = false
+    passSourceJarFromCommandLine: Boolean = false,
+    checkTestTarget: Boolean = false
   ): Unit = {
     val jarSources    = os.rel / "jarStuff"
     val mainSources   = os.rel / "src"
@@ -1427,10 +1428,11 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
         (_, _, remoteServer) =>
           async {
             val buildTargetsResp = await(remoteServer.workspaceBuildTargets().asScala)
-            val Some(mainTarget) = buildTargetsResp
+            val targets = buildTargetsResp
               .getTargets
               .asScala
-              .collectFirst { case t if !t.getId.getUri.contains("-test") => t }
+            val Some(mainTarget) = targets.find(!_.getId.getUri.contains("-test"))
+            val Some(testTarget) = targets.find(_.getId.getUri.contains("-test"))
             // ensure that the project compiles
             val compileRes = await(remoteServer.buildTargetCompile(
               new b.CompileParams(List(mainTarget.getId).asJava)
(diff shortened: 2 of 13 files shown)


modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -1446,6 +1452,17 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
     }
   }
 
+  test("source jars handled correctly from the command line") {
+    testSourceJars(passSourceJarFromCommandLine = true)
+  }
+
+  test("source jars handled correctly from a using directive") {
```
lwronski: It seems to me that this test could be a unit test to verify if the value for sourceJar is properly set in `BuildOptions`. The parsing of using directives can be tested in a unit test. What do you think?

modules/directives/src/main/scala/scala/build/preprocessing/directives/ClasspathUtils.scala
```diff
@@ -0,0 +1,7 @@
+package scala.build.preprocessing.directives
+
+object ClasspathUtils {
+  extension (classpathItem: os.Path) {
+    def hasSourceJarSuffix: Boolean = classpathItem.last.endsWith("-sources.jar")
```
lwronski, line 5: Is this a new syntax specific to scala-cli, or is it used in another similar tool? If we assume it is added in scala-cli, there should also be a flag to disable it.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
Gedochao: the source jar passed via directive has to be packaged first, so I don't think a unit test is suitable.
additionally, I think it makes sense to double-check we pass sources correctly in BSP `buildTarget/dependencySources`
So I'd rather keep it as an integration test

modules/directives/src/main/scala/scala/build/preprocessing/directives/ClasspathUtils.scala
Gedochao: it's new syntax specific to `scala-cli`, although `*-sources.jar` is a widely used naming scheme for source JARs..
I'd rather not add a flag, as it'd be troublesome to respect it in the `using` directive 😐

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
lwronski: For the unit test, we can create a dummy source JAR, which can be a simple text file with the .jar extension.

I think the integration test can be moved to a unit test:

- `"source jars handled correctly from a using directive"`
- `"source jars handled correctly from the command line smartly assuming a *-sources.jar is a source jar"`

modules/directives/src/main/scala/scala/build/preprocessing/directives/ClasspathUtils.scala
lwronski: Indeed, adding a flag may not be a good idea in this case.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
Gedochao: Done, I converted those 2 ITs to unit tests in the last commit

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -1427,10 +1428,13 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
             val buildTargetsResp = await(remoteServer.workspaceBuildTargets().asScala)
-            val Some(mainTarget) = buildTargetsResp
+            val targets = buildTargetsResp
               .getTargets
               .asScala
-              .collectFirst { case t if !t.getId.getUri.contains("-test") => t }
+            val Some(mainTarget) =
+              targets.collectFirst { case t if !t.getId.getUri.contains("-test") => t }
+            val Some(testTarget) =
+              targets.collectFirst { case t if t.getId.getUri.contains("-test") => t }
```
MaciejG604: LGTM

NIT:
Using `find` instead of `collectFirst` in this line and above would be much cleaner

lwronski (approved):
LGTM

MaciejG604 (approved):

PR #2124, opened 2023-05-17
Add main class to jar manfiest in assembly

romanowski:


lwronski (approved):

Gedochao (approved):

Gedochao:
I just rebased this to include the fix from #2123, so that `docs-tests` can pass on the CI.

PR #2127, opened 2023-05-19
Bump toolkit to `v0.1.7` and handle the separate `toolkit-test` dependency

### LINKED ISSUE #2060: Using toolkit should add toolkit-test dependecy for test scope
Issue in Toolkit repo: https://github.com/scala/toolkit/issues/13
Using the toolkit directive (`using toolkit ...`) should add a dependency on `toolkit-test` artifact in the test scope.
The same behavior should be present for the command line option.

Gedochao:
Fixes #2060 

- bumps `toolkit` to `v0.1.7`
- since `toolkit` now consists of 2 dependencies (`toolkit` and `toolkit-test`), both are now added when the `--toolkit` option is passed or the `//> using toolkit latest` directive is declared
- for the `//> using toolkit` directive, `toolkit-test` is added to the test scope only (so to use `munit` one has to run a test scope source)
- for the `--toolkit` option `toolkit-test` is added to the main scope
  - adding it to the test scope only like in the `using` directive would require a large refactor and won't be done in this PR
  - if we decide to bring it in sync with the `using` directive, we should track it under a separate ticket and plan the time for it
  - the effort necessary would mostly cover adding a `--test-dependency` option, as we'd have to add API for passing `WithBuildRequirements[BuildOptions]` from the `cli` module
- the `//> using test.toolkit` directive puts both of the dependencies (`toolkit` and `toolkit-test`) in the test scope (as does putting the `//> using toolkit` directive in a test scope source)
-  backwards compatibility for `toolkit` `v0.1.6` and older is kept

cc @szymon-rd


szymon-rd:
Feature-wise, it seems great! And to be honest, right now we do not need to care that much about keeping backward compat.

modules/directives/src/main/scala/scala/build/preprocessing/directives/Toolkit.scala
```diff
@@ -32,32 +35,70 @@ final case class Toolkit(
...
+          )
   val handler: DirectiveHandler[Toolkit] = DirectiveHandler.derive
-  def buildOptions(t: Positioned[String]): Either[BuildException, BuildOptions] =
-    Right {
-      BuildOptions(
-        classPathOptions = ClassPathOptions(
-          extraDependencies = ShadowingSeq.from(List(resolveDependency(t)))
-        )
-      )
+  private def buildOptionsWithScopeRequirement(
```
lwronski, line 94: What do you think about this approach? I tried to simplify this function

https://github.com/VirtusLab/scala-cli/commit/6183b76aae8371cb7367101cbd4346a1cd164616

Gedochao: I don't think your approach will work... for the standard `//> using toolkit` directive, we need `toolkit` in the main scope and `toolkit-test` in the test scope, while your solution puts both in `main`.
My code gets complex partially because I need to merge both back into a single `BuildOptions` instance for `//> using test.toolkit`.
try it out with my tests if it's unclear.

lwronski: Right, It fixed it and now it should work - https://github.com/lwronski/scala-cli/commit/eb3eb6c0e86b81c0fdd1cfc7968286def922570c

MaciejG604 (approved):
LGTM!

lwronski (approved):
LGTM

PR #2133, opened 2023-05-22
Improve directive parsing errors & special-case `toolkit` directive version parsing

### LINKED ISSUE #2129: Fix error message on //> using toolkit
When somebody just writes:
```
//> using toolkit
```
He is met with this error message:
```
[error]  expected 1 for the  using directive key at [...]; but got 1 values, instead.
```
Some users may find it hard to fix and understand. Please, add error like this:
```
[error] You have to provide version for the toolkit. For example, use: //> using toolkit "latest"
```

Gedochao:
Fixes #2129 
~Depends on #2127~

This fixes directive parsing errors to include the actual key of the directive for which the error occurred.
Additionally, the `//> using toolkit` directive is now special-cased to produced a better error message specifically for the Scala Toolkit.

```bash
scala-cli -e '//> using toolkit'
# [error]  Encountered an error when parsing the `toolkit` using directive.
# Expected a version to be passed.
# Example: `//> using toolkit latest`
```

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
@@ -9,6 +9,8 @@ import scala.build.tests.util.BloopServer
 import build.Ops.EitherThrowOps
 import dependency.AnyDependency
 
+import scala.build.errors.ToolkitDirectiveMissingVersionError
+
 class DirectiveTests extends munit.FunSuite {
 
   val buildThreads = BuildThreads.create()
@@ -87,11 +89,11 @@ class DirectiveTests extends munit.FunSuite {
     }
   }
 
-  test("resolve toolkit & toolkit-test dependency") {
+  test(s"resolve toolkit & toolkit-test dependency with version passed") {
     val testInputs = TestInputs(
       os.rel / "simple.sc" ->
-        """//> using toolkit latest
-          |""".stripMargin
+        s"""//> using toolkit latest
+           |""".stripMargin
     )
     testInputs.withBuilds(baseOptions, buildThreads, bloopConfigOpt) {
       (_, _, maybeBuilds) =>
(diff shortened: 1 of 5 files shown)


sjrd:
> Additionally, I added a specialcase enabling for the `toolkit` version not to be passed when declaring the directive.
> 
> In other words, this is now legal and will use the latest Scala `toolkit`:
> 
> ```scala
> //> using toolkit
> println(os.pwd)
> ```

I really don't think this is a good idea. There is already way too much emphasis in the scala-cli "documentation ecosystem" on using `latest.release`. It will be detrimental to the robustness of the Scala ecosystem. With that mindset, scala-cli will introduce in the Scala ecosystem a very big problem that it does not currently have.

Enshrining this mindset with an even shorter form in the definition of build files will only accentuate this phenomenon.

It would be much, much better to have a command-line command `scala-cli install toolkit` that adds the relevant `//> using toolkit "1.2.3"` line with the specific version available at the time of running the command. And the same for any other library, for that matter.

mpkocher:
> Fixes #2129 ~Depends on #2127~
> 
> This fixes directive parsing errors to include the actual key of the directive for which the error occurred. Additionally, I added a specialcase enabling for the `toolkit` version not to be passed when declaring the directive.
> 
> In other words, this is now legal and will use the latest Scala `toolkit`:
> 
> ```scala
> //> using toolkit
> println(os.pwd)
> ```

It's really not the greatest idea to randomly change the spec of how directives are defined. Where was the discussion on changing the semantics/spec?

Coming at it from another angle, what is the point of defining use cases and requirements/spec, if someone is just going to fundamentally change the semantics in a random PR.

szymon-rd:
It's a problematic solution as it introduces unexpected non-reproducibility of builds. We decided not to go in that direction.

Gedochao:
@mpkocher @sjrd I adjusted the code to return a meaningful error for when the `toolkit` version is missing. More discussion happened in #2129.
The TL;DR of it is that it was prevously discussed we shouldn't have `latest` as the implicit version.
Sorry for the confusion, it's on me.

sjrd:
Thank you :)

lwronski (approved):
LGTM

PR #2136, opened 2023-05-23
Add object wrapper flag and directive

### LINKED ISSUE #2117: Allow to use object as a script wrapper
**Version(s)**
Only from 1.0.0-RC2, everything is fine with previous releases.

**Describe the bug**
The script compile, then a `*** RUN ABORTED***` message is printed
image.

*(Of course I could use the new `test.dep` and `scl test myscript.sc` approach but it should run as before the old way)*

**To Reproduce**
- KO script : 
  - https://gist.github.com/dacr/fa515384be730103dbd37e7b3815b58f  
```scala
// ---------------------
//> using scala  "3.2.2"
//> using dep "org.scalatest::scalatest:3.2.15"
// ---------------------

import org.scalatest.*, flatspec.*, matchers.*

class PiTest extends AnyFlatSpec with should.Matchers {
  "pi calculus" should "return a precise enough pi value" in {
    math.Pi shouldBe 3.14158d +- 0.001d
  }
}
org.scalatest.tools.Runner.main(Array("-oDF", "-s", classOf[PiTest].getName))
```
- OK script : 
  - https://gist.github.com/dacr/78bb969a26d7394900396fd8ebd0142b

**Expected behaviour**
running fine without this error message and with this output : 
image

### LINKED ISSUE #2128: Update to 1.0.0-RC2 triggers 'assertion failed: asTerm called on not-a-Term val <none>'
**Version(s)**
1.0.0-RC2

**Describe the bug**
After upgrading to `1.0.0-RC2` the script below fails to compile with the following error:
```
Error: Unexpected error when compiling project_8be8567734-de847ff703: 'assertion failed: asTerm called on not-a-Term val <none>'
```

System: Tried Ubuntu 23.04 and 22.04 so far

**To Reproduce**
```scala
//> using scala "3.2.2"
//> using jvm "17"
//> using lib "com.github.alexarchambault::case-app:2.1.0-M24"

import caseapp._

case class Options(
  domain: String = "development"
)

object BuildAndRegister extends CaseApp[Options] {
  override def run(options: Options, remainingArgs: RemainingArgs): Unit = {}
}
```

```bash
scala-cli compile script.sc
```

Note that the bug only appears to occur on Scala 3 in combination with a script:

```bash
scala-cli --cli-version 1.0.0-RC2 compile script.sc # error
scala-cli --cli-version 1.0.0-RC1 compile           # works
scala-cli compile --scala-version 2.13              # also works

mv script.sc script.scala
scala-cli compile script.sc # works
```

Also, if we remove the default arg in the options case class, it also works again:

```diff
case class Options(
-  domain: String = "development"
+ domain: String
)
```

I haven't been able to minimize the example further yet.

One hunch is that the new class based wrapping triggers something here as it happens only in scripts.

**Expected behaviour**
The script should compile like in earlier versions.

MaciejG604:
Addresses #2128 and #2117

Adds cli flag `--object-wrapper` and `using objectWrapper` for forcing object-wrapping for script.

Additionally fixes wrong mainClass name being passed for scripts when filename contains special characters e.g. `-`

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
@@ -160,16 +160,18 @@ case object ScriptPreprocessor extends Preprocessor {
     * @return
     *   code wrapper compatible with provided BuildOptions
     */
-  def getScriptWrapper(buildOptions: BuildOptions): CodeWrapper = {
-    val scalaVersionOpt = for {
-      maybeScalaVersion <- buildOptions.scalaOptions.scalaVersion
-      scalaVersion      <- maybeScalaVersion.versionOpt
-    } yield scalaVersion
-    buildOptions.scalaOptions.platform.map(_.value) match {
-      case Some(_: Platform.JS.type)                      => ObjectCodeWrapper
-      case _ if scalaVersionOpt.exists(_.startsWith("2")) => ObjectCodeWrapper
-      case _                                              => ClassCodeWrapper
+  def getScriptWrapper(buildOptions: BuildOptions): CodeWrapper =
+    buildOptions.scriptOptions.forceObjectWrapper match {
+      case Some(true) => ObjectCodeWrapper
+      case _ =>
+        val scalaVersionOpt = for {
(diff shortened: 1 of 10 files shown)


lwronski (approved):
LGTM

PR #2135, opened 2023-05-22
Bump typelevel-toolkit to 0.0.11, configure toolkit-test

armanbilge:
Follow-up to https://github.com/VirtusLab/scala-cli/pull/2127 with same changes for typelevel-toolkit. cc @toniogela

PR #2137, opened 2023-05-23
Add a guide doc for toolkits & enforce all toolkits to have a separate `toolkit-test`

Gedochao:
A followup for #2135, #2127 and #2025 

With these changes, we now enforce all toolkits to have the `toolkit` and `toolkit-test` dependency separation. 
Since `Typelevel` supports it now, we might as well treat it as the norm.

I also added some misc refactors and a dedicated guide doc for using toolkits.


Gedochao:
cc @armanbilge @TonioGela @szymon-rd

MaciejG604 (approved):
LGTM

PR #2138, opened 2023-05-23
Fix updating toolkit dependencies

Gedochao:
`dependency-update` has been breaking because of the extra `toolkit-test` dependency in toolkits.
This should fix the `main` branch CI failure on M1: https://github.com/VirtusLab/scala-cli/actions/runs/5054418646/jobs/9069673783

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/DependencyUpdateTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/DependencyUpdateTests.scala
@@ -61,7 +61,7 @@ class DependencyUpdateTests extends ScalaCliSuite {
     }
   }
 
-  test("update typelevel toolkit dependence") {
+  test("update typelevel toolkit dependency") {
     val toolkitVersion = "0.0.1"
     val testInputs = TestInputs(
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/scala/scala/cli/commands/dependencyupdate/DependencyUpdate.scala
```diff
@@ -5,6 +5,7 @@ import caseapp.core.help.HelpFormat
 
 import scala.build.actionable.ActionableDependencyHandler
 import scala.build.actionable.ActionableDiagnostic.ActionableDependencyUpdateDiagnostic
+import scala.build.internal.Constants
```
MaciejG604:
```suggestion
```

All changes to this file could be reverted

MaciejG604 (approved):

modules/cli/src/main/scala/scala/cli/commands/dependencyupdate/DependencyUpdate.scala
Gedochao: fixed

PR #2152, opened 2023-05-25
Use external directive-handler library

alexarchambault:
This makes Scala CLI use directive handler classes from [an external project](https://github.com/scala-cli/directive-handler), rather than from its sources. This allows other projects to handle using directives with the same high level API as Scala CLI, rather than the rather low level [using_directives](https://github.com/VirtusLab/using_directives). I'm trying to [use this library from Almond](https://github.com/almond-sh/almond/pull/1124), for example.


build.sc
```diff
@@ -461,7 +461,8 @@ trait Directives extends ScalaCliSbtModule with ScalaCliPublishModule with HasTe
     Deps.bloopConfig,
     Deps.jsoniterCore,
     Deps.pprint,
-    Deps.usingDirectives
+    Deps.usingDirectives,
+    ivy"sh.almond::directive-handler:0.13.14-33-77bfc73b-SNAPSHOT"
```
tgodzik: Ideally this would be under the same org as the normal using directives, no?

alexarchambault: Yeah, my bad, that's stale code. Just changed that.

alexarchambault:
I'd be ok giving you access to [the directive-handler repo](https://github.com/scala-cli/directive-handler), or having you fork it under the VirtusLab org, and manage it from there.

tgodzik:
> I'd be ok giving you access to [the directive-handler repo](https://github.com/scala-cli/directive-handler), or having you fork it under the VirtusLab org, and manage it from there.

I think it would be best to fork it under VirtusLab since it's a core part of ScalaCLI and it would be easier for use manage access for people in the team

tgodzik:
@romanowski could you fork the repository?

romanowski:
Sure: https://github.com/VirtusLab/directive-handler

tgodzik:
@alexarchambault where do we stand with this? Could we instead publish the library from the usual Scala CLI repo?

I think that's what we discussed at the call, or am I totally mistaken?

Gedochao:
Closing this, there'll be another approach to this soon.
To be tracked under #2443.

PR #2153, opened 2023-05-26
Fix - add test to output from name of script example

lwronski:


website/docs/guides/scripts.md
```diff
@@ -185,9 +185,13 @@ chmod +x script.sc
 ```
 
 ```text
```
Gedochao, line 187:
````suggestion
```bash
````

lwronski: It is output from the previous command, so there's no need to execute the script again.

Gedochao: ahhhh, ok, my bad

Gedochao (approved):

PR #2168, opened 2023-05-29
Fix conflicts when watch and interactive try to read StdIn

### LINKED ISSUE #2121: Multi-main project run in interactive with watch fails after first selection
**Version(s)**
1.0.0-RC2

**Describe the bug**
When running 

**To Reproduce**
1. create:
**bug.scala**
```scala
object Run1 extends App {}
object Run2 extends App {}
```
2. run:
```
scala-cli run --watch --interactive bug.scala
```
3. Select `Run1` with `0`
4. Press enter to run again
5. Select `Run1` with `0` again.
6. From now on scala-cli will be stuck in a loop with the following error:
```
Unable to parse input: integer value from 0 to 1 is expected.
```

**Expected behaviour**
Second (and other runs) should continue correctly

MaciejG604:
Fixes #2121

The original problem was caused by watch mode reading StdIn for `ENTER` and then immediately doing `StdIn.read` again, which stole the first character from what interactive mode should have read.

However there's another issue with watch-interactive interaction, sometimes the watch mode reruns when sources change and should no longer read `StdIn`, but the `StdIn.read` is not interruptible, so again it will steal first character from interactive mode.

The solution from this PR fixes both problems.

### DIFF
--- a/modules/options/src/main/scala/scala/build/interactive/Interactive.scala
+++ b/modules/options/src/main/scala/scala/build/interactive/Interactive.scala
@@ -1,6 +1,6 @@
 package scala.build.interactive
 
-import scala.io.StdIn
+import scala.build.internal.StdInConcurrentReader
 
 sealed abstract class Interactive extends Product with Serializable {
   def confirmOperation(msg: String): Option[Boolean]                = None
@@ -16,7 +16,7 @@ object Interactive {
   private def readLine(): String =
     interactiveInputsOpt match {
       case None =>
-        StdIn.readLine()
+        StdInConcurrentReader.waitForLine().getOrElse("")
       case Some(interactiveInputs) =>
         synchronized {
           interactiveInputs match {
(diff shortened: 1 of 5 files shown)


modules/options/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -35,10 +35,7 @@ object Interactive {
     private sealed abstract class Action[V] extends Product with Serializable {
       def msg: String
       def action: Option[V]
-      final def run: Option[V] =
-        if (interactiveInputsOpt.nonEmpty || coursier.paths.Util.useAnsiOutput())
-          action
-        else None
```
lwronski: Why was this code removed?

MaciejG604: There's a condition inside that interfered with testing `--interactive` with stdin input (namely `if (System.console() == null) return false`), after consideration this check is required. I added an environment variable specifically for testing to override this behaviour.

modules/options/src/main/scala/scala/build/interactive/Interactive.scala
```diff
@@ -36,7 +36,11 @@ object Interactive {
       def msg: String
       def action: Option[V]
       final def run: Option[V] =
-        if (interactiveInputsOpt.nonEmpty || coursier.paths.Util.useAnsiOutput())
+        if (
+          interactiveInputsOpt.nonEmpty ||
+          coursier.paths.Util.useAnsiOutput() ||
+          System.getenv("SCALA_CLI_INTERACTIVE_TEST") != null
```
lwronski:
```suggestion
          System.getenv("SCALA_CLI_INTERACTIVe") != null
```

This flag could be useful not only in the tests.

lwronski (approved):
LGTM

PR #2171, opened 2023-05-30
Bsp wrapper fixes

### LINKED ISSUE #2158: Top and bottom wrappers broken after recent changes
**Version(s)**
1.0.0

**Describe the bug**
Metals uses `WrappedSourcesResult` to get the approximate look of a script, so that the compiler can understand it properly.

Previously we would get:
```
package foo
                object hello {
```

for top and

```
def args = hello_sc.args$
                }
                object hello_sc {
                  private var args$opt = Option.empty[Array[String]]
                  def args$ = args$opt.getOrElse(sys.error("No arguments passed to this script"))
                    args$opt = Some(args)
                    hello
                  }
                }

```

For the bottom.

Now we only have:

`package` for the top and nothing for the bottom.

This makes Metals barely understand Scala CLI scripts and most of the time they stop working.

**To Reproduce**
This is the reason update PR fails in https://github.com/scalameta/metals/pull/5276

**Expected behaviour**

The wrappers could probably stay the same.

### LINKED ISSUE #2165: No semanticdb for scripts
**Version(s)**
1.0.0

**Describe the bug**
Previously, we would get a proper semanticdb file for script files, which are needed for definition, references etc.

**To Reproduce**
Try using two scripts and reuse definitions from another one. Or alternatively try to look for refrences.

**Expected behaviour**
Semanticdb files are properly generated for script files

This should have been detected by tests, but it seems that we might be missing any tests here, so the script behaviour in Metals became very unstable.

MaciejG604:
Fixes #2158 and #2165

Refactored propagating the value of topWrapperLineCount into a bigger structure `WrapperParams` that has more information.


lwronski (approved):
LGTM

modules/core/src/main/scala/scala/build/internals/CodeWrapper.scala
```diff
@@ -40,3 +49,10 @@ object CodeWrapper {
   def mainClassObject(className: Name): Name =
     Name(className.raw ++ "_sc")
 }
+
+case class WrapperParams(
+  topWrapperLineCount: Int,
+  userCodeLineCount: Int,
+  bottomWrapperLineCount: Int,
+  userCodeNestingLevel: Int
```
tgodzik:
```suggestion
  userCodeNestingLevel: Int
```
Is this used?

tgodzik (approved):
Thanks for working on this! Just one question from me

modules/core/src/main/scala/scala/build/internals/CodeWrapper.scala
MaciejG604: Actually, it's never used. Good catch, i've removed it

PR #2174, opened 2023-05-31
refactor: Remove JavaInterface, which causes compilation issues with Bloop

tgodzik:
I don't think it's used anywhere right now, but we could just move it to a separate module if needed.


tgodzik:
@alexarchambault do we need that interface anywhere else or is it safe to remove? I can also just move it to another module.

Gedochao (approved):
LGTM

alexarchambault:
> do we need that interface anywhere else or is it safe to remove? I can also just move it to another module.

~Yes, it's used by coursier bootstraps, IIRC. They're written in Java, and can read the proxy config and the default repository list from the Scala CLI config using this API.~ My bad, they don't rely on this anymore (they must have at some point, possibly only locally on my machine). I guess this can be removed.

PR #2196, opened 2023-06-09
Support nativeTarget to build Scala Native static/shared libraries

keynmol:
Scala Native CLI supports building dynamic and shared libraries using `--build-target` argument.

So far it hasn't been supported. This PR adds the support, by

- Adding nativeTarget directive and native-target CLI option
- Splitting PackageType.Native into 3 different types
- Modifying the logic to not throw when main class is not found if we are building a library
- **IMPORTANT**: use `-Wl,install_name` argument to rename the library inside its own metadata, otherwise it points to `.scala-build/project_asdkljadljk190/main` - instead of `test.dylib` that we want


modules/options/src/main/scala/scala/build/options/ScalaNativeOptions.scala
```diff
@@ -20,13 +38,27 @@ final case class ScalaNativeOptions(
...
 
+  def target(): Option[ScalaNativeTarget] =
+    buildTargetStr.flatMap(ScalaNativeTarget.fromString)
+
+  private def targetCliOption(): List[String] =
+    import ScalaNativeTarget.*
+    val targ = target().map {
+      case Application    => "application"
+      case LibraryDynamic => "library-dynamic"
+      case LibraryStatic  => "library-static"
```
lolgab, line 57: Have you considered `dynamic-library` and `static-library`?

keynmol: These values actually have to match what Scala Native CLI uses: https://github.com/scala-native/scala-native-cli/blob/9d541e5c3a107b006cd61e7ece88df5cda7072e0/cli/src/main/scala/scala/scalanative/cli/options/NativeConfigOptions.scala#L40

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -85,6 +89,19 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
         configDb.get(Keys.actions).getOrElse(None)
       )
 
+    def detectNativePackageType(
+      platform: Platform,
+      sn: ScalaNativeOptions
+    ): Option[Native & scala.deriving.Mirror.Singleton] =
```
lwronski: Can you explain how `scala.deriving.Mirror.Singleton` works? Do we need to use it here?

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -101,12 +118,15 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
       ) { res =>
         res.orReport(logger).map(_.main).foreach {
           case s: Build.Successful =>
-            s.copyOutput(options.shared)
```
lwronski: Why was this code removed?

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -101,12 +118,15 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
...
             val mtimeDestPath = doPackage(
               logger = logger,
               outputOpt = options.output.filter(_.nonEmpty),
               force = options.force,
-              forcedPackageTypeOpt = options.forcedPackageTypeOpt,
+              forcedPackageTypeOpt =
+                options.forcedPackageTypeOpt orElse detectNativePackageType(
+                  s.options.platform.value,
+                  s.options.scalaNativeOptions
+                ),
```
lwronski: I think that instead of executing `detectNativePackageType` here, it can be handled in this [section](https://github.com/VirtusLab/scala-cli/blob/d062ea14f3286adbad722117bbac45cb69284085/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala#L222)

lwronski (commented):
It looks promising to me, and I would be happy to merge it. Unfortunately, there is one new test that is failing. You can find more details [here](https://github.com/VirtusLab/scala-cli/actions/runs/5219458233/jobs/9421507356?pr=2196#step:5:2708).

To fix the `checks` and `reference-doc` jobs, run the following commands before pushing your changes:
```
./mill -i __.fix
```

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
keynmol: Apologies, this is a an artefact of a Metals import - Singleton is not needed it here.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
keynmol: Human error :)

keynmol:
@lwronski, thanks, that's exactly the commands I was looking for :) 

If you are happy with the overall approach I will start fixing and cleaning up the PR

lwronski:
> @lwronski, thanks, that's exactly the commands I was looking for :)
> 
> If you are happy with the overall approach I will start fixing and cleaning up the PR

Yes, I'm satisfied with the overall approach, so I will wait for the final results.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -1016,9 +1073,24 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
...
+    val libraryLinkingOptions: Seq[String] =
+      Option.when(targetType != PackageType.Native.Application) {
+        /* If we are building a library, we make sure to change the name
+         that the linker will put into the loading path - otherwise
+         the built library will depend on some internal path within .scala-build
+         */
+
+        destPath.flatMap(_.lastOpt).toSeq.flatMap { filename =>
+          Seq("--linking-option", s"-Wl,-install_name,$filename")
+        }
```
keynmol: @WojciechMazur let me know if you have thoughts here - without it the libraries produced depend on themselves (located in .scala-build) so they're not portable

We're doing "convention over configuration" here for better user experience.

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -266,7 +266,8 @@ final case class SharedOptions(
       nativeLinking,
       nativeLinkingDefaults,
       nativeCompile,
-      nativeCompileDefaults
+      nativeCompileDefaults,
+      embedResources
```
keynmol: Flagging here - I _think_ this was a bug (embedResources not being passed)

Gedochao:
@keynmol what is the current status on this? can we unblock you somehow?

keynmol:
Hey @Gedochao! I didn't forget about this PR, just was short on time.

I resolved the conflicts (but incorrectly it seems, I will address those failures).

~~Now that I re-ran the tests on Linux, I remember the main problem - this PR relies on passing `-install_name` linker flag, to make sure that the dynamic library loading entries refer to the correct (**relative**) location of the built dynamic library, instead of the location where it's actually built (`.scala-build/.../`).~~

~~Problem is - on Linux this flag doesn't seem to work~~

Just as I was writing those last two paragraphs I found the linux equivalent option 🤦 

I'll pick this up again and hopefully will address remaining test failures.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -469,10 +469,60 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
...
 
-  if (!Properties.isWin && actualScalaVersion.startsWith("2.13"))
+  def libraryNativeTest(shared: Boolean = false): Unit = {
+    val fileName     = "simple.sc"
+    val nativeTarget = if (shared) "dynamic" else "static"
+    val inputs = TestInputs(
+      os.rel / fileName ->
+        s"""
+           |//> using platform scala-native
+           |//> using nativeTarget $nativeTarget
```
keynmol: @Gedochao miraculously, the tests are now green.

One issue I discovered though is that I can't seem to override this directive by passing `--native-target` in the command line.

This (from my side) would be the last thing I'd want to address with this PR

Gedochao: Nice!
I can try looking at it sometime next week, it's not exactly easy to pass it all the way from the command line.
I'll let you know when I do.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
WojciechMazur: I've never actually used this one, but seems to be required. I'm fine with with adding this flag, and wonder should we backport it to the ScalaNative toolchain

keynmol: I think ScalaNative toolchain has same issue:

```
$ otool -L target/scala-3.3.1/libscala-native-seed-project-out.dylib
target/scala-3.3.1/libscala-native-seed-project-out.dylib:
        /private/var/folders/rv/tdh6r2xx2b7by3z590lq8vrm0000gn/T/tmp.iWqfQcKp/scala-native-seed-project/target/scala-3.3.1/libscala-native-seed-project-out.dylib
```

Basically this dylib is not portable - it references itself in the target folder

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -469,10 +469,60 @@ abstract class PackageTestDefinitions(val scalaVersionOpt: Option[String])
...
     test("simple native") {
       simpleNativeTest()
     }
+    test("dynamic library native") {
+      libraryNativeTest(shared = true)
+    }
+
+    // To produce a static library, `LLVM_BIN` environment variable needs to be
+    // present (for `llvm-ar` utility)
+    if (sys.env.contains("LLVM_BIN"))
```
Gedochao, line 533: hey, for this to run on the CI we'd have to have the utility available and have the env variable point to it, right?
Am I assuming correctly that this test is being skipped on the CI right now?

keynmol: Yes, I explicitly ignored this test for now to save you the pain of installing LLVM on CI :D 

Additionally, I'm working on not needing this at all: https://github.com/scala-native/scala-native/pull/3548

Once that's merged and hopefully backported, we can remove the LLVM_BIN check and things should work out of the box

Gedochao: ah, okay.
I was already looking into setting it up on CI ;D
cool, in that case it's okay to skip this test temporarily 👍

keynmol: For posterity, installing LLVM is not horribly difficult, but it takes some time (though on the scale of how long the tests take.. it's insignificant)

https://github.com/indoorvivants/sn-bindgen/blob/main/build/setup-debian.sh#L5-L9

Using their official apt packages: https://apt.llvm.org/

And the LLVM_BIN value will be `/usr/lib/llvm-14/bin` where 14 is the version you install.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
keynmol: We can also go through the package type option (I think I had that originally), but it's a bit confusing with --native-target.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
keynmol: If you'd prefer - I can set up LLVM in only one of the workflows that runs this test (just need to figure out which one), to make sure we don't ignore any CI tests unnecessarily

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -272,7 +272,9 @@ final case class SharedOptions(
       nativeLinking,
       nativeLinkingDefaults,
       nativeCompile,
-      nativeCompileDefaults
+      nativeCompileDefaults,
+      embedResources,
+      nativeTarget
```
Gedochao, line 277: @keynmol this is what seems to have been missing for the command line override

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
Gedochao: I fixed the `--native-target` override in the last commit and added a relevant test, should work correctly now

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
Gedochao: If you have the time to set up the workflow, go ahead.
Although it's also fine to leave it temporarily inactive, since it seems it won't be necessary anyway once https://github.com/scala-native/scala-native/pull/3548 gets merged. Maybe easier that way for now.

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
keynmol: nice!

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
keynmol: > Maybe easier that way for now

Ok, I won't do it then :)

Once we have a timeline/decision on that PR being backported in 0.4.x, I can go ahead and make a separate PR adding LLVM (if necessary).

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaNative.scala
```diff
@@ -41,7 +44,8 @@ final case class ScalaNative(
   nativeClang: Option[String] = None,
   @DirectiveName("nativeClangPp")
     nativeClangPP: Option[String] = None,
-  nativeEmbedResources: Option[Boolean] = None
+  nativeEmbedResources: Option[Boolean] = None,
+    nativeTarget: Option[String] = None,
```
MaciejG604:
```suggestion
  nativeEmbedResources: Option[Boolean] = None,
  nativeTarget: Option[String] = None,
```

modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
```diff
@@ -61,6 +61,11 @@ final case class ScalaNativeOptions(
    @Tag(tags.implementation)
     nativeCompileDefaults: Option[Boolean] = None, //TODO does it even work when we default it to true while handling?
 
+  @Group(HelpGroup.ScalaNative.toString)
+  @HelpMessage("Build target type")
+  @Tag(tags.should)
```
MaciejG604, line 66:
```suggestion
  @Tag(tags.should)
  @ValueDescription("app|static|dynamic")
```

MaciejG604 (approved):
LGTM

Gedochao (approved):

keynmol:
🎉 

Thanks @Gedochao and @MaciejG604

PR #2202, opened 2023-06-09
Support for new command

### LINKED ISSUE #19: Add `new` command?
To create sources with skeletons of simple apps for cats-effect, zio, various testing frameworks, etc.

zetashift:
Adds a `new` command to `scala-cli` to enable creating projects through giter8.

Resolves https://github.com/VirtusLab/scala-cli/issues/19

New to the `scala-cli` codebase, so all feedback welcome as I have no real idea if I'm heading in the right direction.
I'll add docs when the code is in a better shape!


modules/cli/src/main/scala/scala/cli/commands/new/NewOptions.scala
```diff
@@ -0,0 +1,25 @@
+package scala.cli.commands.`new`
+
+import caseapp.*
+import scala.cli.commands.shared.*
+
+@HelpMessage(NewOptions.newMessage, "", NewOptions.detailedNewMessage)
+final case class NewOptions(
+  @Recurse
+  shared: SharedOptions = SharedOptions()
+) extends HasSharedOptions
```
zetashift: Do I really need `HasSharedOptions` here? I kinda added it to just satisfy `ScalaCommand`

Gedochao: actually, I believe you only need `HasGlobalOptions` here

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -0,0 +1,19 @@
...
+
+object New extends ScalaCommand[NewOptions] {
+  override def group: String = HelpCommandGroup.Main.toString
+
+  override def scalaSpecificationLevel = SpecificationLevel.IMPLEMENTATION
+
+  override def runCommand(options: NewOptions, remainingArgs: RemainingArgs, logger: Logger): Unit =
+    val inputArgs = remainingArgs.remaining
+    Giter8.run(inputArgs.toArray)
+    ()
```
zetashift: I should probably handle the exitcode giter8 gives back better, not sure what is idiomatic for scala-cli

lwronski: Hi @zetashift, thank you for your contribution. The overall changes are great, and `scala-cli` team would be happy to merge your changes.

We suggest downloading `giter8` as a JAR and running it using the Java command instead of adding the `giter8` dependency directly to the `scala-cli` code.

Our intention is to keep the current size of scala-cli, which is relatively small. We prefer not to add new dependencies that would significantly increase the size of the `scala-cli` fat JAR, which in this case add about 3-4Mb.

You can take inspiration from this [example](https://github.com/VirtusLab/scala-cli/blob/d062ea14f3286adbad722117bbac45cb69284085/modules/cli/src/main/scala/scala/cli/launcher/LauncherCli.scala#L33). In that case, scala-cli downloads a nightly version of the [scala-cli JVM launcher](https://scala-cli.virtuslab.org/docs/commands/basics/#scala-cli-version) and runs it.
 
Version, organization, and name of `giter8` can be hardcoded in `Constants.scala`, similar to how it was done for [scalafmt](https://github.com/VirtusLab/scala-cli/blob/d062ea14f3286adbad722117bbac45cb69284085/build.sc#L403).

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -0,0 +1,57 @@
...
+  override def group: String = HelpCommandGroup.Main.toString
+
+  override def scalaSpecificationLevel = SpecificationLevel.IMPLEMENTATION
+
+  val giter8Dependency =
+    Seq(dep"${Constants.giter8Organization}::${Constants.giter8Name}:${Constants.giter8Version}")
+
+  override def runCommand(options: NewOptions, remainingArgs: RemainingArgs, logger: Logger): Unit =
+    val fetchedGiter8 = Artifacts.fetch(
+      Positioned.none(giter8Dependency),
```
zetashift, line 23: @lwronski hi! Thank you for the help :D. I kinda just worked with the example you gave, because `Artifacts.fetch` is all function signature.

With my current code I'm running into
```
Got Scala dependency org.foundweekends.giter8::giter8_2.13:0.16.2, but no Scala version is provided
```

I'm not sure what the best way is to "supply" the scala version

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -0,0 +1,57 @@
...
+import scala.build.internal.{Constants, OsLibc, Runner}
+import scala.build.options.{BuildOptions, JavaOptions}
+import scala.build.{Artifacts, Logger, Positioned}
+import scala.cli.commands.ScalaCommand
+import scala.cli.commands.shared.{CoursierOptions, HelpCommandGroup}
+
+object New extends ScalaCommand[NewOptions] {
+  override def group: String = HelpCommandGroup.Main.toString
+
+  override def scalaSpecificationLevel = SpecificationLevel.IMPLEMENTATION
```
lwronski:
```suggestion
  override def scalaSpecificationLevel = SpecificationLevel.EXPERIMENTAL
```

Since this command is not part of SIP, we should mark it as experimental.

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def runCommand(options: NewOptions, remainingArgs: RemainingArgs, logger: Logger): Unit =
+    val fetchedGiter8 = Artifacts.fetch(
+      Positioned.none(giter8Dependency),
+      Seq.empty,
+      None,
+      logger,
+      CoursierOptions().coursierCache(logger.coursierLogger("")),
+      None
+    ) match {
```
lwronski:
```suggestion
   val scalaParameters = ScalaParameters(Constants.defaultScala213Version)
    val fetchedGiter8 = Artifacts.fetch(
      Positioned.none(giter8Dependency),
      Seq.empty,
      Some(scalaParameters),
      logger,
      CoursierOptions().coursierCache(logger.coursierLogger("")),
      None
    ) match {
```

project/deps.sc
```diff
@@ -113,6 +113,7 @@ object Deps {
   // TODO bump once 0.15.5 is out
   def expecty       = ivy"com.eed3si9n.expecty::expecty:0.16.0"
   def fansi         = ivy"com.lihaoyi::fansi:0.4.0"
+  def giter8        = ivy"org.foundweekends.giter8:giter8_2.13:0.16.2"
```
lwronski:
```suggestion
  def giter8        = ivy"org.foundweekends.giter8::giter8:0.16.2"
```

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
lwronski: I will suggest solution in the comments that should resolve your problems.

lwronski:
It would also be beneficial to add some integration tests. You can create a new class `NewTests` in integration test, similar to [FmtTests](https://github.com/VirtusLab/scala-cli/blob/ae4a44eac5e32313f3ed71788e8f3db1ef9d2866/modules/integration/src/test/scala/scala/cli/integration/FmtTests.scala#L7).

zetashift:
> It would also be beneficial to add some integration tests. You can create a new class NewTests in integration test, similar to [FmtTests](https://github.com/VirtusLab/scala-cli/blob/ae4a44eac5e32313f3ed71788e8f3db1ef9d2866/modules/integration/src/test/scala/scala/cli/integration/FmtTests.scala#L7).

Currently trying to add integration tests, but kinda lost in what kind of template to expect.

```scala
package scala.cli.integration

class NewTests extends ScalaCliSuite {
  
}
```

The scalafmt integrations seem quite straightforward, but for the giter8 dep I need some "default" scala-cli giter8 template no? And should this live on my github or can we update (for example) this one: https://github.com/VirtuslabRnD/scala-cli.g8 ?

lwronski:
> > It would also be beneficial to add some integration tests. You can create a new class NewTests in integration test, similar to [FmtTests](https://github.com/VirtusLab/scala-cli/blob/ae4a44eac5e32313f3ed71788e8f3db1ef9d2866/modules/integration/src/test/scala/scala/cli/integration/FmtTests.scala#L7).
> 
> Currently trying to add integration tests, but kinda lost in what kind of template to expect.
> 
> ```scala
> package scala.cli.integration
> 
> class NewTests extends ScalaCliSuite {
>   
> }
> ```
> 
> The scalafmt integrations seem quite straightforward, but for the giter8 dep I need some "default" scala-cli giter8 template no? And should this live on my github or can we update (for example) this one: https://github.com/VirtuslabRnD/scala-cli.g8 ?

I think it is the default [repository](https://github.com/VirtusLab/scala-cli.g8) that scala-cli wants to maintain. So let's use them in the integration test.

zetashift:
> I think it is the default [repository](https://github.com/VirtusLab/scala-cli.g8) that scala-cli wants to maintain. So let's use them in the integration test.

No idea why I couldn't find this, thank you! :D

modules/integration/src/test/scala/scala/cli/integration/NewTests.scala
```diff
@@ -0,0 +1,41 @@
...
+      os.proc(TestUtil.cli, "--power", "new", simpleTemplateName).call(cwd = root)
+      val content = os.read(root / "HelloWorld.scala")
+      expect(content == expectedTemplateContent)
+    }
+  }
+
+  test("error missing template argument") {
+    TestInputs.empty.fromRoot { root =>
+      val result = os.proc(TestUtil.cli, "--power", "new").call(cwd = root, check = false)
+      val lines = removeAnsiColors(result.err.text()).linesIterator.toVector
```
zetashift: This keeps being empty output(empty vector and string) and I can't figure out whyyyyy

lwronski: By default, `os-lib` inherits `stderr` from the shell and prints any errors to it. To override the `stderr` output, the easiest approach is to merge it with the `stdout`. The following code should work and pass the test.

```suggestion
     val result = os.proc(TestUtil.cli, "--power", "new").call(
        cwd = root,
        check = false,
        mergeErrIntoOut = true
      )
      val lines = removeAnsiColors(result.out.text()).linesIterator.toVector
```

zetashift: Thank you! : D I thought `result.err` would allow me to access the `stderr` output.

zetashift:
CI is reporting:

```
scala.cli.integration.SparkTests213:
  + run spark 3.3 standalone 9.271s
  + run spark spark-submit args 8.239s
scala.cli.integration.ExportJsonTestsDefault:
scala.cli.integration.DefaultFileTests:
1 targets failed
integration.test.test 1 tests failed: 
  scala.cli.integration.PackageTests213.simple script scala.cli.integration.PackageTests213.simple script
Error: Process completed with exit code 1.
```

I am quite sure I am not touching this part of the code: https://github.com/VirtusLab/scala-cli/blob/563fdd5ccf4d6d3c9d2e412f506b2eca711fa4bb/modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala#L33

lwronski:
> CI is reporting:
> 
> ```
> scala.cli.integration.SparkTests213:
>   + run spark 3.3 standalone 9.271s
>   + run spark spark-submit args 8.239s
> scala.cli.integration.ExportJsonTestsDefault:
> scala.cli.integration.DefaultFileTests:
> 1 targets failed
> integration.test.test 1 tests failed: 
>   scala.cli.integration.PackageTests213.simple script scala.cli.integration.PackageTests213.simple script
> Error: Process completed with exit code 1.
> ```
> 
> I am quite sure I am not touching this part of the code:
> 
> https://github.com/VirtusLab/scala-cli/blob/563fdd5ccf4d6d3c9d2e412f506b2eca711fa4bb/modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala#L33

Not really, it throws some [timeout](https://github.com/VirtusLab/scala-cli/actions/runs/5311000943/jobs/9614163181?pr=2202#step:5:5176). I will re-run CI.

zetashift:
> Not really, it throws some [timeout](https://github.com/VirtusLab/scala-cli/actions/runs/5311000943/jobs/9614163181?pr=2202#step:5:5176). I will re-run CI.

Ohhh didn't see that 🙈 , thank you :D.

bishabosha:
does this also create the `.bsp/scala-cli.json` file so that IDE import works right after? - would be good to not require another step from the user

zetashift:
> does this also create the `.bsp/scala-cli.json` file so that IDE import works right after? - would be good to not require another step from the user

Currently the command does not do this. I don't mind adding it!
A consideration:
- There is also nothing stopping one from instantiating a g8 template that targets a sbt or mill project, so maybe that option should be a flag?

lwronski:
> does this also create the `.bsp/scala-cli.json` file so that IDE import works right after? - would be good to not require another step from the user

No, it doesn't generate `.bsp` automatically, but it would be a good idea for Scala CLI to do so. I think this idea could be implemented in a separate pull request.

zetashift:
> I think this idea could be implemented in a separate pull request.

Yeah, sounds good to me.

lwronski (approved):
LGTM @zetashift Thanks for your contribution

lwronski:
> > I think this idea could be implemented in a separate pull request.
> 
> Yeah, sounds good to me.

To keep track of this idea, I have created an issue for it - https://github.com/VirtusLab/scala-cli/issues/2230

MaciejG604 (approved):

PR #2211, opened 2023-06-15
Fix dep update error

### LINKED ISSUE #2177: Spurious compilation error on dependency with snapshot version for verbosity level >= 2
**Version(s)**
1.0.0

**Describe the bug**
Compiling a file declaring a dependency on a library in a snapshot version produces a fake error message when `-v -v` flags are present. The entire compilation is successful however.

**To Reproduce**

Lib.scala
```scala
//> using scala 3.3.0
//> using publish.organization io.github.prolativ
//> using publish.name fake-lib
//> using publish.version 0.0.1-SNAPSHOT

object Lib
```

Main.scala
```scala
//> using scala 3.3.0
//> using dep io.github.prolativ::fake-lib:0.0.1-SNAPSHOT

object Main
```

Run:
```shell
scala-cli publish local Lib.scala

scala-cli compile Main.scala -v -v
```

Output:
```scala
...
[error]  No latest version found for io.github.prolativ::fake-lib:0.0.1-SNAPSHOT
```

**Expected behaviour**
No error message should be shown

MaciejG604:
Fixes #2177

### DIFF
--- a/modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
+++ b/modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
@@ -6,14 +6,13 @@ import coursier.parse.RepositoryParser
 import dependency.*
 
 import scala.build.EitherCps.{either, value}
-import scala.build.Positioned
 import scala.build.actionable.ActionableDiagnostic.*
-import scala.build.actionable.errors.ActionableHandlerError
-import scala.build.errors.{BuildException, RepositoryFormatError}
+import scala.build.errors.{BuildException, RepositoryFormatError, Severity}
 import scala.build.internal.Constants
 import scala.build.internal.Util.*
 import scala.build.options.BuildOptions
 import scala.build.options.ScalaVersionUtil.versions
+import scala.build.{Logger, Positioned}
 import scala.concurrent.duration.DurationInt
 
 case object ActionableDependencyHandler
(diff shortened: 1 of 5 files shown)


modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
```diff
@@ -28,62 +27,70 @@ case object ActionableDependencyHandler
...
   /** Versions like 'latest.*': 'latest.release', 'latest.integration', 'latest.stable'
     */
   private def isLatestSyntaxVersion(version: String): Boolean = Latest(version).nonEmpty
   private def findLatestVersion(
     buildOptions: BuildOptions,
-    dependency: AnyDependency
-  ): Either[BuildException, String] = either {
+    setting: Positioned[AnyDependency],
+    loggerOpt: Option[Logger]
+  ): Either[BuildException, Option[String]] = either {
```
lwronski, line 76:
```suggestion
  ): Option[String]  = 
```

lwronski (approved):
LGTM

modules/options/src/main/scala/scala/build/actionable/ActionableDependencyHandler.scala
MaciejG604: Still some operations can fail in this function, so the Either is needed.

PR #2227, opened 2023-06-19
Script wrapper verification

MaciejG604:
Tests for the script wrapper checking the generated source files for the wrapper type.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -63,7 +63,8 @@ sealed class UnwrappedCrossSources(
     *   CrossSources with all the scripts wrapped
     */
   def withWrappedScripts(buildOptions: BuildOptions): CrossSources = {
-    val codeWrapper = ScriptPreprocessor.getScriptWrapper(buildOptions)
+    val sharedOptions = this.sharedOptions(buildOptions)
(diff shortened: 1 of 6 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -63,7 +63,8 @@ sealed class UnwrappedCrossSources(
     *   CrossSources with all the scripts wrapped
     */
   def withWrappedScripts(buildOptions: BuildOptions): CrossSources = {
-    val codeWrapper = ScriptPreprocessor.getScriptWrapper(buildOptions)
+    val sharedOptions = this.sharedOptions(buildOptions)
```
tgodzik, line 66: So shared options should contain the right Scala version?

lwronski:
It would be beneficial to move some of the integration tests to unit tests to avoid a significant increase in job time on CI.

modules/build/src/main/scala/scala/build/CrossSources.scala
MaciejG604: Yes, when this method is passed the buildOptions created from CLI options it will merge them with the buildOptions obtained from using directives, thus we'll get the final version of options here.

tgodzik: Awesome, thanks!

lwronski (approved):
LGTM

PR #2261, opened 2023-07-03
Fix empty position in DependencyFormatErrors

MaciejG604:
During BSP operations errors like: `//> using dep not-a-dep` were not get highlighted, turned out their exceptions were not given postions. This PR fixes that.

### DIFF
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
@@ -62,9 +62,15 @@ object DirectiveUtil {
   extension (deps: List[Positioned[String]]) {
     def asDependencies: Either[BuildException, Seq[Positioned[AnyDependency]]] =
       deps
-        .map {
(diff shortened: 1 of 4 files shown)


modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
```diff
@@ -62,9 +62,15 @@ object DirectiveUtil {
...
-        .map {
-          _.map { str =>
-            DependencyParser.parse(str).left.map(new DependencyFormatError(str, _))
+        .map { positionedDep =>
+          positionedDep.map { str =>
+            DependencyParser.parse(str).left.map { error =>
+              new DependencyFormatError(
+                str,
+                error,
+                positionOpt = positionedDep.positions.headOption
```
lwronski: Why is only the head position passed from the `positionedDep.positions` list? The `DependencyFormatError` should be updated to accept a list of `Position`


```suggestion
                positionOpt = positionedDep.positions
```

lwronski (approved):

PR #2267, opened 2023-07-05
Add support for passing java properties into scala from .scala-jvmopts

### LINKED ISSUE #340: Support for `.jvmopts`
`.jvmopts` file is often used to store JVM options and Scala CLI may read JVM options from there.

lwronski:
Fixes #340

### DIFF
--- a/project/settings.sc
+++ b/project/settings.sc
@@ -836,8 +836,9 @@ trait ScalaCliModule extends ScalaModule {
   }
 }
 
-def workspaceDirName = ".scala-build"
-def projectFileName  = "project.scala"
+def workspaceDirName      = ".scala-build"
+def projectFileName       = "project.scala"
+def jvmPropertiesFileName = ".scala-jvmopts"
 
 final case class License(licenseId: String, name: String, reference: String)
 object License {
--- a/website/docs/under-the-hood.md
+++ b/website/docs/under-the-hood.md
@@ -46,6 +46,19 @@ In order set them, the `-D` command-line flags must be placed as the first optio
 scala-cli -Dfoo1=bar1 -Dfoo2=bar2 run ...
 ```
 
+The Scala CLI can also load Java properties from the `.scala-jvmopts` file present in the current working
+directory and import these Java properties into Scala CLI. Any java options in the `.scala-jvmopts` that are not
+recognizable as Java properties will be ignored.
+
+The example below shows that the Java properties `foo1` and `foo2` from the `.scala-jvmopts` file will be passed
+into the Scala CLI:
+```bash ignore
+$ cat .scala-jvmopts
+-Dfoo1=bar1 
+-Dfoo2=bar2
+$ scala-cli run ...
+```
+
 :::note
 - `scala-cli run . -Dfoo=bar` would pass the java property into your Scala app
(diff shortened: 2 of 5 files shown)


website/docs/under-the-hood.md
```diff
@@ -46,6 +46,19 @@ In order set them, the `-D` command-line flags must be placed as the first optio
 scala-cli -Dfoo1=bar1 -Dfoo2=bar2 run ...
 ```
 
+The Scala CLI can also load Java properties from the `.scala-jvmopts` file present in the current working
+directory and import these Java properties into Scala CLI. Any java options in the `.scala-jvmopts` that are not
+recognizable as Java properties will be ignored.
+
+The example below example show that the Java properties `foo1` and `foo2` from the `.scala-jvmopts` file will be passed
```
MaciejG604:
```suggestion
The example below shows that the Java properties `foo1` and `foo2` from the `.scala-jvmopts` file will be passed
```

website/docs/under-the-hood.md
```diff
@@ -46,6 +46,19 @@ In order set them, the `-D` command-line flags must be placed as the first optio
 scala-cli -Dfoo1=bar1 -Dfoo2=bar2 run ...
 ```
 
+The Scala CLI can also load Java properties from the `.scala-jvmopts` file present in the current working
+directory and import these Java properties into Scala CLI. Any java options in the `.scala-jvmopts` that are not
+recognizable as Java properties will be ignored.
+
+The example below show that the Java properties `foo1` and `foo2` from the `.scala-jvmopts` file will be passed
```
MaciejG604:
```suggestion
The example below shows that the Java properties `foo1` and `foo2` from the `.scala-jvmopts` file will be passed
```

MaciejG604 (approved):

project/settings.sc
```diff
@@ -836,8 +836,9 @@ trait ScalaCliModule extends ScalaModule {
   }
 }
 
-def workspaceDirName = ".scala-build"
-def projectFileName  = "project.scala"
+def workspaceDirName      = ".scala-build"
+def projectFileName       = "project.scala"
+def jvmPropertiesFileName = ".scala-jvmopts"
```
tgodzik, line 841: why `scala-jvmopts` ? That's not a standard name, so might be ahrd to discover.

lwronski: I'd vote for the name `.scala-jvmopts` to prevent potential collision with other tools that use `.jvm-opts`. Our native launcher restricts us from supporting every Java option added in `.jvmopts` and can only pass Java properties during Scala CLI execution.

The possibility of Scala CLI printing warning, unable to load some options like `-Xmx6g` from `.jvm-opts` in `scala-cli` launcher, can lead to confusion. Hence, I suggest using `.scala-jvmopts` name.

tgodzik: > The possibility of Scala CLI printing warning, unable to load some options like -Xmx6g from .jvm-opts in scala-cli launcher, can lead to confusion. Hence, I suggest using .scala-jvmopts name.

I would just mention that this is a native image and it doesn't support those options. Besides I am not sure this would be a common scenario. `.scala-jvmopts` would be very hard to discover and would require the user to dig into documentation, which we should avoid.

tgodzik (approved):

PR #2280, opened 2023-07-11
Bump case-app, add names limit to HelpFormat, move some name aliases, add test

### LINKED ISSUE #1883: Limit aliases for options in `--help`
Scala CLI has numerous aliases for many options. Which is nice, since it accounts for typos and makes it easier to remember the syntax for some, so let's not change that.
However, when viewing the `--help` for some commands, showing **all** of those aliases is overkill. 
I mean, look at this:
```bash
▶ scala-cli compile -h
Usage: scala-cli compile [options]
Compile Scala code.

You are currently viewing the basic help for the compile sub-command. You can view the full help by running: 
   scala-cli compile --help-full
For detailed documentation refer to our website: https://scala-cli.virtuslab.org/docs/commands/compile

Compilation options:
  -p, --print-classpath, --print-class-path  Print the resulting class path
  --test                                     Compile test scope

Scala options:
  -P, --plugin, --compiler-plugin string*        Add compiler plugin dependencies
  -S, --scala, --scala-version version           Set the Scala version (3.2.2 by default)
  --platform scala-js|scala-native|jvm           Specify platform
  -with-compiler, --with-compiler, --with-scala-compiler  Allows to include the Scala compiler artifacts on the classpath.
  -d, --compile-out, --destination, --compile-output, --output-directory, --compilation-output /example/path  Copy compilation results to output directory using either relative or absolute path
  --toolkit, --with-toolkit version|latest       Add toolkit to classPath

Java options:
  -j, --jvm jvm-name                             Use a specific JVM, such as `14`, `adopt:11`, or `graalvm:21`, or `system`
  -cp, --jar, --jars, --class, --classes, -classpath, --classpath, --extra-jar, --extra-jars, --class-path, --extra-class, --extra-classes, --extra-class-path paths  Add extra JARs and compiled classes to the class path
  --resource-dir, --resource-dirs paths          Add a resource directory

Watch options:
  -w, --watch            Run the application in the background, automatically wake the thread and re-run if sources have been changed
  --restart, --revolver  Run the application in the background, automatically kill the process and restart if sources have been changed

Dependency options:
  --dep, --dependency string*       Add dependencies
  -r, --repo, --repository string*  Add repositories

Help options:
  --help-js                                      Show options for ScalaJS
  --help-native                                  Show options for ScalaNative
  --help-doc, --doc-help, --help-scaladoc, --scaladoc-help  Show options for Scaladoc
  --help-repl, --repl-help                       Show options for Scala REPL
  --help-fmt, --fmt-help, --help-scalafmt, --scalafmt-help  Show options for Scalafmt
```
especially lines like
```
-cp, --jar, --jars, --class, --classes, -classpath, --classpath, --extra-jar, --extra-jars, --class-path, --extra-class, --extra-classes, --extra-class-path paths  Add extra JARs and compiled classes to the class path
```
it's a bit much, no?

Only show one or two aliases per option in the short `--help`.
We can show them all in `--full-help`, where the users are explicitly asking for verbosity.

So I'd like `-cp` to be mentioned in `--help` more like this:
```
  -cp, --jar    paths       Add extra JARs and compiled classes to the class path
```

MaciejG604:
Fixes #1883
Related to https://github.com/alexarchambault/case-app/pull/480

Some name aliases were moved to the top of the list of `@Name`s so that they are displayed in the short help. In this way we get full option name displayed and possibly a single letter flag for it.

Rules for limiting name aliases are:
Using the `namesLimit = Some(N)` parameter in HelpFormat would pick N-1 first name aliases + the main name.
For example, for option:
```
    @ExtraName("fourOption")
    @ExtraName("f")
    @ExtraName("four")
    @ExtraName("fOpt")
    fourth: Int = 0
```

and `namesLimit = Some(2)` will result in `--fourth, --four-option` getting displayed
and for full help `-f, --four, --f-opt, --fourth, --four-option`

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/HelpTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/HelpTests.scala
@@ -46,6 +46,23 @@ class HelpTests extends ScalaCliSuite {
       expect(fullHelpOutput.contains("Legacy Scala runner options"))
     }
   }
+
+  test("name aliases limited for standard help") {
+    val help       = os.proc(TestUtil.cli, "run", "--help").call()
+    val helpOutput = help.out.trim()
+
+    expect(TestUtil.removeAnsiColors(helpOutput).contains(
+      "--jar, --extra-jars paths"
+    ))
+  }
+
+  test("name aliases not limited for full help") {
+    val help       = os.proc(TestUtil.cli, "run", "--full-help").call()
+    val helpOutput = help.out.trim()
+    expect(TestUtil.removeAnsiColors(helpOutput).contains(
+      "-cp, --jar, --jars, --class, --classes, -classpath, --extra-jar, --classpath, --extra-jars, --class-path, --extra-class, --extra-classes, --extra-class-path paths"
+    ))
+  }
 }
 
 object HelpTests {
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroupOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/HelpGroupOptions.scala
@@ -22,9 +22,9 @@ case class HelpGroupOptions(
   helpNative: Boolean = false,
   @Group(HelpGroup.Help.toString)
   @HelpMessage("Show options for Scaladoc")
+  @Name("helpDoc")
   @Name("scaladocHelp")
   @Name("docHelp")
-  @Name("helpDoc")
   @Tag(tags.implementation)
   @Tag(tags.inShortHelp)
   helpScaladoc: Boolean = false,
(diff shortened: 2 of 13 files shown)


modules/integration/src/test/scala/scala/cli/integration/HelpTests.scala
```diff
@@ -46,6 +46,31 @@ class HelpTests extends ScalaCliSuite {
       expect(fullHelpOutput.contains("Legacy Scala runner options"))
     }
   }
+
+  def getOptionAliasesForExtraJars(help: String): Seq[String] = {
+    val description = "Add extra JARs and compiled classes to the class path"
+    expect(help.contains(description))
+    val line = help.linesIterator
+      .find(_.contains(description))
+      .get
```
lwronski: Can you add some logging to better understand what's happening? that the description should be updated to reflect the value that is set in the HelpMessage.

lwronski (approved):

modules/integration/src/test/scala/scala/cli/integration/HelpTests.scala
MaciejG604: I simplified the test so this function is now removed

lwronski (approved):

PR #2285, opened 2023-07-12
deps: update mill-scalafix to 0.3.1

ckipp01:
The main reason for this is that you can get rid of the external organize imports rule now that it's inlined.

### DIFF
--- a/project/settings.sc
+++ b/project/settings.sc
@@ -779,13 +779,13 @@ trait FormatNativeImageConf extends JavaModule {
 }
 
 trait ScalaCliScalafixModule extends ScalafixModule {
+
+  override def semanticDbVersion = Deps.Versions.scalaMeta
+
   def scalafixConfig = T {
     if (scalaVersion().startsWith("2.")) super.scalafixConfig()
     else Some(os.pwd / ".scalafix3.conf")
   }
-  def scalafixIvyDeps = super.scalafixIvyDeps() ++ Seq(
-    Deps.organizeImports
-  )
   def scalacPluginIvyDeps = super.scalacPluginIvyDeps() ++ {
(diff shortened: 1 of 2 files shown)


project/settings.sc
```diff
@@ -778,13 +778,13 @@ trait FormatNativeImageConf extends JavaModule {
 }
 
 trait ScalaCliScalafixModule extends ScalafixModule {
+
+  override def semanticDbVersion = T.input(Deps.Versions.scalaMeta)
```
ckipp01: I don't really get why this needs to be an `input` making it re-evaluate every time. In my own plugins I was able to just make this a string, but I'm also on `0.11.x`. So no idea if there is a way around this or not.

lwronski: @ckipp01, there's one difference  to note: `scala-cli` is still using Mill in version `0.10.12`.

ckipp01: > @ckipp01, there's one difference to note: `scala-cli` is still using Mill in version `0.10.12`.

Yup, that's what I meant by mentioning my usage of 0.11.x. Is there any plans to update? I'm assuming with all the breaking changes that it won't be a small task to update this build since it's so large.

lwronski: On Monday mill should be updated to `0.11.x`.

lwronski: The upgrade to Mill version `0.11.1` is currently in progress - https://github.com/VirtusLab/scala-cli/pull/2297. 

It should be merged in the coming days.

lwronski: https://github.com/VirtusLab/scala-cli/pull/2297 was merged. So ScalaCLI uses now Mill `0.11.1` .

ckipp01: Perfect, I just rebased.

lwronski (approved):
@ckipp01 Thanks for your contribution. LGTM

PR #2293, opened 2023-07-14
Run SetupIde command after creating a giter8 template

### LINKED ISSUE #2230: `new` command should generate `.bsp/scala-cli.json` file

When users run the new command, Scala CLI should generate a default .bsp/scala-cli.json file to enable IDE imports to work. This idea is based on the comment - https://github.com/VirtusLab/scala-cli/pull/2202#issuecomment-1597192591.

zetashift:
Resolves https://github.com/VirtusLab/scala-cli/issues/2230

I'll add the integration check that checks for the bsp json after I fix the parsing of options!

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
@@ -58,7 +58,7 @@ object New extends ScalaCommand[NewOptions] {
 
     if (exitCode == 0) {
       val inputs = options.shared.inputs(
-        remainingArgs.remaining,
+        Seq(options.out),
         () => Inputs.default()
       )(
         using invokeData
--- a/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
@@ -58,7 +58,7 @@ object New extends ScalaCommand[NewOptions] {
 
     if (exitCode == 0) {
       val inputs = options.shared.inputs(
-        Seq(options.out),
+        Seq(destPath.toString),
         () => Inputs.default()
       )(
         using invokeData
--- a/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
@@ -44,6 +44,7 @@ object New extends ScalaCommand[NewOptions] {
         jvmIdOpt = Some(OsLibc.baseDefaultJvm(OsLibc.jvmIndexOs, "17")).map(Positioned.none)
       )
     )
+    val destPath = options.out.filter(_.nonEmpty).map(os.Path(_, os.pwd)).getOrElse(os.pwd)
 
     val exitCode =
       Runner.runJvm(
@@ -51,7 +52,7 @@ object New extends ScalaCommand[NewOptions] {
         buildOptions.javaOptions.javaOpts.toSeq.map(_.value.value),
         giter8,
         "giter8.Giter8",
-        remainingArgs.remaining,
+        remainingArgs.remaining ++ Seq("-o", destPath.toString),
         logger,
         allowExecve = true
       ).waitFor()
--- a/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/new/New.scala
@@ -3,11 +3,13 @@ package scala.cli.commands.`new`
 import caseapp.core.RemainingArgs
 import dependency.*
 
+import scala.build.input.Inputs
 import scala.build.internal.{Constants, OsLibc, Runner}
 import scala.build.options.{BuildOptions, JavaOptions}
 import scala.build.{Artifacts, Logger, Positioned}
 import scala.cli.commands.ScalaCommand
-import scala.cli.commands.shared.{CoursierOptions, HelpCommandGroup}
+import scala.cli.commands.setupide.SetupIde
+import scala.cli.commands.shared.{CoursierOptions, HelpCommandGroup, SharedOptions}
 
 object New extends ScalaCommand[NewOptions] {
   override def group: String = HelpCommandGroup.Main.toString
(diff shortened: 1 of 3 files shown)


modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -54,5 +56,22 @@ object New extends ScalaCommand[NewOptions] {
...
+        using invokeData
+      ).orExit(logger)
+
+      SetupIde.runSafe(
+        options.shared,
+        inputs,
+        logger,
+        buildOptions,
+        Some(name),
+        Seq.empty
```
zetashift, line 74: This currently fails with:
```

[error]  zetashift/tlkit.g8: input file not found
Try viewing the relevant help to see the list of available sub-commands and options.
  scala-cli new --help
  
```
It seems that I'm giving the inputs/options wrong, but I'm not sure why :S

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -54,5 +56,22 @@ object New extends ScalaCommand[NewOptions] {
         allowExecve = true
       ).waitFor()
 
+    if (exitCode == 0) {
+      val inputs = options.shared.inputs(
+        remainingArgs.remaining,
```
lwronski: Hello @zetashift, thank you for start working on this issue. I think there's an bug in this line, as inputs expects a list of files and directories, yet `zetashift/tlkit.g8` was  passed.

I suggest the easiest solution that would be to pass the `output` directory where the g8 template was created:

```suggestion
        Seq(outputGiter8Path),
```

The Giter8 accepts the following options:
```
  -o, --out <value>        Output directory
```

I propose adding a new option to `NewOptions`, such as `--out`, `--dest`, `-o`, which would allow us to create a template in this directory. This would further enable us to reuse this path and pass it into `options.shared.inputs`

When you run `giter8.Giter8` before in code, you should specify the output directory for template.

modules/cli/src/main/scala/scala/cli/commands/new/NewOptions.scala
```diff
@@ -7,7 +7,10 @@ import scala.cli.commands.shared.*
 @HelpMessage(NewOptions.newMessage, "", NewOptions.detailedNewMessage)
 final case class NewOptions(
   @Recurse
-  global: GlobalOptions = GlobalOptions()
+  global: GlobalOptions = GlobalOptions(),
+  @Recurse
+  shared: SharedOptions = SharedOptions(),
+  out: String = ".",
```
zetashift: A hunch saying is this is not how I should declare the outdir argument lol

lwronski:
```suggestion
  out: Option[String] = None,
```

modules/cli/src/main/scala/scala/cli/commands/new/New.scala
```diff
@@ -54,5 +56,22 @@ object New extends ScalaCommand[NewOptions] {
         allowExecve = true
       ).waitFor()
 
+    if (exitCode == 0) {
+      val inputs = options.shared.inputs(
+        Seq(options.out),
```
lwronski:
```suggestion
      val inputs = options.shared.inputs(
        Seq(destPath.toString),
```

lwronski (commented):
It would be mandatory to include the output directory as a argument for the `giter8` app.
```
remainingArgs.remaining ++ Seq("-o", destPath.toString),
```

and destPath can be declared for example in this way:

```
val destPath = options.out.filter(_.nonEmpty).map(os.Path(_, os.pwd)).getOrElse(os.pwd)
```

bishabosha:
I've kind of thought a bit, if the g8 command is intended to be ran on any kind of g8 template (e.g. sbt or mill) then isn't generating `.bsp` going to conflict?

Now that metals 1.0.0 defaults to scala-cli (and generates .bsp) when it doesnt detect sbt or mill this seems less useful

zetashift:
> I've kind of thought a bit, if the g8 command is intended to be ran on any kind of g8 template (e.g. sbt or mill) then isn't generating `.bsp` going to conflict?
> 
> Now that metals 1.0.0 defaults to scala-cli (and generates .bsp) when it doesnt detect sbt or mill this seems less useful

I thought about this too!  I use scala-cli for sbt and scala-cli giter8 templates so I prefer it to not auto-generate it.
But I guess the ideal situation would be that it does it under certain circumstances.

lwronski:
> I've kind of thought a bit, if the g8 command is intended to be ran on any kind of g8 template (e.g. sbt or mill) then isn't generating `.bsp` going to conflict?
> 
> Now that metals 1.0.0 defaults to scala-cli (and generates .bsp) when it doesnt detect sbt or mill this seems less useful

@bishabosha, thank you for bringing this issue. The ScalaCli team has discussed it internally and we think that the best solution is not to automatically execute `SetupIde.runSafe` after each template generation. Instead, we could improve the logging.

For example, we can implement logic to identify if a generated template is a ScalaCli project. A simple way to achieve this could be to verify the absence of `build.sc` or `build.sbt`. Then we could then log message which inform users to execute `scala-cli setup-ide ${dest-path}` if they want IDE support.

zetashift:
> For example, we can implement logic to identify if a generated template is a ScalaCli project. A simple way to achieve this could be to verify the absence of build.sc or build.sbt. Then we could then log message which inform users to execute scala-cli setup-ide ${dest-path} if they want IDE support.

If it's okay with you, I'll close this PR and I'll try to wiggle some time to add the improvements to the logging in the coming week(s)?

lwronski:
> > For example, we can implement logic to identify if a generated template is a ScalaCli project. A simple way to achieve this could be to verify the absence of build.sc or build.sbt. Then we could then log message which inform users to execute scala-cli setup-ide ${dest-path} if they want IDE support.
> 
> If it's okay with you, I'll close this PR and I'll try to wiggle some time to add the improvements to the logging in the coming week(s)?

Certainly, that works for me. It's great to hear that you're planning to work on it,

Gedochao:
Closing this for now, based on the preceding discussion it seems that it won't be possible to merge it in its current state.
@zetashift feel free to reopen if you find the time to keep working on this.
Thanks for the work you've already managed to put into it!

PR #2297, opened 2023-07-17
Update mill 0.11.1

lwronski:


project/settings.sc
```diff
@@ -539,7 +540,7 @@ trait HasTests extends SbtModule {
       else Nil
     super.scalacOptions() ++ extraOptions
   }
-  trait Tests extends super.Tests {
+  trait ScalaCliTests extends super.SbtModuleTests {
```
MaciejG604: I think it should be like that to have `ScalaCliTests` mean the same as `Tests with ScalaCliTests` did before.
```suggestion
  trait ScalaCliTests extends ScalaCliModule with super.SbtModuleTests {
```

project/settings.sc
```diff
@@ -816,7 +816,7 @@ trait ScalaCliScalafixModule extends ScalafixModule {
   }
 }
 
-trait ScalaCliCrossSbtModule extends CrossSbtModule with ScalaCliModule
+trait ScalaCliCrossSbtModule extends Cross.Module[String] with CrossSbtModule with ScalaCliModule
```
MaciejG604: This is not needed since `CrossSbtModule` extends `CrossModuleBase` which in turn extends has `Cross.Module[String]` mixed-in.

project/settings.sc
```diff
@@ -704,7 +705,13 @@ private def doFormatNativeImageConf(dir: os.Path, format: Boolean): List[os.Path
...
             val values =
               arr.toVector.groupBy(_("name").str).toVector.sortBy(_._1).map(_._2).map { t =>
-                val entries = t.map(_.obj).reduce(_ ++ _)
+                val entries =
+                  t.map(_.obj).foldLeft(LinkedHashMap[String, ujson.Value]()) {
+                    case (acc, obj) =>
+                      obj.iterator.foldLeft(acc) { case (acc, v) =>
+                        acc.addOne(v)
+                      }
+                  }
```
MaciejG604:
```suggestion
                val entries = t.map(_.obj).reduce(_ addAll _)
```

MaciejG604 (approved):
LGTM!
I can tell this was not a walk in the park judging by how troublesome the review was. Hats off!

PR #2299, opened 2023-07-17
Add support for compile-only dependencies

### LINKED ISSUE #817: Add support for compile-only dependencies (aka "provided" dependencies)
We should allow users to add "compile-only" dependencies, added to the class path only at compile-time:
```scala
//> using compile-library "com.github.plokhotnyuk.jsoniter-scala::jsoniter-scala-macros:2.13.8"
```

There's already some support for compile-only class path internally (see `Artifacts.compileClassPath`). It should be a matter of adding JARs coming from dependency resolution there.

Implementation-wise, note that we should probably run two dependency resolutions when users add compile-only dependencies: one compile resolution with both compile-only and standard dependencies, and a second one with just standard dependencies.

lwronski:
Fixes #817

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala
+++ b/modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala
@@ -41,7 +41,7 @@ final case class SbtProjectDescriptor(
       !options.scalaOptions.addScalaCompiler.contains(true) &&
       sources.paths.forall(_._1.last.endsWith(".java")) &&
       sources.inMemory.forall(_.generatedRelPath.last.endsWith(".java")) &&
-      options.classPathOptions.extraDependencies.toSeq
+      options.classPathOptions.allExtraDependencies.toSeq
         .forall(_.value.nameAttributes == NoAttributes)
 
     val settings =
@@ -274,10 +274,16 @@ final case class SbtProjectDescriptor(
   private def dependencySettings(options: BuildOptions, scope: Scope): SbtProject = {
 
     val depSettings = {
-      val depStrings = options.classPathOptions
-        .extraDependencies.toSeq.toList
-        .map(_.value)
-        .map { dep =>
+      val allDeps =
+        options.classPathOptions.extraDependencies.toSeq.toList.map(_.value).map((_, false)) ++
+          options.classPathOptions.extraCompileOnlyDependencies
+            .toSeq
+            .toList
+            .map(_.value)
+            .map((_, true))
+
+      val depStrings = allDeps.map {
+        case (dep, isCompileOnly) =>
           val org  = dep.organization
           val name = dep.name
           val ver  = dep.version
@@ -293,11 +299,15 @@ final case class SbtProjectDescriptor(
               val sep = "%%"
               (sep, suffixOpt0)
           }
-          val scope0 = if (scope == Scope.Test) "% Test" else ""
+          val scope0 =
+            // FIXME This ignores the isCompileOnly when scope == Scope.Test
+            if (scope == Scope.Test) "% Test"
+            else if (isCompileOnly) "% Provided"
+            else ""
 
           val baseDep = s"""$q$org$q $sep $q$name$q % $q$ver$q $scope0"""
           suffixOpt.fold(baseDep)(suffix => s"($baseDep)$suffix")
-        }
+      }
 
       if (depStrings.isEmpty) Nil
       else if (depStrings.lengthCompare(1) == 0)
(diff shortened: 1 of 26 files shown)


modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala
```diff
@@ -274,10 +274,16 @@ final case class SbtProjectDescriptor(
...
+      val allDeps =
+        options.classPathOptions.extraDependencies.toSeq.toList.map(_.value).map((_, false)) ++
+          options.classPathOptions.extraCompileOnlyDependencies
+            .toSeq
+            .toList
+            .map(_.value)
+            .map((_, true))
+
+      val depStrings = allDeps.map {
+        case (dep, isCompileOnly) =>
```
tgodzik: MAybe a function instead of  zip?

Something like:
```
def toDepString(isCompileOnly: Boolean)(dep: ...) = 
``
and then just map `extraDependencies` and `extraCompileOnlyDependencies` separately?

File: modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala [hunk 2]
```diff
@@ -293,11 +299,15 @@ final case class SbtProjectDescriptor(
               val sep = "%%"
               (sep, suffixOpt0)
           }
-          val scope0 = if (scope == Scope.Test) "% Test" else ""
+          val scope0 =
+            // FIXME This ignores the isCompileOnly when scope == Scope.Test
+            if (scope == Scope.Test) "% Test"
```
tgodzik, line 303: Do we have `TestProvided` scope? Is this even possible? MAybe there shouldn't be an option to define them in test scope?

modules/options/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -310,6 +325,31 @@ object Artifacts {
...
+    }
+    val runtimeRes = {
+      val resolution = fetchRes.resolution.subset(updatedDependencies0.map(_._2._1))
+      // this is actually fetcher.artifacts, which is a private field…
+      val artifacts = coursier.Artifacts()
+        .withCache(fetcher.cache)
+        .withClassifiers(fetcher.classifiers)
+        .withMainArtifactsOpt(fetcher.mainArtifactsOpt)
+        .withArtifactTypesOpt(fetcher.artifactTypesOpt)
+        // .withOtherCaches(fetcher.otherCaches) missing on coursier.Fetch
```
tgodzik:
```suggestion
```

website/docs/commands/compile.md
```diff
@@ -482,4 +482,22 @@ For example, to exclude all files in the `example/scala` directory, add the foll
 
 ```scala title=project.scala
 //> using exclude "example/scala"
+```
+
+## Compile-Only Dependencies
+
+Compile-only dependencies, allowing users to include certain libraries exclusively at compile-time. These
```
tgodzik:
```suggestion
Compile-only dependencies, allowing users to include certain libraries exclusively at the time of the compilation. These
```

website/docs/commands/compile.md
```diff
@@ -482,4 +482,22 @@ For example, to exclude all files in the `example/scala` directory, add the foll
...
+dependencies are added to the class path during compilation, but won't be included during runtime.
```
tgodzik:
```suggestion
dependencies are added to the class path during compilation, but won't be included when the application is run.
```

tgodzik (changes requested):

modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala
lwronski: I've looked into it and I didn't find solution how to define compile-only dependencies only for test scope in SBT. 

It might be sufficient to define this dependency as main scope, which would enable exported project to work.

website/docs/commands/compile.md
```diff
@@ -482,4 +482,22 @@ For example, to exclude all files in the `example/scala` directory, add the foll
 
 ```scala title=project.scala
 //> using exclude "example/scala"
+```
+
+## Compile-Only Dependencies
+
+Compile-only dependencies, allowing users to include certain libraries exclusively at the time of the compilation. These
```
MaciejG604:
```suggestion
Compile-only dependencies, allow to include certain libraries exclusively at the time of the compilation. These
```

website/docs/commands/compile.md
```diff
@@ -482,4 +482,22 @@ For example, to exclude all files in the `example/scala` directory, add the foll
...
+dependencies are added to the class path during compilation, but won't be included when the application is run.
+
+To declare a compile-only dependency, you should use the `compileOnly.lib` directive or `--compile-lib` command line
```
MaciejG604, line 492:
```suggestion
To declare a compile-only dependency, use the `compileOnly.lib` directive or `--compile-lib` command line
```

MaciejG604 (approved):

tgodzik (approved):
LGTM!

PR #2249, opened 2023-06-28
Introduce BuildInfo

MaciejG604:
Addresses #2106

As well as moves the wrapper generation to the step when Sources are created from ScopedSources

PR #2310, opened 2023-07-25
Build info with compute version

### LINKED ISSUE #2106: Support for `BuildInfo`
Discussed in https://github.com/VirtusLab/scala-cli/discussions/2051

<sup>Originally posted by **ckipp01** April 23, 2023</sup>
I have a use case where I'm building a CLI app. A lot of times with apps like this you want to make sure you can display the version of the app via a command, and many times the value of that version is persisted utilizing a form of `BuildInfo` where based on git, or something else a file is generated and used for compilation. Is there any way to mimic something like this with `scala-cli`? All the solutions I'm coming up with request running `scala-cli` to generate a file first, and then running it again on your main project to utilize that created file. However, it'd be great if this could just be done in a single step somehow. Any ideas?</div>

We could have a directive/command-line option exposing `BuildInfo` data in the project.
```scala
//> using build.info
```

This could potentially also be delivered through code generators. (see #610)

**Additional context**
See the discussions in #2051 and #610.

MaciejG604:
Follow-up to #2249
Fixes #2106


website/docs/reference/directives.md
```diff
@@ -38,6 +38,17 @@ Adds compiler plugins
 #### Examples
 `//> using plugin org.typelevel:::kind-projector:0.13.2`
 
+### Compute Version
+
+Method used to compute the version for BuildInfo
+
+`//> using platform `git:tag
```
bishabosha: this doesnt seem right

modules/cli/src/main/scala/scala/cli/commands/shared/SourceGeneratorOptions.scala
```diff
@@ -11,6 +11,9 @@ final case class SourceGeneratorOptions(
   @HelpMessage("Generate BuildInfo for project")
   @Name("buildInfo")
     useBuildInfo: Option[Boolean] = None,
+  @HelpMessage("Method used to compute the project version for BuildInfo")
+  @Tag(tags.experimental)
+    computeVersion: Option[String] = None,
```
lwronski: It looks like that this options are already defined in [PublishParamOptions](https://github.com/VirtusLab/scala-cli/blob/dfd53395e98a548bb0171200b41857026a8262e5/modules/cli/src/main/scala/scala/cli/commands/publish/PublishParamsOptions.scala#L38).

Perhaps the version options from `Publish` can be extracted into `VersionSharedOptions` or something like this.

modules/core/src/main/scala/scala/build/errors/BuildInfoGenerationError.scala
```diff
@@ -0,0 +1,6 @@
+package scala.build.errors
+
+import scala.build.errors.BuildException
```
lwronski:
```suggestion
```

modules/core/src/main/scala/scala/build/errors/BuildInfoGenerationError.scala
```diff
@@ -0,0 +1,6 @@
...
+
+final class BuildInfoGenerationError(cause: Exception)
+    extends BuildException(s"BuildInfo generation error: ${cause.getMessage}", cause = cause)
```
lwronski, line 4: It's weird, but when passing `case` into` BuildException`, it seems that the full stack trace of case isn't being printed. 

It you agree that this is bug, it should fixed in the following PR.

modules/directives/src/main/scala/scala/build/preprocessing/directives/ComputeVersion.scala
```diff
@@ -0,0 +1,45 @@
...
+  DirectiveUsage,
+  HasBuildOptions
+}
+import scala.build.errors.BuildException
+import scala.build.options.{BuildOptions, ComputeVersion => cv, SourceGeneratorOptions}
+import scala.cli.commands.SpecificationLevel
+
+@DirectiveGroupName("Compute Version")
+@DirectiveExamples("//> using computeVersion git:tag")
+@DirectiveExamples("//> using computeVersion command:'echo 1.0.0'")
```
lwronski: Perhaps these examples could be improved, for example - reading a version from a file.

```
//> using computeVersion command:'cat project.version'"
```

If you running `echo 1.0.0`, it might not make much sense. It would be better to use the `version` option.

modules/core/src/main/scala/scala/build/errors/BuildInfoGenerationError.scala
lwronski: I realize now that the error is being printed [here](https://github.com/VirtusLab/scala-cli/blob/dfd53395e98a548bb0171200b41857026a8262e5/modules/cli/src/main/scala/scala/cli/ScalaCli.scala#L110).

modules/cli/src/main/scala/scala/cli/commands/shared/ProjectVersionOptions.scala
```diff
@@ -0,0 +1,19 @@
...
+import caseapp.*
+
+import scala.cli.commands.tags
+
+// format: off
+final case class ProjectVersionOptions(
+  @HelpMessage("Method used to compute the project version")
+  @ValueDescription("git|git:tag|command:...")
+  @Tag(tags.restricted)
+    computeVersion: Option[String] = None
```
lwronski: Move the `version` parameter from `PublishParamsOptions` into `ProjectVersionOptions`. I think it's also directly connected to the project version.

modules/cli/src/main/scala/scala/cli/commands/shared/SharedVersionOptions.scala
```diff
@@ -0,0 +1,26 @@
...
+  @HelpMessage("Method used to compute the project version")
+  @ValueDescription("git|git:tag|command:...")
+  @Tag(tags.restricted)
+  @Tag(tags.inShortHelp)
+    computeVersion: Option[String] = None,
+  @Group(HelpGroup.ProjectVersion.toString)
+  @HelpMessage("Set the project version")
+  @Tag(tags.restricted)
+  @Tag(tags.inShortHelp)
+    projectVersion: Option[String] = None
```
lwronski, line 19: Add an alias `version` to keep backward compatibility.

modules/options/src/main/scala/scala/build/options/ComputeVersion.scala
```diff
@@ -19,11 +19,11 @@ object ComputeVersion {
...
     def get(workspace: os.Path): Either[BuildException, String] = {
-      val res = os.proc(command).call(stdin = os.Inherit, cwd = workspace, check = false)
-      if (res.exitCode == 0)
-        Right(res.out.trim(Codec.default))
+      val maybeRes = Try(os.proc(command).call(stdin = os.Inherit, cwd = workspace, check = false))
+      if (maybeRes.isSuccess && maybeRes.get.exitCode == 0)
+        Right(maybeRes.get.out.trim(Codec.default))
       else
-        Left(new Command.ComputeVersionCommandError(command, res.exitCode))
+        Left(new Command.ComputeVersionCommandError(command, maybeRes.map(_.exitCode).getOrElse(1)))
```
lwronski: To avoid running `.get` on `Option`, consider writing it in this way.

```
      maybeRes match {
        case Success(res) if res.exitCode == 0 =>
          Right(res.out.trim(Codec.default))
        case _ =>
          Left(new Command.ComputeVersionCommandError(
            command,
            maybeRes.map(_.exitCode).getOrElse(1)
          ))
      }
```

lwronski (approved):

modules/cli/src/main/scala/scala/cli/commands/shared/SharedVersionOptions.scala
MaciejG604: Can't do that, after refactoring this option into SharedOptions the name had to be changed to avoid clashing with `scala-cli --version` as the Default command has a `version` flag.

lwronski: Right, I hadn't checked that. So, it seems fine to me.

tgodzik (approved):

PR #2311, opened 2023-07-25
Accept directive packageType native when using native platform

### LINKED ISSUE #2303: `packaging.packageType native` directive does not work with `platform native`
**Version(s)**

1.0.2

**Describe the bug**

Giving these two directives in conjunction is rejected, though intuitively it should be valid (IMO):

```
//> using platform native
//> using packaging.packageType native
```

The workaround is easy, simply remove the `packaging.packageType` directive. This is merely a UX issue, I'd say.

**To Reproduce**

This is exactly the example given in the docs currently for the Scala Native section of the `package` command doc, except updated to include directives. My intuition was that I should specify `packageType`:

```
$ cat HelloNative.scala
//> using scala 3.3
//> using platform native

//> using packaging.packageType native
//> using packaging.output hello

object Hello {
    println("Hello")
}

$ scala-cli --power package HelloNative.scala
Compiling project (Scala 3.3.0, Scala Native)
Compiled project (Scala 3.3.0, Scala Native)
[error]  Unsupported package type: Native for Scala Native.
```

**Expected behaviour**

Specifying both platform and packaging type together as native should be allowed without error. Even if redundant perhaps, it should not hinder the user's intention.

lwronski:
Fixes #2303

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -194,43 +194,7 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
           .map(_ => None)
       }
     else {
-      val packageType: PackageType = value {
-        val basePackageTypeOpt = build.options.notForBloopOptions.packageOptions.packageTypeOpt
-        lazy val validPackageScalaJS =
-          Seq(PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
-        lazy val validPackageScalaNative =
-          Seq(PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
-
-        forcedPackageTypeOpt -> build.options.platform.value match {
-          case (Some(forcedPackageType), _) => Right(forcedPackageType)
-          case (_, _) if build.options.notForBloopOptions.packageOptions.isDockerEnabled =>
-            for (basePackageType <- basePackageTypeOpt)
-              Left(new MalformedCliInputError(
-                s"Unsupported package type: $basePackageType for Docker."
-              ))
-            Right(PackageType.Docker)
-          case (_, Platform.JS) =>
-            val validatedPackageType =
-              for (basePackageType <- basePackageTypeOpt)
-                yield
-                  if (validPackageScalaJS.contains(basePackageType)) Right(basePackageType)
-                  else Left(new MalformedCliInputError(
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -194,43 +194,7 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
...
-        val basePackageTypeOpt = build.options.notForBloopOptions.packageOptions.packageTypeOpt
-        lazy val validPackageScalaJS =
-          Seq(PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
-        lazy val validPackageScalaNative =
-          Seq(PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
-
-        forcedPackageTypeOpt -> build.options.platform.value match {
-          case (Some(forcedPackageType), _) => Right(forcedPackageType)
-          case (_, _) if build.options.notForBloopOptions.packageOptions.isDockerEnabled =>
-            for (basePackageType <- basePackageTypeOpt)
```
MaciejG604, line 207: Let's improve this code while we're touching it.
This for loop has no effect, the result of this `case` will always be `Right(PackageType.Docker)`

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -1068,4 +1032,43 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
...
+    val basePackageTypeOpt = build.options.notForBloopOptions.packageOptions.packageTypeOpt
+    lazy val validPackageScalaJS =
+      Seq(PackageType.Js, PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
+    lazy val validPackageScalaNative =
+      Seq(PackageType.Native, PackageType.LibraryJar, PackageType.SourceJar, PackageType.DocJar)
+
+    forcedPackageTypeOpt -> build.options.platform.value match {
+      case (Some(forcedPackageType), _) => Right(forcedPackageType)
+      case (_, _) if build.options.notForBloopOptions.packageOptions.isDockerEnabled =>
+        for (basePackageType <- basePackageTypeOpt)
```
MaciejG604: Let's improve this code while we're touching it.
This `for` loop has no effect, the result of this `case` will always be `Right(PackageType.Docker)`

MaciejG604 (approved):

PR #2313, opened 2023-07-25
Load virtual data as byte arrays without encoding using UTF-8

### LINKED ISSUE #2306: VirtualData corrupts binary files
**Version(s)**
v1.0.2

**Describe the bug**

Loading a gist with binary resources results in the data being corrupted.
I believe right now the resources are being loaded as [`VirtualData`](https://github.com/VirtusLab/scala-cli/blob/5afc7f39c3c76e9a5ee0c4f8882c2fba4eaf84c5/modules/build/src/main/scala/scala/build/input/Element.scala#L37), which later assumes that the content is [UTF-8 encoded](https://github.com/VirtusLab/scala-cli/blob/50815da25f02d53207215f993f461287f3a67722/modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala#L28).

**To Reproduce**

Attempt to run `scala-cli --power --enable-markdown https://gist.github.com/JD557/4bd5b7acd53a08e8a9a306a2bde57a97`. It will fail with `Caused by: java.lang.Exception: Unsupported DIB header size: 12435439`

Downloading the zip and running it locally will work as expected.

**Expected behaviour**

The remote gist should run without corrupted files.

lwronski:
Fixes #2306

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
@@ -25,31 +25,19 @@ case object DataPreprocessor extends Preprocessor {
     input match {
       case file: VirtualData =>
         val res = either {
-          val content = new String(file.content, StandardCharsets.UTF_8)
-          val preprocessedDirectives: PreprocessedDirectives = value {
-            DirectivesPreprocessor.preprocess(
-              content,
-              Left(file.subPath.toString),
-              file.scopePath,
-              logger,
-              allowRestrictedFeatures,
-              suppressWarningOptions,
-              maybeRecoverOnError
-            )
-          }
           val inMemory = Seq(
             PreprocessedSource.InMemory(
               originalPath = Left(file.source),
               relPath = file.subPath,
-              code = content,
+              content = file.content,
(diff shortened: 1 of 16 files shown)


modules/build/src/main/scala/scala/build/preprocessing/DataPreprocessor.scala
```diff
@@ -25,31 +25,19 @@ case object DataPreprocessor extends Preprocessor {
     input match {
       case file: VirtualData =>
         val res = either {
-          val content = new String(file.content, StandardCharsets.UTF_8)
-          val preprocessedDirectives: PreprocessedDirectives = value {
```
MaciejG604, line 29: This is a rather big change in behaviour although for very niche use cases.
I wonder if we should give it more thought.

However, I agree that this part is mostly not needed for this type of Elements.

lwronski: According to our documentation, we only process using directives from Scala sources - https://scala-cli.virtuslab.org/docs/guides/using-directives/. 

It seems it is bug that we process using directives from `VirtualData`.

MaciejG604 (approved):

PR #2317, opened 2023-07-28
Set globally Java properties for ScalaCLI launcher

lwronski:
Enable users to globally set Java properties for ScalaCLI launcher using the `config` command.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
@@ -16,14 +16,7 @@ import scala.cli.commands.publish.ConfigUtil.*
 import scala.cli.commands.shared.HelpGroup
 import scala.cli.commands.util.JvmUtils
 import scala.cli.commands.{ScalaCommand, SpecificationLevel}
-import scala.cli.config.{
-  ConfigDb,
-  Keys,
-  PasswordOption,
-  PublishCredentials,
-  RepositoryCredentials,
-  Secret
-}
+import scala.cli.config._
 import scala.cli.util.ArgHelpers.*
 import scala.cli.util.ConfigDbUtils
 object Config extends ScalaCommand[ConfigOptions] {
@@ -285,6 +278,8 @@ object Config extends ScalaCommand[ConfigOptions] {
                       else
                         values
 
+                    checkIfAskForUpdate(entry, finalValues, db, options)
+
                     db.setFromString(entry, finalValues)
                       .wrapConfigException
                       .orExit(logger)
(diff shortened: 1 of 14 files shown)


modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
```diff
@@ -298,4 +300,32 @@ object Config extends ScalaCommand[ConfigOptions] {
       }
     }
   }
+
+  /** If the new value is the same as the previous value, ask user for confirmation or suggest to
```
MaciejG604: Something's mixed up in this scaladoc - if values are the same then no confirmation nor --force flag is needed, right?

lwronski: Right! Good catch.

modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
```diff
@@ -285,6 +278,15 @@ object Config extends ScalaCommand[ConfigOptions] {
                       else
                         values
 
+                    val previousValue = db.get(entry).wrapConfigException.orExit(logger)
+
+                    confirmUpdatedValue(
```
MaciejG604: I think this makes almost all config writes interactive now (except for `Keys.repositoryCredentials` and `Keys.publishCredentials`). We could limit that to only Java props key, or only to `StringListEntry` keys.

IMO for keys that have just a single value, so our `BooleanEntry`, `PasswordEntry` or `StringEntry`, asking for confirmation is not that useful.
WDYT?

website/docs/under-the-hood.md
```diff
@@ -46,6 +46,11 @@ In order set them, the `-D` command-line flags must be placed as the first optio
 scala-cli -Dfoo1=bar1 -Dfoo2=bar2 run ...
 ```
 
+:::note
+- `scala-cli run . -Dfoo=bar` would pass the java property into your Scala app
+- `scala-cli -Dfoo=bar run .` would pass the java property into `scala-cli.`
```
MaciejG604:
```suggestion
- `scala-cli -Dfoo=bar run .` would pass the java property into `scala-cli`
```

website/docs/under-the-hood.md
```diff
@@ -59,7 +64,15 @@ $ cat .scala-jvmopts
 $ scala-cli run ...
 ```
 
+In the Scala CLI, you can set Java properties globally for the Scala CLI launcher using the `config` command.
```
MaciejG604:
```suggestion
You can set Java properties globally for the Scala CLI launcher using the `config` command.
```

website/docs/under-the-hood.md
```diff
@@ -59,7 +64,15 @@ $ cat .scala-jvmopts
...
+The example belows shows how to set the Java properties `-Djavax.net.ssl.trustStore=cacerts` and `-Dfoo=bar2`:
```
MaciejG604:
```suggestion
The example below shows how to set the Java properties `-Djavax.net.ssl.trustStore=cacerts` and `-Dfoo=bar2`:
```

website/docs/under-the-hood.md
```diff
@@ -59,7 +64,15 @@ $ cat .scala-jvmopts
...
+```
+
 :::note
-- `scala-cli run . -Dfoo=bar` would pass the java property into your Scala app
-- `scala-cli -Dfoo=bar run .` would pass the java property into `scala-cli.`
-:::
\ No newline at end of file
+Please note that if you need to modify the Java properties, you have to redefine all the properties. It's not possible
+to update just a single Java properties via the `config` command. Each update effectively replaces the entire Java properties
+list.
```
MaciejG604:
```suggestion
Please note that if you need to modify the Java properties, you have to redefine all of them. It's not possible
to update just a single value via the `config` command. Each update effectively replaces the entire Java properties list.
```

modules/cli/src/main/scala/scala/cli/commands/config/Config.scala
lwronski: Yes, that sounds like a good idea. so, I limit it only to `StringListEntry`.

website/docs/under-the-hood.md
MaciejG604: I think this is still unresolved

MaciejG604 (approved):

PR #2318, opened 2023-07-28
Resolve semanticDB for older scala version

lwronski:
As `semanticdb-scalac` is published only for the latest four patch versions of Scala `2.12` and `2.13`, we need a mechanism to find the recent supported `semanticdb-scalac` version for older Scala version. This would apply when users are trying to generate a semanticDb for  `2.13.4` Scala  version.

A similar implementation was done in Bloop, which you can see here: https://github.com/scalacenter/bloop/pull/2097.

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
@@ -123,14 +123,32 @@ final case class BuildOptions(
   }
 
   private def semanticDbPlugins: Either[BuildException, Seq[AnyDependency]] = either {
-    val generateSemDbs = scalaOptions.generateSemanticDbs.getOrElse(false) &&
-      value(scalaParams).exists(_.scalaVersion.startsWith("2."))
-    if (generateSemDbs)
-      Seq(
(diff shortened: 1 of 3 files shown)


MaciejG604 (approved):
LGTM

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -123,14 +123,32 @@ final case class BuildOptions(
...
+    */
+  def findSemanticDbVersion(scalaVersion: String): String = {
+    val (_, versions) =
+      finalCache.logger.use {
+        coursier.complete.Complete(finalCache)
+          .withScalaVersion(scalaVersion)
+          .withScalaBinaryVersion(scalaVersion.split('.').take(2).mkString("."))
+          .withInput(s"org.scalameta:semanticdb-scalac_$scalaVersion:")
+          .complete()
+          .unsafeRun()(finalCache.ec)
```
tgodzik, line 148: Can this actually hang? Wouldn't we want to cancel this if taking too long?

lwronski: I added timeout in the following PR - https://github.com/VirtusLab/scala-cli/pull/2322.

PR #2322, opened 2023-07-31
Add timeout for resolving semanticDbVersion

lwronski:


modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -139,14 +140,21 @@ final case class BuildOptions(
...
           .complete()
-          .unsafeRun()(finalCache.ec)
+          .future()(finalCache.ec)
+      }
+
+    val versions =
+      try
+        Await.result(versionsFuture, FiniteDuration(10, "s"))._2
+      catch {
+        case _: Throwable => Nil
```
tgodzik: could you log the issue?

lwronski: Fixed.

tgodzik (approved):

PR #2327, opened 2023-08-03
Add test for actionable diagnostics from compiler

### LINKED ISSUE #2307: Actionable Diagnostics from compiler not visible
**Version(s)**
1.0.2

**Describe the bug**

Actions reported by the compiler are not visible with metals 1.0.0 and scala cli

**To Reproduce**
If I use scala `3.3.2-RC1-bin-20230723-5afe621-NIGHTLY` version, and write the following code in VS Code and metals 1.0.0

```scala
//> using scala 3.3.2-RC1-bin-20230723-5afe621-NIGHTLY
def foo(): Int = 23

def test: Int = foo
//              ^^^ error: missing parentheses
```

I'll get a compile error, however I'll have no "quick fixes" to turn `foo` => `foo()`

but if I run the same with sbt 1.9.2, (and sbt bloopInstall) then I'll have quick fixes.

**Expected behaviour**
I should have the same quick fixes as with bloop + sbt

MaciejG604:


modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -1523,6 +1523,101 @@ abstract class BspTestDefinitions(val scalaVersionOpt: Option[String])
...
+              |         },
+              |         "newText": "()"
+              |       }
+              |     ]
+              |   }
+              |}
+              |]
+              |}""".stripMargin
+              .replaceAll("\\s", "")
+              .replaceAll("_", " ")
```
lwronski: Can you try to deserialize this JSON object into `bsp4j.ScalaDiagnostic` as [here](https://github.com/VirtusLab/scala-cli/blob/5dd3559719e9c54498061b40b0a8055764fcf709/modules/build/src/test/scala/scala/build/tests/BuildTests.scala#L378)?. You can use Gson for the deserialization.

lwronski (approved):

PR #2336, opened 2023-08-07
1.0.3 release notes

lwronski:


website/docs/release_notes.md
```diff
@@ -7,6 +7,103 @@ import ReactPlayer from 'react-player'
...
+
+Scala CLI allows users to globally set Java properties for its launcher using the `config` command. This will simplify the JVM properties management process, eliminating the need to pass these properties with each `scala-cli` execution.
+
+To set global Java properties execute the following command:
+
+```
+scala-cli config java.properties Djavax.net.ssl.trustStore=cacerts Dfoo=bar2
+```
+
+When modifying Java properties, remember that you must redefine all of them. It's not possible to update just a single property. Essentially, each time you use the `config` command for Java properties, you replace the entire list of properties.
```
tgodzik, line 58: Also add something along the lines of:

```
Whenever overwriting existing properties Scala CLI will let you know what was the previous value and in interactive mode ensure that you are ok with replacing them
```

tgodzik (changes requested):

website/docs/release_notes.md
lwronski: It's make sense. I updated release notes.

tgodzik (approved):
LGTM!

PR #2350, opened 2023-08-16
Remove ComputeVersion.Command, make ComputeVersion classes positioned

### LINKED ISSUE #2106: Support for `BuildInfo`

MaciejG604:
Remove ComputeVersion.Command since it may be unsafe.
Add positioning to ComputeVersion instances to be able to point out the erroneous directive.
+ Documentation for computing version from #2310 (also for publishing)

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/ComputeVersion.scala
+++ b/modules/options/src/main/scala/scala/build/options/ComputeVersion.scala
@@ -18,34 +18,6 @@ sealed abstract class ComputeVersion extends Product with Serializable {
 }
 
 object ComputeVersion {
-
-  final case class Command(command: Seq[String], positions: Seq[Position]) extends ComputeVersion {
-    def get(workspace: os.Path): Either[BuildException, String] = {
-      val maybeRes = Try(os.proc(command).call(stdin = os.Inherit, cwd = workspace, check = false))
-      maybeRes match {
-        case Success(res) if res.exitCode == 0 =>
-          Right(res.out.trim(Codec.default))
-        case _ =>
-          Left(new Command.ComputeVersionCommandError(
-            command,
-            maybeRes.map(_.exitCode).getOrElse(1),
-            positions
-          ))
-      }
-    }
-  }
-
-  object Command {
-    final class ComputeVersionCommandError(
-      command: Seq[String],
-      exitCode: Int,
-      positions: Seq[Position]
-    ) extends BuildException(
-          s"Error running command ${command.mkString(" ")} (exit code: $exitCode)",
-          positions = positions
-        )
-  }
-
   final case class GitTag(
     repo: os.FilePath,
     dynVer: Boolean,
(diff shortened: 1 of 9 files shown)


modules/options/src/main/scala/scala/build/options/ComputeVersion.scala
```diff
@@ -5,26 +5,29 @@ import com.github.plokhotnyuk.jsoniter_scala.macros.*
...
 sealed abstract class ComputeVersion extends Product with Serializable {
   def get(workspace: os.Path): Either[BuildException, String]
+
+  val positions: Seq[Position]
 }
 
 object ComputeVersion {
 
-  final case class Command(command: Seq[String]) extends ComputeVersion {
+  final case class Command(positions: Seq[Position], command: Seq[String]) extends ComputeVersion {
```
lwronski:
```suggestion
final case class Command(command: Seq[String], positions: Seq[Position] = Nil) extends ComputeVersion {
```

Maybe it is worth adding positions as the last argument, so in some cases, we don't have to pass `Nil` explicitly. This suggestion refers to every class that extends from `ComputeVersion`.

website/docs/commands/publishing/publish.md
```diff
@@ -81,6 +81,17 @@ To override this default value, set the `publish.computeVersion` directive, like
...
 ```
 
+Please note that only tags that follow the semantic versioning are taken into consideration.
+
+Values available for project version configuration are:
+- `git:tag` or `git`: use the latest stable git tag, if it is older than HEAD then try to increment it
+  and add a suffix `-SNAPSHOT`, if no tag is available then use `0.1.0-SNAPSHOT`
+- `git:dynver`: use the latest (stable or unstable) git tag, if it is older than HEAD then use the output of
+  `-{distance from last tag}-g{shortened version of HEAD commit hash}-SNAPSHOT`, if no tag is available then use `0.1.0-SNAPSHOT`
+- `command:cmd`: use the output of the command `cmd` as version
```
lwronski:
```suggestion
- `command:cmd`: use the output of the command `cmd` as version. It runs the cmd in your shell under the hood. This is not the recommended way since the downloaded script can contain any commands
```

modules/options/src/main/scala/scala/build/options/ComputeVersion.scala
MaciejG604: I placed the `positions` at the end of the argument list, since it is more logical this way, but I'd opt for keeping the positioning explicit with no default values. `positions = Nil` is not a typical case, rather everywhere there should be either Poisition.File or Position.CommandLine.

lwronski (approved):

PR #2356, opened 2023-08-21
Read java props from env vars

### LINKED ISSUE #2354: JAVA_OPTS / JDK_JAVA_OPTIONS not respected
**Version(s)**
1.0.4

**Describe the bug**
I have my Java proxies set in `JAVA_OPTS` and `JDK_JAVA_OPTIONS`, but scala-cli seems to ignore or reset them.  I had to put the arguments on the scala-cli command line for them to work.

I first used scala-cli through the metals project, and it (metals) failed because scala-cli couldn't download any files.  In cases like this, where the user is not even aware that they are using scala-cli, it would be nice if scala-cli would use the standard environment variables for proxies.

I'm not really sure if this is a bug or just a feature request... I think it depends on whether the current behavior is intentional.

**To Reproduce**
`env JAVA_OPTS='-Dhttp.proxy=4.4.4.4 ...' JDK_JAVA_OPTIONS='-Dhttp.proxy=4.4.4.4 ...' scala-cli`

**Expected behaviour**
I expected the proxies in `JAVA_OPTS` or `JDK_JAVA_OPTIONS` to be used.

MaciejG604:
Fixes #2354

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -846,6 +846,11 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
     def runClasspath = T {
       super.runClasspath() ++ Seq(localRepoJar())
     }
+
+    // Required by the reflection usage in modules/cli/src/test/scala/cli/tests/SetupScalaCLITests.scala
+    override def forkArgs: T[Seq[String]] = T {
+      super.forkArgs() ++ Seq("--add-opens=java.base/java.util=ALL-UNNAMED")
+    }
   }
 }
 
--- a/modules/cli/src/test/scala/cli/tests/SetupScalaCLITests.scala
+++ b/modules/cli/src/test/scala/cli/tests/SetupScalaCLITests.scala
@@ -2,12 +2,12 @@ package cli.tests
 
 import com.eed3si9n.expecty.Expecty.expect
 
-import java.util.Properties
+import java.util.{Collections, Properties}
 
 import scala.build.internal.Constants
 import scala.build.tests.TestInputs
 import scala.cli.ScalaCli
-import scala.util.Properties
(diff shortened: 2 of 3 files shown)


modules/cli/src/test/scala/cli/tests/SetupScalaCLITests.scala
```diff
@@ -32,4 +32,36 @@ class SetupScalaCLITests extends munit.FunSuite {
...
+          field.setAccessible(true)
+          val obj = field.get(env)
+          val map = obj.asInstanceOf[java.util.Map[String, String]]
+          map.clear()
+          map.putAll(newEnv.asJava)
+        }
+    }
+
+    val javaOptsValues       = "  -Xmx2048m   -Dhttp.proxy=4.4.4.4   -Xss8m"
+    val jdkJavaOptionsValues = " -Xmx1024m -Dscala-cli=true  -Xss4m"
```
lwronski: I suggest removing the `Xmx` option in the tests. Using it might lead to misconceptions, making one believe that scala-cli supports `Xmx` options, when it doesn't.

build.sc
```diff
@@ -846,6 +846,10 @@ trait Cli extends SbtModule with ProtoBuildModule with CliLaunchers
     def runClasspath = T {
       super.runClasspath() ++ Seq(localRepoJar())
     }
+
+    override def forkArgs: T[Seq[String]] = T {
+      super.forkArgs() ++ Seq("--add-opens=java.base/java.util=ALL-UNNAMED")
```
lwronski, line 852: Add an explanation of why this option is needed.

lwronski (approved):

PR #2363, opened 2023-08-28
Add support for repository credentials

### LINKED ISSUE #1629: scala-cli with private gitlab repos
Sadly I can not use scala-cli as I need to pull/push artifacts against private, self-hosted gitlab repos. In the case of using SBT there's a plugin available [sbt-gitlab](https://github.com/gilcloud/sbt-gitlab) which set the needed **custom http header for authentication** with gitlab. I think the best solution would be to patch coursier as we could get rid of the plugin for SBT and at the same time we could use scala-cli.

MaciejG604:
Fixes #1629


MaciejG604:
After consideration Scala CLI will not support any credential interpolation in the repository addresses. Considering that one could then embed reading of arbitrary user's env vars or files, which later would be sent to an unverified web address it has been deemed unsafe.

PR #2364, opened 2023-08-28
Add decoding of class names before showing them to the user with interactive

### LINKED ISSUE #2362: Pretty main names for markdown scripts/interactive mode

When writing tutorials in Github Gists, one of the simplest ways to make sure that the examples are ordered is to prefix each markdown with a number (e.g. `01-intro.md`, `02-first-example.md`, ...).

To allow users to quickly see the examples, it's nice that they can just call `scala-cli --power --enable-markdown --interactive <gist_url>` (OK, I guess it would be nicer without all those flags 😅 ) and get started.

However, they are then greeted with main class names such as `02$minusfirst$minusexample_md`, which is not a best experience.

Here's a concrete example of a gist in that format https://gist.github.com/JD557/8778cbb4a7ed3ad7eda5bffc72fdfbcf

Either:
 - Have the interactive mode demangle the class name
 - Add a directive to manually set the exported main class name (at least I couldn't find one)

I considered manually defining the main class, but this does not play well with literate programming. Namely, I cannot write markdown inside the class, only Scala comments.

MaciejG604:
Fixes #2362

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -17,7 +17,7 @@ import scala.build.errors.*
 import scala.build.input.VirtualScript.VirtualScriptNameRegex
 import scala.build.input.*
 import scala.build.internal.resource.ResourceMapper
-import scala.build.internal.{Constants, MainClass, Util}
+import scala.build.internal.{Constants, MainClass, Name, Util}
 import scala.build.options.ScalaVersionUtil.asVersion
 import scala.build.options.*
 import scala.build.options.validation.ValidationException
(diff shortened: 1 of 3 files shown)


MaciejG604:
I've spotted a similar UX problem when there's a standard `.scala` containing for example a class defined as so:
```
object `Main-2` extends App { println("Hello World") }
```
This main class would have to be specified using its encoded name:
`scala-cli run . --main-class Main$minus2`
This won't work:
`scala-cli run . --main-class Main-2`

@tgodzik do you think we should automatically encode class names passed with --main-class?

tgodzik:
> do you think we should automatically encode class names passed with --main-class?

Makes sense or as a fallback when none is found.

tgodzik (approved):
Looks good! The tests seem to be failing on windows, could that be related to this PR? (I don't see hoow :thinking: )

PR #2370, opened 2023-08-30
Pin the versions of Github CI runners

MaciejG604:


tgodzik:
Looks like that's what we should do, I talked with Wojciech and he said it's safer to have a stable image. You can make it a normal PR

MaciejG604:
Okay, I've made the PR and commits merge-ready, also added pinned version of the macOS runner [(relevant Github docs)](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)

tgodzik (approved):
Thanks!

PR #2375, opened 2023-09-04
Documentation updates

### LINKED ISSUE #1629: scala-cli with private gitlab repos

MaciejG604:
Repositories credentials documentation and tests for the interaction with Coursier API:
#1629

Java properties documentation:
#2356
#2317
#2267


modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -1628,4 +1628,161 @@ abstract class RunTestDefinitions(val scalaVersionOpt: Option[String])
...
+        testOrg,
+        "--name",
+        testName,
+        "--project-version",
+        testVersion
+      )
+        .call(cwd = root, stdin = os.Inherit, stdout = os.Inherit, env = extraEnv)
+
+      TestUtil.serveFilesInHttpServer(repoPath, user, password, realm) { (host, port) =>
+        // test("No repository credentials passed to coursier")
```
tgodzik: Why is this commented out actually? That's the standard in the tests?

website/docs/guides/repositories.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Repositories and HTTP Proxies ⚡️
+sidebar_position: 51
+---
+
+Scala CLI downloads the dependencies declared in your projects using Coursier.
```
tgodzik:
```suggestion
Scala CLI downloads the dependencies declared in your projects using [Coursier](https://get-coursier.io/).
```

tgodzik (approved):
Looks good! Just two comments

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
MaciejG604: I wanted to split the test cases for readability but that would require starting the HTTP server and publishing to it 3 times and I wanted to avoid that. We can't do nested tests so I've left those comments to indicate that each of the code blocks represent a test case.

I'll update the comments so there's no doubt they are just descriptions.

PR #2399, opened 2023-09-18
Accept --power from anywhere

MaciejG604:
With those changes, invocations like:
`scala-cli run . --power` or `scala-cli run --power .`
will work


modules/cli/src/main/scala/scala/cli/launcher/PowerOptions.scala
```diff
@@ -0,0 +1,21 @@
+package scala.cli.launcher
+
+import caseapp.*
+
+import scala.cli.commands.shared.HelpGroup
+import scala.cli.commands.tags
+
+// format: off
```
Gedochao: no need to turn formatting off here, IMO

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -230,9 +230,23 @@ object ScalaCli {
...
+            if launcherOpts.powerOptions.power then
               isSipScala = false
-            args0.toArray
+              args0.toArray
+            else
+              // .ignoreUnrecognized drops '--'
+              val programArgs = args0.dropWhile(_ != "--")
+              val optionArgs  = args0.takeWhile(_ != "--")
+              // Parse again to consume --power at any position
+              PowerOptions.parser.ignoreUnrecognized.parse(optionArgs) match {
```
Gedochao: would it be possible to put `--power` into `GlobalOptions`, so that we don't need to parse for it a third, dedicated time?

modules/cli/src/main/scala/scala/cli/launcher/PowerOptions.scala
MaciejG604: Actually `format: off` helps in our option classes since fmt does not indent the option name further thatn annotations and I think it's hard to read with many options:
preferred:
```
case class PowerOptions(
   @Group(HelpGroup.Launcher.toString)
   @HelpMessage("When called as 'scala', allow to use power commands too")
   @Tag(tags.must)
     power: Boolean = false
 )
```

fmt format:
```
case class PowerOptions(
   @Group(HelpGroup.Launcher.toString)
   @HelpMessage("When called as 'scala', allow to use power commands too")
   @Tag(tags.must)
   power: Boolean = false
 )
```

Gedochao: it's been agreed that `format:off` shouldn't be used in new options definitions.
we should clean up the old ones.
sure, you can indent manually, but there's no validation if it's consistent across files (or even if it's sane indentation).

modules/integration/src/test/scala/scala/cli/integration/ConfigTests.scala
```diff
@@ -16,7 +16,8 @@ class ConfigTests extends ScalaCliSuite {
     val name       = "Alex"
     TestInputs.empty.fromRoot { root =>
       val before =
-        os.proc(TestUtil.cli, "--power", "config", "publish.user.name").call(
+        // Test --power placed after subcommand name
```
Gedochao, line 19: it'd be good to have a dedicated test for this in `SipScalaTests`

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
Gedochao: Yeah, I can see that although it'd be possible, we'd have to at the very least do some work in `case-app` to make it easier to modify help messages from inside of a command's code.

I suppose we'll have to accept the extra, third parsing pass for now.

modules/cli/src/main/scala/scala/cli/launcher/PowerOptions.scala
```diff
@@ -0,0 +1,18 @@
+package scala.cli.launcher
+
+import caseapp.*
+
+import scala.cli.commands.shared.HelpGroup
+import scala.cli.commands.tags
+
+case class PowerOptions(
+  @Group(HelpGroup.Launcher.toString)
+  @HelpMessage("When called as 'scala', allow to use power commands too")
```
Gedochao:
```suggestion
  @HelpMessage("Allows to use restricted & experimental features")
```

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
Gedochao: How about we had both - `--power` in `GlobalOptions` to make it available from subcommands and prevent option alias clashes, as well as an extra parsing of it here?

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -230,9 +230,21 @@ object ScalaCli {
...
+            else
+              // Parse again to register --power at any position
+              // Don't consume it, GlobalOptions parsing will do it
+              PowerOptions.parser.ignoreUnrecognized.parse(args0) match {
+                case Right((powerOptions, _)) =>
+                  if powerOptions.power then
+                    isSipScala = false
+                  args0.toArray
+                case Left(e) =>
+                  System.err.println(e.message)
```
Gedochao, line 245: This shouldn't ever happen, no? since we're ignoring unrecognised.
it's fine to leave this as a failsafe, just mentioning

modules/cli/src/main/scala/scala/cli/launcher/PowerOptions.scala
Gedochao: bump, since we'll now show `--power` in `--help` outputs

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -1,5 +1,6 @@
 package scala.cli.integration
 
+import com.eed3si9n.expecty.Expecty
 import com.eed3si9n.expecty.Expecty.expect
```
Gedochao: NIT or just
```suggestion
import com.eed3si9n.expecty.Expecty._
```

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
MaciejG604: I think so, there's no error to be returned really

Gedochao (approved):

PR #2404, opened 2023-09-19
Totally offline mode

MaciejG604:
Introduce `--offline` flag, which changes Coursier's cache policy to LocalOnly. This way Scala CLI won't download anything, but instead will fail with a message similiar to:
```
[error] ./WithDeps.scala:2:15
[error] Error downloading com.lihaoyi:os-lib_3:0.9.1
[error]   not found: .../ivys/ivy.xml
[error]   not found: ...cache/https/repo1.maven.org/maven2/com/lihaoyi/os-lib_3/0.9.1/os-lib_3-0.9.1.pom
[error]   not found: ...Caches/ScalaCli/local-repo/1.0.4-27-7b90be-DIRTY9091f1a9/com.lihaoyi/os-lib_3/0.9.1/ivys/ivy.xml
[error]   No fallback URL found
[error] //> using dep com.lihaoyi::os-lib:0.9.1
[error]               ^^^^^^^^^^^^^^^^^^^^^^^^^
```

Added support for `COURSIER_MODE=offline` env var and `coursier.mode=offline` java prop (https://github.com/coursier/coursier/issues/688)

### DIFF
--- a/modules/build/src/main/scala/scala/build/compiler/BloopCompilerMaker.scala
+++ b/modules/build/src/main/scala/scala/build/compiler/BloopCompilerMaker.scala
@@ -4,20 +4,24 @@ import bloop.rifle.{BloopRifleConfig, BloopServer, BloopThreads}
 import ch.epfl.scala.bsp4j.BuildClient
 
 import scala.build.Logger
+import scala.build.errors.{FetchingDependenciesError, Severity}
 import scala.build.internal.Constants
+import scala.build.internal.util.WarningMessages
 import scala.concurrent.duration.DurationInt
+import scala.util.Try
 
 final class BloopCompilerMaker(
   config: BloopRifleConfig,
   threads: BloopThreads,
-  strictBloopJsonCheck: Boolean
+  strictBloopJsonCheck: Boolean,
+  offline: Boolean
 ) extends ScalaCompilerMaker {
   def create(
     workspace: os.Path,
     classesDir: os.Path,
     buildClient: BuildClient,
     logger: Logger
-  ): BloopCompiler = {
+  ): ScalaCompiler = {
     val createBuildServer =
       () =>
         BloopServer.buildServer(
@@ -30,6 +34,23 @@ final class BloopCompilerMaker(
           threads,
           logger.bloopRifleLogger
         )
-    new BloopCompiler(createBuildServer, 20.seconds, strictBloopJsonCheck)
+
+    Try(new BloopCompiler(createBuildServer, 20.seconds, strictBloopJsonCheck))
+      .toEither
+      .left.flatMap {
+        case e if offline =>
+          e.getCause match
+            case _: FetchingDependenciesError =>
+              logger.diagnostic(
+                WarningMessages.offlineModeBloopNotFound,
+                Severity.Warning
+              )
+              Right(
+                SimpleScalaCompilerMaker("java", Nil)
+                  .create(workspace, classesDir, buildClient, logger)
+              )
+            case _ => Left(e)
+        case e => Left(e)
+      }.fold(t => throw t, identity)
   }
 }
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/CoursierOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/CoursierOptions.scala
@@ -47,11 +47,15 @@ final case class CoursierOptions(
       baseCache = baseCache.withTtl(ttl0)
     for (loc <- cache.filter(_.trim.nonEmpty))
       baseCache = baseCache.withLocation(loc)
-    for (isOffline <- offline if isOffline)
+    for (isOffline <- getOffline() if isOffline)
       baseCache = baseCache.withCachePolicies(Seq(CachePolicy.LocalOnly))
 
     baseCache
   }
+
+  def getOffline(): Option[Boolean] = offline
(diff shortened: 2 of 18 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/CoursierOptions.scala
```diff
@@ -26,7 +26,12 @@ final case class CoursierOptions(
   @HelpMessage("Enable checksum validation of artifacts downloaded by coursier")
   @Tag(tags.implementation)
   @Hidden
-    coursierValidateChecksums: Option[Boolean] = None
+    coursierValidateChecksums: Option[Boolean] = None,
+
+  @Group(HelpGroup.Dependency.toString)
+  @HelpMessage("Disable using the network to download artifacts, use only the cache")
```
Gedochao:
```suggestion
  @HelpMessage("Disable using the network to download artifacts, use the local cache only")
```

modules/build/src/main/scala/scala/build/compiler/BloopCompilerMaker.scala
```diff
@@ -30,6 +33,20 @@ final class BloopCompilerMaker(
...
         )
-    new BloopCompiler(createBuildServer, 20.seconds, strictBloopJsonCheck)
+
+    Try(new BloopCompiler(createBuildServer, 20.seconds, strictBloopJsonCheck))
+      .toEither
+      .left.flatMap {
+        case e if offline =>
+          e.getCause match
+            case _: FetchingDependenciesError =>
+              logger.debug("Couldn't fetch Bloop from cache, fallback to using scalac")
```
Gedochao: I wonder if it wouldn't be better to print this as a warning, instead of a debug message... I'd want to know if `scalac` was used instead of `bloop` implicitly in my build.
Also, I'd be explicit that this change happens because `--offline` was passed.

So something along the lines of:
```suggestion
              logger.message("Warning: Bloop could not be fetched from the local cache, as offline mode is on, using scalac as fallback")
```

modules/cli/src/main/scala/scala/cli/commands/shared/CoursierOptions.scala
```diff
@@ -42,8 +47,15 @@ final case class CoursierOptions(
       baseCache = baseCache.withTtl(ttl0)
     for (loc <- cache.filter(_.trim.nonEmpty))
       baseCache = baseCache.withLocation(loc)
+    for (offlineTrue <- getOffline().filter(identity))
```
Gedochao: NIT
```suggestion
    for (isOffline <- getOffline().filter(identity))
```

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -536,18 +541,25 @@ final case class SharedOptions(
...
+      bloopRifleConfig() match {
+        case Right(config) =>
+          Right(new BloopCompilerMaker(
+            config,
+            threads.bloop,
+            strictBloopJsonCheckOrDefault,
+            coursier.getOffline().getOrElse(false)
+          ))
+        case Left(ex) if coursier.getOffline().contains(true) =>
+          logger.debug("Couldn't fetch a JVM for Bloop from cache, fallback to using scalac")
```
Gedochao: Similarly, I'd promote this to a warning.
```suggestion
          logger.message("Warning: JVM for Bloop could not be fetched from the local cache; offline mode is on, so using scalac as fallback")
```

modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
```diff
@@ -188,4 +190,41 @@ object Update extends ScalaCommand[UpdateOptions] {
 
     classesDir.contains(binRepoDir)
   }
+
+  // from https://github.com/coursier/coursier/blob/7b7c2c312aea26e850f0cd2cf15e688d0777f819/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala#L489-L497
+  private def closeConn(conn: URLConnection): Unit = {
```
Gedochao, line 195: NIT: we could extract this to a util object or an extension on `URLConnection`
```suggestion
  private def closeConnection(conn: URLConnection): Unit = {
```

modules/cli/src/main/scala/scala/cli/commands/util/JvmUtils.scala
```diff
@@ -67,38 +67,43 @@ object JvmUtils {
...
-        try javaHomeManager.get(jvmId).unsafeRun()
-        catch {
-          case NonFatal(e) => throw new Exception(e)
-        }
-      }
-      os.Path(path)
+
+    val javaHomePathOrError = Try(javaHomeManager.get(jvmId).unsafeRun()) match {
+      case Success(path) => Right(path)
+      case Failure(e)    => Left(JvmDownloadError(jvmId, e))
```
Gedochao, line 80: Great improvement, this ❤️

modules/cli/src/main/scala/scala/cli/commands/update/Update.scala
MaciejG604: I'd say there's no other need to download stuff outside of coursier, so keeping this private in Update is a reasonable choice for now .

Gedochao (approved):

PR #2405, opened 2023-09-19
Fix default scala version in docs

MaciejG604:
Not caught in #2392 :(


MaciejG604:
Changes added to #2396

PR #2407, opened 2023-09-20
Remove bloop timeouts in tests

MaciejG604:
I've tested this  and it seems to make the tests less flaky on the CI.


tgodzik (approved):

lwronski (approved):
I think we can merge these changes. However, I'm uncertain if it will address the issue. By removing the extra options, we back to the default timeout, which is lower than what was set in these options.

PR #2411, opened 2023-09-21
Make explicitly passed scala version use the latest release, not the default one

### LINKED ISSUE #2394: Default Scala versions are out of date
I noticed today that even though Scala 2.13.12 and Scala 3.3.1 are both out now, `scala-cli -S 2` still runs 2.13.11, and `scala-cli -S 3` still runs 3.3.0.

Is there some way we can expedite updating that when new Scala versions come out?

As the Scala 2 release manager, I'd be happy to add this to the [Scala 2 release steps](https://github.com/scala/scala-dev/blob/scala-dev/.github/ISSUE_TEMPLATE/release.md) if action from our end is needed.

One possibility would be simply be for me to open an issue in this repo whenever there's a new Scala 2 release, to remind y'all to do something. But I could also PR a change directly if you point me in the right direction.

Pardon if we've discussed this before and I'm not remembering. but let's not depend on my memory, let's enshrine something in the release steps. (And perhaps @Kordyjan could add something to the Scala 3 release steps.)

MaciejG604:
E.g. make `-S 2.13` expand to `2.13.12` even though our default for now is `2.13.11`
Fixes #2394

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/BuildOptionsTests.scala
@@ -1,8 +1,9 @@
 package scala.build.tests
 
-import com.eed3si9n.expecty.Expecty.{assert => expect}
+import com.eed3si9n.expecty.Expecty.assert as expect
 import coursier.Repositories
 import coursier.cache.FileCache
+import coursier.core.Version
 import dependency.ScalaParameters
 
 import scala.build.Ops.*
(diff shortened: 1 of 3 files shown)


MaciejG604:
Tests run: https://github.com/VirtusLab/scala-cli/actions/runs/6258371552/job/16992276646?pr=2411#step:8:252

SethTisue (approved):
This felt weird to me at first. But after some thought, I think I've convinced myself it's actually best. We don't want the default behavior to require doing a remote query (which might take time, or be blocked by a firewall or a container setting or what have you), and we also don't want someone's Scala version to suddenly change on them unless they either explicitly requested dynamic behavior (e.g. `-S 2` or the like) or they upgraded scala-cli itself.

cc @lrytz

lrytz:
@SethTisue I don't really follow; you say "we don't want the default behavior to require doing a remote query", but IIUC this PR does that, it will query to find out what's the latest release?

SethTisue:
> @SethTisue I don't really follow; you say "we don't want the default behavior to require doing a remote query", but IIUC this PR does that, it will query to find out what's the latest release?

I meant what happens if you don't `-S` at all.

lwronski (approved):
LGTM

PR #2432, opened 2023-09-28
Warn about transitive using file directive

### LINKED ISSUE #2425: "using file" directive fails with transitive dependencies
**Version(s)**

1.0.4

**Describe the bug**

`using file ...` directive allows specifying file dependencies but transitive dependencies do not work, for example:

- Compiling `A` fails when `A`  depends on `B` which depends on `C` 
- Compiling `A` works when `A` depends on `B` and `C` (this is not used), `B` depends on `C`.

**To Reproduce**

- Run `scala-cli A.scala`, which throws:

  ```
  [error] ./B.scala:6:5
  [error] Not found: C
  [error]     C.funC()
  [error]     ^
  ```

- `A.scala`:

  ```scala
  //> using file B.scala
  
  object A {
      println("Fun A")
      B.funB()
    }
  }
  ```

- `B.scala`:

  ```scala
  //> using file C.scala
  
  object B {
    def funB(): Unit = {
      println("Fun B")
      C.funC()
    }
  }
  ```

- `C.scala`:

  ```scala
  object C {
    def funC(): Unit = println("Fun C")
  }
  ```

**Expected behaviour**

`scala-cli A.scala` should work, still, we need to add `//> using file C.scala` to `A.scala` to get this example working.

MaciejG604:
Documents #2425

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -1,6 +1,7 @@
 package scala.build
 
 import java.io.File
+
 import scala.build.CollectionOps.*
 import scala.build.EitherCps.{either, value}
 import scala.build.Ops.*
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -1,7 +1,6 @@
 package scala.build
 
 import java.io.File
-
 import scala.build.CollectionOps.*
 import scala.build.EitherCps.{either, value}
 import scala.build.Ops.*
@@ -10,12 +9,13 @@ import scala.build.errors.{
   BuildException,
   CompositeBuildException,
   ExcludeDefinitionError,
-  MalformedDirectiveError
+  MalformedDirectiveError,
+  Severity
 }
 import scala.build.input.ElementsUtils.*
 import scala.build.input.*
 import scala.build.internal.Constants
-import scala.build.internal.util.RegexUtils
+import scala.build.internal.util.{RegexUtils, WarningMessages}
 import scala.build.options.{
   BuildOptions,
   BuildRequirements,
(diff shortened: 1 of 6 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -462,4 +466,21 @@ object CrossSources {
...
+  def warnAboutChainedUsingFileDirectives(
+    sourcesAddedWithDirectives: Seq[PreprocessedSource],
+    logger: Logger
+  ): Unit = for {
+    additionalSource           <- sourcesAddedWithDirectives
+    buildOptions               <- additionalSource.options
+    transitiveAdditionalSource <- buildOptions.internal.extraSourceFiles
+  } do
+    logger.diagnostic(
+      WarningMessages.chainingUsingFileDirective,
```
Gedochao, line 482: NIT: the warning string could potentially include the name for the `transitiveAdditionalSource` that'd be skipped

Gedochao (approved):
LGTM, commented on potential minor improvements

modules/build/src/main/scala/scala/build/CrossSources.scala
MaciejG604: That's a good suggestion, but I think passing the position parameter to the logger makes it explicit already, see the test assertions.

Gedochao: would it list all the transitive dependencies individually like this?
if so, then y're right, that should be sufficient already

MaciejG604: Will look like this:
```
[warn] .../Bar.scala:1:16
[warn] Chaining the 'using file' directive is not supported, the source won't be included in the build.
[warn] //> using file Foo.scala
[warn]                ^^^^^^^^^
[warn] .../Bar.scala:2:16
[warn] Chaining the 'using file' directive is not supported, the source won't be included in the build.
[warn] //> using file Bar.scala
[warn]                ^^^^^^^^^
```

Gedochao: yeah, that's fine then

PR #2435, opened 2023-10-03
Add more configuration for publish

MaciejG604:
Since 0.1.5 coursier-publish allows for retrying of sonatype's staging repository operations, the number of retries can be configured.
Since 0.1.6 it allows for setting timeouts for uploading files.
Those options could not be configured from Scala CLI, this PR fixes that.

Added for uploading:
- connection timeout - SAME OPTION USED IN BOTH
- the number of retries on connection timeout
- timeout when waiting for server response - should be useful when uploading large packages

Added for stagin repo:
- connection timeout - SAME OPTION USED IN BOTH
- number of retries on failure
- base wait time between the retries

New configuration options reside in PublishConnectionOptions.scala


modules/cli/src/main/scala/scala/cli/commands/publish/PublishConnectionOptions.scala
```diff
@@ -0,0 +1,45 @@
...
+import caseapp.*
+
+import scala.cli.commands.shared.{HelpGroup, SharedVersionOptions}
+import scala.cli.commands.tags
+
+// format: off
+final case class PublishConnectionOptions(
+  @Group(HelpGroup.Publishing.toString)
+  @HelpMessage("Connection timeout, in seconds.")
+  @Tag(tags.restricted)
```
Gedochao, line 12: Shouldn't all publish-related options be experimental?
```suggestion
  @Tag(tags.experimental)
```

MaciejG604: None of them are and I gues it's to not dispaly more `experimental` warnings than necessary.
Only warnings about using the command and all publish-related using directives are displayed right now.
Should we change it?

Gedochao: tbh it feels like we should, if only to be consistent...
doesn't make sense to have `restricted` options for an `experimental` command

MaciejG604: Let's get into it in another PR

Gedochao (approved):

PR #2445, opened 2023-10-04
Fix 'JVM too old' as bsp

### LINKED ISSUE #2412: Constant recompilation and error when you have a mismatched JVM
**Version(s)**
1.0.4

**Describe the bug**
So this happened when in my code I decided to make the JVM version explicit. I was using 17 and then added the following using directive to my build:

```
//> using jvm 19
```

I didn't notice this until I had an error and saw it constantly disappearing and repairing.

When you look in the Metals logs you can finally see:

```
2023.09.21 11:11:39 INFO  BSP server: [warn] [current bloop jvm] /Users/ckipp/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17%252B35/OpenJDK17-jdk_aarch64_mac_hotspot_17_35.tar.gz/jdk-17+35/Contents/Home
2023.09.21 11:11:39 INFO  BSP server: [warn]                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2023.09.21 11:11:39 INFO  BSP server: [warn]  JVM that is hosting bloop is older than the requested runtime. Please run command `bloop exit`, and then use `--jvm` flag to restart Bloop
2023.09.21 11:11:39 INFO  BSP server: [warn] COMMAND_LINE
2023.09.21 11:11:39 INFO  BSP server: [warn] [current bloop jvm] /Users/ckipp/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17%252B35/OpenJDK17-jdk_aarch64_mac_hotspot_17_35.tar.gz/jdk-17+35/Contents/Home
2023.09.21 11:11:39 INFO  BSP server: [warn]                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2023.09.21 11:11:40 INFO  time: indexed workspace in 0.73s
2023.09.21 11:11:40 INFO  Disconnecting from scala-cli session...
2023.09.21 11:11:40 INFO  Shut down connection with build server.
2023.09.21 11:11:40 INFO  Cancelling compilation on scala-cli server
2023.09.21 11:11:40 INFO  Attempting to connect to the build server...
```

The tricky thing is that for a normal user, they will have no idea what is going on. Since even if you look in the Metals doctor it will be constantly refreshing as the build server reconnects, so you can't even see what the issue if it is listed there. When this happens there should be a better way to signify to the user what is happening. Maybe these warnings should be errors? Maybe on the Metals side these should be surfaced to the user?

Also note that even if the user sees this message and tries a `bloop exit` it won't fix the issue. Also trying a `scala-cli bloop exit` still doesn't fix the issue.  The only way I was able to fix the issue was switching my system jvm to 19, then killing everything, and then restarting.

**To Reproduce**
This might be hard to reproduce, but just have a minimal project, make sure bloop is started with 17, and then add in a using directive to use 19. You should see this behavior.

**Expected behaviour**
I'd expect a clear message surfaced to the user about how to fix the issue directly from within the editor.

MaciejG604:
Fixes #2412

Root cause:
- our Bsp command didn't take into consideration the options set with `using directives`, so the `using jvm 19` could never be taken into consideration when downloading jvm for bloop
- adding a different version of the jvm to the build made [Scala CLI add a different `--release` flag to the project](https://github.com/VirtusLab/scala-cli/blob/d8a4d59308b4f65368fac36d68b720e425170469/modules/build/src/main/scala/scala/build/Build.scala#L1028), the build client gets notified about this fact with [OnBuildTargetDidChange: notification](https://build-server-protocol.github.io/docs/specification/#onbuildtargetdidchange-notification), in case of metals the response for it is to restart the build server thus the constant recompilation. 

EDIT: The `-release` flag should not be set if the jvm of bloop is too low, thus it should not result in the buildTargetChange notification. However one of our functions (here )did not get passed a compiler argument thus it didn't check if the '-release' flag should be set. This resulted in the flag being set and reset which resulted in the buildTargetChange notification.

Since we already preprocess sources with `CrossSources.forInputs` there's no significant additional time complexity of the solution proposed.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedCompilationServerOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedCompilationServerOptions.scala
@@ -276,7 +276,7 @@ final case class SharedCompilationServerOptions(
       javaOpts =
         (if (bloopDefaultJavaOpts) baseConfig.javaOpts
          else Nil) ++ bloopJavaOpt ++ bloopDefaultJvmOptions(logger).getOrElse(Nil),
-      minimumBloopJvm = javaV.getOrElse(8),
+      minimumBloopJvm = javaV.getOrElse(17),
       retainedBloopVersion = retainedBloopVersion
     )
   }
--- a/scala-cli-src
+++ b/scala-cli-src
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 set -euo pipefail
 SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
-LAUNCHER="$(cd "$SCRIPT_DIR" && ./mill show cli.launcher </dev/null | jq -r . | sed 's/ref:[a-z0-9]*://')"
+LAUNCHER="$(cd "$SCRIPT_DIR" && ./mill show cli.launcher </dev/null | jq -r . | sed 's/ref:[a-z0-9]*:[a-z0-9]*://')"
 exec "$LAUNCHER" "$@"
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -197,6 +197,9 @@ object Build {
     def diagnostics: None.type   = None
   }
 
+  /** If some options are manually overridden, append a hash of the options to the project name
+    * Using only the command-line options not the ones from the sources.
+    */
   def updateInputs(
     inputs: Inputs,
     options: BuildOptions,
--- a/modules/build/src/main/scala/scala/build/bsp/Bsp.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/Bsp.scala
@@ -7,7 +7,7 @@ import scala.build.input.{Inputs, ScalaCliInvokeData}
 import scala.concurrent.Future
 
 trait Bsp {
-  def run(initialInputs: Inputs): Future[Unit]
+  def run(initialInputs: Inputs, initialBspOptions: BspReloadableOptions): Future[Unit]
   def shutdown(): Unit
 }
 
--- a/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspImpl.scala
@@ -456,10 +456,9 @@ final class BspImpl(
     *   the initial input sources passed upon initializing the BSP connection (which are subject to
     *   change on subsequent workspace/reload requests)
     */
-  def run(initialInputs: Inputs): Future[Unit] = {
-    val reloadableOptions = bspReloadableOptionsReference.get
-    val logger            = reloadableOptions.logger
-    val verbosity         = reloadableOptions.verbosity
+  def run(initialInputs: Inputs, initialBspOptions: BspReloadableOptions): Future[Unit] = {
+    val logger    = initialBspOptions.logger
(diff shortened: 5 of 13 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/SharedCompilationServerOptions.scala
```diff
@@ -276,7 +276,7 @@ final case class SharedCompilationServerOptions(
       javaOpts =
         (if (bloopDefaultJavaOpts) baseConfig.javaOpts
          else Nil) ++ bloopJavaOpt ++ bloopDefaultJvmOptions(logger).getOrElse(Nil),
-      minimumBloopJvm = javaV.getOrElse(8),
```
Gedochao, line 279: weren't we set on allowing to run Bloop with older Java versions? I have a vague recollection it should still run on 11 🤔 
I could be wrong

Gedochao:
Could we add an anti-regression test for this?
should be possible to set up an integration test with BSP, I think

modules/cli/src/main/scala/scala/cli/commands/shared/SharedCompilationServerOptions.scala
MaciejG604: Right now, we very restricitvely keep the JVM for bloop at least version 17, so I thought using 17 as minimum here is reasonable.
If the minimum is not set there's a possibility that an existing bloop server won't be restarted if it's jvm is older than requested. Actually that was part of the problem of the issue.

Gedochao: Alright then, let's bump to 17 👍

scala-cli-src
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 set -euo pipefail
 SCRIPT_DIR="$( cd "$( dirname "$0" )" && pwd )"
-LAUNCHER="$(cd "$SCRIPT_DIR" && ./mill show cli.launcher </dev/null | jq -r . | sed 's/ref:[a-z0-9]*://')"
+LAUNCHER="$(cd "$SCRIPT_DIR" && ./mill show cli.launcher </dev/null | jq -r . | sed 's/ref:[a-z0-9]*:[a-z0-9]*://')"
```
Gedochao, line 4: ?

Gedochao: right... `scala-cli-src` was broken somewhere along the way, since `./mill show cli.launcher` output changed.
ok.

Gedochao (approved):

PR #2460, opened 2023-10-12
Support Scala Native 0.5.x changes in publishing artifacts

WojciechMazur:
Upcoming Scala Native 0.5.x changes how some artifacts are published: 
1. Scala Standard Library (scalalib) is now published for each Scala version, instead of being published once for Scala binary version. It is required to properly handle possible changes in sources between versions (mostly in Scala3) and to prepare for [SIP-51](https://github.com/scala/improvement-proposals/pull/54/) and potential unfreezing of Scala 2.13 stdlib. 
2. `scalalib` does no longer contain transitive dependency on Java Standard Library (javalib) re-implementation. This dependency is now handled explicitly.

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/ScalaNativeOptions.scala
+++ b/modules/options/src/main/scala/scala/build/options/ScalaNativeOptions.scala
@@ -115,11 +115,19 @@ final case class ScalaNativeOptions(
   def platformSuffix: String =
     "native" + ScalaVersion.nativeBinary(finalVersion).getOrElse(finalVersion)
 
-  def nativeDependencies(scalaVersion: String): Seq[AnyDependency] =
+  def nativeDependencies(scalaVersion: String): Seq[AnyDependency] = {
+    // https://github.com/scala-native/scala-native/pull/3326
+    val scalalibVersion =
+      if (finalVersion.startsWith("0.4.")) finalVersion
+      else s"$scalaVersion+$finalVersion"
+    // Since 0.5.x Scala Native requires explicit dependency on javalib
+    // See https://github.com/scala-native/scala-native/pull/3566
+    val javalib = dep"org.scala-native::javalib::$finalVersion"
     if (scalaVersion.startsWith("2."))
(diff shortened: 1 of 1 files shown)


WojciechMazur:
Change was tested locally based on published 0.5.0-SNAPSHOT versions. I don't think it's good idea to add dependency on snapshots in tests

modules/options/src/main/scala/scala/build/options/ScalaNativeOptions.scala
```diff
@@ -115,11 +115,19 @@ final case class ScalaNativeOptions(
   def platformSuffix: String =
     "native" + ScalaVersion.nativeBinary(finalVersion).getOrElse(finalVersion)
 
-  def nativeDependencies(scalaVersion: String): Seq[AnyDependency] =
+  def nativeDependencies(scalaVersion: String): Seq[AnyDependency] = {
+    // https://github.com/scala-native/scala-native/pull/3326
+    val scalalibVersion =
+      if (finalVersion.startsWith("0.4.")) finalVersion
```
Gedochao, line 121: should this be the case for `0.4.x` only, or for earlier versions as well?

Gedochao:
@WojciechMazur I wonder if we can't wait for a stable `0.5.0` release, so that we can test for it?

modules/options/src/main/scala/scala/build/options/ScalaNativeOptions.scala
WojciechMazur: Earlier, that is 0.3.x are not supported at all (they've used Scala 2.10 and 2.11 only) and are long time deprecated so we don't need to worry about them.

WojciechMazur:
> I wonder if we can't wait for a stable 0.5.0 release, so that we can test for it? 

We can wait, but there are some early adopters using 0.5.0-SNAPSHOTS who'd like to use it. I'll this decision to the ScalaCLI team

Gedochao (approved):
LGTM, let's make sure we test for this when we bump Scala Native to `0.5.0`

PR #2479, opened 2023-10-19
Make script object wrapper for Scala 2 extend DelayedInit

### LINKED ISSUE #2470: scala-cli stalls when awaiting every single result from a list of Futures.
**Version(s)**

**Describe the bug**

**To Reproduce**

**Expected behaviour**

MaciejG604:
Fixes #2470

Thanks to @rochala for suggesting the solution of using a wrapper that extends `App`.

This simple fix seems to have been overlooked in the past, probably because it has some quirks when it comes to different Scala versions.

For Scala 2.12 and 2.13 an object that extends App avoids the deadlock when background threads are started from its initialization clause.
However, this is not true for Scala 3, it is a known bug that has not been fixed yet - https://github.com/lampepfl/dotty/issues/16149

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala
@@ -12,7 +12,7 @@ import coursier.publish.signing.logger.InteractiveSignerLogger
 import coursier.publish.signing.{GpgSigner, NopSigner, Signer}
 import coursier.publish.sonatype.SonatypeApi
 import coursier.publish.upload.logger.InteractiveUploadLogger
-import coursier.publish.upload.{DummyUpload, FileUpload, HttpURLConnectionUpload}
+import coursier.publish.upload.{DummyUpload, FileUpload, HttpURLConnectionUpload, Upload}
 import coursier.publish.{Content, Hooks, Pom, PublishRepository}
 
 import java.io.{File, OutputStreamWriter}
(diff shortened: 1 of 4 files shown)


build.sc
```diff
@@ -395,7 +395,7 @@ trait Core extends ScalaCliSbtModule with ScalaCliPublishModule with HasTests
          |
          |  def runnerOrganization = "${runner(Scala.runnerScala3).pomSettings().organization}"
          |  def runnerModuleName = "${runner(Scala.runnerScala3).artifactName()}"
-         |  def runnerVersion = "${runner(Scala.runnerScala3).publishVersion()}"
+         |  def runnerVersion = "1.0.5-SNAPSHOT"
```
Gedochao: dev leftover

modules/build/src/main/scala/scala/build/Sources.scala
```diff
@@ -86,14 +86,22 @@ object Sources {
...
+      Seq(
+        InMemory(
+          originalPath,
+          generatedRelPath,
+          content.getBytes(StandardCharsets.UTF_8),
+          Some(wrapperParams))
+      ) ++ wrapper.additionalSourceCode.map { sourceCode =>
+        InMemory(
+          Left("script-wrapper"),
+          generatedRelPath / os.up / "script-wrapper.scala",
```
Gedochao: might want to make this path unique per-script.
maybe smth like:
```suggestion
          generatedRelPath / os.up / s"${generatedRelPath.last.replace(".", "_")}_script-wrapper.scala",
```

MaciejG604: RIght now all the script-wrapper.scala files contain the same code, so naming them the same allows for filtering them out with `.distinct`. Although it would be good to name them per-wrapper so we could have different code generated for class and object wrappers.

PR #2484, opened 2023-10-22
Load local ivy path from `ivy.home` and `user.home` system properties

### LINKED ISSUE #2483: Publish local is not respecting user.home
**Version(s)**
1.0.3

**Describe the bug**

When publishing locally, if the local ivy repo is not defined, scala cli falls back to `~/.ivy2/local`: https://github.com/VirtusLab/scala-cli/blob/89c275885b14d600ee760034499bc74344cc43fa/modules/cli/src/main/scala/scala/cli/commands/publish/RepoParams.scala#L182-L183

However, when resolving libraries it takes the `user.home` java property into account, so it tries to resolve the dependencies from `{user.home}/.ivy2/local`

**To Reproduce**

- Set `JAVA_OPTS=-Duser.home=~/somewhere/`
- Run `scala-cli --power publish local foo.sc`, you'll see that it was published to `~/.ivy2/local`
- Try to import the local library in another script, you'll get a `not found: ~/somewhere/.ivy2/local/...`

**Expected behaviour**

The fallback should be `user.home` and, only if that property is not defined, it would fallback to `~`

Looking at the coursier source code, it might also be worth it to look at the `ivy.home` property:

https://github.com/coursier/sbt-coursier/blob/e2de6e30e32d9eccfab99b38ff4e4d281e00817e/modules/lm-coursier/src/main/scala/lmcoursier/internal/ResolutionParams.scala#L93-L101

JD557:
Fixes #2483

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/RepoParams.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/RepoParams.scala
@@ -179,7 +179,10 @@ object RepoParams {
   }
 
   def ivy2Local(ivy2HomeOpt: Option[os.Path]) = {
-    val home = ivy2HomeOpt.getOrElse(os.home / ".ivy2")
+    val home = ivy2HomeOpt
+      .orElse(sys.props.get("ivy.home").map(prop => os.Path(prop)))
+      .orElse(sys.props.get("user.home").map(prop => os.Path(prop) / ".ivy2"))
+      .getOrElse(os.home / ".ivy2")
     val base = home / "local"
     // not really a Maven repo…
     RepoParams(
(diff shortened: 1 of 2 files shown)


Gedochao:
@JD557 would you have the time to add an anti-regression test as well?
We have the relevant integration tests in `scala.cli.integration.PublishLocalTestDefinitions`: https://github.com/VirtusLab/scala-cli/blob/7376b2f858326d9055f67e3de5beae1a9850dc5c/modules/integration/src/test/scala/scala/cli/integration/PublishLocalTestDefinitions.scala#L5

JD557:
Added some tests in b8d5023.

I hope this is what you meant, I mostly just copied and adapted the other test in that file.

Gedochao (approved):
LGTM

PR #2488, opened 2023-10-23
Ensure docker is logged in on the M1 CI

Gedochao:
This should fix the `generate-linux-arm64-native-launcher` job on the `main` branch


tgodzik (approved):
Let's check it out!

PR #2492, opened 2023-10-23
Add docker logging and keychain unlocking to CI

MaciejG604:
Follow-up to #2488 #2489 #2490


tgodzik (approved):
We can try, though I think that keychain needs to be created

PR #2508, opened 2023-11-02
bugfix: Don't try to always get system jvm first

tgodzik:
Previously, we would always first specify `system|` when downloading default JVM, which would coursier just return java from JAVA_HOME without any checks even if we needed to download never Java. Now, we only use specific versions, since java home is handled before.

Trying to figure out the tests in https://github.com/VirtusLab/scala-cli/pull/2507

### DIFF
--- a/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
+++ b/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
@@ -74,14 +74,7 @@ object OsLibc {
   }
 
   def defaultJvm(os: String): String = {
-    val hasEmptyJavaHome = Option(System.getenv("JAVA_HOME")).exists(_.trim.isEmpty)
-    val defaultJvm0      = baseDefaultJvm(os, defaultJvmVersion)
-    if (hasEmptyJavaHome)
-      // Not using the system JVM if JAVA_HOME is set to an empty string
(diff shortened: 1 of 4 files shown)


Gedochao (approved):

modules/core/src/main/scala/scala/build/internals/OsLibc.scala
```diff
@@ -74,14 +74,7 @@ object OsLibc {
   }
 
   def defaultJvm(os: String): String = {
-    val hasEmptyJavaHome = Option(System.getenv("JAVA_HOME")).exists(_.trim.isEmpty)
-    val defaultJvm0      = baseDefaultJvm(os, defaultJvmVersion)
-    if (hasEmptyJavaHome)
-      // Not using the system JVM if JAVA_HOME is set to an empty string
-      // (workaround for https://github.com/coursier/coursier/issues/2292)
```
MaciejG604, line 81: While we're at it, I think we should also have this workaround at `JavaOptions.scala line: 153`.

tgodzik: Added nonEmpty to be on the safe side, though I don't know when it would fail, so I haven't added a test.

Gedochao:
I rebased this on top of `main`, that should fix the CI

Gedochao (approved):

PR #2512, opened 2023-11-06
Simplify build

lolgab:
These `sources` and `resources` were either adding the directories already present, or adding directories that didn't contain any sources (like `modules/src/main/scala`, which doesn't exist).


Gedochao (approved):
Indeed, thanks for the contribution!

PR #2528, opened 2023-11-10
Fix publishing with mill `0.11.5`

Gedochao:
This should address the problems with publishing after the `0.11.5` mill bump.
Note that passing `--disable-callgraph-invalidation` explicitly is now necessary until https://github.com/com-lihaoyi/mill/issues/2844 gets resolved.
Another note: `-i` requires to always be the first argument when passed to the `mill` script.
For the initial workaround for the `0.11.5` bump, check the bump PR #2446


Gedochao:
Closing this, feels pointless to hack any further.
We'll wait with the bump until https://github.com/com-lihaoyi/mill/issues/2844 gets resolved.
For now, the bump will get reverted in #2529

PR #2535, opened 2023-11-13
Make object wrapper extend DelayedInit

### LINKED ISSUE #2470: scala-cli stalls when awaiting every single result from a list of Futures.

MaciejG604:
Fixes https://github.com/VirtusLab/scala-cli/issues/2470

> Thanks to @rochala for suggesting the solution of using a wrapper that extends App.
> 
> This simple fix seems to have been overlooked in the past, probably because it has some quirks when it comes to different Scala versions.
> 
> For Scala 2.12 and 2.13 an object that extends App avoids the deadlock when background threads are started from its initialization clause.
> However, this is not true for Scala 3, it is a known bug that has not been fixed yet - https://github.com/lampepfl/dotty/issues/16149

Object wrapper extending `DelayedInit` can be chosen with `--delayed-init` and `//> using delayedInit.wrapper`.
This wrapper will not suffer from deadlocks in Scala 2.13 and 2.12 when background threads are used.

Since `DelayedInit` is deprecated, this mechanism won't be the default.
Using `App` instead of `DelayedInit` requires a bit more work as it introduces an extra main function, which messes up with Scala CLI's main class inference from scripts.


build.sc
```diff
@@ -395,7 +395,7 @@ trait Core extends ScalaCliSbtModule with ScalaCliPublishModule with HasTests
          |
          |  def runnerOrganization = "${runner(Scala.runnerScala3).pomSettings().organization}"
          |  def runnerModuleName = "${runner(Scala.runnerScala3).artifactName()}"
-         |  def runnerVersion = "${runner(Scala.runnerScala3).publishVersion()}"
+         |  def runnerVersion = "1.0.5-SNAPSHOT"
```
Gedochao: to be dropped, I assume

modules/build/src/main/scala/scala/build/internal/ObjectCodeWrapper.scala
```diff
@@ -4,66 +4,92 @@ package scala.build.internal
...
+         |$aliasObject
+         |""".stripMargin
+    )
 
     (top, bottom)
   }
+
+  override def additionalSourceCode: Option[(os.RelPath, String)] = Option.when(useDelayedInit)(
+    os.RelPath("delayed-init-wrapper.scala") ->
+      // This is copied from Scala 2 implementation of App trait, but with no main method and no argument handling
```
Gedochao, line 79: might want to include the commit hash of the copied code from the Scala 2 repo, for future context (even if the code isn't subject to much changes between versions, as I'd presume)

modules/cli/src/main/scala/scala/cli/commands/shared/SharedScriptOptions.scala
```diff
@@ -0,0 +1,21 @@
+package scala.cli.commands.shared
+
+import caseapp.*
+
+import scala.cli.commands.tags
+
+// format: off
```
Gedochao: NIT it's ok to enable formatting

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScriptWrapper.scala
```diff
@@ -0,0 +1,37 @@
...
+  "//> using delayedInitWrapper",
+  """`//> using objectWrapper`
+    |
+    |`//> using delayedInitWrapper`
+    |""".stripMargin
+)
+@DirectiveLevel(SpecificationLevel.RESTRICTED)
+@DirectiveDescription("Set parameter for the code wrapper for scripts")
+final case class ScriptWrapper(
+  @DirectiveName("object.wrapper")
```
Gedochao, line 20: IMO `wrapper.*` should be the prefix here
```suggestion
  @DirectiveName("wrapper.object")
```

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScriptWrapper.scala
```diff
@@ -0,0 +1,37 @@
...
+  objectWrapper: Boolean = false,
+  @DirectiveName("delayedInit.wrapper")
```
Gedochao: Similarly here. Also, I believe we were to avoid camel case in directives' keys,
so
```suggestion
  @DirectiveName("wrapper.delayed.init")
```
or
```suggestion
  @DirectiveName("wrapper.delayed-init")
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -174,8 +174,9 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
         "--java-prop", "scala.colored-stack-traces=false"
       )
       // format: on
-      val res            = os.proc(cmd).call(cwd = root, check = false, mergeErrIntoOut = true)
-      val output         = res.out.lines()
+      val res    = os.proc(cmd).call(cwd = root, check = false, mergeErrIntoOut = true)
+      val output = res.out.lines()
+      pprint.err.log(TestUtil.removeAnsiColors(output.mkString("\n")))
```
Gedochao: ?

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScriptWrapper.scala
MaciejG604: I also think so, but introducing this change now will break compatibility unfortunately :(

Gedochao (approved):
LGTM

som-snytt:
> This simple fix seems to have been overlooked in the past

Previous `REPL wrap is App` https://github.com/scala/scala/pull/4246

Uncharacteristically, I submitted it three times. I don't remember why it did not fly.

MaciejG604:
May be replaced by #2556

MaciejG604:
Replaced by #2556

PR #2550, opened 2023-11-15
Chore/group warnings about directives in multiple files

### LINKED ISSUE #2377: Omit triggering warning about directives in multple files for "//> using file ..." directives

I'm playing with moving much of my sysadmin-ish scripting to Scala / scala-cli. I'd like to implement most of it with `shebang` style scripts.

The scripts would make use of published libraries, but also local out-factorings of shared code, for which I would like to retain script-like ergonomics. For example, I might bring them into a `shebang` script with `//> using file shared.scala`, and work with the imported code.

This all works! The problem is just a warning:

```
[warn] ./shared.scala:1:1                                                                                                                                                                                                                                                                        
[warn] Using directives detected in multiple files. It is recommended to keep them centralized in the /Users/swaldman/Sync/BaseFolders/development-why/playpen/scala-cli-directives-in-scripts/project.scala file.                                                                               
[warn] ./script2:1:1                                                                                                                                                                                                                                                                             
[warn] Using directives detected in multiple files. It is recommended to keep them centralized in the /Users/swaldman/Sync/BaseFolders/development-why/playpen/scala-cli-directives-in-scripts/project.scala file.
``` 

This is in general a good warning! I can centralize nearly all of my directives in `project.scala`, and bring them in as `//> using file project.scala`. So in general, my scripts can just have...

```
//> using file project.scala
//> using file shared.scala
```

and `shared.scala` can just contain

```
//> using file project.scala
```

The settings that describe everything other than the source set for the script can and should be centralized into `project.scala`.

But I need the `//> using file ...` in multiple files. So I still have the warning.

The easiest solution might be just to omit `//> using file ...` from the set of directives that triggers the `Using directives detected in multiple files.` warning. `//> using file ...` directives can't be centralized in that way when the entry point to the project is a script.

Another approach might be an Ammonite-like or C-pre-processor-style textual include. Top-level `shebang` scripts could declare their full directive set, and shared source would be brought into a single logical file. But this seems less elegant.

**Additional context**

Very much a work in progress, but for an example of the kind of script-set I'm trying to build, see https://github.com/swaldman/mchange-sysadmin-scripts/tree/main/taskbin

MaciejG604:
Answer to #2377
I don't think we can omit the warning, but this makes it more tidy.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -347,21 +347,16 @@ object CrossSources {
         .flatten
         .filter((path, _) => ScopePath.fromPath(path) != ScopePath.fromPath(projectFilePath))
         .pipe { pathsToReport =>
-          def coreMessage(paths: Iterable[String] = Nil) =
-            val detectedMsg = "Using directives detected in multiple files"
-            val recommendedMsg =
-              s"It is recommended to keep them centralized in the $projectFilePath file."
-            if paths.isEmpty then
-              s"$detectedMsg. $recommendedMsg"
-            else
-              s"""$detectedMsg:
-                 |${paths.mkString("- ", s"${System.lineSeparator}- ", "")}
-                 |$recommendedMsg
-                 |""".stripMargin
+          val diagnosticMessage = WarningMessages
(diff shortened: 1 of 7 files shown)


Gedochao (approved):
LGTM

PR #2552, opened 2023-11-15
Fix list of licenses URL

JD557:
The old URL now redirects to https://docs.fedoraproject.org/en-US/legal/license-approval/

I was trying to find a list there and I think this is it? https://spdx.org/licenses/


tgodzik (approved):
Looks good! Thanks!

PR #2556, opened 2023-11-16
Make object wrapper extend App

### LINKED ISSUE #2470: scala-cli stalls when awaiting every single result from a list of Futures.

MaciejG604:
Fixes https://github.com/VirtusLab/scala-cli/issues/2470

Thanks to @rochala for suggesting the solution of using a wrapper that extends App.

This simple fix seems to have been overlooked in the past, probably because it has some quirks when it comes to different Scala versions.

For Scala 2.12 and 2.13 an object that extends App avoids the deadlock when background threads are started from its initialization clause.
However, this is not true for Scala 3, it is a known bug that has not been fixed yet - https://github.com/lampepfl/dotty/issues/16149

------

RESTRICTIONS of the new wrapper:
- scripts named `main.sc` can't be accessed from other scripts that use the App wrapper since the name `main` clashes with their `main` method.


modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
```diff
@@ -165,18 +157,27 @@ case object ScriptPreprocessor extends Preprocessor {
...
     */
-  def getScriptWrapper(buildOptions: BuildOptions): CodeWrapper =
+  def getScriptWrapper(buildOptions: BuildOptions): CodeWrapper = {
+    val scalaVersionOpt = for {
+      maybeScalaVersion <- buildOptions.scalaOptions.scalaVersion
+      scalaVersion      <- maybeScalaVersion.versionOpt
+    } yield scalaVersion
+
+    def objectCodeWrapperForScalaVersion =
+      // AppObjectWrapper only introduces the 'main.sc' restriction when used in Scala 3, there's no gain in using it
```
Gedochao:
```suggestion
      // AppObjectWrapper only introduces the 'main.sc' restriction when used in Scala 3, there's no gain in using it with Scala 3
```
wasn't clear to me

Gedochao (approved):
LGTM, very cool hack

PR #2570, opened 2023-11-21
Add log separators for integration and build tests

MaciejG604:
@Gedochao you speak, you have


Gedochao (approved):
LGTM
This is beautiful, why didn't we do this before 😍

PR #2571, opened 2023-11-21
Default to publish repository configured for local machine when infering publish.ci.repository

### LINKED ISSUE #2563: Misleading error message in publish setup when the publish repository has a typo
**Version(s)**
v1.0.6

**Describe the bug**

If `publish.repository` has a typo, `publish setup` fails with the following error message:

```
[error]  Missing publish user for publishing, specify one with  or with a '//> using publish.credentials' directive or by setting publish.credentials in the configuration
[error]  Missing publish password for publishing, specify one with  or with a '//> using publish.credentials' directive or by setting publish.credentials in the configuration
```

This is particularly annoying, because if one calls `scala-cli --power config publish.credentials`, they'll see the credentials as expected.

I believe that happens because in https://github.com/VirtusLab/scala-cli/blob/main/modules/cli/src/main/scala/scala/cli/commands/publish/checks/UserCheck.scala#L122-L131 the generic "publish user not found" error is used no matter what happens.

**To Reproduce**

Write some gibberish in the `publish.repository` setting and run `scala-cli publish setup .`

**Expected behaviour**

It would be nice if the message at least mentioned the repository (e.g. "Missing publish user for repository "gibberish" for publishing, specify one with  or with a '//> using publish.credentials' directive or by setting publish.credentials in the configuration"), that would make it simpler to spot typos.

It could also be nice if the error message also showed the mapping between the repository name and expected host.

### LINKED ISSUE #2566: GitHub Actions setup `publish.ci.repository` should default to `publish.repository`
**Version(s)**
v1.0.6

**Describe the bug**

Even if a project has a custom repository setup (say `//> using publish.repository "central"`) running the [GitHub Actions setup](https://scala-cli.virtuslab.org/docs/commands/publishing/publish-setup#github-actions-setup) still sets the `publish.ci.repository` to `s1-central`.

At least personally this was quite surprising. I had everything working locally, I could see the secrets were in GitHub, but I was still getting a 403.

**To Reproduce**

Follow the usual `[publish setup` instructions](https://scala-cli.virtuslab.org/docs/commands/publishing/publish-setup), but change the `publish.repository` to something else before setting the GitHub actions.

**Expected behaviour**

I think defaulting to the `publish.repository` should be a safer option, mostly because I don't think there's any place more dangerous to publish to than sonatype releases (e.g. if I have a company project that I publish to the company repo and suddenly add GitHub options, I surely won't want it to be published there).

This is a bit of bikeshedding, though. If you don't think this should be the default behaviour, then maybe a warning should be triggered? Not sure.

MaciejG604:
Fixes #2566 and #2563

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/checks/UserCheck.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/checks/UserCheck.scala
@@ -21,54 +21,28 @@ final case class UserCheck(
   def fieldName     = "user"
   def directivePath = "publish" + (if (options.publishParams.setupCi) ".ci" else "") + ".user"
 
-  private def repoOpt(pubOpt: BPublishOptions): Option[String] =
-    options.publishRepo.publishRepository
-      .orElse {
-        pubOpt.contextual(options.publishParams.setupCi).repository
-      }
-      .orElse {
-        if (options.publishParams.setupCi)
-          pubOpt.contextual(isCi = false).repository
-        else
-          None
-      }
-
-  private def hostOpt(pubOpt: BPublishOptions): Option[String] =
-    repoOpt(pubOpt).flatMap { repo =>
-      RepoParams(
-        repo,
-        pubOpt.versionControl.map(_.url),
-        workspace,
-        None,
-        false,
-        null,
-        logger
-      ) match {
-        case Left(ex) =>
-          logger.debug("Caught exception when trying to compute host to check user credentials")
-          logger.debug(ex)
-          None
-        case Right(params) =>
-          Some(new URI(params.repo.snapshotRepo.root).getHost)
-      }
-    }
-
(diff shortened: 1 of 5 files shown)


modules/cli/src/main/scala/scala/cli/commands/publish/checks/UserCheck.scala
```diff
@@ -21,27 +21,37 @@ final case class UserCheck(
...
+        null,
+        logger
+      ) match {
+        case Left(ex) =>
+          logger.debug("Caught exception when trying to compute host to check user credentials")
+          logger.debug(ex)
+          None
+        case Right(params) =>
+          Some(new URI(params.repo.snapshotRepo.root).getHost)
+      }
```
Gedochao: these 2 functions seem like the same ones as in `PasswordCheck`, should be able to abstract

Gedochao (approved):
LGTM

PR #2573, opened 2023-11-22
Workaround for home.dir property not being set

MaciejG604:
This will allow us to fix the issue before https://github.com/com-lihaoyi/os-lib/pull/239 gets merged and released.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -20,6 +20,11 @@ import scala.util.Properties
 
 object ScalaCli {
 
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -20,6 +20,11 @@ import scala.util.Properties
 
 object ScalaCli {
 
+  if (scala.util.Try(os.Path(System.getProperty("user.home"))).isFailure) {
```
Gedochao, line 24: maybe add a `TODO` comment with a link to the `os-lib` issue?

MaciejG604:
Linked to #2585

Gedochao (approved):

PR #2576, opened 2023-11-24
Skip validation for default Scala versions, add build test

MaciejG604:
Validation requires maven-metadata.xml to be downloaded which can't always be dowloaded even if maven-central itself can be accessed from environment.

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
+++ b/modules/options/src/main/scala/scala/build/options/BuildOptions.scala
@@ -312,10 +312,20 @@ final case class BuildOptions(
     repositories: Seq[Repository] = Nil
   ): Either[BuildException, Option[ScalaParameters]] = either {
 
+    val defaultVersions = Set(
+      Constants.defaultScalaVersion,
+      Constants.defaultScala212Version,
+      Constants.defaultScala213Version
+    )
+
     val svOpt: Option[String] = scalaOptions.scalaVersion match {
       case Some(MaybeScalaVersion(None)) =>
         None
+      // Do not validate Scala version in offline mode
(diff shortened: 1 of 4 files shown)


modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -312,11 +312,21 @@ final case class BuildOptions(
...
+    val defaultVersions = Set(
+      Constants.defaultScalaVersion,
+      Constants.defaultScala212Version,
+      Constants.defaultScala213Version
+    )
+
     val svOpt: Option[String] = scalaOptions.scalaVersion match {
       case Some(MaybeScalaVersion(None)) =>
         None
+      // v Do not validate Scala version
```
Gedochao: NIT these comments aren't entirely readable to me.
```suggestion
         // do not validate the default Scala version when none has been passed explicitly
        None
```

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -312,11 +312,21 @@ final case class BuildOptions(
...
       case Some(MaybeScalaVersion(Some(svInput))) if internal.offline.getOrElse(false) =>
         Some(svInput)
+      case Some(MaybeScalaVersion(Some(svInput))) if defaultVersions.contains(svInput) =>
+        Some(svInput)
+      // ^ Do not validate Scala version
```
Gedochao: NIT
```suggestion
        // do not validate explicitly passed default Scala versions, either
        Some(svInput)
```

Gedochao (approved):
LGTM

PR #2578, opened 2023-11-24
Remove logs for some integration tests

MaciejG604:


MaciejG604:
Not valid, that was a sanity check for our CI problems

PR #2579, opened 2023-11-24
Bump MacOS CI to `macOS-13`

Gedochao:
Okay, so here's some context.
We've been running our `macOS` tests on `macOS-12` (not to confuse with `macOS-m1` tests, which use a self-hosted runner)
It seems the `macOS-12` has become very flaky between the `15.11.2023` and `29.10.2023` versions.
As GitHub runners only guarantee the major versions of the image running on the runner node, it seems there's 2 kinds of nodes in the flock:
- https://github.com/actions/runner-images/releases/tag/macOS-12%2F20231029.1
`macOS-12/20231029.1` <- this runs our CI fine with MacOS 12.6.9, version from `29.10.2023` (older)
- https://github.com/actions/runner-images/releases/tag/macOS-12%2F20231115.2
`macOS-12/20241115.2` <- this fails a couple of our tests with invalid `libsodiumjni`, macOS 12.7.1, version from `15.11.2023` (seems to be the latest `macOS-12`)

https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
`macOS-13` seems to be tagged as a beta release as of the time of this PR, but I suspect it should get promoted to stable soon.
For the purposes of our CI, it seems to be a lot more stable than `macOS-12`.
The flakiness detected in this PR (https://github.com/electron-userland/electron-builder/issues/7137, as noted by @MaciejG604) actually happens on `macos-12` as well.

TL;DR bumping to `macOS-13` seems to be the way to go, even though it's still considered to be in beta.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
@@ -63,11 +63,7 @@ class NativePackagerTests extends ScalaCliSuite {
         }
       }
     }
-
-    // building dmg package sometimes fails with:
-    // 'hdiutil: couldn't eject "disk2" - Resource busy'
-    test("building dmg package".flaky) {
-
+    def testBuildingDmgPackage(): Unit =
       testInputs.fromRoot { root =>
 
         val appName = helloWorldFileName.stripSuffix(".scala").toLowerCase()
@@ -109,6 +105,15 @@ class NativePackagerTests extends ScalaCliSuite {
           )
         }
       }
+
+    // FIXME: building dmg package sometimes fails with:
+    // 'hdiutil: couldn't eject "disk2" - Resource busy'
+    if (TestUtil.isM1 || !TestUtil.isCI)
+      test("building dmg package") {
+        testBuildingDmgPackage()
+      }
+    else test("building dmg package".flaky) {
+      testBuildingDmgPackage()
     }
   }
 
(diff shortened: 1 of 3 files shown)


MaciejG604:
I've added a flaky label, let's see the results.
The case of ejecting the disk with the `hdiutil` command seems to fail randomly https://github.com/electron-userland/electron-builder/issues/7137

We could add a work-around, similar to the one that fixesdthe issue linked above, to https://github.com/VirtusLab/scala-packager

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -107,6 +105,15 @@ class NativePackagerTests extends ScalaCliSuite {
...
       }
+
+    // FIXME: building dmg package sometimes fails with:
+    // 'hdiutil: couldn't eject "disk2" - Resource busy'
+    if (TestUtil.isM1 || !TestUtil.isCI)
+      test("building dmg package") {
+        testBuildingDmgPackage()
+      }
+    else test("building dmg package".flaky) {
+      testBuildingDmgPackage()
```
MaciejG604, line 116: There is a dedicated munit solution for this, something along the lines of:
```
override def munitFlakyOK: Boolean = Properties.isMac && !TestUtil.isM1 && TestUtil.isCI
```
Docs: https://scalameta.org/munit/docs/tests.html#tag-flaky-tests

But this is fine

MaciejG604 (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
Gedochao: I think it might be confusing to set flakiness according to platform, as the other tests in this file would also be affected.
So if someone tagged something as `.flaky` in another PR, he might be taken by surprise.
We can try to refactor this later, I'll merge this for now.

Issue #2585, opened 2023-11-27
Remove user.home hack when fix gets released in os-lib


MaciejG604:
When this gets merged and released https://github.com/com-lihaoyi/os-lib/pull/239

PR #2587, opened 2023-11-28
Add env for configuring home directory overriding

MaciejG604:
Allow to override the home directory set, using environment variable `SCALA_CLI_HOME_DIR_OVERRIDE`
This, together with #2573 should be removed in #2585

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -19,11 +19,38 @@ import scala.cli.util.ConfigDbUtils
 import scala.util.Properties
 
 object ScalaCli {
-
   // TODO: Remove this part once fix is released in os-lib (Issue #2585)
-  if (scala.util.Try(os.Path(System.getProperty("user.home"))).isFailure) {
-    System.err.println("Warning: user.home property is not set, setting it to user.dir")
-    System.setProperty("user.home", System.getProperty("user.dir"))
+  sys.env.get("SCALA_CLI_HOME_DIR_OVERRIDE")
+    .filter(_.nonEmpty)
+    .filter(homeDir => scala.util.Try(os.Path(homeDir)).isSuccess)
+    .foreach { homeDirOverride =>
+      System.err.println(
+        s"Warning: user.home property overridden with the SCALA_CLI_HOME_DIR_OVERRIDE env var to: $homeDirOverride"
+      )
+      System.setProperty("user.home", homeDirOverride)
+    }
+  private def getInvalidPropMessage(homeDirOpt: Option[String]): String = homeDirOpt match {
+    case Some(value) => s"not valid: $value"
+    case None        => "not set"
+  }
+  sys.props.get("user.home") -> sys.props.get("user.dir") match {
+    case (Some(homeDir), _)
+        if scala.util.Try(os.Path(homeDir)).isSuccess => // all is good, do nothing
+    case (invalidHomeDirOpt, Some(userDir)) if scala.util.Try(os.Path(userDir)).isSuccess =>
+      System.err.println(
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -23,7 +23,9 @@ object ScalaCli {
   // TODO: Remove this part once fix is released in os-lib (Issue #2585)
   if (scala.util.Try(os.Path(System.getProperty("user.home"))).isFailure) {
     System.err.println("Warning: user.home property is not set, setting it to user.dir")
-    System.setProperty("user.home", System.getProperty("user.dir"))
+    val homeDirOverride = Option(System.getenv("SCALA_CLI_HOME_DIR_OVERRIDE"))
```
Gedochao: I suppose if we have an override, it should work arbitrarily.
as in, even if the `user.home` property is present, the env override should still be applied outside of this if clause.

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -21,9 +21,14 @@ import scala.util.Properties
...
-    System.err.println("Warning: user.home property is not set, setting it to user.dir")
-    System.setProperty("user.home", System.getProperty("user.dir"))
+  Option(System.getenv("SCALA_CLI_HOME_DIR_OVERRIDE")).orElse {
+    scala.util.Try(os.Path(System.getProperty("user.home"))) match
+      case _: scala.util.Success[os.Path] => None
+      case _ =>
+        System.err.println("Warning: user.home property is not set, setting it to user.dir")
+        Option(System.getProperty("user.dir"))
+  }.foreach { homeDirOverride =>
+    System.setProperty("user.home", homeDirOverride)
```
Gedochao: I don't think the override works right now.
If the `SCALA_CLI_HOME_DIR_OVERRIDE` is a valid path, then it will just be a `Success[os.Path]` and won't be propagated to `user.home`, if I'm not mistaken.
Also, we never validate the `user.dir` here, which theoretically might also be invalid (my trust in the JDK is dropping).

How about:
```suggestion
      sys.env.get("SCALA_CLI_HOME_DIR_OVERRIDE")
      .filter(_.nonEmpty)
      .filter(homeDir => scala.util.Try(os.Path(homeDir)).isSuccess)
      .foreach { homeDirOverride =>
        System.err.println(s"Warning: user.home property overriden with the SCALA_CLI_HOME_DIR_OVERRIDE env var to: $homeDirOverride")
        System.setProperty("user.home", homeDirOverride)
      }

    sys.props.get("user.home") -> sys.props.get("user.dir") match {
      case (Some(homeDir), _) if scala.util.Try(os.Path(homeDir)).isSuccess => // all is good, do nothing
      case (invalidHomeDir, Some(userDir)) if scala.util.Try(os.Path(userDir)).isSuccess =>
        System.err.println(s"Warning: user.home property is not valid: $invalidHomeDir, setting it to user.dir value: $userDir")
        System.setProperty("user.home", userDir)
      case (invalidHomeDir, invalidUserDir) =>
        System.err.println(s"Error: user.home property is not valid: $invalidHomeDir")
        System.err.println(s"Error: user.dir property is not valid: $invalidUserDir")
        System.err.println("Scala CLI cannot work correctly without a valid home or user directory.")
        System.err.println("Consider overriding it to a valid directory path with the SCALA_CLI_HOME_DIR_OVERRIDE environment variable.")
        sys.exit(1)
    }
```

Gedochao: I know this is temporary, but a test for the env variable override might be a good idea, so that we're sure it works

Gedochao (approved):

PR #2590, opened 2023-11-29
Migrate to Docusaurus v3, add local search plugin

MaciejG604:
Updated dependencies to use docusaurus 3, made necessary changes according to this guide: https://docusaurus.io/docs/migration/v3#common-mdx-problems

PR #2591, opened 2023-11-29
Add local search plugin to website

MaciejG604:
Stay on docusaurus v2 and add plugin for local search.
We should merge this or #2590


MaciejG604:
Superseded by #2590

PR #2601, opened 2023-12-01
Pass java opts to scalac

### LINKED ISSUE #2477: Add on option to pass Java options to the compiler process when running with `--server=false`
When running with `--server=false` scala-cli runs a scala compiler process (main class `dotty.tools.dotc.Main` for scala 3), it would be nice to be able to pass Java options (like `-Xmx`) to that process.

A new CLI option, something like `--compiler-java-opt` or `--standalone-compiler-java-opt` analogous to [--bloop-java-opt](https://scala-cli.virtuslab.org/docs/reference/cli-options#--bloop-java-opt) and [--java-opt](https://scala-cli.virtuslab.org/docs/reference/cli-options#--java-opt)

MaciejG604:
Fixes #2477

Should we leave passing options from `--javac-opts` flag to scalac? I can't really backtrace why it is like that now.


Gedochao:
> Should we leave passing options from `--javac-opts` flag to scalac? I can't really backtrace why it is like that now.

I think we should. The Scala compiler acts as the Java compiler in this case, so it makes sense.

Gedochao (approved):
LGTM

Issue #2592, opened 2023-11-29
Improve the UX of the docs website


Gedochao:
- [x] fix the search feature
- [x] add categories level for guides & cookbooks
- [x] shorten guides & cookbooks titles in the links tree
- [ ] generate sitemap & analyse it
- [ ] look for other improvements

MaciejG604:
Some sections from Reference could probably be merged with Guides e.g.:
- [ ] `Reference > Proxy authentication` and `Guides > Repositories and HTTP Proxies`
- [ ] `Reference > Dependency format` and `Guides > Managing dependencies / Updating dependencies` 


Also `Under the hood` and `Guides > Internals` have some similiar elements

PR #2612, opened 2023-12-05
Add a subcategories layer for guides & cookbooks

Gedochao:
Relevant ticket: #2592 

image

Guides & cookbooks are now divided into subcategories in the sidebar. 
The division is done according to the existing index docs.
Some recent docs haven't been present in the indexes, so I added them.

I also cleaned up some other stuff, like consistently using relative links where able (so that IDEs can do the busywork for us when we move docs around) and other minor things.


website/docs/cookbooks/ide/intellij.md
```diff
@@ -3,7 +3,7 @@ title: IntelliJ IDEA setup
 sidebar_position: 11
 ---
 
-import {ChainedSnippets} from "../../src/components/MarkdownComponents.js";
+import {ChainedSnippets} from "../../../src/components/MarkdownComponents.js";
```
tgodzik, line 6: Would it make sense to join the Intellij cookbooks into one?

website/docs/cookbooks/introduction/scala-scripts.md
tgodzik, line 1: Should we join instant startup with scripts?

website/docs/guides/intro.md
```diff
@@ -15,38 +15,41 @@ the [cookbooks section](../cookbooks/intro.md).
 
 These few guides are a good starting point when learning how to use Scala CLI.
 
-- [Configuration](./configuration.md) - learn how to configure various options, and what configuration styles are best
+- [Configuration](introduction/configuration.md) - learn how to configure various options, and what configuration styles are best
```
tgodzik, line 18: It's hard to figure out now, what is the difference between a guide and cook book. Isn't it basically the same?

tgodzik (approved):
Looks good! I added some comments about possible further improvements. If you think they are worth considering, we can add them to the issue.

website/docs/guides/intro.md
MaciejG604: Good point, the difference is quite slim, maybe we should consider merging them?
Although some of them won't really fit the guides, those could be left in a separate `Examples` section.
WDYT?
Of course it's for a potential separate ticket.

Gedochao: I believe the idea is for cookbooks to be simple step-by-step posts with clear examples, while guides are meant to be more in-depth.
Perhaps `Examples` would be a more readable category, rather than cookbooks...
But then, we might need a lot more of them, cause such a change would make more people to look for examples in there, rather than in command docs or guides.

Gedochao: alternatively, we could indeed merge them and only have guides... dunno, might be worth discussing it

website/docs/cookbooks/ide/intellij.md
Gedochao: we could, but that would make it complex enough to make it a guide.

Gedochao: also, besom's contributing doc seems to be linking to one of them, if we end up merging these into a guide, then we should update it in there: https://github.com/VirtusLab/besom/blob/develop/CONTRIBUTING.md#intellij-setup

website/docs/cookbooks/introduction/scala-scripts.md
Gedochao: similar to the IntelliJ ones, it'd make the result a guide rather than a cookbook

Issue #2184, opened 2023-06-03
Docs: How do you full link a Scala.js app when packaging?


davesmith00000:
**Version(s)**
1.0.0, but may not be relevant.

**Describe what needs to be done and why**
Missing docs on how to package a Scala.js file file that has been compressed by the closure compiler.

**Additional context**
Equivalent to `sbt fullLinkJS`. I'm sure this is doable I just can't find any information on it. I can add the docs for you if you give me the incantation. :smile:

tgodzik:
Does `scala-cli package --js-mode release` option work for you? That's what I found when doing ` scala package --help-js  `

tgodzik:
We no have more self explanatory options

PR #2614, opened 2023-12-06
Remove semantics Compliant for asInstaceOf

### LINKED ISSUE #2602: The default semantic of `asInstanceOf`s on Scala.js should not be `Compliant`
https://github.com/VirtusLab/scala-cli/blob/9385014b9a0207f3234cd1d4cf378aba5f46555e/modules/options/src/main/scala/scala/build/internal/ScalaJsLinkerConfig.scala#L93-L95

This contradicts the [documented default behavior](https://www.scala-js.org/doc/semantics.html#undefined-behaviors) for Scala.js.

> By default, undefined behaviors are in `Fatal` mode for `fastLinkJS` and in `Unchecked` mode for `fullLinkJS`

In particular, they note that enabling `Compliant` behavior:
> will have (potentially major) performance impacts.

This is highly surprising. This default makes scala-cli poorly suited for developing efficient webapps or JS lambdas.

MaciejG604:
Fixes #2602
Connected to https://github.com/VirtusLab/scala-js-cli/pull/40

Also added aliases for optimisation mode for Scala JS, to have the same as in the sbt plugin `fastLinkJs` and `fullLinkJs` to avoid questions such as in #2184 .

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -690,19 +690,20 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
     destPath: os.Path,
     mainClass: Option[String],
     logger: Logger
-  ): Either[BuildException, os.Path] = {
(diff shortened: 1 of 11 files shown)


Gedochao (approved):
lgtm

PR #2619, opened 2023-12-07
Docs/java opts for compiler

MaciejG604:
Docs for #2601

### DIFF
--- a/website/docs/commands/compile.md
+++ b/website/docs/commands/compile.md
@@ -289,11 +289,10 @@ Certain compiler options allow to view relevant help. Inputs aren't required whe
 Those include:
 
 - `-help`
-- all options prefixed with:
-    - `-V`
-    - `-W`
-    - `-X`
-    - `-Y`
+- `-V`
+- `-W`
+- `-X`
+- `-Y`
 
 <ChainedSnippets>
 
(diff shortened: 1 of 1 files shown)


website/docs/commands/compile.md
```diff
@@ -269,6 +269,13 @@ but they will not be recognised if they're not supported in a given Scala versio
 In such a case, refer to the `--scalac-help` output while passing the appropriate version with `-S`.
 :::
 
+### Java options for the compiler
+
+There are two ways to pass Java options to the compiler:
+- `--bloop-java-opt` when using the build server, which is the default, e.g. `--bloop-java-opt -XX:MaxHeapSize=8g`
+- `//> using options` or `--scalac-opt` with arguments prefixed by `-J`, e.g. `//> using options -J-XX:MaxHeapSize=8g`,
+this will work only when the build server is disabled (with `--server=false`).
```
MaciejG604, line 277: Should we unify the behaviour of `--scalac-opt` to also pass java opts to bloop?
This will require a larger refactor for `//> using options` to work as generating the compiler happens before preprocessing. But we can display warnings.
We could also display warnings when java opts are ignored when using bloop.

WDYT?

tgodzik: I wouldn't change it currently unless someone reports it being a problem. This a power user thing, so might be enough to have it well documented.

tgodzik (approved):

PR #2622, opened 2023-12-11
Add deprecation reporting mechanism for using directives

### LINKED ISSUE #2426: Deprecate `latest` for `toolkit` & add a relevant actionable diagnostic
**Version(s)**
1.0.4

**Describe the bug**

The default versions of Scala Native (and Scala.js) are behind the latest releases. Because they are not guaranteed to be forward-compatible, that means that libraries and toolkits built with the latest Native/JS releases may not work with scala-cli out-of-the-box.

**To Reproduce**

```scala
//> using toolkit typelevel:latest
//> using platform native

import cats.effect.*

object App extends IOApp.Simple:
  def run = IO.unit
```

```scala
$ scala-cli foo.scala 
Compiling project (Scala 3.3.0, Scala Native)
Compiled project (Scala 3.3.0, Scala Native)
[info] Linking (11491 ms)
[info] Checking intermediate code (quick) (300 ms)
[info] Discovered 1744 classes and 11262 methods
[info] Optimizing (debug mode) (6478 ms)
[info] Generating intermediate code (6400 ms)
[info] Produced 16 files
[info] Compiling to native code (4600 ms)
[error] /usr/bin/ld: /workspace/sandbox/.scala-build/sandbox_05adbf4d1c_919f107979-b0bbeb3ef8/native/native-code-nativelib_native0.4_3-0.4.15-1/scala-native/gc/immix/Marker.c.o: in function `Object_IsWeakReference':
[error] Marker.c:(.text+0x88): undefined reference to `__weak_ref_ids_min'
[error] /usr/bin/ld: Marker.c:(.text+0xa2): undefined reference to `__weak_ref_ids_max'
[error] clang: error: linker command failed with exit code 1 (use -v to see invocation)
Exception in thread "main" scala.scalanative.build.BuildException: Failed to link /workspace/sandbox/.scala-build/sandbox_05adbf4d1c_919f107979-b0bbeb3ef8/native/main
        at scala.scalanative.build.LLVM$.link(LLVM.scala:131)
        at scala.scalanative.build.Build$.$anonfun$build$1(Build.scala:106)
        at scala.scalanative.build.Logger.time(Logger.scala:35)
        at scala.scalanative.build.Logger.time$(Logger.scala:32)
        at scala.scalanative.cli.utils.FilteredLogger.time(FilteredLogger.scala:7)
        at scala.scalanative.build.Build$.build(Build.scala:61)
        at scala.scalanative.cli.ScalaNativeLd$.$anonfun$runLd$3(ScalaNativeLd.scala:73)
        at scala.scalanative.util.Scope$.apply(Scope.scala:32)
        at scala.scalanative.cli.ScalaNativeLd$.runLd(ScalaNativeLd.scala:72)
        at scala.scalanative.cli.ScalaNativeLd$.main(ScalaNativeLd.scala:42)
        at scala.scalanative.cli.ScalaNativeLd.main(ScalaNativeLd.scala)
Error: scala.build.errors.ScalaNativeBuildError: Error compiling with Scala Native
For more details, please see '/workspace/sandbox/.scala-build/stacktraces/1695712231-10233067241258950535.log'
```

Adding `//> using nativeVersion 0.4.15` fixes it.

**Expected behaviour**

IMHO a latest toolkit should Just Work:tm: with latest scala-cli. If a user has to explicitly lookup and add a version then it undermines the convenience factor and generally breaks user expectation.

- https://github.com/typelevel/toolkit/issues/49

There are two possible strategies to address this.

1. scala-cli should update its defaults to latest Scala.js and Scala Native versions as soon as they are released.

2. Toolkits should not release with libraries that require newer Native/JS versions than what scala-cli ships by default. In practice, this means the libraries aggregated in toolkits should also hold back on JS/Native updates, otherwise it might be impossible to release a toolkit with an important bug fix to one of those libraries.

cc @toniogela

### LINKED ISSUE #2473: //> using lib should emit a warning about deprecation
**Version(s)**
1.0.4

**Describe the bug**
`//> using lib` is deprecated in favour of `//> using dep`

**Expected behaviour**
Warning should be emitted. Ideally, with actionable diagnostics to change it to dep.

MaciejG604:
Fixes #2473 and #2426

After merge:
We'll need to make changes to the docs of known toolkits to use 'default'


modules/build/src/main/scala/scala/build/internal/util/WarningMessages.scala
```diff
@@ -123,4 +123,11 @@ object WarningMessages {
 
   val mainScriptNameClashesWithAppWrapper =
     "Script file named 'main.sc' detected, keep in mind that accessing it from other scripts is impossible due to a clash of `main` symbols"
+
+  def deprecatedWarning(old: String, `new`: String) =
+    s"Using `$old` is deprecated, use `${`new`}` instead"
```
Gedochao: NIT use `'` instead of ` in command line logs
```suggestion
    s"Using '$old' is deprecated, use '${`new`}' instead"
```

modules/build/src/main/scala/scala/build/preprocessing/DeprecatedDirectives.scala
```diff
@@ -0,0 +1,96 @@
...
+    handler.keys.find(_.nameAliases.contains(key))
+      .toSeq
+      .flatMap(_.nameAliases)
+
+  private def allKeysFrom(handler: DirectiveHandler[_]): Seq[String] =
+    handler.keys.flatMap(_.nameAliases)
+
+  private val deprecatedCombinationsAndReplacements = Map[DirectiveTemplate, WarningAndReplacement](
+    DirectiveTemplate(Seq("lib"), None)  -> keyReplacement("dep")(deprecatedWarning("lib", "dep")),
+    DirectiveTemplate(Seq("libs"), None) -> keyReplacement("dep")(deprecatedWarning("libs", "dep")),
```
Gedochao:
```suggestion
    DirectiveTemplate(Seq("libs"), None) -> keyReplacement("deps")(deprecatedWarning("libs", "deps")),
```

modules/build/src/main/scala/scala/build/preprocessing/DeprecatedDirectives.scala
```diff
@@ -0,0 +1,96 @@
...
+    DirectiveTemplate(Seq("compileOnly.lib"), None) -> keyReplacement("compileOnly.dep")(
+      deprecatedWarning("compileOnly.lib", "compileOnly.dep")
+    ),
+    DirectiveTemplate(Seq("compileOnly.libs"), None) -> keyReplacement("compileOnly.dep")(
+      deprecatedWarning("compileOnly.libs", "compileOnly.dep")
+    ),
```
Gedochao:
```suggestion
    DirectiveTemplate(Seq("compileOnly.libs"), None) -> keyReplacement("compileOnly.deps")(
      deprecatedWarning("compileOnly.libs", "compileOnly.deps")
    ),
```

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -60,14 +63,26 @@ object ExtractedDirectives {
...
+        )
+
+        directives.getAst match
+          case uds: UsingDefs => uds.getUsingDefs.asScala.toSeq.map(toStrictDirective)
+          case ud: UsingDef   => Seq(toStrictDirective(ud))
+          case _              => Nil // There should be nothing else here than UsingDefs or UsingDef
+      }
 
-      Right(ExtractedDirectives(strictDirectives, directivesPositionOpt))
+      Right(ExtractedDirectives(strictDirectives.reverse, directivesPositionOpt))
```
Gedochao, line 85: ...wait, why `.reverse`?

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
```diff
@@ -60,14 +63,26 @@ object ExtractedDirectives {
...
+        def toStrictDirective(ud: UsingDef) = StrictDirective(
+          ud.getKey(),
+          toStrictValue(ud.getValue()),
+          ud.getPosition().getColumn()
+        )
+
+        directives.getAst match
+          case uds: UsingDefs => uds.getUsingDefs.asScala.toSeq.map(toStrictDirective)
+          case ud: UsingDef   => Seq(toStrictDirective(ud))
+          case _              => Nil // There should be nothing else here than UsingDefs or UsingDef
```
Gedochao: NIT
```suggestion
          case _              => Nil // There should be nothing else here other than UsingDefs or UsingDef
```

modules/options/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -654,6 +653,17 @@ object Artifacts {
     dependencies: Seq[Positioned[coursier.Dependency]],
     resolutionError: coursier.error.ResolutionError.Simple
   ) = resolutionError match {
+    // FIXME This if may not be enough when new toolkits are added
```
Gedochao: maybe add some context on what needs to be done here? potentially link to a new ticket?

modules/build/src/main/scala/scala/build/preprocessing/ExtractedDirectives.scala
MaciejG604: The new way of creating `StrictDirective` objects traverses the tree of using directives instead of calling `getFlattenedMap`. This is done to be able to write the start position of the key to the `StrictDirective` since with `getFlattenedMap` we lose this information (key is returned as a String with no binding to the tree nor its positions). 
Because of that it seems the order of returned `StrictDirective`s has been reversed, so to keep the old behaviour I used `.reverse` at the end. 

However this still changes the order slightly (probably only for `target` using directives) as can be seen in [this test output](https://github.com/VirtusLab/scala-cli/pull/2622/files#diff-dddbf7feed1e70ad65fb0129a911fb3868985b44225bce659f21134062154833)

modules/options/src/main/scala/scala/build/Artifacts.scala
MaciejG604: Actually I think my comment is not valid since the names of the artifacts for toolkits should always the same, no matter the library author. Am I right?

EDIT:
This is the case and can be easily checked in the method for parsing the values of options/directives in `Toolkit.scala`, since it always returns:

```
... dep"$org::${Constants.toolkitName}::$v,toolkit" ...
... dep"$org::${Constants.toolkitTestName}::$v,toolkit" ...
```

Gedochao: yeah, I think so.
Then it seems the `// FIXME` comment is invalid.

Gedochao (approved):

PR #2625, opened 2023-12-12
Update test framework versions

mbovel:


website/docs/commands/test.md
```diff
@@ -95,9 +95,9 @@ see [the `using` directives guide](../guides/introduction/using-directives.md#di
 In order to run tests with a test framework, add the framework dependency to your application.
 Some of the most popular test frameworks in Scala are:
 
-- [munit](https://scalameta.org/munit): `org.scalameta::munit::0.7.29`
-- [utest](https://github.com/com-lihaoyi/utest): `com.lihaoyi::utest::0.8.1`
-- [ScalaTest](https://www.scalatest.org): `org.scalatest::scalatest::3.2.15`
+- [munit](https://scalameta.org/munit): `org.scalameta::munit::1.0.0-M10`
```
Gedochao: I'd leave it at `0.7.29`, as `1.0.0-M10` is a milestone release (it's not considered a stable release just yet, really, even though it is already widely used)

Gedochao (commented):

BTW these changes are docs-only, so it'd be better to target them at the `stable` branch.
docs changes merged to `main` will only get updated on the website during the next release.

website/docs/commands/test.md
Gedochao: reverted this line to `0.7.29`

Gedochao (approved):

PR #2632, opened 2023-12-14
Downgrade Xcode on macos CI runners

MaciejG604:


modules/integration/src/test/scala/scala/cli/integration/PublishSetupTests.scala
```diff
@@ -89,6 +90,9 @@ class PublishSetupTests extends ScalaCliSuite {
       .toMap
   }
 
+  override def munitFlakyOK: Boolean =
+    TestUtil.isNativeCli && TestUtil.isCI && Properties.isMac && !TestUtil.isM1
```
Gedochao: I have mixed feelings about this, as if we add more flaky tests in this suite for other setups, this will have to be refactored... but eh, maybe it's ok for now.
Hopefully we won't get more flaky `publish setup` tests... 😕

Gedochao (approved):
LGTM

MaciejG604:
Marking tests as flaky should not be needed now and we can come back to using latest macos runners instead of the beta macos-13.
The root cause of the issue was a supposedly incompatible Xcode version on the runner, libsodiumjni was released using version 12.3, while the runners use 15 by default.
We should try to release an updated version of libsodium jni in the near future.

.github/workflows/ci.yml
```diff
@@ -288,6 +288,9 @@ jobs:
     - uses: VirtusLab/scala-cli-setup@5b1a6c5ca98b5642996812781f6ad9b2d5935f85
       with:
         jvm: "temurin:17"
+    - run: |
+        sudo xcode-select -s "/Applications/Xcode_13.1.0.app" || ( ls /Applications; exit 1 )
+        sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
```
Gedochao: can we put this in a script? since it seems we're reusing it a couple times.
even if we end up getting rid of it later.

PR #2650, opened 2023-12-21
Revert xcode version downgrade

MaciejG604:
Revert the changes that do not fix the flakiness of the tests, mark tests using libsodium as flaky


Gedochao (approved):
LGTM

PR #2652, opened 2023-12-21
Move `integration.TestInputs` to the test scope

Gedochao:
Given some gimmick in our Mill build configuration, IntelliJ had trouble finding the `TestInputs` class in the `integration` module. This is fixed when we move the class to test sources (the only place it was used was the `integration.docker-slim.test` module, so I now made it depend on the integration tests).

This requires putting an empty placeholder `package.scala` file in its stead to the main scope of the `integration` module, as otherwise SemanticDb complains about... well, no classes being compiled there (as `TestInputs`) was the only thing in there. An placeholder file solves that.

In case someone has an idea for a refactor, this is how the stacktrace looks like: 
```text
integration.semanticDbData java.nio.file.NoSuchFileException: /Users/pchabelski/IdeaProjects/scala-cli/out/integration/semanticDbData.dest/classes
    os.walk$stream$.attrs(ListOps.scala:205)
    os.walk$stream$.apply(ListOps.scala:166)
    os.walk$.apply(ListOps.scala:96)
    mill.scalalib.SemanticDbJavaModule$.copySemanticdbFiles(SemanticDbJavaModule.scala:224)
    mill.scalalib.ScalaModule.$anonfun$semanticDbData$6(ScalaModule.scala:585)
    mill.api.Result$Success.map(Result.scala:33)
    mill.api.Result$Success.map(Result.scala:32)
    mill.scalalib.ScalaModule.$anonfun$semanticDbData$2(ScalaModule.scala:584)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:71)
```
It's likely possible to refactor the `integration` module definition to get rid of the placeholder (and if not, we should raise an issue for it in the `Mill` repo), but for now I think the placeholder workaround is fine.


Gedochao:
Closing this for now, the gain is small and realistically we need a more global fix for the IDE integration

PR #2655, opened 2024-01-08
Add missing Scala.js mode aliases

Gedochao:
Follow-up to #2630 

Adds missing `fullLinkJS` and `fastLinkJS` Scala.js mode aliases.


MaciejG604 (approved):
LGTM

PR #2657, opened 2024-01-09
Prevent the toolkit `latest` deprecation warning from being logged more than once

Gedochao:
Adding the toolkit dependency with the deprecated `latest` version causes the warning to be logged no less than 6 times (sic!)
```bash
scala-cli --toolkit latest -e 'println(os.pwd)'
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Compiling project (Scala 3.3.1, JVM (17))
# Compiled project (Scala 3.3.1, JVM (17))
# ~/IdeaProjects/scala-cli-tests
```

This scales it down to a reasonable and desired once.

```bash
scala-cli --toolkit latest -e 'println(os.pwd)'
# Using 'latest' for toolkit is deprecated, use 'default' to get more stable behaviour:
#  --toolkit default
# Compiling project (Scala 3.3.1, JVM (17))
# Compiled project (Scala 3.3.1, JVM (17))
# ~/IdeaProjects/scala-cli-tests
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -12,6 +12,7 @@ import dependency.parser.DependencyParser
 
 import java.io.{File, InputStream}
 import java.nio.file.Paths
+import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.build.EitherCps.{either, value}
 import scala.build.Ops.EitherOptOps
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -720,14 +722,18 @@ object SharedOptions {
         }
       }
 
+  // TODO: remove this state and prevent shared options from being created multiple times
```
tgodzik: Link to an issue?

tgodzik (approved):

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
Gedochao: https://github.com/VirtusLab/scala-cli/issues/2658

Gedochao: linked to the issue in the TODO comment

MaciejG604 (approved):

Issue #2621, opened 2023-12-08
Scala option `-Xplugin` doesn't work


SpriteOvO:
**Version(s)**
```
$ scala-cli -version
Scala CLI version: 1.0.5
Scala version (default): 3.3.1
Your Scala CLI version is outdated. The newest version is 1.1.0
It is recommended that you update Scala CLI through the same tool or method you used for its initial installation for avoiding the creation of outdated duplicates.
```

**Describe the bug**

`scala-cli run` doesn't load the specified compiler plugin.

**To Reproduce**

1. Compile a `divbyzero` compiler plugin by `scalac` according to the documentation [here](https://docs.scala-lang.org/overviews/plugins/index.html).

2. `scalac` outputs the expected compilation error.
   ```
   $ scalac -Xplugin:divbyzero.jar test.scala
   test.scala:3: error: definitely division by zero
     val amount = five / 0
                       ^
   one error found
   ```
3. `scala-cli` doesn't output the compilation error, which means the compiler plugin is not loaded.
   ```
   $ scala-cli --scala-version=2.13.12 --scalac-option="-Xplugin:divbyzero.jar" --server=false test.scala
   Exception in thread "main" java.lang.ExceptionInInitializerError
           at Test.main(test.scala)
   Caused by: java.lang.ArithmeticException: / by zero
           at Test$.<clinit>(test.scala:3)
           ... 1 more
   ```
4. More debug information
   ```
   $ scalac -Xplugin:divbyzero.jar -Xshow-phases test.scala | grep divbyzero
        divbyzero  10

   $ scalac -Xplugin:divbyzero.jar -Xplugin-list test.scala | grep divbyzero
   divbyzero - checks for division by zero

   $ scala-cli --scala-version=2.13.12 --scalac-option="-Xplugin:divbyzero.jar" --scalac-option="-Xshow-phases" --server=false test.scala | grep divbyzero
        divbyzero   9

   $ scala-cli --scala-version=2.13.12 --scalac-option="-Xplugin:divbyzero.jar" --scalac-option="-Xplugin-list" --server=false test.scala | grep divbyzero
   [error]  No main class found
   ```

**Expected behaviour**

`scala-cli` should work like `scalac` when used with `-Xplugin`.

sequencer:
[...]

PR #2666, opened 2024-01-12
Fix repeatable compiler options handling from the command line

### LINKED ISSUE #2653: scalac `-P` option is not passed to scalac from the CLI
**Version(s)**
1.1.0

**Describe the bug**
When trying to use the continuations compiler via scala-cli, the required `-P:continuations:enable` argument is not being passed scalac when provided to scala-cli via the CLI.

**To Reproduce**
```
scala-cli \
  --scala 2.12.2 \
  --compiler-plugin org.scala-lang.plugins:::scala-continuations-plugin:1.0.3 \
  --dependency org.scala-lang.plugins::scala-continuations-library:1.0.3 \
  -P:continuations:enable \
  myscript.sc
```

**Expected behaviour**
The code should compile and run, but compilation fails with `this code must be compiled with the Scala continuations plugin enabled`.  Running scala-cli with `-v -v` shows that the `-P` argument is not being passed to scalac.

**Workaround**
Setting the `-P` argument as a directive works correctly:

```
//> using option -P:continuations:enable
```

Gedochao:
Fixes #2653

Repeatable compiler options (`-P:*:*` and `-Xplugin:*`, specifically) were being filtered out on some code paths. 
This was happening as we do treat them differently, as they can't be shadowed. We have special handling for compiler options with keys which can be passed multiple times, as those cannot be overridden from the CLI when they're also being declared in a `using` directive.

@sequencer This may already be enough to cover your usecase from #2621. It still doesn't seem to work correctly with Bloop, but I can no longer replicate the problem with `--server=false` (although it may be more complicated in your `mill` build, so let's verify it after this gets merged).

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/util/ScalacOptionsUtil.scala
@@ -1,6 +1,7 @@
 package scala.cli.commands.util
 
 import scala.build.Logger
+import scala.build.options.ScalacOpt.filterScalacOptionKeys
 import scala.build.options.{ScalacOpt, ShadowingSeq}
 import scala.cli.commands.bloop.BloopExit
 import scala.cli.commands.default.LegacyScalaOptions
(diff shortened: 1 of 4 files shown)


tgodzik (approved):

sequencer:
Thank you @Gedochao, is possible to release a new minor version of scala-cli for this bug fix?

PR #2680, opened 2024-01-16
Ensure Java home is set when creating Bloop JSON

### LINKED ISSUE #2663: Platform should be set when non default JVM is used.
**Version(s)**
1.1.1

**Describe the bug**
Currently, if the user specifies JDK 11 as the JVM to use as in:

The platform section used for Bloop should show that, but it doesn't:
`"platform": { "name": "jvm", "config": { "options": [] }, "mainClass": [] },`

This means when running the wrong JVM will be used, the one Bloop runs on.

**To Reproduce**

```
//> using scala 3.3.1
//> using jvm adopt:1.11

@main
def main = 
    println(s"Hello java ${sys.props("java.home")}!")

```

prints:
`Hello java /usr/lib/jvm/graal17!` in my case, since I have graal17.

**Expected behaviour**
Prints:

`Hello java <coursier-java-path>!`

Gedochao:
Fixes #2663 
Note: since we only use Bloop for compilation (and not for running on the JVM), setting Java Home in the platform section of the Bloop project setup doesn't actually change anything.
This change is only done for data consistency, for when external tools may want to read the Bloop JSON (or expect the valid Java Home to be present i.e. in BSP messages, as tested in `BspTestDefinitions`)

### DIFF
--- a/modules/build/src/main/scala/scala/build/Project.scala
+++ b/modules/build/src/main/scala/scala/build/Project.scala
@@ -38,8 +38,14 @@ final case class Project(
       case (None, None) =>
         val baseJvmConf = bloopJvmPlatform
         baseJvmConf.copy(
-          // We don't pass jvm home here, because it applies only to java files compilation
-          config = baseJvmConf.config.copy(home = None)
+          // We don't pass jvm home to the general config
+          // it applies only to java files compilation
+          config = baseJvmConf.config.copy(home = None),
+          runtimeConfig = baseJvmConf.runtimeConfig
+            .orElse(Some(BloopConfig.JvmConfig.empty))
+            // the java home used here is not truly necessary, as we don't use Bloop for JVM runtime
+            // we only pass it here to make the Bloop JSON valid for BSP/external tools
+            .map(_.copy(home = javaHomeOpt.map(_.toNIO)))
         )
       case (Some(jsConfig), _) => BloopConfig.Platform.Js(config = jsConfig, mainClass = None)
       case (_, Some(nativeConfig)) =>


tgodzik (approved):
LGTM :rocket:

modules/build/src/main/scala/scala/build/Project.scala
```diff
@@ -38,8 +38,14 @@ final case class Project(
       case (None, None) =>
         val baseJvmConf = bloopJvmPlatform
         baseJvmConf.copy(
-          // We don't pass jvm home here, because it applies only to java files compilation
-          config = baseJvmConf.config.copy(home = None)
+          // We don't pass jvm home to the general config
+          // it applies only to java files compilation
+          config = baseJvmConf.config.copy(home = None),
```
Gedochao, line 43: It seems changing Java Home here makes Bloop actually use it while compiling, which breaks compatibility for when Bloop is being run on an older Java version (which isn't something we actively support, but neither should we actively break it if not strictly necessary)

modules/build/src/main/scala/scala/build/Project.scala
```diff
@@ -38,8 +38,14 @@ final case class Project(
...
+          runtimeConfig = baseJvmConf.runtimeConfig
+            .orElse(Some(BloopConfig.JvmConfig.empty))
+            // the java home used here is not truly necessary, as we don't use Bloop for JVM runtime
+            // we only pass it here to make the Bloop JSON valid for BSP/external tools
+            .map(_.copy(home = javaHomeOpt.map(_.toNIO)))
```
Gedochao, line 48: the problem with `runtimeConfig's  Java Home is that it isn't passed in the data tied to a build target in BSP, which limits its usefulness

Gedochao:
Moving this back to a draft, as it seems we need to figure out how to expose the Java Home in BSP without breaking compat tests

Gedochao:
Closing this for now, it doesn't seem like the way to go in the end.

PR #2692, opened 2024-01-18
Add a Scala/Java version agnostic option for setting the SemanticDB target root  and source root directories

### LINKED ISSUE #2646: It should be possible for SemanticDB `targetroot` to be different from `projectWorkspace/.scala-build`
**Version(s)**
1.1.0

**Describe the bug**
Currently, if we specify for Metals the flag for changing the location for `scala-build` the sourceroot option for semanticdb will also be changed to that directory. This might cause issues when reading semanticdb as sourceroot will not be a parent of the sources, which is expected.

**To Reproduce**
Open Metals after running setup-ide with changing build directory to some temporary folder.

**Expected behaviour**
Sourceroot should be sent to some parent directory.

https://github.com/scalameta/metals/issues/5748

Gedochao:
Fixes #2646

This adds `--semanticdb-targetroot` and `--semanticdb-sourceroot` options. 
`--semanticdb-targetroot` accepts a path to the directory, where SemanticDB files should be written.
`--semanticdb-sourceroot` accepts a path to the directory, from which the relative paths for SemanticDB files should be calculated.

Technically this could have already been done by passing the correct compiler options, but those were very much dependent on a number of things, including Scala/Java version and the manner the semanticdb plugin is passed to the compiler. 
The options added here provide unified syntax for all of these cases.

When the target directory is not set, the default follows compiler defaults, which means the semanticdb files are saved alongside compiled class files.

When the source directory is not set, the default follows the Scala CLI project root.
Please note that the Scala CLI project root is not necessarily a parent directory for all project sources.


modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -842,21 +844,25 @@ object Build {
...
+          ).map(ScalacOpt(_))
+        val semanticDbScalacOptions: Seq[ScalacOpt] =
+          if generateSemanticDbs then
+            semanticDbTargetRootOptions ++ (
+              if params.scalaVersion.startsWith("2.") then
+                Seq(
+                  "-Yrangepos",
+                  "-P:semanticdb:failures:warning",
+                  "-P:semanticdb:synthetics:on",
+                  s"-P:semanticdb:sourceroot:${inputs.workspace}"
```
tgodzik: Setting sourceroot, should be enough here, no? Do we need targetRoot? I don't think it's easy to set it manually since this will be somewhere deep in bloop directories

Gedochao: You may notice that we have been setting the `sourceroot` value to `inputs.workspace`, I did not change that.
However, `sourceroot` does not control where SemanticDB files go, it's merely `Used to relativize source file paths`. as per [SemanticDB ](https://github.com/scalameta/scalameta/blob/main/semanticdb/semanticdb3/guide.md)docs.
What controls SemanticDB files is `targetroot`, which defaults to the compiler's compiled classes directory (or in other words, the `-d` compiler option).
If we need SemanticDB files to go elsewhere, I don't think there is a way to do that without passing an explicit `targetroot`.

Gedochao: https://github.com/scalameta/scalameta/blob/main/semanticdb/semanticdb3/guide.md#producing-semanticdb

Gedochao: Wait... do you mean that you'd like to be able to override `sourceroot` as well?
I think I may have misunderstood where the issue is from Metals' perspective...
It's not just about where the SemanticDB files go, but also how the relative paths are calculated, is it?

Gedochao: I can give `sourceroot` the same treatment as `targetroot` is getting in this PR, where it could be overridden separately from `--workspace`, if that'd help.

I cannot change the `sourceroot` to "some parent directory", as because of how Scala CLI inputs may be spread, such a thing may not even exist (and we need to be internally consistent), not to mention looking for a parent directory between source directories may catch paths the user would like to isolate from his project. Remember that a project can be spread between separate drives.

The project root/workspace is thus assumed as a best approximation of what the `sourceroot` default can be.

tgodzik: I don't think targetroot would help, as we would still get relativized paths the files might end up in weird places. If we set sourceroot properly, we will get proper relative paths and the default targetroot should work fine.

Gedochao: I gave `sourceroot` the same treatment, so that there's a Java/Scala version agnostic way to override either or both. This should satisfy all cases, I think.

tgodzik (approved):
Thanks! LGTM

PR #2699, opened 2024-01-22
Update ScalaCliDockerFile

i10416:
The `stable` tag is not clear about which debian version it uses.

```sh
docker run --rm -it --entrypoint cat  virtuslab/scala-cli:1.1.2 /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

```sh
docker run --rm -it --entrypoint cat  virtuslab/scala-cli:1.0.0-RC1 /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### DIFF
--- a/.github/scripts/docker/ScalaCliDockerFile
+++ b/.github/scripts/docker/ScalaCliDockerFile
@@ -1,4 +1,4 @@
-FROM debian:stable-slim
+FROM debian:bookworm-slim
 RUN apt update && apt install build-essential libz-dev clang procps -y
 ADD scala-cli /usr/bin/
 RUN \


Gedochao:
@i10416 Any reason why we'd want to lock the docker image at the `bookworm-slim` tag? (Debian 12)
I believe it's [generally advised](https://www.debian.org/doc/manuals/debian-faq/choosing.en.html#s3.1.1) to use `stable` (`stable-slim` in our case).
If you need a particular debian version in your image, you can always add Scala CLI to it yourself, no?
I'd rather keep our image at `stable-slim` TBH...

i10416:
Thank you for feedback.

> Any reason why we'd want to lock the docker image at the bookworm-slim tag

It is because the current latest image uses bookworm-slim.

If you mean "why I want to pin the base image to a specific version(regardless of current stable)", it is because I had trouble finding the reason why my application, which is a small http server written in Scala Native built in a Docker container based on virtuslab/scala-cli, couldn't find libcrypto.so at runtime(not build time as I added libssl-dev to the image) and it turned out that runtime does not contain a specific version of openssl or libssl-dev. The version of openssl is different between Debian version and I thought it is better to use absolute Debian version or name rather than relative one in terms of reproducibility.
Ididn't know that practice you linked above and no one but me complains about it, so I close this issue.

Thank you.

PR #2709, opened 2024-02-02
Add `3.lts` and `lts` scala version option

### LINKED ISSUE #2645: Additional ways for specifying Scala version
It was suggested that maybe it would be useful to have additional `3.lts` and `3.next` to have up to date version for 3.3.x line and the  3.4.x+ one

3.lts should lead to the current Scala 3 lts version and 3.next to the latest version.

Specify it manually.

**Additional context**

https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/2645

### DIFF
--- a/project/deps.sc
+++ b/project/deps.sc
@@ -5,6 +5,7 @@ object Scala {
   def scala213     = "2.13.12"
   def runnerScala3 = "3.0.2" // the newest version that is compatible with all Scala 3.x versions
   def scala3       = "3.3.1"
+  def scala3Lts    = "3.3" //the full version should be resolved later
 
   // The Scala version used to build the CLI itself.
   def defaultInternal = sys.props.get("scala.version.internal").getOrElse(scala3)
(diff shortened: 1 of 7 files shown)


project/deps.sc
```diff
@@ -5,6 +5,7 @@ object Scala {
   def scala213     = "2.13.12"
   def runnerScala3 = "3.0.2" // the newest version that is compatible with all Scala 3.x versions
   def scala3       = "3.3.1"
+  def scala3Lts    = "3.3" //the full version should be resolved later
```
Gedochao, line 8: I wonder if we could avoid hardcoding this in the future... perhaps LTS compiler releases could be tagged in the POM smh 🤔 
just thinking out loud, should be ok for now

Gedochao (approved):
LGTM

PR #2710, opened 2024-02-02
remove `user.home` hack

### LINKED ISSUE #2585: Remove user.home hack when fix gets released in os-lib

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/2585

Upstream fix made `os.pwd` not fail when `os.home` is invalid. But I wonder if `os.home` shouldn't always be wrapped in a try clause, if it can be invalid.


Gedochao:
https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala#L2142-L2167
We should also adjust the relevant integration test to be sure it's not needed anymore, even without the env override.

Gedochao (approved):
LGTM

PR #2713, opened 2024-02-04
Add information about --preamble in assembly packaging documentation

spacebanana420:
This is a short modification to scala-cli's website documentation. The changed file in question is ```website/docs/commands/package.md```.

The changed area is the part about assembly packaging. I added information on how you can use ```--preamble=false``` to generate an assembly JAR that is not self-executable and therefore can be run with Java directly. This information is followed by a very similar example to the already existing one.

### DIFF
--- a/website/docs/commands/package.md
+++ b/website/docs/commands/package.md
@@ -127,7 +127,7 @@ Hello
 Hello
 -->
 
-By default assemblies are self-executable, just like the default package format. With the argument ```--preamble=false``` you can build an assembly that just contains the JAR and does not contain any built-in Bash code, and therefore can be launched directly with Java and is more portable:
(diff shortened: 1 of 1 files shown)


website/docs/commands/package.md
```diff
@@ -127,6 +127,25 @@ Hello
 Hello
 -->
 
+By default assemblies are self-executable, just like the default package format. With the argument ```--preamble=false``` you can build an assembly that just contains the JAR and does not contain any built-in Bash code, and therefore can be launched directly with Java and is more portable:
```
Gedochao:
````suggestion
By default assemblies are self-executable, just like the default package format. With the `--preamble=false` option, you can build an assembly that just contains the JAR and does not contain any built-in Bash code, and therefore can be launched directly with Java and is more portable:
````

Gedochao (approved):
Hey, thanks for the contribution! LGTM, I just commented on a minor nitpick.

PR #2721, opened 2024-02-08
Added info about repl with toolkit

yadavan88:
Hi @Gedochao 
It would be nice to add the information that the scala-cli REPL can be started with toolkit enabled. I could not find this info in the website. So, I created a new PR with this info in the REPL page. 
Please merge it if you think this is useful.

Thanks 
Yadu

### DIFF
--- a/website/docs/commands/repl.md
+++ b/website/docs/commands/repl.md
@@ -80,6 +80,29 @@ scala> :quit
 
 </ChainedSnippets>
 
+## Using Toolkit in REPL
+It is also possible to start the scala-cli REPL with [toolkit](https://scala-cli.virtuslab.org/docs/guides/introduction/toolkit/) enabled
+
+<ChainedSnippets>
+```bash ignore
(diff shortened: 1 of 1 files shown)


website/docs/commands/repl.md
```diff
@@ -80,6 +80,30 @@ scala> :quit
...
+scala> import os._
+                                                                                
+scala> os.pwd
+val res0: os.Path = /Users/yadukrishnan/test
+scala> :quit
+
+```
+</ChainedSnippets>
+
+Since we started the repl with toolkit enbled, we can use the libraries included in the toolkit direclty. In the above example, `os-lib` library from the toolkit is used to print the current path.
```
SethTisue:
```suggestion
Since we started the repl with toolkit enabled, we can use the libraries included in the toolkit directly. In the above example, the `os-lib` library from the toolkit is used to print the current path. 
```

Gedochao (approved):
LGTM

PR #2737, opened 2024-02-15
Add the ability to remap EsModule imports at link time

### LINKED ISSUE #2698: Offer a way to rewrite scalaJS imports at linktime
Proposal: Offer a way to rewrite scalaJS imports at linktime. 

Here's the prior art
https://github.com/armanbilge/scalajs-importmap

Some prior discussions for context
https://github.com/VirtusLab/scala-cli/discussions/1968#discussioncomment-5446977

https://github.com/VirtusLab/scala-cli/issues/297#issuecomment-1485161038

My goal is to see if I get agreement on 
1. This being desirable in scala-cli (I think it is) - i.e. that a PR turning the CI green with this would get merged
2. Agreement on a solution sketch

My aim, would be to be able to write something like;

project.scala
```
//> using scala 3.3.1
//> using platform js
//> using jsImportMap @stdlib/blas/base=https://cdn.jsdelivr.net/gh/stdlib-js/blas@esm/index.mjs
```

facade.scala
```
import scala.scalajs.js
import scala.scalajs.js.annotation.JSImport
import scala.scalajs.js.typedarray.Float64Array

@js.native
@JSImport("@stdlib/blas/base", JSImport.Default)
object blas extends BlasArrayOps

@js.native
trait BlasArrayOps extends js.Object:
    def dnrm2(N: Int, x: Float64Array, strideX: Int): Double = js.native
end BlasArrayOps

@main def run =
  println(blas.dnrm2(3,Float64Array(1.0, 2.0, 3.0), 0).mkString(","))
```
The expected behaviour, would be for scala-cli to rewrite the import at link time, and rely on the JS module system to resolve the dependancy. Hopefully, this sweeps away all of the bundling / build tool complexity of leveraging the wider JS ecosystem.

I intend to try and PR it myself, but I would doubt I can do it without (a lot of) help. My concrete aim in this issue is for the two top points to be agreed and clearly laid out, so that others have a chance to follow the why and how breadcrumbs, too, should they wish. 

Here's my best stab at conretely, what I think would need to happen. 

Firstly, Introduce https://github.com/armanbilge/scalajs-importmap as a dependancy to scala-cli. Then, we need to be able to feed it what it is, that we want to remap. The most "scala-cli" like approach, would appear to be to lean into the using directives. i.e. add a jsImportMap directive here?
https://github.com/VirtusLab/scala-cli/blob/main/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala

Then, need to figure out where in scala-cli, we have access to the JS IR (<-- help!), then wheel in Armans library and the values set in the new directive. 

Then... unit test (where? <--- help!) and iterate...

Quafadas:
WIP is a rather kind description. A cry for help, perhaps. 

This attempts to follow up on https://github.com/VirtusLab/scala-js-cli/pull/47, and close https://github.com/VirtusLab/scala-cli/issues/2698 - i.e. introduce a using directive and / or command line option, to enable the work done in scala-js-cli. 

This introduces a failing test, which I can run with;
`./mill -i 'integration.test' 'scala.cli.integration.RunTests213.remap imports' `

My current headscratcher, is that I don't seem to be able to modify the behaviour of scala-cli itself though. I've liberally put some printlns everywhere - which aren't printed. So I think I'm suffering from some sort of new contributor syndrome. 

Would anyone have a hint, on why the intregration test, could be ignoring my attempts to introspect it?

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
@@ -37,44 +37,58 @@ final case class ScalaJsOptions(
 
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.should)
-  jsCheckIr: Option[Boolean] = None,
+    jsCheckIr: Option[Boolean] = None,
 
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("Emit source maps")
   @Tag(tags.should)
     jsEmitSourceMaps: Boolean = false,
+
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("Set the destination path of source maps")
   @Tag(tags.should)
    jsSourceMapsPath: Option[String] = None,
+
+  @Group(HelpGroup.ScalaJs.toString)
+  @HelpMessage("A file relative to the root directory containing import maps for ES module imports")
+  @Tag(tags.experimental)
+    jsEsModuleImportMap: Option[String] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.should)
   @HelpMessage("Enable jsdom")
     jsDom: Option[Boolean] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.should)
   @HelpMessage("A header that will be added at the top of generated .js files")
     jsHeader: Option[String] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.implementation)
   @HelpMessage("Primitive Longs *may* be compiled as primitive JavaScript bigints")
     jsAllowBigIntsForLongs: Option[Boolean] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.implementation)
   @HelpMessage("Avoid class'es when using functions and prototypes has the same observable semantics.")
     jsAvoidClasses: Option[Boolean] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.implementation)
   @HelpMessage("Avoid lets and consts when using vars has the same observable semantics.")
     jsAvoidLetsAndConsts: Option[Boolean] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.implementation)
   @HelpMessage("The Scala.js module split style: fewestmodules, smallestmodules, smallmodulesfor")
     jsModuleSplitStyle: Option[String] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.implementation)
   @HelpMessage("Create as many small modules as possible for the classes in the passed packages and their subpackages.")
     jsSmallModuleForPackage: List[String] = Nil,
+
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.should)
   @HelpMessage("The Scala.js ECMA Script version: es5_1, es2015, es2016, es2017, es2018, es2019, es2020, es2021")
@@ -86,18 +100,21 @@ final case class ScalaJsOptions(
   @Tag(tags.implementation)
   @Hidden
     jsLinkerPath: Option[String] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage(s"Scala.js CLI version to use for linking (${Constants.scalaJsCliVersion} by default).")
   @ValueDescription("version")
   @Tag(tags.implementation)
   @Hidden
     jsCliVersion: Option[String] = None,
+
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("Scala.js CLI Java options")
   @Tag(tags.implementation)
   @ValueDescription("option")
   @Hidden
     jsCliJavaArg: List[String] = Nil,
+
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("Whether to run the Scala.js CLI on the JVM or using a native executable")
   @Tag(tags.implementation)
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
@@ -51,7 +51,7 @@ final case class ScalaJsOptions(
 
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("A file relative to the root directory containing import maps for ES module imports")
-  @Tag(tags.experimental)
+  @Tag(tags.restricted)
     jsEsModuleImportMap: Option[String] = None,
 
   @Group(HelpGroup.ScalaJs.toString)
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
@@ -87,13 +87,7 @@ final case class ScalaJs(
     )
 
     def absFilePath(pathStr: String): Either[ImportMapNotFound, Path] = {
-      Try {
-        os.Path(pathStr)
-      }.orElse(
-        Try{
-          os.Path(pathStr, base = os.pwd)
-        }
-      ).toEither.fold(ex =>
+      Try(os.Path(pathStr, os.pwd)).toEither.fold(ex =>
         Left(ImportMapNotFound(s"""Invalid path to EsImportMap. Please check your "using jsEsModuleImportMap xxxx" directive. Does this file exist $pathStr ?""", ex)),
         path =>
           os.isFile(path) && os.exists(path) match {
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
@@ -1,11 +1,15 @@
 package scala.build.preprocessing.directives
 
+import os.Path
+
 import scala.build.EitherCps.{either, value}
+import scala.build.Ops.EitherOptOps
 import scala.build.directives.*
 import scala.build.errors.BuildException
 import scala.build.options.{BuildOptions, JavaOpt, ScalaJsMode, ScalaJsOptions, ShadowingSeq}
 import scala.build.{Logger, Positioned, options}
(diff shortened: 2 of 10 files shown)


Gedochao:
One thing this made me realise is that we never bumped `scala-js-cli` to 1.15.0.1, so I'm doing it in #2738... (I know you're overriding the version in the test, just mentioning, still need to look into your code)

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -291,6 +291,50 @@ trait RunScalaJsTestDefinitions { _: RunTestDefinitions =>
...
+           |}""".stripMargin,
+      os.rel / importmapFile -> """{"imports": {"@stdlib/linspace": "https://cdn.skypack.dev/@stdlib/linspace"}}""".stripMargin
+    )
+
+    inputs.fromRoot { root =>
+      val absOutDir = root / outDir
+      val outFile = absOutDir / "main.js"
+      os.makeDir.all(absOutDir)
+      os.proc(TestUtil.cli, "--power","package", fileName, "--js", "--js-module-kind", "ESModule", "--js-cli-version", "1.15.0.1", "-o", outFile, "-f", "--js-es-module-import-map", "importmap.json" ).call(cwd = root).out.trim()
+      os.copy.over(outFile, os.home / "Code" / "main.js")
```
Gedochao: wait, what?
I mean, of course this will fail on the CI, there's nothing to overwrite.
why would you need to copy the output file?

Gedochao:
```
Compiling project (Scala 3.3.1, Scala.js 1.15.0)
Compiled project (Scala 3.3.1, Scala.js 1.15.0)
Downloading https://github.com/virtusLab/scala-js-cli/releases/download/v1.15.0.1/scala-js-ld-aarch64-apple-darwin.gz
Downloaded https://github.com/virtusLab/scala-js-cli/releases/download/v1.15.0.1/scala-js-ld-aarch64-apple-darwin.gz
Wrote /Users/pchabelski/IdeaProjects/scala-cli/out/integration/tmpDirBase.dest/working-dir/run-1335615095/test-2/out/main.js,
```
the package is being called correctly, btw.

Gedochao:
> My current headscratcher, is that I don't seem to be able to modify the behaviour of scala-cli itself though. I've liberally put some printlns everywhere - which aren't printed. So I think I'm suffering from some sort of new contributor syndrome.

Can you point me at where you're trying to introspect, which prints don't show?

> Would anyone have a hint, on why the intregration test, could be ignoring my attempts to introspect it?

a wild guess is that you are printing on `stdout`, while `stdout` is being piped in the test...
https://github.com/com-lihaoyi/os-lib/blob/6603eb0e7434c86721806f2c1922452d9d9e49fc/os/src-jvm/ProcessOps.scala#L63-L65
If you want to print the logs when running the integration test (as opposed to when you call a freshly built launcher directly), then you need 
```scala
os.proc(...).call(cwd = root, stdout = os.Inherit)
```
as `stdout` defaults to `os.Pipe`.
alternatively, you can print on `stderr`, which defaults to `os.Inherit`.
No idea if this was the question 😅

Gedochao:
beyond the above, I'd recommend debugging the launcher directly, instead of printing stuff in the integration tests.
```bash
./mill -i integration.test.jvm 'scala.cli.integration.RunTests213.remap imports' --debug
# (...)
# --debug option has been passed. Listening for transport dt_socket at address: 5005
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
Quafadas: Yeah, this was me copying the file to a persistent location locally - scala-cli is quite good about tidying up after its test runs, and I wanted to inspect it manually. This will need to go before a merge, clearly.

Quafadas:
Would you have a hint on how to fix this?

I think it's somehow checking the documentation...

image

Quafadas:
Another Q: 

I built a launcher with;
`./mill -i show cli.nativeImage `

image

But it doesn't like the arguments I'm giving it. However, the integration test runs through, and it tests, exactly that flag? I'm not totally sure how that can happen?

Gedochao:
> Would you have a hint on how to fix this?

Yes.
https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md#rules-for-a-well-formed-pr
You need to regenerate the reference docs, among other things.
```bash
```

Also, you may want to rebase on top of `main`, rather than cherry picking individual commits.

Quafadas:
@Gedochao @armanbilge  This actually ... works ... to my satisfaction. 

The happy path works well.

I have a question on error messaging, which is that if I (deliberately) break it, for example by entering a bad file path to my importmap, it produces the error message I'd expect, but it's hidden at the top of the error stack, above scala-cli help. My eyeballs are where the green line is, but the message I need, is where the red line is. 

Given that I'm naively throwing an exception, could I be engaging with scala-cli's error handling the wrong way?

Screenshot 2024-02-16 at 09 48 30

Gedochao:
> Given that I'm naively throwing an exception, could I be engaging with scala-cli's error handling the wrong way?

@Quafadas Rather than throwing an exception, you need to return a `Left[BuildException]`. 
You'll need to define your own exception type extending `BuildException` (refer to the existing errors that are returned).

Quafadas:
- [x] Tighten error messages
- [x] Tests for error conditions
- [x] Tests for happy path
- [x] Manual testing
- [x] Squash
- [x] Docs
- [x] CI green

Gedochao:
```bash
./mill -i __.fix
```
you need this to satisfy the `checks` CI job.

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
```diff
@@ -76,14 +83,37 @@ final case class ScalaJs(
...
-      scalaJsOptions = scalaJsOptions
+
+    def absFilePath(pathStr: String): Either[ImportMapNotFound, Path] = {
+      Try {
+        os.Path(pathStr)
+      }.orElse(
+        Try{
+          os.Path(pathStr, base = os.pwd)
+        }
+      ).toEither.fold(ex =>
```
Gedochao:
```suggestion
      Try(os.Path(pathStr, os.pwd)).toEither.fold(ex =>
```

modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
```diff
@@ -37,44 +37,58 @@ final case class ScalaJsOptions(
...
     jsEmitSourceMaps: Boolean = false,
+
   @Group(HelpGroup.ScalaJs.toString)
   @HelpMessage("Set the destination path of source maps")
   @Tag(tags.should)
    jsSourceMapsPath: Option[String] = None,
+
+  @Group(HelpGroup.ScalaJs.toString)
+  @HelpMessage("A file relative to the root directory containing import maps for ES module imports")
+  @Tag(tags.experimental)
```
Gedochao: Should be fine to tag this as restricted
```suggestion
  @Tag(tags.restricted)
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -291,6 +291,152 @@ trait RunScalaJsTestDefinitions { _: RunTestDefinitions =>
...
+      val outFile = absOutDir / "main.js"
+      os.makeDir.all(absOutDir)
+      os.proc(
+        TestUtil.cli,
+        "--power",
+        "package",
+        fileName,
+        "--js",
+        "--js-module-kind", "ESModule",
+        "--js-cli-version", "1.15.0.1",
```
Gedochao: I think this shouldn't be needed after the rebase, `1.15.0.1` should now be used by default. 
It'd be best if the test picked up subsequent versions automatically in the future.
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -291,6 +291,152 @@ trait RunScalaJsTestDefinitions { _: RunTestDefinitions =>
...
+      val outFile = absOutDir / "main.js"
+      os.makeDir.all(absOutDir)
+      val result = os.proc(
+        TestUtil.cli,
+        "--power",
+        "package",
+        fileName,
+        "--js",
+        "--js-module-kind", "ESModule",
+        "--js-cli-version", "1.15.0.1",
```
Gedochao:
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -291,6 +291,152 @@ trait RunScalaJsTestDefinitions { _: RunTestDefinitions =>
...
+      val outFile = absOutDir / "main.js"
+      os.makeDir.all(absOutDir)
+      os.proc(
+        TestUtil.cli,
+        "--power",
+        "package",
+        fileName,
+        "--js",
+        "--js-module-kind", "ESModule",
+        "--js-cli-version", "1.15.0.1",
```
Gedochao:
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
Quafadas: Agreed

modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
Quafadas: agreed

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaJs.scala
Quafadas: I originally had the idea, to be able to specify a file at any location in a sfilesystem. I think that isn't a good idea, and that the import map shoudl be part of a repository. So agreed.

Gedochao (approved):
LGTM

PR #2753, opened 2024-02-22
Bump Scala 2.12 to 2.12.19

Gedochao:


modules/integration/src/test/scala/scala/cli/integration/ReplTestDefinitions.scala
```diff
@@ -51,19 +60,20 @@ abstract class ReplTestDefinitions(val scalaVersionOpt: Option[String])
...
+      }
     }
   }
 
-  test("ammonite") {
-    ammoniteTest()
-  }
-
-  test("ammonite scalapy") {
+  def ammoniteScalapyTest(useMaxAmmoniteScalaVersion: Boolean): Unit = {
```
kasiaMarek, line 75: Why do you have this `useMaxAmmoniteScalaVersion` flag? Is there a use case where you want to set it to `false`?

kasiaMarek (approved):

modules/integration/src/test/scala/scala/cli/integration/ReplTestDefinitions.scala
Gedochao: Yes. Check `ReplTestsDefault`.
When the default Scala version for users is higher than the max supported for ammonite, we still want to run the test to make sure it falls back to the max gracefully, rather than crashing.

kasiaMarek: Oh, I see what you mean. Somehow I missed that, thanks.

PR #2781, opened 2024-03-06
Fix handling for `-Xlint:help`

### LINKED ISSUE #2779: Cannot always print help from compiler
Version 1.1.3

Using `-X` is ok, but `-Xlint:help` not.

```
➜ sandbox scala-cli compile -S 2 -X
Usage: scalac <options> <source files>

Available advanced options:
  -Xasync                          Enable the async phase for scala.async.Async.{async,await}. [false]
...


➜ sandbox scala-cli compile -S 2 -Xlint:help
[error]  No inputs provided (expected files with .scala, .sc, .java or .md extensions, and / or directories).
➜ sandbox
```

It works when using plain `scala-cli` without the `compile` subcommand.

Gedochao:
Fixes #2779 
Also, improves the way we test compiler print options.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
@@ -58,7 +58,13 @@ object ScalacOptions {
     * message instead.
     */
   val ScalacPrintOptions: Set[String] =
-    scalacOptionsPurePrefixes ++ Set("-help", "-Xshow-phases", "-Xplugin-list", "-Vphases")
+    scalacOptionsPurePrefixes ++ Set(
+      "-help",
+      "-Xshow-phases",
+      "-Xplugin-list",
+      "-Xlint:help",
+      "-Vphases"
+    )
 
   /** This includes all the scalac options which are redirected to native Scala CLI options. */
   val ScalaCliRedirectedOptions = Set(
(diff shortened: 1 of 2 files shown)


kasiaMarek (approved):

modules/cli/src/main/scala/scala/cli/commands/shared/ScalacOptions.scala
```diff
@@ -58,7 +58,13 @@ object ScalacOptions {
     * message instead.
     */
   val ScalacPrintOptions: Set[String] =
-    scalacOptionsPurePrefixes ++ Set("-help", "-Xshow-phases", "-Xplugin-list", "-Vphases")
+    scalacOptionsPurePrefixes ++ Set(
+      "-help",
+      "-Xshow-phases",
+      "-Xplugin-list",
+      "-Xlint:help",
+      "-Vphases"
```
lrytz, line 66: There are others, like `-opt:help`, `-Xsource:help`, `-Xmixin-force-forwarders:help`. Would it be possible to forward any `...:help`?

Or a more general change where the compiler is invoked in case there are no source files?

Somewhat related

```
➜ sandbox scala-cli compile --server=false -S 2 -Vprint-args a.txt A.scala
[error]  a.txt: unrecognized source type (expected .scala or .sc extension, or a directory).
If a.txt is meant to be treated as a script, add a shebang header in its top line.
When a shebang header is provided, the script can then be run with the 'shebang' sub-command, even if no file extension is present.
  scala-cli shebang a.txt
```

PR #2784, opened 2024-03-07
Fix `--semanticdb-targetroot` & `--semanticdb-sourceroot` for scripts

### LINKED ISSUE #2780: Incorrect behaviour of `--semantic-db-source-root` flag for scripts
**Version(s)**
`1.1.3`

**To Reproduce**
Having the following structure:
```
~/tmp2
~/tmp
  -/demo.sc
```
Run `scala-cli compile --workspace ~/tmp2 --semantic-db-source-root ~/tmp --semantic-db ~/tmp`.

See there is no `~/tmp2/.scala-build/tmp2_.../classes/main/META-INF`, semanticdb seems to be generated for the generated `demo.scala` at `~/tmp2/.scala-build/tmp2_.../src_generated/main/demo.scala.semanticdb` instead.

**Expected behaviour**
semanticdb for `demo.sc` is generated at `~/tmp2/.scala-build/tmp2_.../classes/main/META-INF/semanticdb/demo.sc.semanticdb`

Gedochao:
Fixes #2780

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -252,7 +252,8 @@ object Build {
             logger,
             inputs.workspace,
             updateSemanticDbs = true,
-            scalaVersion = sv
+            scalaVersion = sv,
+            buildOptions = build.options
           ).left.foreach(_.foreach(logger.message(_)))
       case _ =>
     }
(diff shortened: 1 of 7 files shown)


Gedochao:
`--semanticdb-targetroot` & `--semanticdb-sourceroot` are already working correctly with scripts.

This PR also now includes a fix to support the unusual case of a `--workspace` separate from `--semanticdb-sourceroot`, which is needed in Metals. However, the fix currently only works on Scala 3.

We may want to try to fix it for Scala 2.13 & 2.12, but I think it should perhaps be done in a separate PR, so that we don't grow the changes in this one.

kasiaMarek (approved):

PR #2800, opened 2024-03-14
Add -unchecked to the list of options that don't require -O

joan38:


Gedochao:
I can see the reference docs were broken, so I pushed a commit to fix it.
When a change affects the reference docs, please remember to regenerate them with the following command:
```bash
```

Gedochao (approved):
LGTM

joan38:
Thanks!

PR #2815, opened 2024-03-25
Adjust actionable diagnostics for scripts

### LINKED ISSUE #2812: Insert missing pat-mat cases action changes wrong lines
**Version(s)**
1.2.0

**Describe the bug**
Given a following code:
```scala
//> using scala 3.3.3

enum Animal {
  case Cat(name: String)
  case Dog(name: String)
}

object Main {
  def foo(a: Animal) = {
    a match
      case Animal.Cat(name) => println(name)

  }
}
```
scala3 compiler reports pattern not exhaustive diagnostic together with an actionable item to fix it. Applying that actionable adds the missing case statement into an incorrect position. This works correctly when running from within an sbt project.

Result of application of the actionable:
```scala
//> using scala 3.3.3

enum Animal {
  case Cat(name: String)
  case Dog(name: String)
}

object Main {
  def foo(a: Animal) = {
    a match
      case Animal.Cat(name) => println(name)

  }
}

      case Animal.Dog(_) => ???
```

**To Reproduce**
Paste the snippet into `*.sc` file and import it.

**Expected behaviour**
I would expect the actionable to insert the missing case in the correct place as it does in case of sbt project.

rochala:
Fixes https://github.com/VirtusLab/scala-cli/issues/2812

### DIFF
--- a/modules/build/src/main/scala/scala/build/bsp/BspClient.scala
+++ b/modules/build/src/main/scala/scala/build/bsp/BspClient.scala
@@ -2,7 +2,7 @@ package scala.build.bsp
 
 import ch.epfl.scala.bsp4j.{ScalaAction, ScalaDiagnostic, ScalaTextEdit, ScalaWorkspaceEdit}
 import ch.epfl.scala.bsp4j as b
-import com.google.gson.Gson
+import com.google.gson.{Gson, JsonElement}
 
 import java.lang.Boolean as JBoolean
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

MaciejG604 (approved):
🎊

Gedochao (approved):
LGTM

PR #2819, opened 2024-03-28
Fix publishing of `runner` & `test-runner` artifacts

### LINKED ISSUE #2809: Possible change in dependency publication method or in dependency solving method
**Version(s)**
1.2.0

**Describe the bug**
Sorry for the possibly vague and confused issue title, but I'm confused as well.
In the typelevel toolkit CI we rely on the `org.virtuslab.scala-cli::cli` [dependency](https://github.com/typelevel/toolkit/blob/main/build.sbt#L59) to run unit tests against all the (versions, platform) combinations we support [invoking scala cli directly](https://github.com/typelevel/toolkit/blob/main/tests/shared/src/test/scala/org/typelevel/toolkit/ScalaCliProcess.scala#L33) and not relying on the command line launcher, a process that I also described [here](https://toniogela.dev/testing-typelevel-toolkit/) as it was really fun to implement.

Since the `cli_2.13` artifact isn't available for most of the scala cli versions we've always relied on `cli_3`, using `CrossVersion.for2_13Use3` ([here](https://github.com/typelevel/toolkit/blob/main/build.sbt#L59))

Since 1.2.0, we're getting [a dependency resolution error](https://github.com/typelevel/toolkit/actions/runs/8222884766/job/22882084451):
```
 [error]  Error downloading org.virtuslab.scala-cli:test-runner_2.13:1.2.0
  not found: /home/runner/.ivy2/local/org.virtuslab.scala-cli/test-runner_2.13/1.2.0/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/org/virtuslab/scala-cli/test-runner_2.13/1.2.0/test-runner_2.13-1.2.0.pom
  No fallback URL found
```
Taking a look [on maven central](https://repo1.maven.org/maven2/org/virtuslab/scala-cli/test-runner_2.13/) I've noticed that `test-runner_2.13` isn't published for 1.2.0, but neither is `runner_2.13`, so I can't really tell where the dependency resolution problem is (or even if there is one). What confuses me the most is what I wrote in the `To Reproduce` section later.

**To Reproduce**
I've minimized the (possible) bug here:

(1.1.3, calling "compile") WORKS:
```scala
//> using scala 2.13.12
//> using option -Ytasty-reader
//> using dep org.virtuslab.scala-cli:cli_3:1.1.3

import scala.cli.ScalaCli
object Main {
    def main(args: Array[String]): Unit = ScalaCli.main(Array("compile", "."))
}
```
(1.1.3, calling "test") WORKS:
```scala
//> using scala 2.13.12
//> using option -Ytasty-reader
//> using dep org.virtuslab.scala-cli:cli_3:1.1.3

import scala.cli.ScalaCli
object Main {
    def main(args: Array[String]): Unit = ScalaCli.main(Array("test", "."))
}
```

(1.2.0, calling "compile") WORKS:
```scala
//> using scala 2.13.12
//> using option -Ytasty-reader
//> using dep org.virtuslab.scala-cli:cli_3:1.2.0

import scala.cli.ScalaCli
object Main {
    def main(args: Array[String]): Unit = ScalaCli.main(Array("compile", "."))
}
```

(1.2.0, calling "test") FAILS:
```scala
//> using scala 2.13.12
//> using option -Ytasty-reader
//> using dep org.virtuslab.scala-cli:cli_3:1.2.0

import scala.cli.ScalaCli
object Main {
    def main(args: Array[String]): Unit = ScalaCli.main(Array("test", "."))
}
```
and the reported error is
```
Downloading 2 dependencies and one internal dependency
[error]  Error downloading org.virtuslab.scala-cli:test-runner_2.13:1.2.0
  not found: /Users/toniogela/.ivy2/local/org.virtuslab.scala-cli/test-runner_2.13/1.2.0/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/org/virtuslab/scala-cli/test-runner_2.13/1.2.0/test-runner_2.13-1.2.0.pom
  No fallback URL found
```

I should also report that switching to the latest scala 3 version makes the four snippet compile and run successfully, so it seems related to the way deps are resolved for scala 2.13, but just for tests as "compile" works correctly despite `runner_2.13` version 1.2.0 was [NOT published](https://repo1.maven.org/maven2/org/virtuslab/scala-cli/runner_2.13/).

Also, ofc, running `scala-cli test X` using the command line launcher where X is a `.test.scala` file containing the directive `//> using scala 2.13.12` works flawlessly (and this adds even more confusion)

**Expected behaviour**
I won't claim that I expect all four aforementioned snippets to work flawlessly, as this is clearly a non-standard use case that relies on dependencies that weren't meant to be used by the general public, but I would love to be enlightened about what it might be the possible root cause, and if this is due to a design decision that we should cope with to keep testing the typelevel toolkit.

Thanks in advance!
//cc @armanbilge

Gedochao:
Fixes #2809

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1256,7 +1256,7 @@ jobs:
       env:
         PGP_SECRET: ${{ secrets.PGP_SECRET }}
     - run: ./mill -i ci.setShouldPublish
-    - run: ./mill -i publishSonatype '{__[],_}.publishArtifacts'
+    - run: ./mill -i publishSonatype '{__[],_,test-runner[2.13.13],test-runner[2.12.19],runner[2.13.13],runner[2.12.19]}.publishArtifacts'
       if: env.SHOULD_PUBLISH == 'true'
       env:
         PGP_PASSWORD: ${{ secrets.PGP_PASSPHRASE }}
--- a/build.sc
+++ b/build.sc
@@ -93,7 +93,7 @@ object `build-module` extends Cross[Build](Scala.scala3MainVersions)
     with CrossScalaDefaultToInternal
 object runner        extends Cross[Runner](Scala.runnerScalaVersions) with CrossScalaDefaultToRunner
 object `test-runner` extends Cross[TestRunner](Scala.testRunnerScalaVersions)
-    with CrossScalaDefaultToInternal
+    with CrossScalaDefaultToRunner
 object `tasty-lib` extends Cross[TastyLib](Scala.all) with CrossScalaDefaultToInternal
 // Runtime classes used within native image on Scala 3 replacing runtime from Scala
 object `scala3-runtime` extends Cross[Scala3Runtime](Scala.scala3MainVersions)
--- a/build.sc
+++ b/build.sc
@@ -39,8 +39,17 @@ implicit def millModuleBasePath: define.Ctx.BasePath =
 
 object cli extends Cross[Cli](Scala.scala3MainVersions) with CrossScalaDefaultToInternal
 
-trait CrossScalaDefaultToInternal { _: mill.define.Cross[_] =>
-  def defaultCrossSegments = Seq(Scala.defaultInternal)
+trait CrossScalaDefault { _: mill.define.Cross[_] =>
+  def crossScalaDefaultVersion: String
+  def defaultCrossSegments = Seq(crossScalaDefaultVersion)
+}
+
+trait CrossScalaDefaultToInternal extends CrossScalaDefault { _: mill.define.Cross[_] =>
+  def crossScalaDefaultVersion: String = Scala.defaultInternal
+}
+
+trait CrossScalaDefaultToRunner extends CrossScalaDefault { _: mill.define.Cross[_] =>
+  def crossScalaDefaultVersion: String = Scala.runnerScala3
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1256,7 +1256,7 @@ jobs:
       env:
         PGP_SECRET: ${{ secrets.PGP_SECRET }}
     - run: ./mill -i ci.setShouldPublish
-    - run: ./mill -i publishSonatype '{__[],_}.publishArtifacts'
+    - run: ./mill -i publishSonatype '{__[],_,test-runner[2.13.13],test-runner[2.12.19],runner[2.13.13],runner[2.12.19]}.publishArtifacts'
```
Gedochao, line 1259: I wish there was a simpler way to do this

tgodzik: My heart bleeds :<

Could we add a custom task that does it instead?

tgodzik (approved):

.github/workflows/ci.yml
Gedochao: perhaps.
I'll try to find the time to refactor this in a separate PR later.

PR #2820, opened 2024-03-28
Add release notes for `v1.2.1`

Gedochao:
release issue: https://github.com/VirtusLab/scala-cli/issues/2790

### DIFF
--- a/website/docs/release_notes.md
+++ b/website/docs/release_notes.md
@@ -8,6 +8,74 @@ import ReactPlayer from 'react-player'
 
 # Release notes
 
+## [v1.2.1](https://github.com/VirtusLab/scala-cli/releases/tag/v1.2.1)
+
+### Support for Scala 3.4.1
+This Scala CLI version adds support for Scala 3.4.1.
+
+```bash
+scala-cli version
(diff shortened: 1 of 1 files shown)


website/docs/release_notes.md
```diff
@@ -8,6 +8,47 @@ import ReactPlayer from 'react-player'
 
 # Release notes
 
+## [v1.2.1](https://github.com/VirtusLab/scala-cli/releases/tag/v1.2.1)
+
+### Support for Scala.js 1.16.0
```
tgodzik, line 34: Don't we also start supporting 0.5.0 via Bloop upgrade? Maybe worth mentioning, probably also worth a test

tgodzik (changes requested):

website/docs/release_notes.md
Gedochao: 0.5.0? of what?

tgodzik (approved):
Ok, it seems scala native 0.5 works at least in command line with older scala cli versions so maybe not worth mentioning it.

tgodzik (approved):

Issue #2841, opened 2024-04-07
Add support for all kinds of JVM args via `.scala-jvmopts` file and `JAVA_OPTS` env var, not only java properties


nehaev:
**Describe what needs to be done and why**
I'm trying to run a local Spark job as a shebang script. On modern JDKs, Spark requires certain args to be provided to JVM  (e.g., `--add-exports java.base/sun.nio.ch=ALL-UNNAMED`). The current implementation of scala-cli can only propagate java properties (using `JAVA_OPTS` or `.scala-jvmopts`), but not other vm args. I receive the following warning:

```
Warning: Only java properties are supported in .scala-jvmopts file. Other options are ignored: --add-exports java.base/sun.nio.ch=ALL-UNNAMED
```

It's not the case for sbt, for example, as it supports such kind of args both in `JAVA_OPTS` and in `.jvmopts` file.

**Describe alternatives you've considered**
I've considered `scala-cli --power run --spark-standalone`, it starts to download JDK 8, but it's not what I want.

It's not about running Spark at all costs, or inability to downgrade JVM. I believe scala-cli should be able to propagate all kinds of args to JVM.

Gedochao:
JVM args other than Java properties indeed aren't supported in the `JAVA_OPTS` env at the moment, as well as in the `.scala-jvmopts` file.

A warning was missing for when invalid JVM args are being passed in the env vars, I linked a quick fix PR (I just add the missing warning, so it doesn't solve this feature request).

Java options can instead be passed via command line options, [check the relevant doc](https://scala-cli.virtuslab.org/docs/commands/run#jvm-options).

```bash
scala-cli . --java-opt --add-exports --java-opt java.base/sun.nio.ch=ALL-UNNAMED
```

PR #2843, opened 2024-04-09
Log a warning when invalid java properties are being passed by env vars

Gedochao:
Relevant for #2841

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -184,7 +184,7 @@ object ScalaCli {
       }
       if otherOpts.nonEmpty then
         System.err.println(
-          s"Warning: Only java properties are supported in .scala-jvmopts file. Other options are ignored: ${otherOpts.mkString(", ")} "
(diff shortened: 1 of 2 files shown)


kasiaMarek (approved):
lgtm

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -201,14 +201,21 @@ object ScalaCli {
...
+          opt.stripPrefix("-D").split("=", 2) match {
+            case Array(key, value) =>
+              System.setProperty(key, value)
+              None
+            case ignored => Some(ignored) // non-property opts are ignored here
+          }
+        }.flatten
+    if ignoredJavaOpts.nonEmpty then
+      System.err.println(
+        s"Warning: Only java properties are supported in JAVA_OPTS and JDK_JAVA_OPTIONS environment variables. Other options are ignored: ${ignoredJavaOpts.mkString(", ")}"
```
lrytz, line 217: I'm now getting this message every time I invoke scala-cli. What would be the recommendation? I would like to keep that env var because sbt supports it (I'm using `-Xmx1536m -Xms1536m -Xss5m ...`).

PR #2888, opened 2024-05-06
Adjust TASTY bump warnings to respect overridden Scala version defaults

### LINKED ISSUE #2881: error when cleaning up TASTy of script file with future TASTy version
**Version(s)**
1.3.0

**Describe the bug**

**To Reproduce**
```
echo 'println(23)' >> local/foo.sc

scala-cli --cli-default-scala-version 3.5.0-RC1-bin-20240429-e2c456f-NIGHTLY \
  local/foo.sc \
  --server=false

Scala CLI (v1.3.0) cannot post process TASTY files from Scala 3.5.0-RC1-bin-20240429-e2c456f-NIGHTLY.
This is not a fatal error since post processing only cleans up source paths in TASTY file.
It should not affect your application.
You may be getting this warning because you are using a newer version of Scala than the one supported by Scala CLI (v1.3.0).
Make sure your Scala CLI is up-to-date.
You may need to wait for 3.5.0-RC1-bin-20240429-e2c456f-NIGHTLY support in a future version of Scala CLI.

23
```

**Expected behaviour**
I understand that this error is a problem, but if we want to do the strategy for SIP-46 of downloading a launcher then this will occur whenever we switch to a new minor release (because Scala CLI is not built yet with the new minor)

Gedochao:
Fixed #2881


Gedochao:
cc @bishabosha

bishabosha (approved):

tgodzik (approved):

PR #2908, opened 2024-05-14
Fix support of multiline shebang

### LINKED ISSUE #2901: script files (.sc) do not strip multiline shebang
**Version(s)**
1.3.1

**Describe the bug**
If a script file (`.sc`) has a multiline shebang, then only the first line is stripped. Meaning the rest of the shebang is parsed as code by the Scala Compiler, leading to errors.

**To Reproduce**

make a file `shebangTest.sc` with the following content, and then run it with `scala-cli run shebangTest.sc`

```scala
#!/usr/bin/env fake-program-to-test-shebang-removal
# comment
STUFF=nada
!#
// everything above this point should be ignored by the compiler
  args.zipWithIndex.foreach { case (arg,i) => printf("arg %d: [%s]\n",i,arg) }
  System.err.printf("mainClassFromStack: %s\n",mainFromStack)
  assert(mainFromStack.contains("sheBang"),s"fromStack[$mainFromStack]")

  lazy val mainFromStack:String = {
    val result = new java.io.StringWriter()
    new RuntimeException("stack").printStackTrace(new java.io.PrintWriter(result))
    val stack = result.toString.split("[\r\n]+").toList
    if verbose then for( s <- stack ){ System.err.printf("[%s]\n",s) }
    stack.filter { str => str.contains(".main(") }.map {
      // derive main class name from stack when main object is NOT declared in source
      _.replaceAll("[.].*","").
      replaceAll("\\s+at\\s+","")
    }.distinct.take(1).mkString("")
  }

  lazy val verbose = Option(System.getenv("DOTC_VERBOSE")) match
    case None => false
    case _ => true
```

you should get errors like so:
```scala
-- Error: /Users/jamie/workspace/dotty/compiler/test-resources/scripting/.scala-build/scripting_6199f560ec-7ad3f4e164/src_generated/main/sheBang.scala:8:0
8 |# comment
  |^
  |end of statement expected but '#' found
-- [E006] Not Found Error: /Users/jamie/workspace/dotty/compiler/test-resources/scripting/.scala-build/scripting_6199f560ec-7ad3f4e164/src_generated/main/sheBang.scala:9:0
9 |STUFF=nada
  |^^^^^
  |Not found: STUFF
  |
  | longer explanation available when compiling with `-explain`
-- [E006] Not Found Error: /Users/jamie/workspace/dotty/compiler/test-resources/scripting/.scala-build/scripting_6199f560ec-7ad3f4e164/src_generated/main/sheBang.scala:9:6
9 |STUFF=nada
  |      ^^^^
  |      Not found: nada
  |
  | longer explanation available when compiling with `-explain`
-- [E006] Not Found Error: /Users/jamie/workspace/dotty/compiler/test-resources/scripting/.scala-build/scripting_6199f560ec-7ad3f4e164/src_generated/main/sheBang.scala:10:0
10 |!#
   |^^
   |Not found: !# - did you mean !=? or perhaps != or ## or ##?
   |
   | longer explanation available when compiling with `-explain`
4 errors found
Compilation failed
```

**Expected behaviour**
no errors, because the shebang should be stripped by scala-cli when wrapping the script source

sierikov:
I expanded a regular expression to support multiline shebang (#2901) and included the appropriate test case. Please see the short demo:

anim-opt

### DIFF
--- a/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
@@ -3,7 +3,7 @@ package scala.build.preprocessing
 import scala.util.matching.Regex
 
 object SheBang {
-  private val sheBangRegex: Regex = s"""(^(#!.*(\\r\\n?|\\n)?)+(\\s*!#.*)?)""".r
+  private val sheBangRegex: Regex = s"""(^(#!.*(\\X)?)+(\\X*!#.*)?)""".r
 
   def isShebangScript(content: String): Boolean = sheBangRegex.unanchored.matches(content)
 
--- a/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
@@ -318,6 +318,13 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
       os.rel / "something5.scala" ->
         """#!/usr/bin/scala-cli
           |
+          |println("Hello World #!")""".stripMargin,
+      os.rel / "multiline.scala" ->
+        """#!/usr/bin/scala-cli
+          |# comment
+          |VAL=1
+          |!#
+          |
           |println("Hello World #!")""".stripMargin
     )
     val expectedParsedCodes = Seq(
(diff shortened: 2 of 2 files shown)


Gedochao:
To fix `docs-tests` failing on the CI (`sclicheck.DocTests.cookbook introduction/instant-startup-scala-scripts`) it should be enough to rebase this on top of the newest `main`. 
It's a bug that got fixed in https://github.com/VirtusLab/scala-cli/pull/2905

Gedochao (approved):
Thanks for the contribution!
LGTM

bishabosha:
Actually it would be good to test this in script files (`.sc`) as reported in https://github.com/VirtusLab/scala-cli/issues/2901 because the scala compiler itself strips multiline shebang from the top of a source, but script files avoid this because their content is copy pasted in the middle of the wrapper code

PR #2931, opened 2024-05-23
Improve launcher options handling

### LINKED ISSUE #1263: `.bsp/scala-cli.json`'s generation ignores wrapper scripts
**Version(s)**
0.1.11

**Describe the bug**
When using a wrapper script for scala-cli, you'll see that wrapper script ignored when `.bsp/scala-cli.json` is generated.

For example, in Nixpkgs, the `scala-cli` executable maps to the following wrapper:

```bash
#! /nix/store/db7n4xkfng8y4c3klp6bjppaws3rphq6-bash-5.1-p16/bin/bash -e
export JAVA_HOME='/nix/store/qaf9wyzymdij19qxyw7208s72lajz65b-zulu17.34.19-ca-jdk-17.0.3'
exec "/nix/store/jdrnvm9rq3cniy2g0x70k0rfvx1h68rs-scala-cli-0.1.11/bin/.scala-cli-wrapped"  "$@" 
```

(`.scala-cli-wrapped` is the actual binary from GitHub releases)

and the bsp config gets generated as:

```json
{
  "name": "scala-cli",
  "argv": [
    "/nix/store/jdrnvm9rq3cniy2g0x70k0rfvx1h68rs-scala-cli-0.1.11/bin/.scala-cli-wrapped",
    "bsp",
    "--json-options",
    "/Users/kubukoz/dev/scala-scripts/.scala-build/ide-options-v2.json",
    "/Users/kubukoz/dev/scala-scripts"
  ],
  "version": "0.1.11",
  "bspVersion": "2.0.0",
  "languages": [
    "scala",
    "java"
  ]
}
```

The point is that when you install scala-cli from Nixpkgs, it'll always use a Nix-managed Java 17. Personally I rely on this mostly to avoid the x86 release of scala-cli fetching an x86 JVM on an M1 Mac.

**Why this is an issue**

When I shutdown the Bloop instance started by scala-cli, and then in the future I go back to a Metals project using scala-cli, I lose the `JAVA_HOME` variable. scala-cli will then start Bloop using a Coursier-fetched Java 17, which will be an x86 binary, which are much much slower to run via Rosetta than an ARM-native JVM.

**To Reproduce**

1. Wrap scala-cli in a script, any script
2. Run `scala-cli compile .` in a project
3. Look at `.bsp/scala-cli.json` and see the wrapped program being used

**Expected behaviour**

Either a dealiased `scala-cli` is used as the first `argv` element, or just "scala-cli" (the command I ran when the bsp config was supposed to be generated).

Alternative - The bsp file always keeps a reference to the JVM that was available when generating it.

### LINKED ISSUE #1683: BSP doesn't work with `nightly`/`snapshot` versions of the CLI
**Version(s)**
nightly (0.1.18-32-g5bb00445-SNAPSHOT)

**Describe the bug**
A project setup with `--cli-version nightly` doesn't import in an IDE via BSP.
The resulting connection json points to the snapshot jar, which is not executable.
By default, the IDE complains about the permissions to the jar, but adding executable permissions of course doesn't fix a thing, as the jar isn't executable.

image

Relevant BSP log (nothing interesting here, the jar can't be executed, so there's no response when trying to connect to bsp from it):
```
[Trace - 08:35:15 AM] Sending request 'build/initialize - (1)'
Params: {
  "rootUri": "file:/Users/pchabelski/IdeaProjects/scala-cli-tests/intellij-bsp-reload/",
  "displayName": "IntelliJ-BSP",
  "version": "2022.3.888",
  "bspVersion": "2.0",
  "capabilities": {
    "languageIds": [
      "scala",
      "java"
    ]
  },
  "data": {
    "clientClassesRootDir": "file:/Users/pchabelski/IdeaProjects/scala-cli-tests/intellij-bsp-reload/out",
    "supportedScalaVersions": []
  }
}


[Trace - 08:35:35 AM] Sending notification '$/cancelRequest'
Params: {
  "id": "1"
}


[Trace - 08:35:35 AM] Sending request 'build/shutdown - (2)'
Params: null
```

**To Reproduce**
Just setup a project for the IDE with a nightly CLI version.
```bash
scala-cli --cli-version nightly setup-ide .
```
or any snapshot version, really.
```bash
scala-cli --cli-version 0.1.18-32-g5bb00445-SNAPSHOT setup-ide
```
The resulting BSP connection file (`.bsp/scala-cli.json`) is borked.
```json
{
  "name": "scala-cli",
  "argv": [
    "/Users/pchabelski/Library/Caches/Coursier/v1/https/oss.sonatype.org/content/repositories/snapshots/org/virtuslab/scala-cli/cli_3/0.1.18-32-g5bb00445-SNAPSHOT/cli_3-0.1.18-32-g5bb00445-SNAPSHOT.jar",
    "bsp",
    "--json-options",
    "/Users/pchabelski/IdeaProjects/scala-cli-tests/intellij-bsp-reload/.scala-build/ide-options-v2.json",
    "/Users/pchabelski/IdeaProjects/scala-cli-tests/intellij-bsp-reload"
  ],
  "version": "0.1.18-32-g5bb00445-SNAPSHOT",
  "bspVersion": "2.1.0-M3",
  "languages": [
    "scala",
    "java"
  ]
}
```

**Expected behaviour**
It should be possible to try out the most recent fixes to the BSP support with nightlies (or try out BSP with snapshots in general, though I believe forcing nightlies to work is key).

Gedochao:
Fixes #1263


Gedochao:
Kudos for @bishabosha for setting up the repo with Scala 3.5.0-RC1-fakeversion-bin-SNAPSHOT jars:
- https://github.com/dotty-staging/maven-test-repo

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/1683

modules/integration/src/test/scala/scala/cli/integration/BspTests3NextRc.scala
```diff
@@ -1,3 +1,95 @@
...
+import com.eed3si9n.expecty.Expecty.expect
+
+import scala.async.Async.{async, await}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.jdk.CollectionConverters.*
+import scala.util.Properties
+
+class BspTests3NextRc extends BspTestDefinitions with BspTests3Definitions with Test3NextRc {
+  test("BSP respects --cli-default-scala-version & --predefined-repository launcher options") {
+    // 3.5.0-RC1-fakeversion-bin-SNAPSHOT has too long filenames for Windows.
```
bishabosha, line 13: wow just learning this also, apparently Microsoft say you can prefix with `\\?\` to get a longer length limit https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

Gedochao: I'll try it in a separate PR later, nice!

modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
```diff
@@ -29,6 +30,14 @@ object Bsp extends ScalaCommand[BspOptions] {
         val content = os.read.bytes(os.Path(optionsPath, os.pwd))
         readFromArray(content)(SharedOptions.jsonCodec)
       }.getOrElse(options.shared)
+  private def latestLauncherOptions(options: BspOptions): LauncherOptions =
+    options.jsonLauncherOptions
+      .map(path => os.Path(path, os.pwd))
+      .filter(path => os.exists(path) && os.isFile(path))
+      .map { optionsPath =>
```
tgodzik, line 37:
```suggestion
      .collect { case optionsPath if os.exists(optionsPath) && os.isFile(optionsPath)) =>
```
super nitpick

tgodzik (approved):
LGTM

modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
Gedochao: will address separately, got no time to restart the CI on this 👍

PR #2943, opened 2024-05-26
scala-cli: 2804- Support .jar files in -d option

### LINKED ISSUE #2804: Support .jar files in -d option
`scalac` allows putting `.jar` files in `-d` option.
Invoking`scalac file.scala -d foo.jar` creates `foo.jar` file.
`scala-cli compile file.scala -d foo.jar` creates `foo.jar` folder instead.
I'd expect the same behavior of those two commands.

kannupriyakalra:
Issue- https://github.com/VirtusLab/scala-cli/issues/2804

-added code changes required for replicating scalac commands behaviour into scala-cli
-added test (scala-cli compile -d option package into a jar)


modules/integration/src/test/scala/scala/cli/integration/RunScalacCompatTestDefinitions.scala
```diff
@@ -206,6 +206,43 @@ trait RunScalacCompatTestDefinitions {
...
+        "-p",
+        preCompiledInput,
+        mainInput,
+        "-d",
+        outputJar.toString,
+        extraOptions
+      ).call(cwd = root / preCompileDir)
+
+      // ensure the precompiled JAR exists
+      assert(os.exists(root / preCompileDir / outputJar), s"Precompiled JAR not found at $outputJar")
```
bishabosha, line 231: I think we should validate two conditions:
- [x] that `outputJar` exits.
- [ ] `outputJar` is a file, not a directory (i.e. what issue #2804 complains about)

modules/integration/src/test/scala/scala/cli/integration/RunScalacCompatTestDefinitions.scala
```diff
@@ -206,6 +206,43 @@ trait RunScalacCompatTestDefinitions {
...
+    val preCompiledInput = "Message.scala"
+    val mainInput = "Main.scala"
+    val expectedOutput = "Hello"
+    TestInputs(
+      os.rel / preCompileDir / preCompiledInput -> "case class Message(value: String)",
+      os.rel / preCompileDir / mainInput -> s"""object Main extends App { println(Message("$expectedOutput").value) }"""
+    ).fromRoot { (root: os.Path) =>
+      val outputJar = os.rel / "outParentDir" / "out.jar"
+
+      // first, precompile to an explicitly specified output directory with -d
```
bishabosha:
```suggestion
      // first, precompile to an explicitly specified output jar file with -d
```

modules/integration/src/test/scala/scala/cli/integration/RunScalacCompatTestDefinitions.scala
```diff
@@ -206,6 +206,43 @@ trait RunScalacCompatTestDefinitions {
...
   }
 
+  test("scala-cli compile -d option package into a jar") {
+    val preCompileDir = "PreCompileDir"
+    val preCompiledInput = "Message.scala"
+    val mainInput = "Main.scala"
+    val expectedOutput = "Hello"
+    TestInputs(
+      os.rel / preCompileDir / preCompiledInput -> "case class Message(value: String)",
+      os.rel / preCompileDir / mainInput -> s"""object Main extends App { println(Message("$expectedOutput").value) }"""
```
bishabosha: minor point - it's unnecessary to have these two source files because you compile them together - its probably fine to drop the `Message` class altogether and just call println with expectedOutput

kannupriyakalra:
@bishabosha Thank you for reviewing my PR, I have incorporated changes in the integration test based on the comments. PR is up for re-review.

modules/integration/src/test/scala/scala/cli/integration/RunScalacCompatTestDefinitions.scala
```diff
@@ -206,6 +206,43 @@ trait RunScalacCompatTestDefinitions {
...
+        "-p",
+        mainInput,
+        "-d",
+        outputJar.toString,
+        extraOptions
+      ).call(cwd = root / preCompileDir)
+
+      // ensure the precompiled JAR exists and is a file ( not a directory )
+      assert(os.exists(root / preCompileDir / outputJar), s"Precompiled JAR not found at $outputJar")
+      assert(!os.isFile(root / preCompileDir / outputJar), s"$outputJar is not a file")
```
bishabosha, line 232:
```suggestion
      assert(os.isFile(root / preCompileDir / outputJar), s"$outputJar is not a file")
```
this matches up with the comments and assertion message

Gedochao: this assertion is passing, because the produced outputJar is a directory, rather than a JAR file.
the directory contains compiled classes, so passing them as an inputs does, in fact, work.
it does not fulfill the requirements of what we want here, however.
```
run-891283530/test-2/PreCompileDir
▶ cd outParentDir 

test-2/PreCompileDir/outParentDir 
▶ ls
out.jar

test-2/PreCompileDir/outParentDir
▶ cd out.jar

PreCompileDir/outParentDir/out.jar
▶ ls
Main$.class Main.class  Main.tasty
```

modules/integration/src/test/scala/scala/cli/integration/RunScalacCompatTestDefinitions.scala
Gedochao: this comment hasn't been resolved. As mentioned by @bishabosha , we need the `outputJar` to be a JAR file, rather than a directory containing compiled classes.

Gedochao (changes requested):
As mentioned in earlier comments, this solution isn't working just yet, despite the CI being green. We need an actual `.jar` file produced, rather than a directory with `.jar` in its name.

I'm not sure what would be the optimal way to achieve this without diving into the code myself, but I'd guess we can either let the compiler create the JAR for us (since `-d` is a compiler option in this context) or actually use the internals of the `package` sub-command (which is more complex, but might be the optimal solution, since that allows us to wire in all the magic that's done there).

Gedochao:
Closing this due to inactivity. 
Feel free to open another attempt.

PR #2968, opened 2024-06-17
Scalafix command for scala-cli with basic options and tests

### LINKED ISSUE #647: scalafix command to run library migration & hygiene rules

Scalafix provides a [CLI](https://scalacenter.github.io/scalafix/docs/users/installation.html#command-line) which is rather easy to use to run syntactic rules (using on-demand scalameta parsers under the hood, like scalafmt) but is a hassle to setup to run semantic rules (requiring prior compilation output, namely class & semanticdb files). That's why [we usually recommend using build tools plugins instead](https://github.com/scalacenter/scalafix/issues/1536#issuecomment-1031758593) (either the official sbt plugin or the community-driven maven, gradle & mill plugins).

For projects managed by Scala CLI, using the Scalafix CLI is currently the only option. Providing a built-in integration with Scalafix inside Scala CLI would be useful to cover the typical Scalafix use cases on such projects:
1. helping library clients to handle deprecation and breaking changes ("migration" use-case)
1. enforcing conventions and preventing bad smells ("hygiene" use-case)

A `scalafix` command could take care of
- triggering semanticdb-enabled compilation (I assume this is already in place)
- fetching Scalafix itself & [community rules](https://scalacenter.github.io/scalafix/docs/rules/external-rules.html) if provided (relying on [`scalafix-interfaces`](https://github.com/scalacenter/scalafix/blob/main/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java))
- invoking Scalafix with the right arguments

**Additional context**

If Scala CLI adoption picks up, I assume https://github.com/scala-steward-org/scala-steward will eventually support projects managed by Scala CLI and be able to update directives. A support for [running Scalafix migrations just like in the other build tools](https://github.com/scala-steward-org/scala-steward/blob/main/docs/scalafix-migrations.md) would then be welcome, and could rely on that built-in `scalafix` command.

Automating the bump of dependencies by updating directives might be implemented as part of Scala CLI itself in the future, but in any case integrating Scalafix (potentially without a user-facing command) will be useful to leverage [existing migration rules](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/scalafix-migrations.conf).

Vigorge:
Resolves Issue #647


modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+
+    val scalaVersion =
+      options.buildOptions.orExit(logger).scalaParams.orExit(logger).map(_.scalaVersion)
+        .getOrElse(Constants.defaultScalaVersion)
+    val configFilePathOpt = options.scalafixConf.map(os.Path(_, os.pwd))
+    val relPaths          = sourcePaths.map(_.toNIO.getFileName)
+    val toolClasspath     = options.shared.dependencies.compileOnlyDependency
+
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance("2.13")
```
tgodzik: This should be either 2.12 for 2.12.x or 2.13 for 2.13.x and 3.x.y

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+import scala.cli.config.Keys
+import scala.cli.util.ArgHelpers.*
+import scala.cli.util.ConfigDbUtils
+import scala.jdk.CollectionConverters.*
+import scala.jdk.OptionConverters.*
+
+object Scalafix extends ScalaCommand[ScalafixOptions] {
+  override def group: String = HelpCommandGroup.Main.toString
+  override def sharedOptions(options: ScalafixOptions): Option[SharedOptions] = Some(options.shared)
+  override def scalaSpecificationLevel: SpecificationLevel = SpecificationLevel.SHOULD
```
Gedochao:
```suggestion
  override def scalaSpecificationLevel: SpecificationLevel = SpecificationLevel.EXPERIMENTAL
```

Gedochao:
I'll make sure to review this as soon as I am available, but given the volume of code included, I might need a couple days to get to it.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+    val toolClasspath     = options.shared.dependencies.compileOnlyDependency
+
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance("2.13")
+      .newArguments()
+      .withParsedArguments(options.scalafixArg.asJava)
+      .withWorkingDirectory(workspace.toNIO)
+      .withPaths(relPaths.asJava)
+      .withRules(options.rules.asJava)
+      .withConfig(Optional.ofNullable(configFilePathOpt.map(_.toNIO).orNull))
```
Gedochao: with 
```scala
import scala.jdk.OptionConverters.*
```
you can just
```suggestion
      .withConfig(configFilePathOpt.map(_.toNIO).toJava)
```

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests.scala
```diff
@@ -0,0 +1,59 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+class ScalafixTests extends ScalaCliSuite {
```
Gedochao: It'd be nice to run this suite for different Scala versions.
Refer to how other integration tests are setup, i.e. `CompileTestDefinitions`

Gedochao:
NOTE: After this is cleaned up and merged, we'll want to merge the `scalafix` sub-command with the `fix` sub-command (there's no point to keep them separate).
Individual `fix` features would probably be enabled/disabled by flags.
**We'll do that as a follow-up, no need to worry about it in this PR.**

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Pass rules directly to scalafix.")
+  @Tag(tags.inShortHelp)
+  rules: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.should)
+  @HelpMessage("Check if sources are well organized")
```
bjaglin: This message looks very specific to some rules.
```suggestion
  @HelpMessage("Fail the invocation instead instead of applying fixes")
```

project/deps.sc
```diff
@@ -100,6 +100,7 @@ object Deps {
     def signingCli           = "0.2.3"
     def signingCliJvmVersion = 17
     def javaClassName        = "0.1.3"
+    def scalafix             = "0.12.0"
```
bjaglin:
```suggestion
    def scalafix             = "0.12.1"
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+      options.buildOptions.orExit(logger).scalaParams.orExit(logger).map(_.scalaVersion)
+        .getOrElse(Constants.defaultScalaVersion)
+    val configFilePathOpt = options.scalafixConf.map(os.Path(_, os.pwd))
+    val relPaths          = sourcePaths.map(_.toNIO.getFileName)
+    val toolClasspath     = options.shared.dependencies.compileOnlyDependency
+
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance("2.13")
+      .newArguments()
+      .withParsedArguments(options.scalafixArg.asJava)
```
bjaglin: Having this last would give the option to force custom values for options injected by the client (to workaround a bug for example).

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+      .withRules(options.rules.asJava)
+      .withConfig(Optional.ofNullable(configFilePathOpt.map(_.toNIO).orNull))
+      .withScalaVersion(scalaVersion)
+      .withToolClasspath(Seq.empty.asJava, toolClasspath.asJava)
+      .withScalacOptions(Seq("-Wunused", "-Wunused:imports", "-P:semanticdb:synthetics:on").asJava)
+
+    val rulesThatWillRun = scalafix.rulesThatWillRun().asScala
+
+    logger.debug(
+      s"Rewriting and linting ${sourcePaths.size} Scala sources against ${rulesThatWillRun.size} rules"
```
bjaglin: "Rewriting" could be misleading as sources are not rewritten when `check` is used

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+      .fetchAndClassloadInstance("2.13")
+      .newArguments()
+      .withParsedArguments(options.scalafixArg.asJava)
+      .withWorkingDirectory(workspace.toNIO)
+      .withPaths(relPaths.asJava)
+      .withRules(options.rules.asJava)
+      .withConfig(Optional.ofNullable(configFilePathOpt.map(_.toNIO).orNull))
+      .withScalaVersion(scalaVersion)
+      .withToolClasspath(Seq.empty.asJava, toolClasspath.asJava)
+      .withScalacOptions(Seq("-Wunused", "-Wunused:imports", "-P:semanticdb:synthetics:on").asJava)
```
bjaglin: Can these be retrieved from the build? otherwise it will trigger false negatives/positives when rules check their pre-requisites...

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+
+      val classPaths = successfulBuildOpt.map(_.fullClassPath).getOrElse(Seq.empty)
+      val externalDeps =
+        options.shared.dependencies.compileOnlyDependency ++ successfulBuildOpt.map(
+          _.options.classPathOptions.extraCompileOnlyJars
+        ).getOrElse(Seq.empty).map(_.toNIO.toString)
+
+      scalafix
+        .withClasspath(classPaths.map(_.toNIO).asJava)
+        .withToolClasspath(Seq.empty.asJava, externalDeps.asJava)
```
bjaglin: Same comment as above regarding toolClasspath, I am not sure what's the intent here?

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+      if (evaluation.isSuccessful)
+        evaluation.getFileEvaluations.foldLeft(List.empty[String]) {
+          case (errors, fileEvaluation) =>
+            val problemMessage = fileEvaluation.getErrorMessage.toScala.orElse(
+              fileEvaluation.previewPatchesAsUnifiedDiff.toScala
+            )
+            errors ++ problemMessage
+        }
+      else
+        evaluation.getErrorMessage.toScala.toList
```
bjaglin: This looks untested.

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests.scala
```diff
@@ -0,0 +1,59 @@
...
+  override def group: ScalaCliSuite.TestGroup = ScalaCliSuite.TestGroup.First
+
+  val confFileName = ".scalafix.conf"
+
+  val emptyInputs: TestInputs = TestInputs(os.rel / ".placeholder" -> "")
+
+  val simpleInputsUnformattedContent: String =
```
bjaglin: Referring to unformatted /formatted content is a bit confusing as scalafix is not a formatting tool. I would suggest original/rewritten.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+      } yield sOpt
+
+      val classPaths = successfulBuildOpt.map(_.fullClassPath).getOrElse(Seq.empty)
+      val externalDeps =
+        options.shared.dependencies.compileOnlyDependency ++ successfulBuildOpt.map(
+          _.options.classPathOptions.extraCompileOnlyJars
+        ).getOrElse(Seq.empty).map(_.toNIO.toString)
+
+      scalafix
+        .withClasspath(classPaths.map(_.toNIO).asJava)
```
bjaglin: This looks untested. Running `ExplicitResultTypes` is an easy way to make sure everything is set (semanticdb generation & classpath injection).

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,174 @@
...
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance("2.13")
+      .newArguments()
+      .withParsedArguments(options.scalafixArg.asJava)
+      .withWorkingDirectory(workspace.toNIO)
+      .withPaths(relPaths.asJava)
+      .withRules(options.rules.asJava)
+      .withConfig(Optional.ofNullable(configFilePathOpt.map(_.toNIO).orNull))
+      .withScalaVersion(scalaVersion)
+      .withToolClasspath(Seq.empty.asJava, toolClasspath.asJava)
```
bjaglin: `toolClasspath` is a way to extend scalafix class path in order to run non-built-in rules. I feel like it's not the intent here. What is this used for?

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Pass an argument to scalafix.")
+  @Tag(tags.inShortHelp)
+  scalafixArg: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Pass rules directly to scalafix.")
```
bjaglin:
```suggestion
  @HelpMessage("Run rule(s) explicitly, overriding the configuration file default.")
```

This looks untested.

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Custom path to the scalafix configuration file.")
+  @Tag(tags.inShortHelp)
+  scalafixConf: Option[String] = None,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Pass an argument to scalafix.")
```
bjaglin:
```suggestion
  @HelpMessage("Pass extra argument(s) to scalafix.")
```

This looks untested.

bjaglin (commented):
Thanks for starting this 🤩 

I am not familiar with scala-cli at all, but as the maintainer of sbt-scalafix and scalafix, here is some feedback FWIW!

tgodzik:
Thanks @bjaglin!

@Vigorge if you are at any point overwhelmed by amount of comments, we are here to help. You already did a lot of great work :heart_eyes:

Vigorge:
> Thanks @bjaglin!
> 
> @Vigorge if you are at any point overwhelmed by amount of comments, we are here to help. You already did a lot of great work 😍

Thanks a lot for such an impactful feedback! These comments are really helpful and I'll surely take them into account. I really hope that I will resolve them in a few days and add some more features. Unfortunately, I had to take a little break, but now am fully committed to work more on this PR now

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Thanks, the other `toolClasspath` was excessive, but here the intent is to use non-built-in rules

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+import scala.util.Properties
+
+// format: off
+@HelpMessage(ScalafixOptions.helpMessage, "", ScalafixOptions.detailedHelpMessage)
+final case class ScalafixOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+  @HelpMessage("Custom path to the scalafix configuration file.")
+  @Tag(tags.inShortHelp)
+  scalafixConf: Option[String] = None,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+  @HelpMessage("Pass extra argument(s) to scalafix.")
+  @Tag(tags.inShortHelp)
+  scalafixArg: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.implementation)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+  @HelpMessage("Run rule(s) explicitly, overriding the configuration file default.")
+  @Tag(tags.inShortHelp)
+  rules: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.should)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

Gedochao:
Not sure why didn't the CI run, but a rebase from `main` seems to be necessary, anyway.

SethTisue:
Great to see activity on this — I think this could be big for Scalafix, in terms of people being able to use it more easily.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Hi, as I mentioned in a thread above, the intend is indeed to run non-built-in rules. This `withToolClasspath` is not valid, so it was deleted, but the other one is added for that purpose.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Thanks, added a test for semantic rule `RemoveUnused`. It has to close this case + check passing scalar argument to scalafix from using directive.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Added a test for external rules to demonstrate this

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Thanks a lot, added a test.

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests.scala
```diff
@@ -0,0 +1,183 @@
...
+        |    greetMany(name = "John", times = 42)
+        |}
+        |""".stripMargin
+    }
+
+    externalRuleInputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "scalafix", "--power", ".", "-S", "2").call(cwd = root)
+      val updatedContent = noCrLf(os.read(root / "Hello.scala"))
+      println(updatedContent)
+      println(expectedContent)
```
dos65:
```suggestion
```

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests.scala
Vigorge: Thank you. Looked it up and resolved in https://github.com/VirtusLab/scala-cli/pull/2968/commits/61b18f7bb05fd1035d9fdb6d1aded707b3869c3d

Vigorge:
Hi, sorry for taking such a long time to resolve all those threads and thanks a lot again for such feedback. 

The PR is ready for review.

bjaglin:
Thanks for the follow-ups @Vigorge 👍 I will take a look at the PR tomorrow.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+      sys.exit(1)
+    }
+  }
+
+  private def prepareErrorMessage(error: ScalafixError): String = error match
+    case ParseError => "A source file failed to be parsed"
+    case CommandLineError =>
+      "A command-line argument was parsed incorrectly"
+    case MissingSemanticdbError =>
+      "A semantic rewrite was run on a source file that has no associated META-INF/semanticdb/.../*.semanticdb"
```
bjaglin: nit: all scalafix rules are not rewrites
```suggestion
      "A semantic rule was run on a source file that has no associated META-INF/semanticdb/.../*.semanticdb"
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+            val problemMessage = fileEvaluation.getErrorMessage.toScala.orElse(
+              fileEvaluation.previewPatchesAsUnifiedDiff.toScala
+            )
+            errors ++ problemMessage
+        }
+      else
+        evaluation.getErrorMessage.toScala.toList
+    }
+    else
+      customScalafixInstance.run().map(prepareErrorMessage).toList
```
bjaglin: Scalafix already outputs human readable messages to stdout by default, so users will get duplicate error messages.

```
$ launcher --power scalafix Test.scala
The `scalafix` sub-command is experimental
No rules requested to run
No rules were provided to Scalafix so nothing happened
```

The **scalafix framework** output (not to be mixed up with the scalafix rules effective output, see comment above) can be silenced or re-routed via `withPrintStream()`. As the internal log entries can carry more information than the status code, I would recommend removing the custom error messages, and either mapping a PrintStream delegating to the scala-cli logger (that's [what sbt-scalafix does](https://github.com/scalacenter/sbt-scalafix/blob/main/src/main/scala/scalafix/internal/sbt/LoggingOutputStream.scala#L10)) or actually simply keeping the [default PrintStream (stdout](https://github.com/scalacenter/scalafix/blob/c02ef92dcaa839b2e11812ee81f3f7029750b6b9/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java#L119-L123)), for the same reason as in the comment above.

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests213.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import com.eed3si9n.expecty.Expecty.expect
+
+class ScalafixTests213 extends ScalafixTestDefinitions with Test213 {
+  override val unusedRuleOption: String = "-Wunused"
+
+  test("external rule") {
+    val unnamedParamsInputsContent: String =
+      """//> using options -P:semanticdb:synthetics:on
+        |//> using compileOnly.dep "com.github.jatcwang::scalafix-named-params:0.2.4"
```
bjaglin: Will possibility of loading custom rules via `compileOnly.dep` be obvious to a scala-cli user? It wasn't to me, so a mention of it in the docs may be useful.

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Run rule(s) explicitly, overriding the configuration file default.")
+  @Tag(tags.inShortHelp)
+  rules: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Fail the invocation instead of applying fixes")
```
bjaglin: nit: some rules won't emit rewrites (they are just linters), in which case `--check` does not change anything, so I would be more specific here
```suggestion
  @HelpMessage("Fail the invocation if rewrites are needed")
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+) extends HasSharedOptions {
+  def buildOptions: Either[BuildException, BuildOptions] = shared.buildOptions()
+
+}
+object ScalafixOptions {
+  implicit lazy val parser: Parser[ScalafixOptions] = Parser.derive
+  implicit lazy val help: Help[ScalafixOptions]     = Help.derive
+
+  val cmdName             = "scalafix"
+  private val helpHeader  = "Fixes Scala code according to scalafix rules."
```
bjaglin: nit: adding the linter aspect of scalafix to the description
```suggestion
  private val helpHeader  = "Run Scalafix rules to lint or rewrite Scala code."
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+  @Tag(tags.experimental)
+  @HelpMessage("Pass extra argument(s) to scalafix.")
+  @Tag(tags.inShortHelp)
+  scalafixArg: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Run rule(s) explicitly, overriding the configuration file default.")
+  @Tag(tags.inShortHelp)
+  rules: List[String] = Nil,
```
bjaglin, line 37: These 2 flags look untested.

It would be trivial to do so by removing the `.scalafix.conf` for the UseNamedParameters test and use CLI args.

With scalafix-cli, this works
```
scalafix --rules UseNamedParameters --settings.UseNamedParameters.minParams=2
```

so I assume with scala-cli it would be something like this ?
```
scala --power scalafix --rules=UseNamedParameters --scalafix-arg=--settings.UseNamedParameters.minParams=2
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+@HelpMessage(ScalafixOptions.helpMessage, "", ScalafixOptions.detailedHelpMessage)
+final case class ScalafixOptions(
+  @Recurse
+    shared: SharedOptions = SharedOptions(),
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Custom path to the scalafix configuration file.")
+  @Tag(tags.inShortHelp)
+  scalafixConf: Option[String] = None,
```
bjaglin, line 25: Untested? It's not very used IMHO, so maybe it can be removed, and re-added with tests in a follow-up PR?

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
```diff
@@ -0,0 +1,66 @@
...
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Pass extra argument(s) to scalafix.")
+  @Tag(tags.inShortHelp)
+  scalafixArg: List[String] = Nil,
+
+  @Group(HelpGroup.Format.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Run rule(s) explicitly, overriding the configuration file default.")
+  @Tag(tags.inShortHelp)
```
bjaglin, line 36: nit: can we add a short flag to mimic scalafix-cli?
```suggestion
  @Tag(tags.inShortHelp)
  @Name("r")
```

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
```diff
@@ -0,0 +1,103 @@
...
+        cwd = root,
+        check = false
+      )
+      expect(res.exitCode == 1)
+      val updatedContent = noCrLf(os.read(root / "Hello.scala"))
+      expect(updatedContent == noCrLf(simpleInputsOriginalContent))
+    }
+  }
+
+  test("semantic rule") {
```
bjaglin, line 64: I would suggest adding a test with another popular built-in rule, ExplicitResultTypes (only for Scala 2.x for now) since it is more complex and would confirm that classpath is properly injected (not just semanticdb files). I saw [this problem in mill-scalafix recently](https://github.com/joan38/mill-scalafix/issues/174#issuecomment-1879468983).

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+    )
+
+    val rulesThatWillRun: Either[ScalafixException, mutable.Buffer[ScalafixRule]] =
+      try
+        Right(scalafix.rulesThatWillRun().asScala)
+      catch
+        case e: ScalafixException => Left(e)
+    val needToBuild: Boolean = rulesThatWillRun match
+      case Right(rules) => rules.exists(_.kind().isSemantic)
+      case Left(_)      => true
```
bjaglin: nit: if `rulesThatWillRun` failed, chances are high that `run` will as well, so I wonder if it's worth building?

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+      val successfulBuildOpt = for {
+        build <- builds.get(Scope.Main)
+        sOpt  <- build.successfulOpt
+      } yield sOpt
+
+      val classPaths = successfulBuildOpt.map(_.fullClassPath).getOrElse(Seq.empty)
+      val externalDeps =
+        options.shared.dependencies.compileOnlyDependency ++ successfulBuildOpt.map(
+          _.options.classPathOptions.extraCompileOnlyDependencies.values.flatten.map(_.value.render)
+        ).getOrElse(Seq.empty)
```
bjaglin: _Not blocking for this PR as the experimental flag allows iterating_

Even if that's likely the most harmless way to get scala-cli to resolve an artifact (either via a directive or the corresponding CLI flag I assume), this approach has several limitations
- it might impact the build by evicting some transient dependencies
- the fact that this is not in the base class probably shows, but it won't allow loading custom rules when running against Scala 3 sources (unless there is some kind of fallback to 2.13 artifacts built-in?), since for now Scalafix is not cross-published to Scala 3, as the mapping above shows https://github.com/VirtusLab/scala-cli/blob/3e08eb23dbc5dcb409c9721f62f608fee4590f51/modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala#L90

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+      case _ =>
+        logger.error("Unsupported scala version " + scalaVersion)
+        sys.exit(1)
+    val configFilePathOpt = options.scalafixConf.map(os.Path(_, os.pwd))
+    val relPaths          = sourcePaths.map(_.toNIO.getFileName)
+
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance(scalaBinaryVersion)
+      .newArguments()
+      .withWorkingDirectory(workspace.toNIO)
```
bjaglin: `workspace` seems to reference an ephemeral directory that does not contain (or no longer contains?) the source file when I run the output of `./mill -i show 'cli[].launcher'` against a source in a random directory

```
➜  /tmp cat Test.scala
//> using options -Wunused:all

package foo

object Hello {
    val name = "John"
    println("Hello")
  }
}
```

```
➜  /tmp /Users/brice.jaglin/git/scala/scala-cli/out/cli/3.3.3/launcher.dest/launcher --power scalafix --rules=RemoveUnused ./Test.scala
Some utilized features are marked as experimental:
 - `scalafix` sub-command
 - `--rules=RemoveUnused` option
[warn] ./Test.scala:7:9
[warn] unused local definition
[warn]     val name = "John"
[warn]         ^^^^
error: /Users/brice.jaglin/Library/Caches/ScalaCli/virtual-projects/a7/project-c4a6da38/Test.scala is not a file
Something unexpected happened running Scalafix
```

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+      )
+
+    val (sourcePaths, workspace, _) =
+      if (args.all.isEmpty)
+        (Seq(os.pwd), os.pwd, None)
+      else {
+        val s = inputs.sourceFiles().collect {
+          case sc: Script          => sc.path
+          case sc: SourceScalaFile => sc.path
+        }
```
bjaglin: nit: scalafix is doing the same check ([I think](https://github.com/tgodzik/scalafix/blob/e6b677278bc9d8489102ae8e4e9c5a03a53ccf0d/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala#L446-L447)), so I'd remove this, in case we add support for other extensions upstream

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+      case v if v.startsWith("2.13.") => "2.13"
+      case v if v.startsWith("3.")    => "2.13"
+      case _ =>
+        logger.error("Unsupported scala version " + scalaVersion)
+        sys.exit(1)
+    val configFilePathOpt = options.scalafixConf.map(os.Path(_, os.pwd))
+    val relPaths          = sourcePaths.map(_.toNIO.getFileName)
+
+    val scalafix = ScalafixInterface
+      .fetchAndClassloadInstance(scalaBinaryVersion)
```
bjaglin: Unsurprisingly, this fails when using the native-image bundle because of the dynamic nature of scalafix-interfaces.

```
➜  /tmp /Users/brice.jaglin/git/scala/scala-cli/out/cli/3.3.3/base-image/nativeImage.dest/scala-cli --power scalafix Test.scala
The `scalafix` sub-command is experimental
Error: scalafix.interfaces.ScalafixException: Failed to load 'scalafix-interfaces.properties' to lookup versions
For more details, please see '/private/tmp/.scala-build/stacktraces/1723547209-8851074228942378765.log'
```

```
➜  /tmp cat /private/tmp/.scala-build/stacktraces/1723546008-8191378505818715998.log
scalafix.interfaces.ScalafixException: Failed to load 'scalafix-interfaces.properties' to lookup versions
  scalafix.interfaces.Scalafix.fetchAndClassloadInstance(Scalafix.java:126)
  scalafix.interfaces.Scalafix.fetchAndClassloadInstance(Scalafix.java:85)
  scala.cli.commands.scalafix.Scalafix$.runCommand(Scalafix.scala:100)
  scala.cli.commands.scalafix.Scalafix$.runCommand(Scalafix.scala:56)
  scala.cli.commands.ScalaCommand.run(ScalaCommand.scala:388)
  scala.cli.commands.ScalaCommand.run(ScalaCommand.scala:369)
  caseapp.core.app.CaseApp.main(CaseApp.scala:157)
  scala.cli.commands.ScalaCommand.main(ScalaCommand.scala:354)
  caseapp.core.app.CommandsEntryPoint.main(CommandsEntryPoint.scala:169)
  scala.cli.ScalaCliCommands.main(ScalaCliCommands.scala:126)
  scala.cli.ScalaCli$.main0(ScalaCli.scala:295)
  scala.cli.ScalaCli$.main(ScalaCli.scala:119)
  scala.cli.ScalaCli.main(ScalaCli.scala)
java.lang.NullPointerException: inStream parameter is null
  java.base@17.0.6/java.util.Objects.requireNonNull(Objects.java:233)
  java.base@17.0.6/java.util.Properties.load(Properties.java:407)
  scalafix.interfaces.Scalafix.fetchAndClassloadInstance(Scalafix.java:124)
  scalafix.interfaces.Scalafix.fetchAndClassloadInstance(Scalafix.java:85)
  scala.cli.commands.scalafix.Scalafix$.runCommand(Scalafix.scala:100)
  scala.cli.commands.scalafix.Scalafix$.runCommand(Scalafix.scala:56)
  scala.cli.commands.ScalaCommand.run(ScalaCommand.scala:388)
  scala.cli.commands.ScalaCommand.run(ScalaCommand.scala:369)
  caseapp.core.app.CaseApp.main(CaseApp.scala:157)
  scala.cli.commands.ScalaCommand.main(ScalaCommand.scala:354)
  caseapp.core.app.CommandsEntryPoint.main(CommandsEntryPoint.scala:169)
  scala.cli.ScalaCliCommands.main(ScalaCliCommands.scala:126)
  scala.cli.ScalaCli$.main0(ScalaCli.scala:295)
  scala.cli.ScalaCli$.main(ScalaCli.scala:119)
  scala.cli.ScalaCli.main(ScalaCli.scala)
```

We could probably fix these [property lookups](https://github.com/scalacenter/scalafix/blob/c02ef92dcaa839b2e11812ee81f3f7029750b6b9/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L120-L132) by [telling GraalVM about it](https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/Resources/), but the [classloading just after](https://github.com/scalacenter/scalafix/blob/c02ef92dcaa839b2e11812ee81f3f7029750b6b9/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L135-L136) would fail anyway.

I am afraid we will need to run Scalafix through the scalafix-cli launcher (like scalafmt apparently?) instead of scalafix-interfaces. It would be a bit more cumbersome to pass CLI args, and a couple of advanced features wouldn't be available, but it should work overall. Or maybe there is a way to classload Scalafix within bloop somehow @tgodzik ? It would have the benefit of keeping classloaders warm across invocations.

Anyway, to unlock this PR, I guess we could add an early return if the native-image runtime is detected, and handle native-image compatibility in a later PR? I assume the native-image bundle is the most popular distribution channel though... 😢

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
```diff
@@ -0,0 +1,197 @@
...
+    val errors = if (options.check) {
+      val evaluation = customScalafixInstance.evaluate()
+      if (evaluation.isSuccessful)
+        evaluation.getFileEvaluations.foldLeft(List.empty[String]) {
+          case (errors, fileEvaluation) =>
+            val problemMessage = fileEvaluation.getErrorMessage.toScala.orElse(
+              fileEvaluation.previewPatchesAsUnifiedDiff.toScala
+            )
+            errors ++ problemMessage
+        }
```
bjaglin: With this implementation, `--check` effectively ignores diagnostics (linters)

```
➜  /tmp cat Test.scala
package foo

final object Hello {
    val name = null
  }
}
```

Expected failure (no `--check`)
```
➜  /tmp /Users/brice.jaglin/git/scala/scala-cli/out/cli/3.3.3/launcher.dest/launcher --power scalafix --rules=DisableSyntax --scalafix-arg=--settings.DisableSyntax.noNulls=true Test.scala && echo OK
Some utilized features are marked as experimental:
 - `scalafix` sub-command
 - `--rules=DisableSyntax` option
 - `--scalafix-arg=--settings.DisableSyntax.noNulls=true` option
/private/tmp/Test.scala:5:16: error: [DisableSyntax.null] null should be avoided, consider using Option instead
    val name = null
               ^^^^
A Scalafix linter error was reported
```

Unexpected success (with `--check`)
```
➜  /tmp /Users/brice.jaglin/git/scala/scala-cli/out/cli/3.3.3/launcher.dest/launcher --power scalafix --rules=DisableSyntax --scalafix-arg=--settings.DisableSyntax.noNulls=true Test.scala --check && echo OK
Some utilized features are marked as experimental:
 - `scalafix` sub-command
 - `--rules=DisableSyntax` option
 - `--scalafix-arg=--settings.DisableSyntax.noNulls=true` option
 - `--check` option
OK
```


Instead of evaluating manually and reconstructing logs from lints/rewrites in each file (which might result in missing upcoming features), you should call `run()` with the `--check` flag. To capture output, you can redirect the ScalafixMainCallback (effectively the output of the **scalafix rules**) to the scala-cli logger (that's [what sbt-scalafix does](https://github.com/scalacenter/sbt-scalafix/blob/main/src/main/scala/scalafix/internal/sbt/ScalafixLogger.scala)), or probably more pragmatic, just keep [the default which is dumping to stdout](https://github.com/scalacenter/scalafix/blob/c02ef92dcaa839b2e11812ee81f3f7029750b6b9/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java#L148-L152), since moving to scalafix-cli for native-image (see other comment) won't allow this level of integration in the future anyway.

bjaglin (commented):
Hey @Vigorge, thanks for all the follow-ups/additions. Having a semantic rule tested against all Scala versions is a great achievement!

I went through a thorough second review to avoid any surprise in the third/final round. From my very own perspective (again, I am not familiar with scala-cli), here are the outstanding comments to be addressed before this could be released
- workspace / current directory injection seems broken - I did not manage to use the CLI
- diagnostics/linters are ignored under `--check`
- using scalafix-interfaces with native-image is impossible and raises a cryptic error - a more actionnable error message should be added until this is sorted out in a further PR

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
bjaglin: This shows that invocation output (either with or without --check for a linter rule such as DisableSyntax, and with --check for a rewrite rule such as RedundantSyntax) should be asserted in the tests.

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
Vigorge: Ouch! You're right, it seems that it won't make a lot of sence if it  doens't work in native image.
I'll try to replace interfaces by scalafix-cli launcher and see how it is going to turn out.

Gedochao:
@Vigorge how's the progress on this? Can we unblock you somehow?

Vigorge:
> @Vigorge how's the progress on this? Can we unblock you somehow?

HI, still am working on the native image solution and gonna add feature in few days.

Thanks for support! I will comment here if some help will be needed.

dos65:
@Gedochao Hi, we did changes in MR to make scalafix work both in jvm and native builds. Could you approve CI run?

dos65:
Update.  I need one additionial iteration to go trough comments from @bjaglin 
(add a bit more docs and tests) but in general I hope the PR is close to completion.

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
dos65: added test

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
dos65: now they covered

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
dos65: added for explicitRulesTypes too

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: removed this code - it isn't needed after switching on launcher

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: removed this code - it isn't needed after switching on launcher

modules/integration/src/test/scala/scala/cli/integration/ScalafixTests213.scala
dos65: I decided to add a new directive -  `scalafix.dep`.
https://github.com/VirtusLab/scala-cli/blob/d1a7ae9fec5e1dd4d217e8faab42dda3ac5c2f9a/website/docs/commands/scalafix.md?plain=1#L40-L44

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
dos65: applied

modules/cli/src/main/scala/scala/cli/commands/scalafix/ScalafixOptions.scala
dos65: applied

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: isn't actual

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: Fixed this by adding `scalafix.dep` +copied the logic for `3` -> `2.13` resolution

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: tested this loccaly - now it works as expected

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: isn't actual

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: we switched on running scalafix-cli as a process - now native works fine too

modules/cli/src/main/scala/scala/cli/commands/scalafix/Scalafix.scala
dos65: tested this loccaly - now it works as expected

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
```diff
@@ -0,0 +1,303 @@
...
+        scalaVersionArgs
+      ).call(cwd = root)
+      val updatedContent = noCrLf(os.read(root / "Hello.scala"))
+      expect(updatedContent == expectedContent)
+    }
+  }
+
+  test("external rule") {
+    val original: String =
+      """|//> using scalafix.dep "io.github.ghostbuster91.scalafix-unified::unified:0.0.8"
```
bjaglin: nitpicking: this looks quite unmaintained, so there is a risk that this prevents a future upgrade of scalafix. As it stands today, I would favor picking something from https://github.com/xuwei-k/scalafix-rules which is (by far) the most up-to-date (and arguably, popular) set of external rules.

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
```diff
@@ -0,0 +1,303 @@
...
+    inputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "scalafix", ".", "--power", scalaVersionArgs).call(cwd = root)
+      val updatedContent = noCrLf(os.read(root / "Hello.scala"))
+      expect(updatedContent == expectedContent)
+    }
+  }
+
+  test {
+    val name = "explicit-result-types"
+    if (actualScalaVersion.startsWith("3")) name.ignore else munit.TestOptions(name)
```
bjaglin: That changed since [my last comment](https://github.com/VirtusLab/scala-cli/pull/2968/files#r1714991736): `ExplicitResultTypes` is now available against Scala 3 sources, as of [Scalafix 0.13.0](https://github.com/scalacenter/scalafix/releases/tag/v0.13.0). As explained in the release notes, Scala 3 scalafix-cli must be loaded (and it seems to be the case in the current implementation of this PR) so it's definitely a good thing to test. Definitely not a blocker though!

modules/build/src/main/scala/scala/build/ScalafixArtifacts.scala
```diff
@@ -0,0 +1,157 @@
...
+      props
+    }
+
+  private def fetchScalafixInterfaces(
+    extraRepositories: Seq[Repository],
+    logger: Logger,
+    cache: FileCache[Task]
+  ): Either[BuildException, Path] =
+    either {
+      val scalafixInterfaces = dep"ch.epfl.scala:scalafix-interfaces:${Constants.scalafixVersion}"
```
bjaglin, line 104: Since the scalafix version does not seem to be overridable at runtime (yet?), could we just hardcode in the build the scalaVersions cli is published with to avoid fetching/scanning that and reducing a bit the complexity?

modules/build/src/main/scala/scala/build/ScalafixArtifacts.scala
```diff
@@ -0,0 +1,157 @@
...
+    zipFile.close()
+    out
+  }
+
+  private def scalafixPropsKey(scalaVersion: String): Either[BuildException, String] = {
+    val regex = "(\\d)\\.(\\d+).+".r
+    scalaVersion match {
+      case regex("2", "12")              => Right("scala212")
+      case regex("2", "13")              => Right("scala213")
+      case regex("3", x) if x.toInt <= 3 => Right("scala3LTS")
```
bjaglin, line 149: Since this will eventually change when LTS is no longer 3.3, I was wondering if it would be worth exposing that logic (currently hidden in the [classloading helper](https://github.com/scalacenter/scalafix/blob/aadace9120b9454dca4e91e319516bd8df30e6aa/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L123-L137) that we need to avoid for scala native) as a simple method in `Scalafix.java`. I am obviously very open to that for the next patch release of Scalafix.

However, I guess it's the same consideration for my comment as above:
- as long as the scalafix version is static, it's OK to keep that inlined here, and my suggestion might be overkill since we could do without scalafix-interfaces
- but if we want scala-cli to run future versions of scalafix-cli, that inlining could bring some incompatbilities

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
```diff
@@ -0,0 +1,304 @@
...
+    val inputs: TestInputs = TestInputs(
+      os.rel / confFileName ->
+        s"""|rules = [
+            |  EmptyCollectionsUnified
+            |]
+            |""".stripMargin,
+      os.rel / "Hello.scala" -> original
+    )
+    val expectedContent: String = noCrLf {
+      """|//> using scalafix.dep "io.github.ghostbuster91.scalafix-unified::unified:0.0.8"
```
bjaglin: As scalafix is (usually) used across several files, I doubt that a directive is the best UI to inject community rules. The best would be to expose a maven-coordinates-enabled `toolClasspath` CLI option, which would naturally fit upstream, so I filed https://github.com/scalacenter/scalafix/issues/2108 as a follow up.

bjaglin (approved):
Wow, great effort @dos65 :clap:

I added some comments as potential follow-ups/simplifications, but based on a review (I can't test right now), that's looking more than enough for an initial release from a Scalafix perspective :+1:

Gedochao:
@dos65 @Vigorge
Do you have time to address the remaining comments? Do you need help?
I think it's also fine to extract any remaining work to separate issues at this stage.

Other than that, it seems `docs-tests` is still failing, there seems to still be some issues in `website/docs/commands/scalafix.md`. Let me know if you need help with that.

dos65:
@Gedochao I started solving these comments. I hope will  complete that in one-two evenings

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
dos65: thx, fixed!

modules/build/src/main/scala/scala/build/ScalafixArtifacts.scala
dos65: I think we could. I did that to make scalafix upgrades less painful.

Maybe it would be better to move this fetching logic into the build and put values from `scalafix-properties` into `Constants.scalafixScala212` and etc. However, I would turn that into an issue

modules/build/src/main/scala/scala/build/ScalafixArtifacts.scala
dos65: > but if we want scala-cli to run future versions of scalafix-cli, that inlining could bring some incompatbilities

Yep, I was thinking about that but decided not to worry for the initial PR as LTS changing won't be happing often.

> I was wondering if it would be worth exposing that logic (currently hidden in the [classloading helper](https://github.com/scalacenter/scalafix/blob/aadace9120b9454dca4e91e319516bd8df30e6aa/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L123-L137) that we need to avoid for scala native) as a simple method in Scalafix.java

Yeah, that might be good idea.

Also, I was thinking maybe it would make sence to make a scala-independent launcher in scalafix project. It would be an exactly the same thing that we have in this PR: takes arguments, selects the scalaVersion for scalafix-cli, fetches using coursier and runs it. Then this logic might be properly shared only in scalafix project.

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
dos65: thx, swtiched this test  on xuwei-k rules

dos65:
@Gedochao could you approve ci run?

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
dos65: > As scalafix is (usually) used across several files, I doubt that a directive is the best UI to inject community rules.

I think it's fine. In case of having multiply files you don't need to define this dep in every file, having it in a single file will be enough

dos65:
@Gedochao There are some failed native-macos tests. I hope they are flaky. Otherwise I have no idea what might be wrong with them

dos65:
Great, now CI is green.

tgodzik:
Great work @dos65 ! @Gedochao is out seek, so we can wait for him to take a look next week.

Gedochao (approved):
LGTM
I'll merge this once Scala CLI 1.5.3 is out.

scarf005:
Sorry for nitpicking, but the title seems to have a typo:
```diff
- Scalafix command for scala-cli with basic optionns and tests
+ Scalafix command for scala-cli with basic options and tests
```
Really looking forward to this PR!

Gedochao:
> I'll merge this once Scala CLI 1.5.3 is out.

...make that 1.5.4 😩 (https://github.com/VirtusLab/scala-cli/issues/3273#issuecomment-2476641331)

modules/build/src/main/scala/scala/build/ScalafixArtifacts.scala
bjaglin: https://github.com/scalacenter/scalafix/issues/2127

PR #2977, opened 2024-06-24
Tag flaky docker image with scala.js app test

Gedochao:
The test has been failing on all PRs suddenly, we should investigate separately.
This change is to unblock other work.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
@@ -6,10 +6,10 @@ import scala.util.Properties
 
 class NativePackagerTests extends ScalaCliSuite {
   override def group: ScalaCliSuite.TestGroup = ScalaCliSuite.TestGroup.First
-  override def munitFlakyOK: Boolean = TestUtil.isCI
-  val helloWorldFileName             = "HelloWorldScalaCli.scala"
-  val message                        = "Hello, world!"
-  val licencePath                    = "DummyLICENSE"
+  override def munitFlakyOK: Boolean          = TestUtil.isCI
+  val helloWorldFileName                      = "HelloWorldScalaCli.scala"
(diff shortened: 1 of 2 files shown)


modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala
```diff
@@ -6,10 +6,10 @@ import scala.util.Properties
 
 class NativePackagerTests extends ScalaCliSuite {
   override def group: ScalaCliSuite.TestGroup = ScalaCliSuite.TestGroup.First
-
-  val helloWorldFileName = "HelloWorldScalaCli.scala"
-  val message            = "Hello, world!"
-  val licencePath        = "DummyLICENSE"
+  override def munitFlakyOK: Boolean = TestUtil.isM1 || TestUtil.isNativeCli || TestUtil.isCI
```
kasiaMarek: Wouldn't `TestUtil.isCI` be enough here?

Gedochao: it could be, but I don't want to accidentally break the other `.flaky` tests in this file: 
https://github.com/Gedochao/scala-cli/blob/d9055102c6ab55553e784d031969742b3c5419db/modules/integration/src/test/scala/scala/cli/integration/NativePackagerTests.scala#L120-L129

kasiaMarek: But that test requires `TestUtil.isCI` to be true to be marked as `.flaky` or am I missing something?

Gedochao: ahhh, you're right. I misread it by accident.
Lemme fix it.

Gedochao: Done

kasiaMarek (approved):

PR #2985, opened 2024-06-25
Pass `javaHome` to Bloop

### LINKED ISSUE #2663: Platform should be set when non default JVM is used.

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/2663

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
@@ -1,7 +1,6 @@
 package scala.cli.integration
 
 import com.eed3si9n.expecty.Expecty.expect
-import os.Pipe
 
 import scala.cli.integration.util.BloopUtil
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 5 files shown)


Gedochao (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/BloopTests.scala
```diff
@@ -1,6 +1,7 @@
 package scala.cli.integration
 
 import com.eed3si9n.expecty.Expecty.expect
+import os.Pipe
```
Gedochao: NIT don't import stuff from the `os.*` package, we generally use things from that package directly

PR #2988, opened 2024-06-26
fix for 2954 running script in root dir

### LINKED ISSUE #2954: Cannot run a script in the root directory
**Version(s)**
Scala-cli vesion 1.3.2

**Describe the bug**
Run a script with these contents:
```
//> using dep com.lihaoyi::requests:0.8.3
//> using dep com.lihaoyi::os-lib:0.10.2
//> using dep com.lihaoyi::upickle:3.3.1
//> using dep org.scala-lang.modules::scala-parallel-collections:1.0.4
println("Descargadas dependencias mas usadas")
```
localted in the root folder with:
```
scala-cli /myscript.sc
````
It complains like:
```
Error: os.PathError$LastOnEmptyPath: empty path has no last segment
````
When run in any other directory, it works ok as expected.

### PR DESCRIPTION
**Pull request by philwalk**
Adds fix for running script from the root directory.
Fixes #2954.

The problem occurs as a side-effect of calling `workspace.baseName` from within [scala.build.input.Inputs.scala](https://github.com/VirtusLab/scala-cli/blob/main/modules/build/src/main/scala/scala/build/input/Inputs.scala)
which throws an exception if `workspace` equals `os.root`.

See the calls to `.baseName` near lines 155 and 469.

The fix is to set `baseName` to an empty String if `workspace` equals `os.root`.

A change was also made to [scala.cli.ScalaCli.scala](https://github.com/VirtusLab/scala-cli/blob/main/modules/cli/src/main/scala/scala/cli/ScalaCli.scala) to insure that the logfile directory exists.

### DIFF
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -124,6 +124,7 @@ final case class Inputs(
     workspace / Constants.workspaceDirName / projectName / "jar"
   def docJarWorkDir: os.Path =
     workspace / Constants.workspaceDirName / projectName / "doc"
+
 }
 
 object Inputs {
@@ -152,7 +153,7 @@ object Inputs {
       updatedElems,
       defaultMainClassElemOpt,
       workspace,
-      workspace.baseName,
+      baseName(workspace),
       mayAppendHash = needsHash,
       workspaceOrigin = Some(workspaceOrigin),
       enableMarkdown = enableMarkdown,
@@ -384,7 +385,6 @@ object Inputs {
           }
         }.getOrElse((os.pwd, true, WorkspaceOrigin.Forced))
       }
-
       val (workspace, needsHash, workspaceOrigin0) = forcedWorkspace match {
         case None => (inferredWorkspace, inferredNeedsHash, workspaceOrigin)
         case Some(forcedWorkspace0) =>
@@ -466,7 +466,7 @@ object Inputs {
       elements = Nil,
       defaultMainClassElement = None,
       workspace = workspace,
-      baseProjectName = workspace.baseName,
+      baseProjectName = baseName(workspace),
       mayAppendHash = true,
       workspaceOrigin = None,
       enableMarkdown = enableMarkdown,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/cli/src/main/scala/scala/cli/ScalaCli.scala [hunk 1]
```diff
@@ -117,7 +117,7 @@ object ScalaCli {
     try main0(args)
     catch {
       case e: Throwable if !isCI && !printStackTraces =>
-        val workspace = CurrentParams.workspaceOpt.getOrElse(os.pwd)
+        val workspace = logDir(CurrentParams.workspaceOpt.getOrElse(os.pwd))
```
Gedochao: If I'm not mistaken, can just be simplified to:
```suggestion
        val workspace = CurrentParams.workspaceOpt.filter(os.isDir).getOrElse(os.pwd)
```

then the `logDir` function is redundant.

modules/build/src/main/scala/scala/build/input/Inputs.scala
```diff
@@ -466,7 +466,7 @@ object Inputs {
       elements = Nil,
       defaultMainClassElement = None,
       workspace = workspace,
-      baseProjectName = workspace.baseName,
+      baseProjectName = baseName(workspace),
```
Gedochao, line 469: no need to extract this into a function, it's not reused anywhere.
```suggestion
      baseProjectName = if workspace == os.root then "" else workspace.baseName,
```

Gedochao (commented):
Commented on some nitpicks.
I'm thinking if we can test that it works on the CI smh... maybe in a docker container? Perhaps doesn't have to be done in this PR though.

modules/build/src/main/scala/scala/build/input/Inputs.scala
philwalk: It's also used on line 156.

modules/cli/src/main/scala/scala/cli/ScalaCli.scala
philwalk: Yes, much better.

Gedochao (approved):

PR #2997, opened 2024-07-02
Ensure BSP respects --power mode

### LINKED ISSUE #2678: `--power` not reflected via `--json-options` for `bsp` subcommand
**Version(s)**
1.1.1

**Describe the bug**
`scala-cli bsp --power` sets `power` while `scala-cli bsp --json-options jsonOptions.json` where `jsonOptions.json` is
```json
{"powerOptions": {"power":true}}
``` 
does not set `power` to `true`.

**To Reproduce**
1. 
```scala 
//app.scala
@main def run = println("Hello world!")
```

2. 
```scala
//project.scala
//> using scala 3.3.1
//> using buildInfo
```

3. Open the project with an IDE with `metals` support. There should be a generated `.bsp/scala-cli.json` and `.scala-build/ide-options-v2.json`.

4. In `project.scala`, there should be the error on `using buildInfo`
image
that complains about the `power` option. 

5. In the root of the project, do `scala-cli setup-ide . --power`. Look at  content of `.scala-build/ide-options-v2.json`, you should find
```json
{"powerOptions": {"power":true}}
``` 

6. However the error in `project.scala` on `using buildInfo` does not go away although `ide-options-v2.json` is read through the arg `--json-options` in `.bsp/scala-cli.json`.

7. Go to `.bsp/scala-cli.json` and append `--power` to args. Error in `project.scala` disappears.

**Expected behaviour**
`--power` should be picked up when set through `--json-options`

Gedochao:
Fixes #2678 

This ensures the `bsp` sub-command (as called by an IDE) picks up on `--power` mode being set by:
- the `--power` launcher option being passed to `setup-ide`
- the `--power` sub-command option being passed to `setup-ide`
- the `SCALA_CLI_POWER` environment variable being passed to `setup-ide`
- the `power` key of the `config` sub-command

If any single one of these sets power mode on, BSP will pick it up.

Power mode should also be picked up by a running BSP server's `workspace/reload` when `setup-ide` or `config` are called.

Note: once power mode gets picked up by BSP, we don't turn it off at any point, even when explicitly passed to `setup-ide`. This is intentional behaviour, as we don't want to break anyone's existing IDE import to break. If you care for power mode to be turned off in your IDE, you need to explicitly restart the BSP server.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
@@ -10,12 +10,13 @@ import scala.build.bsp.{BspReloadableOptions, BspThreads}
 import scala.build.errors.BuildException
 import scala.build.input.Inputs
 import scala.build.options.{BuildOptions, Scope}
-import scala.cli.CurrentParams
 import scala.cli.commands.ScalaCommand
 import scala.cli.commands.publish.ConfigUtil.*
 import scala.cli.commands.shared.SharedOptions
 import scala.cli.config.{ConfigDb, Keys}
 import scala.cli.launcher.LauncherOptions
+import scala.cli.util.ConfigDbUtils
+import scala.cli.{CurrentParams, ScalaCli}
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
@@ -30,6 +31,7 @@ object Bsp extends ScalaCommand[BspOptions] {
         val content = os.read.bytes(os.Path(optionsPath, os.pwd))
         readFromArray(content)(SharedOptions.jsonCodec)
       }.getOrElse(options.shared)
+
   private def latestLauncherOptions(options: BspOptions): LauncherOptions =
     options.jsonLauncherOptions
       .map(path => os.Path(path, os.pwd))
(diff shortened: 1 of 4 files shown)


modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
```diff
@@ -114,6 +138,7 @@ object Bsp extends ScalaCommand[BspOptions] {
       val envs            = getEnvsFromFile()
       val bspBuildOptions = buildOptions(sharedOptions, launcherOptions, envs)
         .orElse(finalBuildOptions)
+      refreshPowerMode(launcherOptions, sharedOptions, envs)
```
kasiaMarek, line 141: Why are you reloading that also twice before? Is it because the options change and at one of the previous stages `power` might have been set to true and you want to pick that up too?

modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
```diff
@@ -51,6 +53,24 @@ object Bsp extends ScalaCommand[BspOptions] {
...
+  private def refreshPowerMode(
+    latestLauncherOptions: LauncherOptions,
+    latestSharedOptions: SharedOptions,
+    latestEnvs: Map[String, String]
+  ): Unit = {
+    val previousPowerMode = ScalaCli.allowRestrictedFeatures
+    val configPowerMode = ConfigDbUtils.getLatestConfigDbOpt(latestSharedOptions.logger)
+      .flatMap(_.get(Keys.power).toOption)
+      .flatten
+      .getOrElse(false)
```
kasiaMarek, line 65: Repeated logic with what is in `isSipScala`?

Gedochao: The main difference is that this is called multiple times while the BSP server is up, and we care about getting the most recent version of `configDb`, which is why we're using `ConfigDbUtils.getLatestConfigDbOpt`, rather than relying on the lazy value.

modules/cli/src/main/scala/scala/cli/commands/bsp/Bsp.scala
Gedochao: yes, exactly this. 
Whenever we load the latest `sharedOptions`, `launcherOptions` and `envs`, we also want to make sure we don't have to update power mode.

tgodzik (approved):
LGTM

PR #2998, opened 2024-07-03
docs: remove `.` from command snippet

spaceunifyfifty:
`scala-cli` contains in an error message:
```
You can run it with the `--power` flag or turn power mode on globally by running:
  scala-cli config power true.
```

The period `.` at the end makes the snippet no longer a valid terminal command, so copy pasting it will result in an error contrary to what most users will expect.
```
$ scala-cli config power true.
[error]  Config DB error: Malformed value 'true.' for the 'power' entry, expected a single boolean value ('true' or 'false').
```


Gedochao (approved):
Nice catch, thanks!

PR #3003, opened 2024-07-05
Add support for exporting to a Maven project

yadavan88:
Export scala-cli project to Maven build with simple pom.xml. 
~~Currently, implementation is done only for java projects~~


modules/cli/src/main/scala/scala/cli/commands/export0/ExportOptions.scala
```diff
@@ -29,6 +29,11 @@ final case class ExportOptions(
   @Group(HelpGroup.BuildToolExport.toString)
   @Tag(tags.restricted)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

modules/integration/src/test/scala/scala/cli/integration/MavenTestHelper.scala
```diff
@@ -0,0 +1,6 @@
+package scala.cli.integration
+
+trait MavenTestHelper {
+
+  protected def mavenCommand(args: String*): os.proc = os.proc("ls") // os.proc(sbt, args)
```
Gedochao: We'll need to run the exported project with maven here (I know, stating the obvious, I know it's WIP, just noting)

modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -129,17 +136,22 @@ object Export extends ScalaCommand[ExportOptions] {
...
       logger.error(
         s"Error: Cannot export to both mill and sbt. Please pick one build tool to export."
       )
       sys.exit(1)
     }
 
     if (!shouldExportToJson) {
-      val buildToolName = if (shouldExportToMill) "mill" else "sbt"
+      val buildToolName =
+        if (shouldExportToMill) "mill" else if (shouldExportToSbt) "sbt" else "maven"
```
Gedochao:
```suggestion
        if (shouldExportToMill) "mill" else if (shouldExportToMaven) "maven" else "sbt"
```
SBT should stay as the default.

Gedochao (commented):
Note: Make sure to reformat and regenerate reference docs, so that the CI stops complaining about it. 
Also, I know it's WIP, but I don't think the `pom.xml` logic is used atm, you're still generating an SBT project when `--maven` is passed.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescription.scala
```diff
@@ -0,0 +1,236 @@
...
+      null,
+      name,
+      scala.xml.Null,
+      scala.xml.TopScope,
+      minimizeEmpty = false,
+      scala.xml.Text(value)
+    )
+}
+
+final case class MavenProjectDescription(extraSettings: Seq[String], logger: Logger)
```
Gedochao:
```suggestion
final case class MavenProjectDescriptor(extraSettings: Seq[String], logger: Logger)
```

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescription.scala
```diff
@@ -0,0 +1,236 @@
...
+    javacOptionsSettings.toList
+  }
+
+  private def projectArtifactSettings(): MavenProject =
+    // todo: Is it really needed to configure the values, or always use the default?
+    MavenProject(
+      groupId = None,
+      artifactId = None,
+      version = None
+    )
```
Gedochao: we only require these when publishing an artifact in Scala CLI.
a Scala CLI project may have them defined if it's configured for publishing.
We should be able to read that.
https://scala-cli.virtuslab.org/docs/reference/directives#publish

As for defaults, I'm not sure if there's any sane ones.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescription.scala
```diff
@@ -0,0 +1,236 @@
...
+          // TODO dep.attributes
+          val artNameWithPrefix = dep.nameAttributes match {
+            case NoAttributes           => name
+            case s: ScalaNameAttributes => s"${name}_$getScalaMajorPrefix"
+          }
+          val scope0 =
+            // FIXME This ignores the isCompileOnly when scope == Scope.Test
+            if (scope == Scope.Test) Some("test")
+            else if (isCompileOnly) Some("provided")
+            else None
```
Gedochao: as discussed, dependencies in the test scope can't be provided for Maven, it seems.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescription.scala
```diff
@@ -0,0 +1,236 @@
...
+        {opts}
+      </javacArgs>
+    }
+
+    val configurationElements = Seq(compileMode, /*scalaVersionNone,*/ javacOptionsElem)
+
+    MavenPlugin(
+      "net.alchim31.maven",
+      "scala-maven-plugin",
+      "4.9.1",
```
Gedochao: the version of the `scala-maven-plugin` should be configurable via a command line option, and the default should be exported to Mill's `deps.sc` and passed via `Constants` all the way here.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescription.scala
```diff
@@ -0,0 +1,236 @@
...
+      </compilerArgs>
+    }
+
+    val sourceArg = buildNode("source", jdkVersion)
+    val targetArg = buildNode("target", jdkVersion)
+
+    MavenPlugin(
+      "org.apache.maven.plugins",
+      "maven-compiler-plugin",
+      "3.8.1",
```
Gedochao: similarly, this version should be configurable and the defaults passed via `Constants`

modules/cli/src/main/scala/scala/cli/exportCmd/ProjectDescriptor.scala
```diff
@@ -63,4 +65,12 @@ object ProjectDescriptor {
     calls
   }
 
+  def isPureJavaProject(options: BuildOptions, sources: Sources): Boolean =
+    !options.scalaOptions.addScalaLibrary.contains(true) &&
+    !options.scalaOptions.addScalaCompiler.contains(true) &&
+    sources.paths.forall(_._1.last.endsWith(".java")) &&
+    sources.inMemory.forall(_.generatedRelPath.last.endsWith(".java")) &&
```
Gedochao:
```suggestion
    sources.hasJava && !sources.hasScala &&
```

modules/integration/src/test/scala/scala/cli/integration/MavenTestHelper.scala
```diff
@@ -0,0 +1,13 @@
...
+
+trait MavenTestHelper {
+
+  protected def mavenCommand(args: String*): os.proc = os.proc(maven, args)
+
+  protected lazy val maven: os.Shellable =
+    Seq[os.Shellable](
+      "mvn",
+      "clean",
+      "install"
```
Gedochao:
```suggestion
     "mvn", "clean", "compile", "exec:java"
```
Also, unless the main class is defined in the POM (which it maybe could be, if it's defined in a Scala CLI build, this would make defining tests here easier), then we also need the `-Dexec.mainClass` option defined when calling Maven (can be passed in the tests directly)

modules/cli/src/main/scala/scala/cli/commands/export0/ExportOptions.scala
```diff
@@ -27,6 +27,11 @@ final case class ExportOptions(
   @HelpMessage("Sets the export format to SBT")
     sbt: Option[Boolean] = None,
   @Group(HelpGroup.BuildToolExport.toString)
+  @Tag(tags.experimental)
+  @Tag(tags.inShortHelp)
+  @HelpMessage("Sets the export format to Maven")
```
Gedochao, line 32: `--mvn` as an alias seems reasonable
```suggestion
  @Name("mvn")
  @HelpMessage("Sets the export format to Maven")
```

Gedochao (commented):
Added some comments while going through the current progress.

modules/cli/src/main/scala/scala/cli/commands/export0/ExportOptions.scala
Gedochao: ah... I meant to add `mvn` as an alias, while keeping `maven`. see the suggested code.

modules/integration/src/test/scala/scala/cli/integration/ExportMavenTestDefinitions.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  override def buildToolCommand(root: os.Path, args: String*): os.proc = mavenCommand(args*)
+
+  // todo: not sure if this is right. how to get the main class name
+  override val runMainArgs: Seq[String] =
+    if (language == JAVA) Seq("exec:java", "-Dexec.mainClass=main")
+    else Seq("scala:run", "-DmainClass=main")
+  override val runTestsArgs: Seq[String] =
+    if (language == JAVA) Seq("exec:java", "-Dexec.mainClass=main")
+    else Seq("scala:run", "-DmainClass=main")
```
Gedochao: we need to actually run the test suite here

modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -212,13 +226,23 @@ object Export extends ScalaCommand[ExportOptions] {
     }
     else {
       val sbtVersion = options.sbtVersion.getOrElse("1.10.1")
+      // todo: how to use it from deps.sc file here?
```
Gedochao: declare it in the `cli` module's `Constants.scala` file via `build.sc`, and then call `Constants` from here.

build.sc
```diff
@@ -491,6 +491,9 @@ trait Core extends ScalaCliCrossSbtModule
          |  def giter8Organization = "${Deps.giter8.dep.module.organization.value}"
          |  def giter8Name = "${Deps.giter8.dep.module.name.value}"
          |  def giter8Version = "${Deps.giter8.dep.version}"
+         |
+         |  def mavenVersion = "3.8.1"
+         |  def mavenScalaCompilerPluginVersion = "4.9.1"
```
Gedochao:
```suggestion
         |  def mavenVersion = "${Deps.Versions.mavenVersion}"
         |  def mavenScalaCompilerPluginVersion = "${Deps.Versions.mavenScalaCompilerPluginVersion}"
```

modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -211,14 +232,27 @@ object Export extends ScalaCommand[ExportOptions] {
...
         if (shouldExportToMill)
           millProjectDescriptor(options.shared.coursierCache, options.project, logger)
+        else if (shouldExportToMaven)
+          mavenProjectDescriptor(
+            defaultMavenCompilerVersion,
+            defaultScalaMavenCompilerVersion,
+            defaultMavenExecPluginVersion,
+            Nil,
+            logger
+          ) // todo: do we need extraSettings?
```
Gedochao:
```suggestion
          ) 
```
NIT: dev leftover? what do you mean by `extraSettings` here?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProject.scala
```diff
@@ -0,0 +1,147 @@
...
+  artifactId: Option[String] = None,
+  version: Option[String] = None,
+  plugins: Seq[MavenPlugin] = Nil,
+  imports: Seq[String] = Nil,
+  settings: Seq[Seq[String]] = Nil,
+  dependencies: Seq[MavenLibraryDependency] = Nil,
+  mainSources: Seq[(os.SubPath, String, Array[Byte])] = Nil,
+  testSources: Seq[(os.SubPath, String, Array[Byte])] = Nil,
+  resourceDirectories: Seq[String] = Nil
+  // properties: Seq[(String, String)] = Nil // using Seq[Tuple] since derive was failing for Map
```
Gedochao:
```suggestion
```
NIT: dev leftover?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProject.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    val buildMavenContent = MavenModel(
+      "4.0.0",
+      groupId.getOrElse("groupId"),
+      artifactId.getOrElse("artifactId"),
+      version.getOrElse("0.1-SNAPSHOT"),
+      dependencies,
+      plugins,
+      resourceDirectories
+      // properties
```
Gedochao:
```suggestion
```
NIT: dev leftover?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProject.scala
```diff
@@ -0,0 +1,147 @@
...
+
+final case class MavenModel(
+  model: String,
+  groupId: String,
+  artifactId: String,
+  version: String,
+  dependencies: Seq[MavenLibraryDependency],
+  plugins: Seq[MavenPlugin],
+  resourceDirectories: Seq[String]
+  // properties: Seq[(String, String)] //todo: bring back only if it is needed
```
Gedochao:
```suggestion
```

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+    MavenProject(
+      mainSources = mainSources,
+      testSources = testSources
+    )
+  }
+
+  // todo: fill this
+  private def javaOptionsSettings(options: BuildOptions): MavenProject =
+    MavenProject(
+      settings = Nil
```
Gedochao, line 59: Looks like we might need this (we don't have a test for it, do we?)...

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+    javacOptionsSettings.toList
+  }
+
+  private def projectArtifactSettings(): MavenProject =
+    // todo: Is it really needed to configure the values, or always use the default?
+    MavenProject(
+      groupId = None,
+      artifactId = None,
+      version = None
+    )
```
Gedochao: We do need to configure the values, but I suppose we're fine with some default, perhaps derived from the root directory name? (and maybe a warning, what's being used?).
I think we're also missing command line options for configuring this. I think it'd be applicable for Mill & SBT as well, but it's obligatory for Maven.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+    sources: Sources
+  ): MavenProject = {
+
+    val scalaV = getScalaVersion(options)
+    def getScalaMajorPrefix =
+      scalaV match {
+        case s"2.12.${patch}" => "2.12"
+        case s"2.13.${patch}" => "2.13"
+        case s"3.$x.$y"       => "3"
+      }
```
Gedochao: IMO safer to just
```suggestion
      if scalaV.startsWith("3") then "3" 
      else if scalaV.startsWith("2.12") then "2.13"
      else "2.12"
```

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+
+  private def dependencySettings(
+    options: BuildOptions,
+    testOptions: BuildOptions,
+    scope: Scope,
+    sources: Sources
+  ): MavenProject = {
+
+    val scalaV = getScalaVersion(options)
+    def getScalaMajorPrefix =
```
Gedochao:
```suggestion
    def getScalaPrefix =
```
`major` implies the major version component, which would be `2` for `2.12.x`

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+      mainDeps: ShadowingSeq[Positioned[AnyDependency]],
+      isCompileOnly: Boolean
+    ) =
+      mainDeps.toSeq.toList.map(_.value).map { dep =>
+        val org  = dep.organization
+        val name = dep.name
+        val ver  = dep.version
+        // TODO dep.userParams
+        // TODO dep.exclude
+        // TODO dep.attributes
```
Gedochao, line 113: should we cover these in this PR, or extract to an issue?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+        val artNameWithPrefix = dep.nameAttributes match {
+          case NoAttributes           => name
+          case s: ScalaNameAttributes => s"${name}_$getScalaMajorPrefix"
+        }
+        val scope0 =
+          if (scope == Scope.Test) MavenScopes.Test
+          else if (isCompileOnly)
+            MavenScopes.Provided // maven seems to support either test or provided, not both
```
Gedochao: might be worth a warning

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    MavenProject(
+      dependencies = depSettings
+    )
+  }
+
+  private def getScalaVersion(options: BuildOptions): String =
+    options.scalaParams.toOption.flatten.map(_.scalaVersion).getOrElse(
+      ScalaCli.getDefaultScalaVersion // FIXME account for pure Java projects, where Scala version isn't defined
```
Gedochao: should we cover the `FIXME` comment in this PR, or extract to an issue?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+    )
+
+  private def plugins(
+    options: BuildOptions,
+    scope: Scope,
+    jdkVersion: String,
+    sourcesMain: Sources
+  ): MavenProject = {
+
+    // todo: use this method from mill and sbt projects as well
```
Gedochao: .

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
```diff
@@ -0,0 +1,316 @@
...
+  def `export`(
+    optionsMain: BuildOptions,
+    optionsTest: BuildOptions,
+    sourcesMain: Sources,
+    sourcesTest: Sources
+  ): Either[BuildException, MavenProject] = {
+    val jdk =
+      optionsMain.javaOptions.jvmIdOpt.map(_.value).getOrElse(
+        "17"
+      ) // todo: get from constants for default jdk
```
Gedochao, line 300: .

modules/integration/src/test/scala/scala/cli/integration/ExportMavenTestDefinitions.scala
```diff
@@ -0,0 +1,48 @@
...
+      "--mvn",
+      "-o",
+      outputDir.toString,
+      args
+    )
+
+  override def buildToolCommand(root: os.Path, mainClass: Option[String], args: String*): os.proc =
+    mavenCommand(args*)
+
+  // todo: not sure if this is right. how to get the main class name
```
Gedochao:
```suggestion
```
dev leftover

modules/integration/src/test/scala/scala/cli/integration/ExportMavenTestJava.scala
```diff
@@ -0,0 +1,12 @@
+package scala.cli.integration
+
+class ExportMavenTestJava extends ExportMavenTestDefinitions with Test3Lts with MavenJava {
+  // disable running scala tests in java maven export
+  override def runExportTests: Boolean = false
+  test("pure java") {
```
Gedochao:
```suggestion
  if (!Properties.isWin) test("pure java") {
```
The other tests export tests apparently are only being run on mac for now. We can skip Windows here and figure it out in a future PR/issue. It passes as it should on the other platforms as is.

Gedochao (commented):
Looking good already, we just need a couple tweaks. Great contribution!

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
yadavan88: This shouldn't be an issue now. In maven, it is now handled based on purJava variable.

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
yadavan88: I think it is better to extract to separate issue and fix in based on demand. At the initial step, this might be unnecessary as the chance of using exclude etc are less.

modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
yadavan88: I kept it to support for future extension to add more attributes/settings to the pom. I removed the comment, but I think, it is better to leave the argument there as it is for adding more options in future. wdyt?

modules/cli/src/main/scala/scala/cli/exportCmd/MavenProjectDescriptor.scala
yadavan88: created issue in github

Gedochao (approved):
A bunch of things can still be done to polish the solution, but we can do those as a follow-ups.
LGTM, great contribution!

Extracted issues:
- https://github.com/VirtusLab/scala-cli/issues/3082
- https://github.com/VirtusLab/scala-cli/issues/3083

PR #3004, opened 2024-07-05
chore: Bump outdated `javac-semanticdb` plugin to 0.10.0

tgodzik:
This causes some issues when running Java via metals if there is no package defined (java semanticdb has some wierd stuff)

Not sure why this isn't being bumped.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/SemanticDbTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SemanticDbTestDefinitions.scala
@@ -44,7 +44,7 @@ trait SemanticDbTestDefinitions { _: CompileTestDefinitions =>
           .flatMap(opt => List("--javac-opt", opt))
         val javaSemDbOptions = Seq(
           "--javac-plugin",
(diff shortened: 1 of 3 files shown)


modules/integration/src/test/scala/scala/cli/integration/SemanticDbTestDefinitions.scala
```diff
@@ -44,7 +44,7 @@ trait SemanticDbTestDefinitions { _: CompileTestDefinitions =>
           .flatMap(opt => List("--javac-opt", opt))
         val javaSemDbOptions = Seq(
           "--javac-plugin",
-          "com.sourcegraph:semanticdb-javac:0.7.4",
+          "com.sourcegraph:semanticdb-javac:0.10.0",
```
Gedochao: maybe we could pass this version from `deps.sc` via `Constants`?

tgodzik: Done!

Gedochao (approved):
LGTM

PR #3009, opened 2024-07-09
Add Scala to pure Java test builds

### LINKED ISSUE #2940: Scala CLI fails to run tests for a pure Java project
**Version(s)**
`1.3.1`

**Describe the bug**
Running the `test` command for a pure Java project results in an error message like
```text
[error]  Got Scala dependency org.virtuslab.scala-cli::test-runner:1.3.1, but no Scala version is provided
```

**To Reproduce**
This can be reproduced from the [coursier/directories-jvm](https://github.com/coursier/directories-jvm) repository:
```text
$ git rev-parse HEAD
968eb94f7c837e768e00dfc4e9971de7cc4018e1
$ scala-cli test .
…
[error]  Got Scala dependency org.virtuslab.scala-cli::test-runner:1.3.1, but no Scala version is provided
```

**Expected behaviour**
Scala CLI runs tests fine for pure Java projects

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/2940

We generally don't add Scala to pure Java builds to keep them clean. Unfortunately, we do not have a pure Java test runner at the moment, so we still need Scala to run tests properly.

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -1029,7 +1029,8 @@ object Build {
     either {
 
       val options0 =
-        if (sources.hasJava && !sources.hasScala)
+        // FIXME: don't add Scala to pure Java test builds (need to add pure Java test runner)
+        if (sources.hasJava && !sources.hasScala && scope != Scope.Test)
           options.copy(
             scalaOptions = options.scalaOptions.copy(
               scalaVersion = options.scalaOptions.scalaVersion.orElse {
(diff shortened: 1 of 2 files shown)


modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -1029,7 +1029,8 @@ object Build {
     either {
 
       val options0 =
-        if (sources.hasJava && !sources.hasScala)
+        // FIXME: don't add Scala to pure Java test builds (need to add pure Java test runner)
+        if (sources.hasJava && !sources.hasScala && scope != Scope.Test)
```
Gedochao, line 1033: We should add a pure Java test runner at some point, so that we can keep Java test builds pure as well.

Gedochao: Created an issue: 
- https://github.com/VirtusLab/scala-cli/issues/3010

tgodzik (approved):
Btw. Can't we use Bloop in that case? It should work correctly without Scala

Gedochao:
> Btw. Can't we use Bloop in that case? It should work correctly without Scala

might be worth a try... I'll check in a separate PR

PR #3012, opened 2024-07-09
feat: allow to pass scalac options using files

### LINKED ISSUE #2958: java 9+ arguments file not yet supported
A better way to add extra jars to the a script classpath is needed, equivalent to the "arguments file" feature of `java 9+`.

The current support for adding jars that aren't in a repository to a script classpath is restrictive.
It also prevents such scripts from being os-portable, because it requires the `path.separator` to be imbedded in the script.

The following script successfully adds a jar to the classpath:
```
#!/usr/bin/env -S scala-cli shebang --extra-jars C:/Users/username/lib/privateJarA.jar
//> using scala "3.4.2"
sys.props("java.class.path").split(psep).foreach { println }
def psep: String = sys.props("path.separator")
```
This approach works for jar paths that are sufficiently short, but in general, it's inflexible, and it requires a non-portable classpath.

The old script runner supports using an `arguments file` to get the classpath argument.
```
#!/usr/bin/env -S scala @/opt/ue/argsFile
  def psep: String = sys.props("path.separator")
  sys.props("java.class.path").split(psep).foreach { println }

```
It's succinct, flexible and scalable.  It moves the `path.separator` from the script to a local path, resulting in an os-portable script.

I tried to pass the `arguments file` as a `javaOpt`, but it quietly failed (it didn't modify the classpath):
```
//> using scala "3.4.2"
//> using javaOpt @/opt/ue/argsFile
def psep: String = sys.props("path.separator")
sys.props("java.class.path").split(psep).foreach { println }
```

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/2958

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -3,7 +3,6 @@ package scala.cli.commands.shared
 import bloop.rifle.BloopRifleConfig
 import caseapp.*
 import caseapp.core.Arg
-import caseapp.core.argparser.SimpleArgParser
 import caseapp.core.help.Help
 import caseapp.core.util.Formatter
 import com.github.plokhotnyuk.jsoniter_scala.core.*
@@ -13,7 +12,6 @@ import coursier.core.Version
 import coursier.util.{Artifact, Task}
 import dependency.AnyDependency
 import dependency.parser.DependencyParser
-import shapeless.{:: => :*:, HNil}
 
 import java.io.{File, InputStream}
 import java.nio.file.Paths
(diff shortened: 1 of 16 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -289,13 +295,20 @@ final case class SharedOptions(
     )
   }
 
+  lazy val scalacOptionsFromFiles: List[String] =
+    argsFiles.flatMap(argFile =>
+      os.read(os.Path(argFile.file, os.pwd)).split(System.lineSeparator())
```
kasiaMarek: Should there be some special error handling here?

Gedochao (approved):
LGTM

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
Gedochao: I think it's fine.

PR #3017, opened 2024-07-11
Swap `scalameta` `trees` for `semanticdb-shared` & bump `scalameta` to 4.9.8

Gedochao:
Supersedes #3016

### DIFF
--- a/project/deps.sc
+++ b/project/deps.sc
@@ -97,7 +97,7 @@ object Deps {
     def coursierM1Cli                     = coursierDefault
     def jsoniterScala                     = "2.23.2"
     def jsoniterScalaJava8                = "2.13.5.2"
-    def scalaMeta                         = "4.9.7"
+    def scalaMeta                         = "4.9.8"
     def scalaNative04                     = "0.4.17"
     def scalaNative05                     = "0.5.4"
     def scalaNative                       = scalaNative05
--- a/project/deps.sc
+++ b/project/deps.sc
@@ -188,6 +188,7 @@ object Deps {
   def semanticDbScalac         = ivy"org.scalameta:::semanticdb-scalac:${Versions.scalaMeta}"
   def scalametaSemanticDbShared =
     ivy"org.scalameta:semanticdb-shared_${Scala.scala213}:${Versions.scalaMeta}"
+      .exclude("org.jline" -> "jline") // to prevent incompatibilities with GraalVM <23
   def signingCliShared =
     ivy"org.virtuslab.scala-cli-signing::shared:${Versions.signingCli}"
       // to prevent collisions with scala-cli's case-app version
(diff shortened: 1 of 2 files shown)


Gedochao:
https://github.com/jline/jline3/issues/954
Ok... so it seems we swallow Jline 3.25.1 along with `scalameta` 4.9.8 here.
That version of JLine seems to require Graal 23+...  so we need to bump Graal first somehow for our native images to get generated properly.
Otherwise we're getting:
```
Error: Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-terminal/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-terminal-jansi/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-terminal-jni/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jansi/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-terminal-jna/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-terminal-ffm/native-image.properties'. Use -H:PrintFlags= to list all available options.
Could not find option 'UnlockExperimentalVMOptions' from 'jar:file:///data/cp/jline-3.25.1.jar!/META-INF/native-image/org.jline/jline-native/native-image.properties'. Use -H:PrintFlags= to list all available options.
```
And bumping Graal is tricky, since the aarch64 platform has been dropped from CE after the version we're at.
I'll try looking into it separately, but it seems this is blocked for now.

project/deps.sc
```diff
@@ -180,16 +180,15 @@ object Deps {
   def scalaJsLogging   = ivy"org.scala-js:scalajs-logging_2.13:1.1.1"
   // Force using of 2.13 - is there a better way?
   def scalaJsTestAdapter = ivy"org.scala-js:scalajs-sbt-test-adapter_2.13:${Scala.scalaJs}"
-  // Force using of 2.13 - is there a better way?
-  def scalametaTrees = ivy"org.scalameta:trees_2.13:${Versions.scalaMeta}"
```
tgodzik, line 184: So we never really needed trees actually? This actually seems like the wrong dependency was used

tgodzik (approved):
Thanks for digging into this!

project/deps.sc
Gedochao: It seems we only needed it for the SemanticDB stuff, yeah.

PR #3029, opened 2024-07-16
Fix --offline mode for scala-cli as scala installation via coursier

### LINKED ISSUE #3021: can not use explict repo in offline mode when coursier cache has only POM file
**Version(s)**
1.4.0 (problematic in scala 3.5.0-RC3 release when installing from `cs install scala:3.5.0-RC3`)

**Describe the bug**
in offline mode, If the coursier cache is only partially populated by POM, and no jar file (such is the state after running the `cs install` command) - then any explicitly added local repository is ignored and classpath resolution fails

**To Reproduce**
follow steps in https://github.com/coursier/coursier/issues/3024

**Expected behaviour**
I have passed an explicit repo in offline mode, I expect to resolve the classpath from there, the fact that coursier complains underneath is a leaky abstraction.

Gedochao:
Fixes #3021

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
@@ -847,4 +847,61 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
       expect(res.out.trim().contains(scalaVersion))
     }
   }
+
+  if (!Properties.isWin) // FIXME: run this test on Windows
+    test("coursier scala installation works in --offline mode") {
+      TestInputs.empty.fromRoot { root =>
+        val localCache = root / "local-cache"
+        val localBin   = root / "local-bin"
+        val sv         = "3.5.0-RC4"
+        os.proc(
+          TestUtil.cs,
+          "install",
+          "--cache",
+          localCache,
+          "--install-dir",
+          localBin,
+          s"scala:$sv"
+        ).call(cwd = root)
+        val scalaBinary: os.Path = localBin / "scala"
+        val fileBytes            = os.read.bytes(scalaBinary)
+        val shebang              = new String(fileBytes.takeWhile(_ != '\n'), "UTF-8")
+        val binaryData           = fileBytes.drop(shebang.length + 1)
+        val execLine             = new String(binaryData.takeWhile(_ != '\n'), "UTF-8")
+        val scriptPathRegex      = """exec "([^"]+/bin/scala).*"""".r
+        val scalaScript = execLine match { case scriptPathRegex(extractedPath) => extractedPath }
+        val scalaScriptPath = os.Path(scalaScript)
+        val lineToChange    = "eval \"${SCALA_CLI_CMD_BASH[@]}\" \\"
+        // FIXME: the way the scala script calls the launcher currently ignores the --debug flag
+        val newContent = os.read(scalaScriptPath).replace(
+          lineToChange,
+          s"""SCALA_CLI_CMD_BASH=(\"\\\"${TestUtil.cliPath}\\\"\")
+             |$lineToChange""".stripMargin
+        )
+        os.write.over(scalaScriptPath, newContent)
+        val r =
+          os.proc(
+            scalaScript,
+            "--offline",
+            "--power",
+            "--with-compiler",
+            "-e",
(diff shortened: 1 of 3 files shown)


modules/options/src/main/scala/scala/build/Artifacts.scala
```diff
@@ -624,9 +624,11 @@ object Artifacts {
     val forceVersion = forceScalaVersions ++ forcedVersions
 
     // FIXME Many parameters that we could allow to customize here
+    val defaultFetcher = coursier.Fetch()
```
tgodzik, line 627: Should we use it in the line below?

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -847,4 +847,61 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
...
+        val scalaBinary: os.Path = localBin / "scala"
+        val fileBytes            = os.read.bytes(scalaBinary)
+        val shebang              = new String(fileBytes.takeWhile(_ != '\n'), "UTF-8")
+        val binaryData           = fileBytes.drop(shebang.length + 1)
+        val execLine             = new String(binaryData.takeWhile(_ != '\n'), "UTF-8")
+        val scriptPathRegex      = """exec "([^"]+/bin/scala).*"""".r
+        val scalaScript = execLine match { case scriptPathRegex(extractedPath) => extractedPath }
+        val scalaScriptPath = os.Path(scalaScript)
+        val lineToChange    = "eval \"${SCALA_CLI_CMD_BASH[@]}\" \\"
+        // FIXME: the way the scala script calls the launcher currently ignores the --debug flag
```
tgodzik, line 875: Do we have an issue for that?

tgodzik (approved):
Lgtm!

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: it's the mill flag, and only affects this particular test.
I don't think it warrants an issue.

Gedochao: the `scala-cli` user-facing `--debug` flag is working as expected.

modules/options/src/main/scala/scala/build/Artifacts.scala
Gedochao: Ah yes, missed that.
Fixed.

Gedochao:
Kudos to @kasiaMarek for the initial idea for this hack, btw.

bishabosha (approved):
Can confirm this works very well (`pwd` returns the directory of the coursier installed scala 3.5.0-RC4)-
```bash
/Users/jamie/Workspace/scala-cli/out/cli/3.3.3/standaloneLauncher.dest/launcher \
  --prog-name scala \
  --cli-default-scala-version 3.5.0-RC4 \
  -r file://$(pwd)/maven2 \
  --power -e 'println(sys.props("java.class.path"))' --offline --server=false
The `--offline` option is experimental
/var/folders/s7/dl7h4p057cz0fw4xgrv9snmm0000gn/T/10085725059665397379/.scala-build/10085725059665397379_39f5220048-c1fd01f914/classes/main:/Users/jamie/Library/Caches/Coursier/arc/https/github.com/scala/scala3/releases/download/3.5.0-RC4/scala3-3.5.0-RC4-aarch64-apple-darwin.tar.gz/scala3-3.5.0-RC4-aarch64-apple-darwin/maven2/org/scala-lang/scala3-library_3/3.5.0-RC4/scala3-library_3-3.5.0-RC4.jar:/Users/jamie/Library/Caches/Coursier/arc/https/github.com/scala/scala3/releases/download/3.5.0-RC4/scala3-3.5.0-RC4-aarch64-apple-darwin.tar.gz/scala3-3.5.0-RC4-aarch64-apple-darwin/maven2/org/scala-lang/scala-library/2.13.14/scala-library-2.13.14.jar
```

this is under the same conditions - coursier cache only has POM for scala3-compiler_3:3.5.0-RC4

kasiaMarek:
> Kudos to @kasiaMarek for the initial idea for this hack, btw.

Oh, how nice and thoughtful, though I believe it was actually @tgodzik's idea.

PR #3032, opened 2024-07-16
BugFix - Export issues when full scala version is not used in directives

### LINKED ISSUE #3031: Broken SBT export
**Version(s)**
version: 1.4.0

**Describe the bug**
When exporting the scala-cli project to sbt (and possibly the same for mill), the build configuration for scalaVersion is generating wrongly if the full scala version is not used in the directive

**To Reproduce**
Sample.scala file
```
//> using scala 3
@main def main = println("Hello!")
```

Export using:
`scala-cli --power export Sample.scala --sbt`

try to run this sbt project using `sbt run`. It fails due to the build.sbt having wrong scala version:
```
scalaVersion := "3"
```

**Expected behaviour**
It should be `scalaVersion := "3.3.4"` probably the lts version. It could be having the same issue in mill export as well.

yadavan88:
Bugfix for the issue https://github.com/VirtusLab/scala-cli/issues/3031


modules/options/src/main/scala/scala/build/options/ScalaOptions.scala
```diff
@@ -29,6 +30,20 @@ final case class ScalaOptions(
...
+      .map {
+        case "3"                                     => Constants.defaultScalaVersion
+        case "2.12"                                  => Constants.defaultScala212Version
+        case "2.13"                                  => Constants.defaultScala213Version
+        case ver @ s"$m.$mn.$p"                      => ver
+        case s"3.${lts}" if lts.toLowerCase == "lts" => Constants.defaultScalaVersion
+        case s"3.${any}"                             => Constants.defaultScalaVersion
+        case v                                       => v
+        // todo: how to get current lts and next versions
+      }
```
Gedochao: This is already covered in `BuildOptions.scalaParams`(where the Scala version validation logic is quite complex), we should likely use that.

modules/integration/src/test/scala/scala/cli/integration/ExportCommonTestDefinitions.scala
```diff
@@ -21,6 +21,7 @@ trait ExportCommonTestDefinitions { _: ScalaCliSuite & TestScalaVersionArgs =>
   protected val prepareTestInputs: TestInputs => TestInputs = identity
 
   protected val outputDir: os.RelPath = os.rel / "output-project"
+  def majorScalaVersion: String
```
Gedochao: calling it `majorScalaVersion` feels confusing... by `major` we usually mean the major version component (so the first digit of the version, as opposed to minor/patch).
I'd also not cover it by inheritance, and rather test the versions we want to check in a for loop.
for example, in the current state of the PR, we don't check exporting `3.nightly`.

modules/integration/src/test/scala/scala/cli/integration/ExportCommonTestDefinitions.scala
```diff
@@ -61,6 +61,16 @@ trait ExportCommonTestDefinitions { _: ScalaCliSuite & TestScalaVersionArgs =>
       expect(output.contains("Hello"))
     }
 
+  protected def scalaVersionTest(majorScalaVersion: String): Unit =
```
Gedochao:
```suggestion
  protected def scalaVersionTest(scalaVersion: String): Unit =
```

modules/cli/src/main/scala/scala/cli/exportCmd/MillProjectDescriptor.scala
```diff
@@ -38,9 +38,9 @@ final case class MillProjectDescriptor(
       options.classPathOptions.allExtraDependencies.toSeq
         .forall(_.value.nameAttributes == NoAttributes)
 
-    val sv = options.scalaOptions.scalaVersion
-      .flatMap(_.versionOpt) // FIXME If versionOpt is empty, the project is pure Java
-      .getOrElse(ScalaCli.getDefaultScalaVersion)
+    val sv = options.scalaParams.toOption.flatten.map(_.scalaVersion).getOrElse(
+      ScalaCli.getDefaultScalaVersion
+    )
```
Gedochao: let's not lose the context of this comment, remains valid
```suggestion
    val sv = options.scalaParams.toOption.flatten.map(_.scalaVersion).getOrElse(
      ScalaCli.getDefaultScalaVersion // FIXME account for pure Java projects, where Scala version isn't defined
    )
```

modules/cli/src/main/scala/scala/cli/exportCmd/SbtProjectDescriptor.scala
```diff
@@ -121,9 +121,11 @@ final case class SbtProjectDescriptor(
   private def scalaVersionSettings(options: BuildOptions): SbtProject = {
 
     val scalaVerSetting = {
-      val sv = options.scalaOptions.scalaVersion
-        .flatMap(_.versionOpt) // FIXME If versionOpt is empty, the project is pure Java
-        .getOrElse(ScalaCli.getDefaultScalaVersion)
+
+      val sv = options.scalaParams.toOption.flatten.map(_.scalaVersion).getOrElse(
+        ScalaCli.getDefaultScalaVersion
+      )
```
Gedochao:
```suggestion
      val sv = options.scalaParams.toOption.flatten.map(_.scalaVersion).getOrElse(
        ScalaCli.getDefaultScalaVersion // FIXME account for pure Java projects, where Scala version isn't defined
      )
```

modules/integration/src/test/scala/scala/cli/integration/ExportTestProjects.scala
```diff
@@ -314,4 +314,14 @@ object ExportTestProjects {
            |}
            |""".stripMargin
     )
+
+  def majorScalaVersionTest(scalaVersionMajor: String): TestInputs =
```
Gedochao: The word "major" is confusing here, IMO
```suggestion
  def scalaVersionTest(scalaVersion: String): TestInputs =
```

Gedochao (commented):
Commented on some nitpicks, otherwise LGTM!

Gedochao (approved):

Issue #610, opened 2022-02-03
Support code generators


kubukoz:
**Is your feature request related to a problem? Please describe.**

Build tools generally provide ways to generate source files before compilation. scala-cli doesn't, it was previously mentioned it potentially could.

**Describe the solution you'd like**

Some way to run an arbitrary script before every compilation, probably configured in directives or by convention (i.e. putting a script in a `./scala-scripts` directory).

The interface would be basically `Unit => Unit`, maybe with extra context passed as environment variables.

Some possible things I'd like to see passed:

- a way to gather all the sources used in the build (e.g. `:`-separated list of source paths)
- the path to a directory where I can output files that'll be included in the compilation. Alternatively I could write to anywhere I want and these paths would be added manually in a directive. Not sure how this would work with cross-compilation.
- pwd (or something like a workspace root, in case of bsp? I don't know how exactly that protocol works though)
- build metadata (in theory I could read the directives myself, but having this passed would be much better), e.g. deps, resolved scala version, target platform)

Note that this would be triggered on IDE compilations as well.

There should be a way to customize the script beyond the runnable name. Ideas:

 - Passing args within the directive, e.g. `//> build script echo foo bar`
 - Wrapping the script with another, and using the other script in the directive. E.g. script `echow` runs `echo foo bar`, and I do `//> build script echow`

Unfortunately, any kind of scripts would mean that builds can't be shared via e.g. gists. Personally I'm fine with this.

**Describe alternatives you've considered**

Making my own build tool wrapping scala-cli :)

**Additional context**

Not much to write here.

ckipp01:
Just to tie these together I just had a usecase where having something like this would have been super useful. I wrote about it in [here](https://github.com/VirtusLab/scala-cli/discussions/2051), but to reiterate I essentially needed a `BuildInfo` that held the version of my app so that it could be displayed to users. Since I use a `Makefile` for the project I was able to make sure that any command ran a [script](https://github.com/ckipp01/skan/blob/main/scripts/generate-build-info.scala) before running the actual command to compile. This sort of works fine until you get out of the context of the `Makefile`. For example I just realized now that Scala Steward can no longer run on my project because of this. Having something like the approach outlined by @kubukoz would really help in situations like this.

slabuz:
Hi @kubukoz 

[...]

PR #3035, opened 2024-07-17
[OLD PoC] Source generators

MaciejG604:
This is an old PoC for implementing source generators that was supposed to use Bloops capability to run an arbitrary command that would generate the sources.

Bloop.Project.sourceGenerators accepts a list of objects of this shape:
`SourceGenerator(sourcesGlobs: List[SourcesGlobs], outputDirectory: PlatformFiles.Path, command: List[String])`

By quickly looking [at Bloop's code](https://github.com/scalacenter/bloop/blob/71c5273f561ec6fb2c05e8949526c7b47a4d5c7e/frontend/src/main/scala/bloop/engine/SourceGenerator.scala#L67) the command gets created like this:
`(command :+ outputDirectory.syntax) ++ inputs.keys.map(_.syntax)`, where the inputs are the results of applying sourceGlobs

So trying to work with this, I wanted to run `scala-cli` command on a small app that uses `dep "com.thesamet.scalapb::scalapbc:0.11.13"`, which would generate the sources before Bloop's compilation would run.
Unfortunately this didn't work, I recall that the command never managed to generate sources even though Bloop would run it (I think). Probably there was a problem with the command also calling Bloop.

I thought about a while ago and IMO for it to work we would need to first build the generator app (potentially without Bloop) and pass the run command to Bloop (sth like `java --cp jar1:jar2:scalapbc-jar MainClass outputDir sourceArg1 sourceArg2 ...`
We already do similar stuff in our [Run command](https://github.com/VirtusLab/scala-cli/blob/e1c5c9cbf9fcfdd0badc8eeeb7ffad2244a932ba/modules/cli/src/main/scala/scala/cli/commands/run/Run.scala#L319-L342).

Off the top of my head this generator app could also be released to maven, but I don't see any problems in hardcoding it in our sources.

Additionally I think the easiest way would be to rework the Bloop API to allow more flexibility e.g. for calling ScalaPBC's main without the generator-app from `gen.scala` as it effectively just shuffles arguments around.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -892,6 +892,10 @@ trait Cli extends CrossSbtModule with ProtoBuildModule with CliLaunchers
 
   def localRepoJar = `local-repo`.localRepoJar()
 
+  def forkArgs = T {
+    super.forkArgs() ++ Seq("-agentlib:jdwp=transport=dt_socket,server=y,address=localhost:5050,suspend=y")
+  }
+
   object test extends ScalaCliTests with ScalaCliScalafixModule {
     def moduleDeps = super.moduleDeps ++ Seq(
       `build-module`(crossScalaVersion).test
--- a/modules/build/src/main/scala/scala/build/Project.scala
+++ b/modules/build/src/main/scala/scala/build/Project.scala
@@ -1,15 +1,16 @@
 package scala.build
 
-import _root_.bloop.config.{Config => BloopConfig, ConfigCodecs => BloopCodecs}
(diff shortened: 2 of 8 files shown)


Gedochao:
Linking https://github.com/VirtusLab/scala-cli/issues/610 to have a trace to find this, if needed.
Closing this, as we wouldn't push this to completion.
Can be used as a reference point while closed.

Issue #2751, opened 2024-02-22
doc command: all links in Scaladoc are hardcoded with "https://scala-lang.org/api/3.x/" as the site root


bishabosha:
when I use doc on my projects, such as enum-extensions [docs here](https://javadoc.io/doc/io.github.bishabosha/enum-extensions_3/latest/enumextensions/EnumMirror.html)

it seems literally all links to are prefixed with `https://scala-lang.org/api/3.x/`, even when not in `scala` package

e.g. 
![Screenshot 2024-02-22 at 10 35 11](https://github.com/VirtusLab/scala-cli/assets/13436592/32434451-5eae-4ad6-9526-b94e062083cc)

Maybe this default setting is to blame?
https://github.com/VirtusLab/scala-cli/blob/310895426e322add4ab2cbad2968e8aa1cc814a7/modules/cli/src/main/scala/scala/cli/commands/doc/Doc.scala#L128

bishabosha:
looking at the source code for scaladoc, it seems at first glance this regex examines any part of the path where the tasty file came from, naturally this path will include `.scala-build`, so could this be the reason?

but in general this scheme seems very prone to error, like the user must never have some path with a `scala` fragment anywhere, so no `scala-projects-workspace/foo/.scala-build`, or `projects/scala/foo/.scala-build` etc.

PR #3041, opened 2024-07-19
Fix default scaladoc config, so that id doesn't break all scaladoc links

KacperFKorban:
Right now running the doc command with default arguments in scala-cli always produced broken scaladoc links, because the default regex for external mappings mapped every tasty file that contained 'scala' to use the `scala-lang.org` mapping. And since scala-cli uses '.scala-build' directory for placing tasty files, almost all links were broken.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -116,7 +116,8 @@ object `scala-cli-bsp` extends JavaModule with ScalaCliPublishModule {
 object integration extends CliIntegration {
   object test extends IntegrationScalaTests {
     def ivyDeps = super.ivyDeps() ++ Seq(
-      Deps.jgit
+      Deps.jgit,
+      Deps.jsoup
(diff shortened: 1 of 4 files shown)


Gedochao:
Tied issues:
- https://github.com/scala/scala3/issues/20541
- https://github.com/VirtusLab/scala-cli/issues/2751

Gedochao:
@KacperFKorban can you include an integration test for this?

Gedochao:
cc @Florian3k

Gedochao:
Seems like it's failing on checks only.
Should be enought o re-run the `scalafix` task.
```bash
./mill -i '{__[],_,integration._}.fix'
```

Gedochao (approved):
LGTM

PR #3042, opened 2024-07-22
Add a global config key for `--offline` mode

### LINKED ISSUE #3024: Add global config for `--offline`
### Discussed in https://github.com/VirtusLab/scala-cli/discussions/3023

<sup>Originally posted by **halloleo** July 12, 2024</sup>
I need to run scala-cli as

    scala-cli --power --offline

because my machine is not on the internet.

The option `--power` I can add with

    $ scala-cli config power true

as a default option which is great.

I tried the same for `--offline`:

    $ scala-cli config offline true

but here I get "Error: unrecognized key offline".

How can I add  `--offline` to the default config?

Also, is there a way to suppress the informational output:

> The `--offline` option is experimental
</div>

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/3024

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/bloop/BloopExit.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/bloop/BloopExit.scala
@@ -18,7 +18,7 @@ object BloopExit extends ScalaCommand[BloopExitOptions] {
     import opts.*
     compilationServer.bloopRifleConfig(
       global.logging.logger,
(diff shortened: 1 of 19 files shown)


Gedochao:
Note: we likely need a follow-up to merge the `--offline` mode of build commands with the `--offline` flag for the `version` sub-command.
A launcher option would also make sense.

The scope of this PR is only to make the config affect build commands' (run, compile, etc.) `--offline` mode.

Gedochao:
Superseded by https://github.com/VirtusLab/scala-cli/pull/3216

PR #3043, opened 2024-07-22
Fix the REPL crashing when a dependency's classpath is called by a macro

Gedochao:
Scala CLI's side of https://github.com/scala/scala3/issues/21229

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/ReplTests3StableDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/ReplTests3StableDefinitions.scala
@@ -1,5 +1,9 @@
 package scala.cli.integration
 
+import com.eed3si9n.expecty.Expecty.expect
+
+import scala.util.Properties
+
 trait ReplTests3StableDefinitions { _: ReplTestDefinitions =>
   if (!actualScalaVersion.equals(actualMaxAmmoniteScalaVersion)) {
     lazy val defaultScalaVersionString =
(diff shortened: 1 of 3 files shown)


Gedochao:
kudos to @jchyb for finding the root cause

modules/integration/src/test/scala/scala/cli/integration/ReplTests3StableDefinitions.scala
```diff
@@ -16,4 +20,31 @@ trait ReplTests3StableDefinitions { _: ReplTestDefinitions =>
...
+        .map {
+          if (Properties.isWin)
+            a => if (a.contains(" ")) "\"" + a.replace("\"", "\\\"") + "\"" else a
+          else
+            identity
+        }
+        .flatMap(arg => Seq("--ammonite-arg", arg))
+      // FIXME: test this on standard Scala 3 REPL, rather than just Ammonite
+      val res = os.proc(TestUtil.cli, "repl", ".", "--power", "--amm", ammArgs, extraOptions)
+        .call(cwd = root)
```
Gedochao, line 46: Note: this is currently a test preventing accidentally breaking ammonite with this (which worked before the change), rather than an actual anti-regression test.
The Scala 3 REPL currently doesn't provide API that'd make this reasonably testable.

Gedochao: Following this up needs:
- https://github.com/scala/scala3/issues/21242

tgodzik (approved):

kasiaMarek (approved):

Issue #2936, opened 2024-05-24
Migrate from the custom M1 runner to the official Github one


Gedochao:
**Version(s)**
v1.3.2

**Describe what needs to be done and why**
We are currently using our own custom MacOS M1 runner for `aarch64` jobs. We could and should migrate to the official M1 runner provided by GitHub. This would also enable us to run those jobs outside of the `main` branch, even on external contributor PRs.

**Requirements**
- [x] migrate `scala-cli` jobs to `macos-14`
- [x] implement a script verifying the architecture for MacOS builds (as there is no separate MacOS runner nametag for the new architecture, just a version descriptor)
- [x] migrate [`coursier-m1`](https://github.com/VirtusLab/coursier-m1) jobs to `macos-14`
- [x] migrate https://github.com/VirtusLab/scala-cli-setup jobs to `macos-14`

**Describe alternatives you've considered**
We could stay on the custom runner we are using, but this should make it easier to maintain long-term (as someone would be doing a big chunk of the maintenance for us from then on)

**Additional context**
- https://github.com/actions/runner-images/issues/9254
- https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

mkurz:
Maybe relevant?
- https://github.blog/2024-06-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/

PR #3050, opened 2024-07-25
Switch to GitHub M1/`aarch64` runners on the CI

Gedochao:
https://github.com/VirtusLab/scala-cli/issues/2936

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -380,7 +380,9 @@ jobs:
           submodules: true
       - uses: VirtusLab/scala-cli-setup@v1
         with:
-          apps: ""
+          jvm: "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.2.0/graalvm-ce-java17-darwin-aarch64-22.2.0.tar.gz"
+      - name: Ensure it's running on aarch64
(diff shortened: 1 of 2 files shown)


Gedochao:
Seems like we can't switch building the Linux `arm64` launchers to a github launcher for now, as the MacOS `aarch64` runners don't have access to docker.
https://github.com/actions/runner-images/issues/9254#issuecomment-1917916016

The hope in that matter is twofold:
- this might be solved in a future `macos-15` or higher runner, if the underlying hardware was M3 or newer (https://github.com/actions/runner-images/issues/9254#issuecomment-2169635361)
  - https://developer.apple.com/documentation/virtualization/vzgenericplatformconfiguration/4360553-isnestedvirtualizationsupported?changes=latest_minor
- the more likely solution is when Linux `arm64` runners get out of public beta and become available for open source projects (which is uncertain on the if and when, but they're currently in public beta: https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/)

Still, it's a clear improvement to be able to run MacOS `aarch64` tests on every PR, rather than just on `main`, so that part still seems worth doing here.

tgodzik (approved):

PR #3055, opened 2024-07-29
Add environment variable help under `--envs-help` & refactor environment variable usage

Gedochao:
Changes are as follows:
- refactors the code to have all environment variables we use listed as constants in one spot (which makes it easier to track)
  - this doesn't include all of the envs used by our dependencies, like Bloop or Coursier, so it's not meant to be truly exhaustive from the user's perspectvie;
- limits environment variables saved for `bsp` in `setup-ide` to only those that are necessary;
- adds environment variable dedicated help under `--envs-help`, which lists all of the envs we use with short descriptions:
```bash
scala-cli --env-help --power
# The following is the list of environment variables used and recognized by Scala CLI.
# It should by no means be treated as an exhaustive list
# Some tools and libraries Scala CLI integrates with may have their own, which may or may not be listed here.
# 
# Scala CLI
#   SCALA_CLI_CONFIG              Scala CLI configuration file path
#   SCALA_CLI_HOME                Scala CLI home directory
#   SCALA_CLI_INTERACTIVE         Interactive mode toggle
#   SCALA_CLI_INTERACTIVE_INPUTS  Interactive mode inputs
#   SCALA_CLI_POWER               Power mode toggle
#   SCALA_CLI_PRINT_STACK_TRACES  Print stack traces toggle
#   SCALA_CLI_SODIUM_JNI_ALLOW    Allow to load libsodiumjni
#   SCALA_CLI_VENDORED_ZIS        Toggle io.github.scala_cli.zip.ZipInputStream
# 
# Java
#   JAVA_HOME                     Java installation directory
#   JAVA_OPTS                     Java options
#   JDK_JAVA_OPTIONS              JDK Java options
# 
# Coursier
#   COURSIER_CACHE                Coursier cache location
#   COURSIER_MODE                 Coursier mode (can be set to 'offline')
# 
# Spark
#   SPARK_HOME                    (power) Spark installation directory
# 
# Miscellaneous
#   PATH                          The app path variable
#   DYLD_LIBRARY_PATH             Runtime library paths on Mac OS X
#   LD_LIBRARY_PATH               Runtime library paths on Linux
#   PATHEXT                       Executable file extensions on Windows
#   SHELL                         The currently used shell
#   VCVARSALL                     Visual C++ Redistributable Runtimes
#   ZDOTDIR                       Zsh configuration directory
# 
# Internal
#   CI                            (power) Marker for running on the CI
```


modules/core/src/main/scala/scala/build/internals/EnvsUtil.scala
```diff
@@ -0,0 +1,135 @@
...
+  ) {
+    def valueOpt: Option[String]  = Option(System.getenv(name))
+    override def toString: String = s"$name=${valueOpt.getOrElse("")}"
+    def helpMessage(spaces: String): String = {
+      val powerString =
+        if requiresPower then s"${ScalaCliConsole.GRAY}(power)${Console.RESET} " else ""
+      s"${Console.YELLOW}$name${Console.RESET}$spaces$powerString$description"
+    }
+  }
+  object EnvVar {
```
tgodzik: Do we need to keep it in `EnvUtil` ? It causes each access to have `EnvUtil.EnvVar` which is a bit redundant

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
```diff
@@ -595,8 +596,8 @@ object BuildOptions {
           currentEnv.keys.find(_.equalsIgnoreCase(name)).getOrElse(name)
         else
           name
-      val javaHomeKey = keyFor("JAVA_HOME")
-      val pathKey     = keyFor("PATH")
+      val javaHomeKey = keyFor(EnvsUtil.EnvVar.Java.javaHome.name)
```
tgodzik:
```suggestion
      val javaHomeKey = keyFor(EnvVars.Java.javaHome.name)
```
Would be much cleaner I feel

modules/core/src/main/scala/scala/build/internals/EnvsUtil.scala
Gedochao: moved everything to the `EnvVar` companion object.

modules/options/src/main/scala/scala/build/options/BuildOptions.scala
Gedochao: should be better now.

tgodzik (approved):

PR #3075, opened 2024-08-07
Add more env vars & generate reference docs for them

Gedochao:
A follow-up to https://github.com/VirtusLab/scala-cli/pull/3055

Adds more environment variables constants (tied to Bloop and Coursier implementations, specifically).
Additionally, adds reference docs generation for environment variables defined in the code.

image

### DIFF
--- a/modules/core/src/main/scala/scala/build/internals/EnvVar.scala
+++ b/modules/core/src/main/scala/scala/build/internals/EnvVar.scala
@@ -57,7 +57,7 @@ object EnvVar {
         }""".stripMargin
     }
   }
-  def allGroups: Seq[EnvVarGroup] = Seq(ScalaCli, Java, Coursier, Spark, Misc, Internal)
+  def allGroups: Seq[EnvVarGroup] = Seq(ScalaCli, Java, Bloop, Coursier, Spark, Misc, Internal)
   def all: Seq[EnvVar]            = allGroups.flatMap(_.all)
   def allBsp: Seq[EnvVar]         = all.filter(_.passToIde)
   object Java extends EnvVarGroup {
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/scala/build/internals/EnvVar.scala
```diff
@@ -124,6 +153,40 @@ object EnvVar {
...
+      EnvVar("BLOOP_EXPORT_JAR_CLASSIFIERS", "Export jar classifiers", requiresPower = true)
+    val bloopJavaOpts = EnvVar("BLOOP_JAVA_OPTS", "Bloop Java options", requiresPower = true)
+    val bloopModule   = EnvVar("BLOOP_MODULE", "Bloop default module", requiresPower = true)
+    val bloopPort     = EnvVar("BLOOP_PORT", "Bloop default port", requiresPower = true)
+    val bloopScalaVersion =
+      EnvVar("BLOOP_SCALA_VERSION", "Bloop default Scala version", requiresPower = true)
+    val bloopVersion      = EnvVar("BLOOP_VERSION", "Bloop default version", requiresPower = true)
+    val bloopServer       = EnvVar("BLOOP_SERVER", "Bloop default host", requiresPower = true)
+    val bloopExtraTimeout = EnvVar("SCALA_CLI_EXTRA_TIMEOUT", "Extra timeout", requiresPower = true)
+    val bloopTravis       = EnvVar("TRAVIS", "Travis CI toggle", requiresPower = true)
```
tgodzik: I don't think this should be promoted. Pretty sure it's only used for Bloop tests (and not for a while) and I would just remove it

modules/core/src/main/scala/scala/build/internals/EnvVar.scala
```diff
@@ -124,6 +153,40 @@ object EnvVar {
...
+      bloopTravis
+    )
+    val bloopComputationCores = EnvVar(
+      "BLOOP_COMPUTATION_CORES",
+      "Number of computation cores to be used",
+      requiresPower = true
+    )
+    val bloopDaemonDir = EnvVar("BLOOP_DAEMON_DIR", "Bloop daemon directory", requiresPower = true)
+    val bloopExportJarClassifiers =
+      EnvVar("BLOOP_EXPORT_JAR_CLASSIFIERS", "Export jar classifiers", requiresPower = true)
```
tgodzik: That's for sbt only, not sure how useful here

tgodzik (changes requested):

modules/core/src/main/scala/scala/build/internals/EnvVar.scala
Gedochao: removed

modules/core/src/main/scala/scala/build/internals/EnvVar.scala
Gedochao: removed this, too

tgodzik (approved):

PR #3079, opened 2024-08-08
improvement: allow to run main class from deps with no inputs

### LINKED ISSUE #2840: Allow `run` with no sources / replacing `cs launch`

scala-cli is almost capable of replacing `cs launch` from Coursier: the only thing that appears to be missing, in order to be able to run an arbitrary application from dependencies, is that `scala-cli run` requires at least one source file / directory.

For example:

```
scala-cli run --dep software.amazon.smithy:smithy-cli:1.45.0 . --main-class software.amazon.smithy.cli.SmithyCli
```

If dependencies are provided, relax the requirement of having to pass a file/directory.

In addition, there are a couple nice-to-haves:

- discover main classes from dependencies, maybe with a flag?
- if no sources are passed, don't start a compilation server (assume `--server=false`?)

- Using `cs launch` as usual
- Using a dummy empty directory as the sources

**Additional context**

This would be nice to enable the usage of code generators, such as [smithy4s](https://github.com/disneystreaming/smithy4s/), without waiting for #610. For example, at the moment:

```
scala-cli run --scala 2.13 --dep com.disneystreaming.smithy4s::smithy4s-codegen-cli:0.18.15 . --main-class smithy4s.codegen.cli.Main --server=false -- version
```

which isn't terrible, but still requires that dummy directory (`.`).

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/2840

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
@@ -113,13 +113,22 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
   }
 
   def runCommand(
-    options: RunOptions,
+    options0: RunOptions,
     inputArgs: Seq[String],
     programArgs: Seq[String],
     defaultInputs: () => Option[Inputs],
     logger: Logger,
     invokeData: ScalaCliInvokeData
   ): Unit = {
+    val shouldDefaultServerFalse =
+      inputArgs.isEmpty && options0.shared.compilationServer.server.isEmpty &&
+      !options0.shared.hasSnippets
+    val options = if (shouldDefaultServerFalse) options0.copy(shared =
+      options0.shared.copy(compilationServer =
+        options0.shared.compilationServer.copy(server = Some(false))
+      )
+    )
+    else options0
     val initialBuildOptions = {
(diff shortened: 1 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
```diff
@@ -113,13 +113,22 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
   }
 
   def runCommand(
-    options: RunOptions,
+    options0: RunOptions,
```
Gedochao, line 116: why the rename?

Gedochao (approved):
LGTM

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
kasiaMarek: Because I'm defining `options` in line `126`, where server may be set to false. It felt like to easiest way to rename it like this. Would you rather it was named differently?

Gedochao: ah, I see, missed that.
Nah, it's fine. I'm not a fan of this convention, but we do use it in the codebase in many spots.

PR #3091, opened 2024-08-12
Tweak benchmarking with JMH

Gedochao:
- updates `jmh` to 1.37 (was 1.29)
- fixes IDE import for benchmarking projects (BSP reload is still broken, but the initial setup works)
- lifts `--jmh` and `--jmh-version` to `SharedOptions`, making it available for more sub-commands (used to be just `run`)
  - `compile` will let benchmarking projects compile
  - `package` will now also work with `--jmh`, but it will mostly ignore it past compilation (the resulting artifact won't be running benchmarks, which is something to be fixed separately)
  - `doc` should work with benchmarking projects
  - `bsp` and `setup-ide` work correctly, but `bsp` fails to reload with `--jmh` (better than it was, but still needs fixing)
  - `export` will include the `jmh-core` dependency as compile-only and not fail, which is better than outright failing, but still needs improvements


tgodzik (approved):

modules/integration/src/test/scala/scala/cli/integration/JmhTests.scala
```diff
@@ -2,41 +2,105 @@ package scala.cli.integration
...
+  test("setup-ide") {
+    // TODO fix setting jmh via a reload & add tests for it
+    inputs.fromRoot { root =>
+      os.proc(TestUtil.cli, "setup-ide", "--power", TestUtil.extraOptions, ".", "--jmh")
+        .call(cwd = root)
+    }
+  }
+
+  test("package") {
+    // TODO make package with --jmh build an artifact that actually runs benchmarks
```
tgodzik, line 70: Should this be an issue?

tgodzik (approved):

modules/integration/src/test/scala/scala/cli/integration/JmhTests.scala
Gedochao: Yes, I'll create one.

PR #3104, opened 2024-08-14
[ISSUE-296] Add support for fish autocompletions

### LINKED ISSUE #296: Fish shell completion support
from @alexarchambault 

For now, we only support bash and zsh completion. I'm not sure what is needed to have fish completion work.

Implementing this should mostly involve changes in [case-app](https://github.com/alexarchambault/case-app), where the core of bash and zsh completion support lives.

KristianLentino99:
Hi all, I wanted to contribute a little bit to this fantastic project so I've taken a look at your "Good first Issues" and I've decided to pick up [this one](https://github.com/VirtusLab/scala-cli/issues/296). 
- I've updated the related test and ran all the steps mentioned in the CONTRIBUTING guidelines. 
- I've also updated the Advanced installation doc adding the tab for the new `fish` CLI. 

I've also ran `scala-cli` with my changes locally and installed it from scratch  to test if it works:
image

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/installcompletions/InstallCompletions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/installcompletions/InstallCompletions.scala
@@ -1,7 +1,7 @@
 package scala.cli.commands.installcompletions
 
 import caseapp.*
-import caseapp.core.complete.{Bash, Zsh}
+import caseapp.core.complete.{Bash, Fish, Zsh}
 import caseapp.core.help.HelpFormat
 
 import java.io.File
@@ -47,6 +47,7 @@ object InstallCompletions extends ScalaCommand[InstallCompletionsOptions] {
         )
         System.err.println(s"$name install completions --shell zsh")
         System.err.println(s"$name install completions --shell bash")
+        System.err.println(s"$name install completions --shell fish")
         sys.exit(1)
       }
     }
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
Thanks! I tested it out and look to work except when using completion without any input. That seems to be case app issue though and can be further improved.

tgodzik:
Actually works fine, I had leftover code from a previous attempt. I wonder if we could get also next subcommands, it doesn't seem implemented in case app (or scala-cli?)

tgodzik:
Ad 3. Looks like some subcommands work, while other do not (bloop about is not suggested). Anyway looks like we can improve things further, but this already works awesome!

Gedochao:
> Ad 3. Looks like some subcommands work, while other do not (bloop about is not suggested). Anyway looks like we can improve things further, but this already works awesome!

@tgodzik Can you extract any issues you've found into separate issues?

Gedochao (approved):
Looks good, thanks for the contribution!

tgodzik:
> > Ad 3. Looks like some subcommands work, while other do not (bloop about is not suggested). Anyway looks like we can improve things further, but this already works awesome!
> 
> @tgodzik Can you extract any issues you've found into separate issues?

Funny thing, those commands do not exist in Scala CLI codebase, but they work since they are just forwarded to Bloop, which understands more commands. The question is whether they should be showed by Scala CLI?

This is things like `scala bloop projects` or even `scala bloop compile` both of which will actually not work unless run in scala-build directory :sweat_smile: 

The only one I think we could add is `about`

PR #3105, opened 2024-08-15
bugfix: Exclude both sourcecode and ccollection-compat correctly

tgodzik:
Checked locally this time and it works.


TonioGela:
I've checked the tl-toolkit tests and I can confirm that this solution's working

Gedochao (approved):

PR #3107, opened 2024-08-15
Remove adding test options to the project/build target name hash

MaciejG604:
The hash in the build target name is used to introduce caching when buildOptions are changed from the CLI. Otherwise, if options are changed from sources, Bloop recompiles the project anyway because the sources changed. Adding this new hash that comes from sources possibly creates new build targets every time somebody fiddles with some test scope options. It's  just nonsense.

Also look at the comments in the code, we've made a rule and then stopped following it.

I understand those are changes with potentially weird consequences - let's see if CI is green

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -200,22 +200,15 @@ object Build {
     */
   def updateInputs(
     inputs: Inputs,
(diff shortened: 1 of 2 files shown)


Gedochao:
The CI is green... although maybe adding some tests could increase our confidence about this 🤔

Gedochao (approved):
LGTM

PR #3112, opened 2024-08-17
Fix/changing options from sources should not require reload

MaciejG604:
Right now from Bsp we pass to BspImpl build options that are the options from CLI and sources combined, this behaviour is incorrect since those options get used as options from CLI (so override all else) in `BspImpl.compile()` and `BspImpl.build()`. 

As a result we have incorrect project configuration on every change of options from sources after launching the BSP command! From what I remeber this is covered up by Metals since on every change of project configuration (more specific on a change of bloop config json file) we send `buildTarget/didChange` and receive from Metals a `workspace/reload` request. This allows us to fix everything since the options used in `BspImpl.compile() `and `BspImpl.build()` get reloaded to correct values that represent options passed from CLI. (See `initialBspOptions` and `bspReloadableOptionsReference` in Bsp)

No idea if IntelliJ does the same on our `buildTarget/didChange`, but if not this might be a major problem for IntelliJ users.

Also due to this error our tests needed some fixing. The test with correct bloop JVM version is a weird case since in old behaviour we never send `java-19` path to bloop before the workspace reload. So we were sending `java-11` and the `--release 19` resulted in an error during compilation. New behaviour is we send `java-19` path and the compilation succeeds, but we get some error logs from bloop from whetever it's doing with the compiler jar 🤷. We might need to fix this so that Scala CLI fails the compilation when the log is received from bloop, but I'm not sure if this is doable.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -786,6 +786,10 @@ abstract class BspTestDefinitions extends ScalaCliSuite with TestScalaVersionArg
           |""".stripMargin
     )
 
+    val actualScalaMajorVersion = actualScalaVersion.split("\\.")
+      .take(if (actualScalaVersion.startsWith("3")) 1 else 2)
+      .mkString(".")
+
     withBsp(inputs, Seq(".")) { (root, localClient, remoteServer) =>
       async {
         val buildTargetsResp = await(remoteServer.workspaceBuildTargets().asScala)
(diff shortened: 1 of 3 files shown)


modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -851,6 +851,41 @@ abstract class BspTestDefinitions extends ScalaCliSuite with TestScalaVersionArg
           val resp = await(remoteServer.buildTargetCompile(new b.CompileParams(targets)).asScala)
           expect(resp.getStatusCode == b.StatusCode.OK)
         }
+
+        expect(changeFuture.isCompleted)
```
Gedochao, line 849: I don't think we're awaiting this future anywhere, shouldn't we? otherwise there could be a race here.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -851,6 +851,41 @@ abstract class BspTestDefinitions extends ScalaCliSuite with TestScalaVersionArg
...
+            expect(!foundDepSources.exists(_.startsWith("os-lib_2.13-0.7.8")))
+          }
+          else if (actualScalaVersion.startsWith("2.12")) {
+            expect(foundDepSources.exists(_.startsWith("utest_2.12-0.7.10")))
+            expect(!foundDepSources.exists(_.startsWith("os-lib_2.12-0.7.8")))
+          }
+          else {
+            expect(foundDepSources.exists(_.startsWith("utest_3-0.7.10")))
+            expect(!foundDepSources.exists(_.startsWith("os-lib_3-0.7.8")))
+          }
```
Gedochao:
```suggestion
val actualScalaMajorVersion = actualScalaVersion.split(".").head
expect(foundDepSources.exists(_.startsWith(s"utest_$actualScalaMajorVersion-0.7.10")))
expect(!foundDepSources.exists(_.startsWith(s"os-lib_$actualScalaMajorVersion-0.7.8")))
```

Gedochao:
> New behaviour is we send java-19 path and the compilation succeeds, but we get some error logs from bloop from whetever it's doing with the compiler jar 🤷. We might need to fix this so that Scala CLI fails the compilation when the log is received from bloop, but I'm not sure if this is doable.

@MaciejG604 can you extract this to an issue?

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
MaciejG604: This promise is fulfilled upon receiving `DidChangeBuildTarget`, this message should arrive when we trigger compilation with `remoteServer.buildTargetCompile`. See `BspImpl.scala` line 296 - `buildChanged` flag is set in the result of `prepareBuild()` if bloop config json file is changed for a build target.

Gedochao (approved):
LGTM

PR #3118, opened 2024-08-20
Add directives for JMH

Gedochao:
This adds the directive counterparts for the `--jmh` and `--jmh-version` CLI options.
```scala
//> using jmh
//> using jmhVersion
```

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/JmhTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/JmhTests.scala
@@ -1,13 +1,18 @@
 package scala.cli.integration
 
+import ch.epfl.scala.bsp4j as b
 import com.eed3si9n.expecty.Expecty.expect
 
 import java.nio.file.Files
 
+import scala.async.Async.{async, await}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.jdk.CollectionConverters.*
 import scala.util.Properties
 
-class JmhTests extends ScalaCliSuite with JmhSuite {
-  override def group: ScalaCliSuite.TestGroup = ScalaCliSuite.TestGroup.First
+class JmhTests extends ScalaCliSuite with JmhSuite with BspSuite {
+  override def group: ScalaCliSuite.TestGroup      = ScalaCliSuite.TestGroup.First
+  override protected val extraOptions: Seq[String] = TestUtil.extraOptions
 
   for {
     useDirective <- Seq(None, Some("//> using jmh"))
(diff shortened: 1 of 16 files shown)


modules/options/src/main/scala/scala/build/options/JmhOptions.scala
```diff
@@ -1,9 +1,28 @@
...
+  * @param enableJmh
+  *   toggle for enabling JMH dependency handling in the build (overrides [[runJmh]] when disabled)
+  * @param runJmh
+  *   toggle for whether JMH should actually be runnable from this build
+  */
 final case class JmhOptions(
-  addJmhDependencies: Option[String] = None,
+  jmhVersion: Option[String] = None,
+  enableJmh: Option[Boolean] = None,
   runJmh: Option[Boolean] = None
```
Gedochao, line 16: this is rather ugly, but we do need something temporary to track if JMH is enabled, while the `runJmh` property is toggled on and off in builds.
We should be able to simplify this when running JMH will be extracted out of the `run` sub-command (or out of the `main` scope).

modules/integration/src/test/scala/scala/cli/integration/JmhTests.scala
```diff
@@ -1,106 +1,193 @@
...
+  for {
+    useDirective <- Seq(None, Some("//> using jmh"))
+    directiveString = useDirective.getOrElse("")
+    jmhOptions      = if (useDirective.isEmpty) Seq("--jmh") else Nil
+    testMessage = useDirective match {
+      case None            => jmhOptions.mkString(" ")
+      case Some(directive) => directive
+    }
+  } {
+    test(s"run ($testMessage)") {
```
kasiaMarek, line 26: Maybe we should add tests for having both jmh and main class? I feel like if you have a main class and jhm tests it should run main by default but if you add `--jmh` explicitly it should rather run benchmarking tests.

Gedochao: Yeah, so right now `--jmh` will run benchmarking tests.
I plan a future follow-up, where `--jmh` would either be extracted to a separate scope or a dedicated sub-command, but in the scope of this PR, `--jmh` forces benchmarking to be run, rather than the project's main class.
There's a bunch of TODOs in the `JmhTests.scala` source file that I plan to transform into issues as a follow-up.

kasiaMarek: Okay, adding `//> using jmh`  will cause that too, but if you explicitly specify `--main-class` it will run the correct the chosen main so I think it is okay for now.

modules/options/src/main/scala/scala/build/options/JmhOptions.scala
```diff
@@ -1,9 +1,28 @@
 package scala.build.options
 
+import scala.build.internal.Constants
+
+/** @param jmhVersion
+  *   the version of JMH to be used in the build
+  * @param enableJmh
+  *   toggle for enabling JMH dependency handling in the build (overrides [[runJmh]] when disabled)
+  * @param runJmh
+  *   toggle for whether JMH should actually be runnable from this build
```
kasiaMarek: Could you explain a bit more in the comment what this flag does?

kasiaMarek (approved):

modules/options/src/main/scala/scala/build/options/JmhOptions.scala
Gedochao: done

PR #3119, opened 2024-08-20
fix: try removing release flag

### LINKED ISSUE #3115: scala-cli eagerly sets `-release` / `-java-output-version`, which prevents compiling on new Java versions
`scala-cli compile` always sets `-release N` (aka `-java-output-version` on Scala 3) where N is the Java version.

This prevents compiling on a Java version which is not yet known to the compiler:

```
➜ sandbox cat A.scala
class A { java.io.IO.print(1) }

➜ sandbox scala compile --server=false -S 3 A.scala
23 is not a valid choice for -java-output-version
  scalac -help  gives more information
Compilation failed
```

I can manually set an older `-release` version, but then new API is not available:

```
➜ sandbox scala compile --server=false -S 3 A.scala -O -release -O 22
-- [E008] Not Found Error: /Users/luc/code/scala/scala13/sandbox/A.scala:1:18 --
1 |class A { java.io.IO.print(1) }
  |          ^^^^^^^^^^
  |          value IO is not a member of java.io
1 error found
Compilation failed
```

Invoking the compiler manually works, the compiler writes the classfile using the default version 52 (corresponds to Java 8):

```
➜ sandbox java -cp $(cs fetch -p org.scala-lang::scala3-compiler:3.5.0) dotty.tools.dotc.Main -usejavacp A.scala
➜ sandbox asm A.class && cat A.asm
// class version 52.0 (52)
// access flags 0x21
public class A {

  // compiled from: A.scala
...
```

@Gedochao mentioned that the flag might have been added due to some interaction with bloop (cc @kasiaMarek, @tgodzik)

kasiaMarek:
resolves: https://github.com/VirtusLab/scala-cli/issues/3115

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -862,10 +862,6 @@ object Build {
     val semanticDbSourceRoot =
       options.scalaOptions.semanticDbOptions.semanticDbSourceRoot.getOrElse(inputs.workspace)
 
-    val releaseFlagVersion =
-      if (options.useBuildServer.getOrElse(true)) None
-      else releaseFlag(options, compilerJvmVersionOpt, logger).map(_.toString)
-
     val scalaCompilerParamsOpt = artifacts.scalaOpt match {
       case Some(scalaArtifacts) =>
         val params = value(options.scalaParams).getOrElse {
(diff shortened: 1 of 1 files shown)


Gedochao (approved):
I wonder at what point did we stop needing it. 🤔 
Either way, LGTM

kasiaMarek:
> I wonder at what point did we stop needing it.

Possibly here: https://github.com/VirtusLab/scala-cli/pull/2985. I just deleted it only for bloop not knowing that originally it was added for bloop (I think).

PR #3122, opened 2024-08-22
Add release notes for Scala CLI v1.5.0

Gedochao:
- https://github.com/VirtusLab/scala-cli/issues/3090

### DIFF
--- a/website/docs/release_notes.md
+++ b/website/docs/release_notes.md
@@ -8,6 +8,188 @@ import ReactPlayer from 'react-player'
 
 # Release notes
 
+## [v1.5.0](https://github.com/VirtusLab/scala-cli/releases/tag/v1.5.0)
+
+### Support for Scala 3.5.0
+This Scala CLI version switches the default Scala version to 3.5.0.
+
+```bash
+scala-cli version
+# Scala CLI version: 1.5.0
+# Scala version (default): 3.5.0
+```
+
+Added by [@Gedochao](https://github.com/Gedochao) in [#3093](https://github.com/VirtusLab/scala-cli/pull/3093).
+
+### Support for Scala Native 0.5.5
+This Scala CLI version switches the default Scala Native version to 0.5.5.
+
+```bash
+scala-cli -e 'println("Hello from Scala Native 0.5.5!")' --native
+# Compiling project (Scala 3.5.0, Scala Native 0.5.5)
+# Compiled project (Scala 3.5.0, Scala Native 0.5.5)
+# [info] Linking (multithreadingEnabled=true, disable if not used) (894 ms)
+# [info] Discovered 888 classes and 5407 methods after classloading
+# [info] Checking intermediate code (quick) (31 ms)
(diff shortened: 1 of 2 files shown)


website/docs/release_notes.md
```diff
@@ -8,6 +8,187 @@ import ReactPlayer from 'react-player'
...
+  def foo(): Unit = {
+    (1L to 1000L).sum
+  }
+}
+```
+
+Expect more improvements in this area in the future.
+Also, do play with it and give us feedback about the current implementation!
+
+Added by [@Gedochao](https://github.com/Gedochao) in [#3091](https://github.com/VirtusLab/scala-cli/pull/3091) and [#3118]( and https://github.com/VirtusLab/scala-cli/pull/3118).
```
kasiaMarek:
```suggestion
Added by [@Gedochao](https://github.com/Gedochao) in [#3091](https://github.com/VirtusLab/scala-cli/pull/3091) and [#3118](https://github.com/VirtusLab/scala-cli/pull/3118).
```

website/docs/release_notes.md
```diff
@@ -8,6 +8,187 @@ import ReactPlayer from 'react-player'
...
+```bash
+scala-cli run --dep io.get-coursier:coursier-cli_2.13:2.1.10 -- version
+# 2.1.10
+```
+
+This can be used similarly to [Coursier's `cs launch`](https://get-coursier.io/docs/cli-launch).
+
+Added by [@kasiaMarek](https://github.com/kasiaMarek) in [#3079](https://github.com/VirtusLab/scala-cli/pull/3079).
+
+### (⚡️ experimental) JMH available in various commands and via `using` directives
```
kasiaMarek, line 93: I would be very explicit here that `//> using jmh` or `--jmh` will cause `scala-cli run` to run benchmarks even when there exists a `main` class (I don't think that is very intuitive).

website/docs/release_notes.md
```diff
@@ -8,6 +8,187 @@ import ReactPlayer from 'react-player'
...
+Some improvements have been done to the experimental support for JMH (Java Microbenchmark Harness).
+
+The `--jmh` and `--jmh-version` options can now be passed to a number of commands:
+- `compile`, so that a Scala CLI project with benchmarking can be compiled separately from being run.
```
kasiaMarek:
```suggestion
- `compile`, so that a Scala CLI project with benchmarking can be compiled separately from being run;
```
NIT, but I don't think `.` makes much sense there.

kasiaMarek (approved):

website/docs/release_notes.md
Gedochao: I agree this is not very intuitive, although this behaviour has been with us long before this release. 
Still, let me add some more explanation, sure.

Gedochao: added some extra explanation under commands where `--jmh` is available, it now lists `run` there.

website/docs/release_notes.md
Gedochao: The regex caught this wrong... Making a mental note to fix the automation later.

PR #3126, opened 2024-08-23
Support multiple separate build targets

MaciejG604:
Had some time to fix a big problem in the way of #3085 - #3112.

After that is gone, moving from the PoC to its basic, still exciting version was very tempting. And so here we are.

This PR introduces support for multiple build targets in the BSP mode. Build targets are not in any way related, their classpath and project configuration should be totally separate, except of course for options passed from CLI which will be applied to all.

Setting up of build targets can be done with a `.toml` file where we define all build targets together with their roots - which corresponds to how all Scala CLI projects start their preprocessing - args from CLI. The rest is just for looping over the build targets doing the same stuff we always do.

The toml looks like this:
```
[modules.core]

[modules.utils]
roots = ["Utils.scala", "Utils2.scala"]
```
Where no roots is defined the name is picked and so the root of module `core` will be `core/` in the project workspace.


Gedochao:
Closing this, linked it back at https://github.com/VirtusLab/scala-cli/issues/4170
Any future discussion should go back there.

PR #3129, opened 2024-08-26
Doc: Tips on how to list available JVMs using coursier

jatcwang:


website/docs/reference/cli-options.md
```diff
@@ -639,7 +639,7 @@ Set the Java home directory
 
 Aliases: `-j`
 
-Use a specific JVM, such as `14`, `adopt:11`, or `graalvm:21`, or `system`
+Use a specific JVM, such as `14`, `adopt:11`, or `graalvm:21`, or `system`. scala-cli uses **cousier** to fetch JVMs, so you can use `cs java --available` to list the available JVMs.
```
tgodzik:
```suggestion
Use a specific JVM, such as `14`, `adopt:11`, or `graalvm:21`, or `system`. scala-cli uses [coursier](https://get-coursier.io/) to fetch JVMs, so you can use `cs java --available` to list the available JVMs.
```

tgodzik (changes requested):
Thanks! It might useful to link to it and there is a small typo.

website/docs/reference/cli-options.md
jatcwang: Updated thanks!

tgodzik:
Could you run `./mill -i 'generate-reference-doc[]'.run` ? We have a special reference doc for SIP purposes and it needs to be up to date. Thanks!

tgodzik (approved):
Thanks!

Issue #2708, opened 2024-01-31
plugin option added twice


tgodzik:
**Version(s)**
1.1.2

**Describe the bug**
When using:
```
//project.scala
//> using plugin "org.virtuslab::besom-compiler-plugin:0.1.1-SNAPSHOT"
```
the plugin option is sometimes added twice, but it doesn't seem to reproduce always/

**Expected behaviour**
Plugin option is added once.

Gedochao:
may be related to https://github.com/VirtusLab/scala-cli/issues/2621, as I never really got it working stably with Bloop... 🤔 
@tgodzik is it reproducible with `--server=false`?

lbialy:
--server=false does not generate bloop's json files I'm afraid

Gedochao:
> --server=false does not generate bloop's json files I'm afraid

@lbialy but it does pass plugin options to the compiler

lbialy:
There never was any issue with compilation by scala-cli, the error crashes metals and metals only.

lbialy:
Passed you a repro with minimization via slack.

keynmol:
Tomasz suggests that this is what breaks Metals for me, diagnostics not being delivered.

```scala
//> using scala "3.3.3"
//> using options -Werror -Wunused:all -Wvalue-discard -Wnonunit-statement
//> using plugin "org.virtuslab::besom-compiler-plugin:0.2.2"
//> using dep "org.virtuslab::besom-core:0.2.2"
//> using dep "org.virtuslab::besom-aws:6.23.0-core.0.2"
//> using dep "org.virtuslab::besom-awsx:2.5.0-core.0.2"

import besom.*
import besom.api.aws, besom.api.awsx

@main def main = Pulumi.run {
  val bucket =
    // THIS IS AN ERROR        v  <---- over here
    aws.s3.Bucket("my-bucket", ws.s3.BucketArgs(forceDestroy = true))

  Stack.exports(
    bucketName = bucket.bucket
  )

}

```
![CleanShot 2024-03-28 at 09 25 06](https://github.com/VirtusLab/scala-cli/assets/1052965/90175895-536a-4396-aa11-57aa0804fb3c)

![CleanShot 2024-03-28 at 09 30 28@2x](https://github.com/VirtusLab/scala-cli/assets/1052965/ecbbcd0f-9858-44af-8e16-8d420961ef0e)

lbialy:
[...]

PR #3139, opened 2024-08-30
improvement: Use distinct on ScalacOpt

tgodzik:
I am still getting an issue with multiple same plugin options and my guess is that the distinct doesn't work, especially that it is hard to figure out what can actually be contained there. Moving distinct to the end should be a much safer option.

Related to https://github.com/VirtusLab/scala-cli/issues/2708

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -862,10 +862,10 @@ object Build {
           )
         }
 
-        val pluginScalacOptions = scalaArtifacts.compilerPlugins.distinct.map {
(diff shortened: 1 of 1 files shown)


Gedochao:
I mean, sure, but I'm not sure if this really changes much.
@tgodzik do we have a reproduction of the double plugin flag, so that we could add a test?

Gedochao (approved):
Optimally we'd like a test for this, but it can't hurt to try this.

lbialy:
I can test this immediately as besom compiler plugin breaks any scala-cli build, how can I download a nightly version of scala-cli?

tgodzik:
`scala-cli --cli-version nightly setup-ide .` should do it, but I don't think the nightly is up yet

PR #3152, opened 2024-09-06
Change how help is referenced to avoid initialization oddness & update `case-app` to 2.1.0-M29

coreyoconnor:
This includes the update for `case-app`.

This supersedes pull #2964

Draft because I don't really understand why this only effected some options and not others and why only with this version of case app. That said...

`scala-cli` overrides `val messages`. This override references `name` which is a `def` in `CaseApp` equal to `help.progName`. Which is equal to `messages.progName`. Which is the `val messages` being defined. I'm not entirely sure on that last bit but I suspect Scala is not entirely sure as well. :)

Breaking these recursive dependencies appears to work. That's what this patch tries to do.
 
I suspect there is an issue in how `CaseApp` is constructed. Or perhaps `def help` should be overriden instead of `val messages`.

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/InstallAndUninstallCompletionsTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/InstallAndUninstallCompletionsTests.scala
@@ -69,9 +69,7 @@ class InstallAndUninstallCompletionsTests extends ScalaCliSuite {
   lazy val fishRcScript: String = {
     val progName = "scala-cli"
     val script =
-      s"""
-    complete $progName -a '($progName complete fish-v1 (math 1 + (count (__fish_print_cmd_args))) (__fish_print_cmd_args))'
-         |""".stripMargin
+      s"""complete $progName -a '($progName complete fish-v1 (math 1 + (count (__fish_print_cmd_args))) (__fish_print_cmd_args))'"""
     addTags(script)
   }
 
--- a/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
@@ -34,9 +34,10 @@ import scala.cli.util.ConfigDbUtils.*
 import scala.cli.{CurrentParams, ScalaCli}
 import scala.util.{Properties, Try}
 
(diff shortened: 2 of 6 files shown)


Gedochao:
Did some fixes:
- formatting was off, ran `scalafmt`
- merged changes from `main` into this branch
- seems like your changes have fixed some of our unit tests, which weren't catching duplicate CLI options; fixed those

What's still breaking:
- it seems the completions don't work correctly, still need to investigate the failing test: `scala.cli.integration.InstallAndUninstallCompletionsTests.installing and uninstalling completions`

Gedochao:
Seems it was just whitespace that changed in the fish completions file generation, unsure what is causing this.
I adjusted the associated test.
I also cleaned up the commit structure, so that we won't have to squash when we merge this, the individual changes should be separated.

@tgodzik can you verify `fish` completions still work okay after this? an extra pair of eyes on the code won't hurt, either.

modules/integration/src/test/scala/scala/cli/integration/InstallAndUninstallCompletionsTests.scala
```diff
@@ -69,9 +69,7 @@ class InstallAndUninstallCompletionsTests extends ScalaCliSuite {
   lazy val fishRcScript: String = {
     val progName = "scala-cli"
     val script =
-      s"""
-    complete $progName -a '($progName complete fish-v1 (math 1 + (count (__fish_print_cmd_args))) (__fish_print_cmd_args))'
-         |""".stripMargin
+      s"""complete $progName -a '($progName complete fish-v1 (math 1 + (count (__fish_print_cmd_args))) (__fish_print_cmd_args))'"""
```
tgodzik, line 72: The whitespace is not significant, so this should be fine. I see that there is an extra newline actually generated previously, so I guess this should count as a fix? :sweat_smile:

Gedochao: Yeah, I thought the same.

Gedochao (approved):
@coreyoconnor I tweaked the PR to make the tests pass, hope you don't mind.
Thanks for helping with this!

coreyoconnor:
> @coreyoconnor I tweaked the PR to make the tests pass, hope you don't mind. Thanks for helping with this!

Woohoo! Thanks! Obviously I was away and only getting back to this now. XD Happy to help!

PR #3163, opened 2024-09-10
Add tests for `setup-ide` with `--cli-version`

Gedochao:
Follow-up to https://github.com/VirtusLab/scala-cli/pull/3161

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
+++ b/modules/cli/src/main/scala/scala/cli/ScalaCli.scala
@@ -252,7 +252,9 @@ object ScalaCli {
                   predefinedCliVersion = None,
                   initialLauncherPath = None
                 )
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/scala/scala/cli/ScalaCli.scala
```diff
@@ -252,7 +252,9 @@ object ScalaCli {
                   predefinedCliVersion = None,
                   initialLauncherPath = None
                 )
-              case v if v < Version("1.5.1") && !ver.contains("nightly") =>
+              case v
+                  if v < Version("1.5.0-34-g31a88e428-SNAPSHOT") && v < Version("1.5.1") &&
```
Gedochao, line 256: This is necessary for snapshot versions called explicitly (rather than through `nightly`) to be handled correctly.

tgodzik (approved):

PR #3192, opened 2024-09-20
chore: Update Bloop to 2.0.2

tgodzik:


Gedochao (approved):
We should do a follow-up to fix the `rt.jar` problems, but it seems it's still worth it to do the update as it is.

tgodzik:
Btw. I managed to improve the situation in https://github.com/scalacenter/bloop/pull/2450

PR #3202, opened 2024-09-24
Apply increased verbosity when compiling via BSP

Gedochao:
This passes the `--verbose` argument to Bloop's BSP when verbosity is > 0.

It applies both for passing `-v` to the raw `bsp` sub-command:
```bash
scala-cli bsp . -v
```
as well as the BSP connection loaded by one's IDE after running `setup-ide`:
```bash
scala-cli setup-ide . -v
```


tgodzik (approved):
LGTM!

Issue #2867, opened 2024-04-19
Improve Scala compiler settings integration


lrytz:
Sorry for another ticket like https://github.com/VirtusLab/scala-cli/issues/2779 / https://github.com/VirtusLab/scala-cli/issues/753... I understand there are difficulties how to handle this, but the user experience can probably still be improved.

There are compiler options not forwarded by the compiler, for example

```
➜ sandbox scala-cli compile --server=false C.scala
-- [E007] Type Mismatch Error: /Users/luc/code/scala/scala13/sandbox/C.scala:2:15
...
  | longer explanation available when compiling with `-explain`
1 error found
Compilation failed

➜ sandbox scala-cli compile --server=false C.scala -explain
Unrecognized argument: -explain

To list all available options, run
  scala-cli compile --help
```

When a user installs Scala, runs the compiler, and the compiler tells them to add a flag (`-explain`), they should not need to figure out that they actually need to do `-O -explain`.

The approach currently is for scala-cli to forward known compiler flags. Would it be possible to pass all unrecognized arguments to the compiler by default (for compile / repl / doc)?

As a backstop, a static hint in the "Unrecognized argument" message could help. If `scala-cli compile / repl / doc` gets an unrecognized argument, it could suggest `"Use -O ${unrecognized} to pass the argument to the Scala compiler"`.

-----

Another confusing situation is because the compiler is not invoked when there are no souce files. So `scala-cli compile -S 2 C.scala -opt:help` works, but `scala-cli compile -S 2 -opt:help` doesn't.

---- 

Finally, in Scala 2 at least there are compiler options that support space separated arguments, for example `-Vopt _` or `-Vprint-args args.txt`. The first makes scala-cli hang, the second gives an error message (`[error]  args.txt: unrecognized source type`).

It's again not obvious for the user what's going on.

Gedochao:
> Would it be possible to pass all unrecognized arguments to the compiler by default (for compile / repl / doc)?

I'm not a fan of this, as it may cause unexpected results. Many compiler options take arguments, which then have to be correctly parsed by Scala CLI. I think it would cause more confusion than merit.

A more sustainable solution would be to rely on the Scala compiler `Settings` API, as implemented in https://github.com/scala/scala3/pull/19766
Doing so is the long-term plan for solving this issue.

I suppose this would only solve options for the Scala 3 compiler (which is the priority).
[...]

PR #3245, opened 2024-10-24
Ensure consecutive `-Wconf:*` flags are not ignored

Gedochao:
a fix for the issue noted in https://github.com/scala/scala3/issues/21818#issuecomment-2433047638

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
@@ -515,6 +515,39 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
     }
   }
 
+  test("consecutive -Wconf:* flags are not ignored") {
+    val sv                 = "3.5.2"
+    val sourceFileName     = "example.scala"
+    val warningConfOptions = Seq("-Wconf:cat=deprecation:e", "-Wconf:any:s")
+    TestInputs(os.rel / sourceFileName ->
+      s"""//> using scala $sv
+         |//> using options ${warningConfOptions.mkString(" ")}
+         |object WConfExample extends App {
+         |  @deprecated("This method will be removed", "1.0.0")
+         |  def oldMethod(): Unit = println("This is an old method.")
+         |  oldMethod()
+         |}
+         |""".stripMargin).fromRoot { root =>
+      val localCache = root / "local-cache"
+      val localBin   = root / "local-bin"
+      os.proc(
+        TestUtil.cs,
+        "install",
+        "--cache",
+        localCache,
+        "--install-dir",
+        localBin,
+        s"scalac:$sv"
+      ).call(cwd = root)
+      val cliRes = os.proc(TestUtil.cli, "compile", sourceFileName, "--server=false")
+        .call(cwd = root, check = false, stderr = os.Pipe)
+      val scalacRes = os.proc(localBin / "scalac", warningConfOptions, sourceFileName)
+        .call(cwd = root, check = false, stderr = os.Pipe)
+      expect(scalacRes.exitCode == cliRes.exitCode)
+      expect(scalacRes.err.trim() == cliRes.err.trim())
+    }
+  }
+
   for {
     sv <- Seq(Constants.scala212, Constants.scala213, Constants.scala3NextRc)
     code =
--- a/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
@@ -515,13 +515,19 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
     }
   }
 
-  test("consecutive -Wconf:* flags are not ignored") {
+  for {
+    useDirective <- Seq(true, false)
+    if !Properties.isWin
+    optionsSource = if (useDirective) "using directive" else "command line"
+  } test(s"consecutive -Wconf:* flags are not ignored (passed via $optionsSource)") {
     val sv                 = "3.5.2"
     val sourceFileName     = "example.scala"
     val warningConfOptions = Seq("-Wconf:cat=deprecation:e", "-Wconf:any:s")
+    val maybeDirectiveString =
+      if (useDirective) s"//> using options ${warningConfOptions.mkString(" ")}" else ""
     TestInputs(os.rel / sourceFileName ->
       s"""//> using scala $sv
-         |//> using options ${warningConfOptions.mkString(" ")}
+         |$maybeDirectiveString
          |object WConfExample extends App {
          |  @deprecated("This method will be removed", "1.0.0")
          |  def oldMethod(): Unit = println("This is an old method.")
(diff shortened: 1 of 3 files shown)


Gedochao:
cc @SethTisue @prolativ

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
```diff
@@ -18,7 +18,8 @@ object ScalacOpt {
   private val repeatingKeys = Set(
     "-Xplugin",
     "-P", // plugin options
-    "-language"
+    "-language",
+    "-Wconf"
```
tgodzik, line 22: This one again :/ Feels like the logic is wrong here, the default behaviour should be to repeat things unless it's one of the options that can't be repeated.

Gedochao: I mean, the proper fix is to work on https://github.com/VirtusLab/scala-cli/issues/2867 first.

Gedochao: The problem is that we're emulating `scalac` options as `scala-cli` options, and hardcoding special cases for now.
We can feed it from the compiler settings API, but that requires rewriting a considerable chunk of the logic.

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -515,6 +515,39 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
     }
   }
 
+  test("consecutive -Wconf:* flags are not ignored") {
+    val sv                 = "3.5.2"
+    val sourceFileName     = "example.scala"
+    val warningConfOptions = Seq("-Wconf:cat=deprecation:e", "-Wconf:any:s")
+    TestInputs(os.rel / sourceFileName ->
+      s"""//> using scala $sv
+         |//> using options ${warningConfOptions.mkString(" ")}
```
prolativ: This test checks the behaviour for repeated options added as a directive. Do/can we have a test to verify this works the same for options passed via command line (I checked manually that the behaviour isn't consistent currently, without the changes from this PR)

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
```diff
@@ -515,6 +515,39 @@ class SipScalaTests extends ScalaCliSuite with SbtTestHelper with MillTestHelper
     }
   }
 
+  test("consecutive -Wconf:* flags are not ignored") {
+    val sv                 = "3.5.2"
+    val sourceFileName     = "example.scala"
+    val warningConfOptions = Seq("-Wconf:cat=deprecation:e", "-Wconf:any:s")
```
prolativ, line 525: How about adding some noise here by adding an option which shouldn't change the compiler's behaviour in this context, just to make sure that it's not only the last or the first option that is passed to the compiler, while the rest would be ignored?

tgodzik (approved):
Thanks!

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
jchyb: It seems I failed to include every repeating option, but another issue I see is that this was meant to allow overwriting directives with cli options, but not necessarily cli options with another cli options (or directive with another directive). Apologies for the error!

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
```diff
@@ -10,15 +10,16 @@ final case class ScalacOpt(value: String) {
   /** @return raw key for the option (only if the key can be shadowed from the CLI) */
   private[options] def shadowableKey: Option[String] = key match
     case Some(key)
-        if ScalacOpt.repeatingKeys.exists(rKey => rKey.startsWith(key + ":") || rKey == key) => None
+        if ScalacOpt.repeatingKeys.exists(rKey => rKey.contains(":") || rKey == key) => None
```
prolativ: There seems to be something in this logic that I don't get for some reason. Is `rKey.contains(":")` ever true? None of the keys listed in `ScalacOpt.repeatingKeys` seems to contain `":"`

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
Gedochao: No worries, it's also hard to keep track of it from one compiler version to another. 
We need to fix it properly in #2867.

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
tgodzik: Och right, this should be `key.contains`

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: ...I get the feeling that's overkill, but I guess it won't hurt. 
Let me add that.

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
Gedochao: ah, that's a legacy inconsistency... because the `:` used to be there.
Now it indeed will never come true. 
Lemme adjust it.

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: ...like?

Gedochao: I think this case is sufficient, as both options affect the output in tandem, so it won't work correctly unless both are applied... And `-Wconf` is quite tweaky, so I'd rather not make it more complex, to make referring to this test later easier...

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
Gedochao: Should be fixed

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
prolativ: I mean e.g. currently (with scala-cli 1.5.1) the following code raises an error
```scala
//> using scala 3.5.2
//> using options -Wconf:cat=deprecation:e -Wconf:any:s
object WConfExample extends App {
  @deprecated("This method will be removed", "1.0.0")
  def oldMethod(): Unit = println("This is an old method.")
  oldMethod()
}
```
but this has just a warning
```scala
//> using scala 3.5.2
//> using options -Wconf:cat=unchecked:s -Wconf:cat=deprecation:e -Wconf:any:s
object WConfExample extends App {
  @deprecated("This method will be removed", "1.0.0")
  def oldMethod(): Unit = println("This is an old method.")
  oldMethod()
}
```
In this test it's not really clear if the given behaviour is a result of reshuffling options or discarding one/some of them

modules/integration/src/test/scala/scala/cli/integration/SipScalaTests.scala
Gedochao: turns out it not only it wasn't, but it also uncovered another bug.
Added a test + options from both command line and directives should now follow the same ordering (they weren't)

tgodzik (approved):

modules/options/src/main/scala/scala/build/options/ScalacOpt.scala
Gedochao: ...I ended up reverting this change, as it breaks options shading/overriding from command line.
We should address this properly elsewhere.

prolativ (approved):

PR #3253, opened 2024-10-29
Misc improvements in compiler options handling

Gedochao:
This is largely a (much belated) follow-up to https://github.com/scala/scala3/pull/19766, which revamped how compiler settings can be called. We haven't been supporting some of the new syntax.
The actual follow-up we plan for is https://github.com/VirtusLab/scala-cli/issues/2867, this PR should be treated as the first step.

Changes include
- compiler options can now be passed with `-` or `--` prefixes interchangably (`-Wunused:locals` means exactly the same as `--Wunused:locals`)
- compiler option values can be passed with `:` or a space, interchangably (`-Wunused:locals` means the same as `-Wunused locals`)
  - this applies to repeatable options, so now `-Wunused:locals,privates`, `-Wunused:locals -Wunusedprivates` and `-Wunused locals -Wunused privates` are all synonyms.
- some options which previously required to be passed via `-O` now don't:
  - `-explain`
  - `-experimental`
- some options which should have allowed to be repeatable (as they accept a list of values) now actually do
  - `-Wunused` 
  - `-coverage-exclude-classlikes`
  - `-coverage-exclude-files`
  - `-Wshadow`
  - `-Xlint`
  - `-Xmacro-settings`
  - `-Xplugin-disable`
  - `-Xplugin-require`
  - `-Yimports`
  - `-Yfrom-tasty-ignore-list`
- `--source` (which used to indicate a source jar artifact is to be produced by the `package` sub-command) and `-source` (a compiler option) are now synonyms, as treated by the Scala compiler
  - to produce a source jar, use `--src` or `--sources` instead.
- the changes have been covered with tests where applicable (within reason, we can't test for each compiler option)
  - these tests are also good preparation for the rewrite necessary for https://github.com/VirtusLab/scala-cli/issues/2867


Gedochao:
Note: the changes are rather large, although I tried to split the commits in a way so that it's easy to review. Let me know if I should split it into smaller bits.

modules/integration/src/test/scala/scala/cli/integration/CompileScalacCompatTestDefinitions.scala
```diff
@@ -0,0 +1,208 @@
...
+        val expectedStrictEqualityError =
+          " Values of types Cat and Dog cannot be compared with == or !="
+        expect(errOutput.contains(expectedStrictEqualityError))
+        val expectedNoAutoTuplingError =
+          "too many arguments for method takesTuple: (tpl: Tuple): Nothing"
+        expect(errOutput.trim().contains(expectedNoAutoTuplingError))
+      }
+    }
+
+  // Given the vast number of ways compiler options can be passed from the CLI,
```
tgodzik, line 34: This looks like we should use scalacheck at some point :sweat_smile:

tgodzik (approved):
Looks good!

PR #3255, opened 2024-10-30
Add a `--js-emit-wasm` option and a corresponding `using` directive

Quafadas:
- [x] run `./mill -i generate-reference-doc.run` in order to generate doc correctly
- [x] check directive tags (should vs experimental)
- [x]  directive doc
- [x] test
- [x] pass CI

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
@@ -61,6 +61,11 @@ final case class ScalaJsOptions(
 
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.should)
+  @HelpMessage("Experimental: emit WASM")
+    jsEmitWasm: Option[Boolean] = None,
+
+  @Group(HelpGroup.ScalaJs.toString)
+  @Tag(tags.should)
   @HelpMessage("A header that will be added at the top of generated .js files")
     jsHeader: Option[String] = None,
 
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
@@ -60,7 +60,7 @@ final case class ScalaJsOptions(
     jsDom: Option[Boolean] = None,
 
   @Group(HelpGroup.ScalaJs.toString)
-  @Tag(tags.should)
+  @Tag(tags.experimental)
   @HelpMessage("Experimental: emit WASM")
     jsEmitWasm: Option[Boolean] = None,
 
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
@@ -61,7 +61,7 @@ final case class ScalaJsOptions(
 
   @Group(HelpGroup.ScalaJs.toString)
   @Tag(tags.experimental)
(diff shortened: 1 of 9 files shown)


Quafadas:
@Gedochao I think I might have gotten to the stage where I need some help / advice / review if I may once the CI has run through. 

I added an emits WASM test. To actually run the WASM, I think we'd need node 22. Whilst I was rather gung-ho trampling the scala-ji-cli gha to add node 22, I'm far more hesitant about doing so for Scala CLI itself. 

https://github.com/Quafadas/scala-cli/blob/cb64e000c0dbacc9e7ba831c83b85d82ba70c9b2/modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala#L327-L328

I'm proposing to not actually run the WASM emitted in this test suite, rather only test that the file exists. Execution _is_ tested in the underlying js-cli project so it's arguably a duplication. If you feel strongly about it then I'll have to go through specifying some node versions... 

The second thing, is that I noticed a test failure on windows BSP on the last time out. I don't _think_ that is me, but I'm also not qualified to diagnose what's happening there. I _think_ CI should go green for my set of changes. If it doesn't, I might need some help to figure out what's going on, if that's okay. 

Finally, I'm not sure if any extra documentation is needed beyond what scala-cli demands as the minimal set it will pass CI with? Perhaps not, as sjrd covered that in the release notes of scala js itself?  

Outside of those points, I believe this would be ready for review / merge ...

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -291,6 +291,50 @@ trait RunScalaJsTestDefinitions { _: RunTestDefinitions =>
...
+      expect(os.exists(absOutDir / "main.wasm"))
+
+      // This would require node 22. As it's tested in the scala-cli-js repo, I'm going to claim that is isn't necessary here.
+      // os.proc("node", "--experimental-wasm-exnref", "main.js").call(
+      //   cwd = root / outDir,
+      //   check = true,
+      //   stdin = os.Inherit,
+      //   stdout = os.Inherit,
+      //   stderr = os.Inherit
+      // )
```
Gedochao: Ok... suggestion: let's remove the commented dead code here for now, and leave a TODO comment. Bumping Node to 22 and adding a test can be extracted to a separate issue.

I'd still rather have something run on the CI in `scala-cli`, but it doesn't have to be in the scope of this PR.

modules/cli/src/main/scala/scala/cli/commands/shared/ScalaJsOptions.scala
```diff
@@ -59,6 +59,11 @@ final case class ScalaJsOptions(
   @HelpMessage("Enable jsdom")
     jsDom: Option[Boolean] = None,
 
+  @Group(HelpGroup.ScalaJs.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Experimental: emit WASM")
```
Gedochao: It's sufficient that it's tagged as `experimental`.
```suggestion
  @HelpMessage("Emit WASM")
```

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
Quafadas: Done - I'll report a seperate issue

Gedochao (approved):
LGTM

Gedochao:
@Quafadas thanks a lot for attending to this!

Quafadas:
@Gedochao Pleasure - thanks for the support :-).

PR #3257, opened 2024-11-04
Deprecate the `--source` command line option for the `package` sub-command

Gedochao:
- the `--source` command line option alias (which makes the `package` sub-command produce source JARs) is getting deprecated in preparation for https://github.com/VirtusLab/scala-cli/pull/3253.
- it will likely be removed in Scala CLI v1.6.0, where `--source` will be treated as a compiler option instead.
- the recommended syntax would now be `--src`.

```bash
scala-cli-jvm --power package --source .                       
# [warn] The --source option alias has been deprecated and may be removed in a future version.
(...)
```

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
@@ -3,13 +3,13 @@ package scala.cli.commands
 import caseapp.Name
 import caseapp.core.app.Command
 import caseapp.core.parser.Parser
-import caseapp.core.util.Formatter
+import caseapp.core.util.{CaseUtil, Formatter}
 import caseapp.core.{Arg, Error}
 
 import scala.build.Logger
 import scala.build.input.ScalaCliInvokeData
 import scala.build.internal.util.WarningMessages
-import scala.build.internals.FeatureType
+import scala.build.internals.{ConsoleUtils, FeatureType}
 import scala.cli.ScalaCli
 import scala.cli.util.ArgHelpers.*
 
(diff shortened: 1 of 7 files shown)


modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
```diff
@@ -57,6 +57,16 @@ object RestrictedCommandsParser {
               if arg.isExperimental && !shouldSuppressExperimentalWarnings =>
             logger.experimentalWarning(passedOption, FeatureType.Option)
             r
+          case (r @ Right(Some(_, arg: Arg, _)), passedOption :: _)
+              if arg.isDeprecated =>
+            // TODO implement proper option/option alias deprecation logic, similar to experimental
```
Gedochao: - https://github.com/VirtusLab/scala-cli/issues/3258

tgodzik: Let's add a link to the issue here.

modules/cli/src/main/scala/scala/cli/commands/package0/PackageOptions.scala
```diff
@@ -49,10 +49,11 @@ final case class PackageOptions(
   @Group(HelpGroup.Package.toString)
   @HelpMessage("Generate a source JAR rather than an executable JAR")
   @Name("sources")
-  @Name("src")
+  @Name("source")
+  @Tag(tags.deprecated("source")) // alias to be removed in 1.6.x
   @Tag(tags.restricted)
   @Tag(tags.inShortHelp)
-    source: Boolean = false,
+    src: Boolean = false,
```
tgodzik, line 58: maybe we should actually be more verbose here? `--outputSources` or `--jarSources`. `src` seems too short and it also can be mixed up with `src` used for naming of the sources directory and easily misused. Like `--src ../additional-sources` for example.

Gedochao: `--src` is an existing alias, I merely now made it the default name... no need to remove it IMO.
I can add `--jar-sources` as an extra alias, if needed.

Gedochao: added `--jar-sources` and `--sources-jar` as additional aliases.

modules/cli/src/main/scala/scala/cli/commands/RestrictedCommandsParser.scala
Gedochao: done

tgodzik (approved):
LGTM

SethTisue:
I'm sorry, I know this feedback is coming too late, but could a _more_ different name than `--src` be found? `--src` and `-source` still seem awfully confusable with each other, especially when spoken out loud.

SethTisue:
Maybe `--with-sources`?

tgodzik:
`--src` already existed before and I agree it's easy to confuse. @Gedochao maybe we should deprecate both?

PR #3272, opened 2024-11-12
Make publishing depend on all integration tests & docs tests

Gedochao:
This is to prevent problems similar to https://github.com/VirtusLab/scala-cli/actions/runs/11792000662/job/32855838373
- [ ] depends on https://github.com/VirtusLab/scala-cli/pull/3270

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1592,7 +1592,47 @@ jobs:
         retention-days: 2
 
   publish:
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
Though it might get hard to release because of flaky tests 🤔

Gedochao:
I'm predicting nightlies will become troublesome with this. 
MIght need to expediate the audit of our integration tests after we release 1.5.3...

PR #3279, opened 2024-11-14
Allow shading of single-choice compiler options from the command line regardless of `-`/`--` prefix

Gedochao:
Scala CLI allows (even before this PR) to override using directive settings from the command line, including single-choice (non-repeatable) compiler options.

Given a source file `unused.scala`:
```scala
//> using options -Wunused:all -Werror
@main def main() = {
  val unusedValue = "something"
  println("Hello, world!")
}
```
The unused warning produced by `-Wunused:all` would get raised to an error due to `-Werror`.
```bash
scala-cli compile unused.scala
# Compiling project (Scala 3.5.1, JVM (17))
# [warn] ./repro.scala:3:7
# [warn] unused local definition
# [warn]   val unusedValue = "something"
# [warn]       ^^^^^^^^^^^
# Error: No warnings can be incurred under -Werror (or -Xfatal-warnings)
# Error compiling project (Scala 3.5.1, JVM (17))
# Compilation failed
```

However, `-Werror` can be overridden from the command line with `-Werror:false`, without editing sources:
```bash
scala-cli compile repro.scala -Werror:false                      
# Compiling project (Scala 3.5.1, JVM (17))
# [warn] ./repro.scala:3:7
# [warn] unused local definition
# [warn]   val unusedValue = "something"
# [warn]       ^^^^^^^^^^^
# Compiled project (Scala 3.5.1, JVM (17))
```

This PR allows to perform overriding like this regardless if the option key is preceded with a single `-` (`-Werror`) or double (`--Werror`). 
So this allows for an override like:
```bash
scala-cli compile repro.scala --Werror:false                    
# Compiling project (Scala 3.5.1, JVM (17))
# [warn] ./repro.scala:3:7
# [warn] unused local definition
# [warn]   val unusedValue = "something"
# [warn]       ^^^^^^^^^^^
# Compiled project (Scala 3.5.1, JVM (17))
```

Dependencies:
- [x] https://github.com/VirtusLab/scala-cli/pull/3253

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
@@ -177,9 +177,7 @@ abstract class ScalaCommand[T <: HasGlobalOptions](implicit myParser: Parser[T],
     sharedOptions(options).foreach { so =>
       val scalacOpts = so.scalacOptions.toScalacOptShadowingSeq
       scalacOpts.keys
-        .find(k =>
-          k == ScalacOpt(s"-$YScriptRunnerOption") || k == ScalacOpt(s"--$YScriptRunnerOption")
-        )
+        .find(_.value.noDashPrefixes == YScriptRunnerOption)
         .map(_.value)
         .foreach(k =>
           logger.message(LegacyScalaOptions.yScriptRunnerWarning(k, scalacOpts.getOption(k)))
--- a/modules/options/src/main/scala/scala/build/options/ShadowingSeq.scala
+++ b/modules/options/src/main/scala/scala/build/options/ShadowingSeq.scala
@@ -2,6 +2,7 @@ package scala.build.options
 
 import dependency.AnyDependency
 
+import scala.build.options.ScalacOpt.noDashPrefixes
 import scala.collection.mutable
 
 /** Seq ensuring some of its values are unique according to some key */
@@ -31,10 +32,11 @@ final case class ShadowingSeq[T] private (values: Seq[Seq[T]]) {
       for (group <- values.iterator ++ other.iterator) {
         assert(group.nonEmpty)
         val keyOpt = key.makeKey(group)
-        if (!keyOpt.exists(seen.contains)) {
+        if !keyOpt.exists(k => seen.contains(k.noDashPrefixes))
+        then {
           l += group
           for (key <- keyOpt)
-            seen += key
+            seen += key.noDashPrefixes
         }
       }
 
--- a/modules/integration/src/test/scala/scala/cli/integration/CompileScalacCompatTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompileScalacCompatTestDefinitions.scala
@@ -205,4 +205,36 @@ trait CompileScalacCompatTestDefinitions { _: CompileTestDefinitions =>
         }
       }
   }
+
+  {
+    val prefixes = Seq("-", "--")
+    for {
+      prefix1 <- prefixes
+      prefix2 <- prefixes
+      optionKey = "Werror"
+      option1   = prefix1 + optionKey
+      option2   = prefix2 + optionKey
+      if actualScalaVersion.startsWith("3")
+    } test(
+      s"allow to override $option1 compiler option passed via directive by passing $option2 from the command line"
+    ) {
+      val file = "example.scala"
+      TestInputs(os.rel / file ->
+        s"""//> using options -Wunused:all $option1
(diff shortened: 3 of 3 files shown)


modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
```diff
@@ -175,9 +176,13 @@ abstract class ScalaCommand[T <: HasGlobalOptions](implicit myParser: Parser[T],
     val logger = options.global.logging.logger
     sharedOptions(options).foreach { so =>
       val scalacOpts = so.scalacOptions.toScalacOptShadowingSeq
-      if scalacOpts.keys.contains(ScalacOpt(YScriptRunnerOption)) then
-        logger.message(
-          LegacyScalaOptions.yScriptRunnerWarning(scalacOpts.getOption(YScriptRunnerOption))
+      scalacOpts.keys
+        .find(k =>
+          k == ScalacOpt(s"-$YScriptRunnerOption") || k == ScalacOpt(s"--$YScriptRunnerOption")
```
tgodzik:
```suggestion
          k.value.noDashPrefixes == YScriptRunnerOption
```
?

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -1712,7 +1712,7 @@ abstract class BspTestDefinitions extends ScalaCliSuite with TestScalaVersionArg
         "--power",
         "package",
         jarSources,
-        "--source",
+        "--src",
```
tgodzik: @SethTisue mentioned that we might want to actually use another name and also deprecate src and I think it makes sense. `--jar-sources` was the proposal

modules/options/src/main/scala/scala/build/options/ShadowingSeq.scala
```diff
@@ -31,7 +31,12 @@ final case class ShadowingSeq[T] private (values: Seq[Seq[T]]) {
       for (group <- values.iterator ++ other.iterator) {
         assert(group.nonEmpty)
         val keyOpt = key.makeKey(group)
-        if (!keyOpt.exists(seen.contains)) {
+        if !keyOpt.exists {
+            case k if k.startsWith("--") => seen.contains(k) || seen.contains(k.stripPrefix("-"))
```
tgodzik: Could we stip prefixes on adding to `seen` and compare just with noDashPrefixes ?

tgodzik (changes requested):

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
Gedochao: I can do that in a separate PR.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: we don't care if it's `-` or `--` here.

modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
Gedochao: a bit awkward, since we just deprecated `--source`, defaulting to `--src` 😅 
but okay.

modules/options/src/main/scala/scala/build/options/ShadowingSeq.scala
Gedochao: Let me try that.

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
tgodzik: I mean that's what `noDashPrefixes` is about?  it would just be a simpler code, no?

Gedochao: ahh, ok. yeah, I read your first comment in a rush. lemme fix that.

modules/options/src/main/scala/scala/build/options/ShadowingSeq.scala
Gedochao: Done

modules/cli/src/main/scala/scala/cli/commands/ScalaCommand.scala
Gedochao: done

tgodzik (approved):

PR #3333, opened 2024-11-28
Fix commas being treated as `using` directives value separators & deprecate using them with whitespace

### LINKED ISSUE #3282: Commas shouldn't be interpreted as directive value separators on their own
**Version(s)**
1d6064127a384cf961949f251f538d1a03df8889

**Describe the bug**
The comma (`,`) character is currently parsed as a separator for using directive values, while it should require a space.

**To Reproduce**
```scala
//> using dep tabby:tabby:0.2.3,url=https://github.com/bjornregnell/tabby/releases/download/v0.2.3/tabby_3-0.2.3.jar
```
this is currently parsed as 2 values being passed to the `dep` directive key:
- `tabby:tabby:0.2.3`
- `url=https://github.com/bjornregnell/tabby/releases/download/v0.2.3/tabby_3-0.2.3.jar`
This makes it very easy to make mistakes such as https://github.com/VirtusLab/scala-cli/pull/3281, as dependency modifiers follow `coursier` syntax, where `,` has syntactic meaning and currently needs to be escaped via surrounding the value with double quotes:
```scala
//> using dep "tabby:tabby:0.2.3,url=https://github.com/bjornregnell/tabby/releases/download/v0.2.3/tabby_3-0.2.3.jar"
```

This is equally as bad for repeating compiler options, where it is easy to make mistakes such as:
```scala
//> using options -Werror -Wunused:privates,locals,imports
```
Where what is passed as a compiler option is:
- `-Werror`
- `-Wunused:privates`
- `locals`
- `imports`
The currently correct way to pass all the values to the compiler would be to escape the commas again:
```scala
//> using options -Werror "-Wunused:privates,locals,imports"
```

**Expected behaviour**
- [ ] escaping commas in values should not be necessary, as `,` shouldn't be a separator on its own
- [ ] using `, ` (a comma followed by a space) as a value separator should be deprecated, producing a warning (but we'd probably still honor this syntax for a long while)

Gedochao:
Fixes #3282 

NOTE: these are breaking changes affecting using directives syntax.
It's technically a fix + deprecation, but it is not impossible for existing builds to break.

For example, where
```scala
//> using scala 2.13,3
```
Used to mean that Scala is set to 2.13 and 3 for cross-builds, it will now mean Scala is set to `2.13,3`, which is not a valid version and will produce an error.
The valid syntax for the same would now be:
```scala
//> using scala 2.13 3
```

The fix itself is contained in the `using_directives` bump to 1.1.2:
- https://github.com/VirtusLab/using_directives/releases/tag/v1.1.2

Specifically, it's this change:
- https://github.com/VirtusLab/using_directives/pull/61

What this PR does:
- bumps `using_directives` to 1.1.2
  - using commas (`,`) as separators without whitespace is no longer supported (the comma will now be treated as part of the value)
  - using commas (`,`) with a following whitespace as separator gets deprecated and will produce a warning
- adds tests ensuring the old syntax is no longer honoured (no longer can use commas on their own as value separators)
- adds tests for the deprecation of using commas with whitespace as value separators
- updates the syntax in docs, old code & tests (which contributes the most to the change line count)
- updates the `publish setup` and `fix` sub-commands to use the latest syntax

### DIFF
--- a/gcbenchmark/gcbenchmark.scala
+++ b/gcbenchmark/gcbenchmark.scala
@@ -1,6 +1,6 @@
-//> using dep "com.lihaoyi::os-lib:0.9.1"
-//> using dep "com.lihaoyi::pprint:0.8.1"
-//> using scala "2.13"
+//> using dep com.lihaoyi::os-lib:0.9.1
+//> using dep com.lihaoyi::pprint:0.8.1
+//> using scala 2.13
 
 // Usage: scala-cli gcbenchmark.scala -- <path_to_scala_cli_executable>
 
--- a/gifs/create_missing.sc
+++ b/gifs/create_missing.sc
@@ -1,6 +1,6 @@
 #!/usr/bin/env scala-cli
 
-//> using lib "com.lihaoyi::os-lib:0.7.8"
+//> using lib com.lihaoyi::os-lib:0.7.8
 
 /** Small and handy script to generate stubs for .svg files with nice TODO
   */
--- a/gifs/scenarios/embeddable_scripts.sh
+++ b/gifs/scenarios/embeddable_scripts.sh
@@ -18,7 +18,7 @@ else
   # Put your stuff here
     cat <<EOF | updateFile count_lines.sc
 #!/usr/bin/env scala-cli
-//> using scala "3.0.2"
+//> using scala 3.0.2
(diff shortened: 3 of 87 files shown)


tgodzik (approved):
LGTM!

PR #2701, opened 2024-01-23
Bump actions/upload-artifact & actions/download-artifact from 3 to 4

Gedochao:
Supersedes #2640  & #2641, the bump has to be done in one go.
This is just a squashed commit with the contents of both aforementioned PRs.

PR #3342, opened 2024-12-03
Get rid of duplicate names for uploaded/downloaded artifacts on the CI

Gedochao:
Follow-up after:
- https://github.com/VirtusLab/scala-cli/pull/2701

This should do the trick and fix nightlies on `main` (hopefully).
We don't run `vc-redist` on PRs, which is why it didn't fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
     - uses: actions/upload-artifact@v4
       if: runner.os == 'Linux'
       with:
-        name: launchers
+        name: jvm-launchers
         path: artifacts/
         if-no-files-found: error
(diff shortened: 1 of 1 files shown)


kasiaMarek:
Was it named the same before because you use one or the other?

Gedochao:
> Was it named the same before because you use one or the other?

@kasiaMarek it was named the same before, because it could.
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact
the upgrade to v4 apparently no longer supports merging artifacts when uploading to the same name, so each has to have its own.

kasiaMarek (approved):

PR #3358, opened 2024-12-09
Bump Scala Next to 3.6.2

Gedochao:
- https://github.com/scala/scala3/issues/22167


tgodzik (approved):

modules/integration/src/test/scala/scala/cli/integration/ScalafixTestDefinitions.scala
```diff
@@ -276,6 +276,7 @@ abstract class ScalafixTestDefinitions extends ScalaCliSuite with TestScalaVersi
         s"""|rules = [
             |  ExplicitResultTypes
             |]
+            |ExplicitResultTypes.fetchScala3CompilerArtifactsOnVersionMismatch = true
```
Gedochao, line 279: We need this for Scala minor version bumps (3.5.x => 3.6.x, in this case)

PR #3375, opened 2024-12-17
Run extra tests for main supported JVM versions

Gedochao:
- just some extra tests to ensure the CLI is compatible with all the main Java versions it should be compatible with.
- relevant to https://github.com/scala/docs.scala-lang/pull/3082


modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala
```diff
@@ -0,0 +1,133 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+import scala.util.Properties
+
+trait RunJdkTestDefinitions { _: RunTestDefinitions =>
+  def javaIndex(javaVersion: Int): String =
+    // TODO just passing the version number on arm64 should be enough, needs a fix in cs
```
tgodzik, line 9: Do we have it raised upstream?

modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala
```diff
@@ -0,0 +1,133 @@
...
+          withLauncher(root) { launcher =>
+            val res = os.proc(launcher, "run", ".", extraOptions, "--jvm", javaVersion)
+              .call(cwd = root)
+            expect(res.out.trim() == expectedMessage)
+          }
+        }
+      }
+    }
+
+    if (!Properties.isWin || !useScalaInstallationWrapper) // TODO make this pass on Windows
```
tgodzik, line 79: Why is not passing on windows?

tgodzik (approved):

modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala
Gedochao: seemingly Scala CLI can't find the Java installation passed from `cs` via `--java-home`.
Since the same works when passed via `JAVA_HOME`, I assume the installation itself is valid.
It's likely that `--java-home` is broken on Windows, have to investigate separately.

modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala
Gedochao: We don't. 
I'm not yet 100% sure where should we fix it.
Technically speaking, `cs` defaults to adoptium everywhere, so the error it fails with is valid...
```bash
cs java-home --jvm 16 --architecture arm64
# JVM 16 not found in index: No adoptium version matching '1.16+' found
```
I don't think we should fail in Scala CLi under the same circumstances, however.
Perhaps we should just log that the index wasn't found for `adoptium`, and fallback to smth else (i.e. `zulu`)...

Gedochao: - https://github.com/VirtusLab/scala-cli/issues/3396: raised an issue, let's track it there.

modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala
Gedochao: - https://github.com/VirtusLab/scala-cli/issues/3397: created an issue, let's track it there.

Issue #2973, opened 2024-06-21
Exclude directive is not honored for directories under Windows


nemoo:
Scala-cli 1.2.0
Windows 11

**Describe the bug**
The "exclude" directive is not honored for directories under Windows. 

**To Reproduce**
create Main.scala:
```
//> using exclude "input"

@main def go = println("Hello")
 ```

create input/Source.scala:
```
a
 ```

When running in Windows 11, produced the error: 
```
[error] .\input/Source.scala:1:1
[error] Illegal start of toplevel definition
[error] a
[error] ^
```

**Expected behaviour**
It should not try to compile files underneath the `input` directory because I have explicitly excluded the whole directory via the exclude directive. Also writing it like that is ignored: `//> using exclude "input/*"`

bishabosha:
I dont think its just windows - I also can't get it to work in https://github.com/bishabosha/ops-mirror (im using macOS)

PR #3381, opened 2024-12-19
Fix using directives crashing on `*/` by removing `/* (..) */` comments support in `using_directives`

### LINKED ISSUE #3374: `using_directives` parser crashes on `*/*/Foo.scala`
**Version(s)**
1.5.4

**Describe the bug**
Using a nested wildcard (`*/*/...`) in `exclude` directive results in a crash caused by `java.lang.ArrayIndexOutOfBoundsException`

**To Reproduce**
Having the following project structure:
* ./project.scala:
```scala
//> using exclude */*/Foo.scala
```

* ./dir1/dir2/Foo.scala:
```scala
val foo // invalid code
```
Try to compile the project:
```shell
scala-cli compile .
```
**Expected behaviour**
Should not crash. Also the invalid file `Foo.scala` should be excluded from compilation sources and the compilation should succeed

Gedochao:
Fixes #3374 

Requires:
- [x] https://github.com/VirtusLab/using_directives/pull/67

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
@@ -5,6 +5,7 @@ import com.eed3si9n.expecty.Expecty.expect
 import java.io.File
 
 import scala.cli.integration.util.BloopUtil
+import scala.util.Properties
 
 abstract class CompileTestDefinitions
     extends ScalaCliSuite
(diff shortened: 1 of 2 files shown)


modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -720,4 +720,20 @@ abstract class CompileTestDefinitions
       expect(buildTargetDirs.size == 2)
     }
   }
+
+  if (!Properties.isWin)
+    // TODO: make this work on Windows
```
Gedochao: - likely the same problem as https://github.com/VirtusLab/scala-cli/issues/2973, let's track it there

tgodzik: We should attach it to the TODO

tgodzik (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
Gedochao: sure, added the ticket link to the comment

PR #3386, opened 2024-12-20
docs: document Scala Native flag options

scarf005:
referenced from: https://github.com/scala-native/scala-native/blob/157355f4bcbb4a2873c0c7ccdc9be4be969135e2/docs/user/sbt.md?plain=1#L132

### DIFF
--- a/website/docs/reference/scala-command/runner-specification.md
+++ b/website/docs/reference/scala-command/runner-specification.md
@@ -182,15 +182,15 @@ Set the Scala Native version (0.5.6 by default).
 
 **--native-mode**
 
-Set Scala Native compilation mode
+Set Scala Native compilation mode (debug by default): debug, release-fast, release-size, release-full
 
 **--native-lto**
 
-Link-time optimisation mode
+Link-time optimisation mode (none by default): none, full, thin
 
 **--native-gc**
 
-Set the Scala Native garbage collector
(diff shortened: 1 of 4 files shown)


website/docs/reference/scala-command/runner-specification.md
```diff
@@ -182,15 +182,15 @@ Set the Scala Native version (0.5.6 by default).
 
 **--native-mode**
 
-Set Scala Native compilation mode
```
Gedochao, line 185: Everything under `website/docs/reference` is generated automatically with
```bash
```

This needs to be adjusted in the Scala Native options implementation first:
https://github.com/VirtusLab/scala-cli/blob/4ab322768f7b5813312aeb553fed8a2caf35e4a4/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala#L22

Gedochao (changes requested):

website/docs/reference/scala-command/runner-specification.md
scarf005: applied and force-pushed in 29f790e0546f2ee98ff4da92741672de6f322c5a.

Gedochao (approved):

He-Pin:
Thanks

Issue #3330, opened 2024-11-28
Add a default set of rules for running `scalafix`


Gedochao:
**Is your feature request related to a problem? Please describe.**
We currently require a `.scalafix.conf` file to be present in the project.
We should use a predefined default set of rules, if none are provided.

**Describe the solution you'd like**
- [ ] when `scalafix` is run in a directory with no `.scalafix.conf`, a set of default rules should be run (the temp config file can be kept in the `.scala-build` workspace directory)
- [ ] there should be a command line option for persisting the default set of rules in a `.scalafix.conf` file in the project root directory

Gedochao:
@bjaglin @dos65 @Vigorge @tgodzik 
Any suggestions on what rules should be enabled by default?

tgodzik:
That is an interesting question, since it might be highly subjective. :thinking: 

I think OrganizeImports is a pretty sure one. 

Let's maybe create a list:
- [ ]  [ExplicitResultTypes](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html) <- this still needs some work for Scala 3
- [ ]  [NoAutoTupling](https://scalacenter.github.io/scalafix/docs/rules/NoAutoTupling.html) Inserts explicit tuples for adapted argument lists for compatibility with -Yno-adapted-args <- not super neccessary
- [x] [OrganizeImports](https://scalacenter.github.io/scalafix/docs/rules/OrganizeImports.html) <- 100% and should be safe
- [x] [RemoveUnused](https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html) <- this one is also safe and only removes anything reported by the compiler under -Wunused

- [ ] [DisableSyntax](https://scalacenter.github.io/scalafix/docs/rules/DisableSyntax.html)	Reports an error for disabled features such as var or XML literals. <- I don't think we need this one
- [x]  [LeakingImplicitClassVal](https://scalacenter.github.io/scalafix/docs/rules/LeakingImplicitClassVal.html) - Adds 'private' to val parameters of implicit value classes <- also looks pretty safe
- [x] [NoValInForComprehension](https://scalacenter.github.io/scalafix/docs/rules/NoValInForComprehension.html)	Removes deprecated val inside for-comprehension binders <- should also be safe
[...]

PR #2309, opened 2023-07-25
Add fix command for extracting using directives into `project.scala` file

MaciejG604:
Fixes #2104

Introduces a new subcommand `fix` that extracts all `using directives` that exist througout the project sources (not only files on disk but everything that's linked to the project build). Those extracted directives are then written to `project.scala` file.

This process can be split into two parts (their order is not exactly 1. -> 2. in code):
1. Extraction of directives and writing them into `project.scala`
2. Modification of directives in their original sources.

Rules regarding the 1. part:
- if a source contains any `target using directive` then it's directives are not extracted, since it's a correct way to create projects with cross-configuration (the warning about using directives in multiple files has been modified accordingly)
- directives from sources belonging to the Test scope are extracted and written to `project.scala` using their `test.` equivalents (using dep -> using test.dep) if possible, if not possible they are not written to `project.scala`

Rules regarding the 2. part: 
- if a source is writable (is on local disk) and it is a member of original inputs then the extracted directives are removed from its contents, otherwise no modification takes place
- if a source belonging to the Test scope contains directives that do not have `test.` equivalents they are kept in this source and are not moved to `project.scala`, all the others extracted `using directives` are removed from the source if rule above applies

PR #3400, opened 2024-12-31
Merge `scalafix` into `fix`

### LINKED ISSUE #3327: Merge `scalafix` sub-command into `fix`
I'd like `scalafix` to be integrated with the `fix` command. There's no point to keep those commands separate.

**Requirements**
- [x] merge `scalafix` into `fix`
- [x] add a command line option to run `scalafix` rules alongside Scala CLI rules

**Additional context**
Initial `scalafix` integration implementation:
- https://github.com/VirtusLab/scala-cli/pull/2968

Gedochao:
Closes #3327

This merges the `scalafix` sub-command into `fix`.

```bash
scala-cli fix . --power
# The `fix` sub-command is experimental
# Please bear in mind that non-ideal user experience should be expected.
# If you encounter any bugs or have feedback to share, make sure to reach out to the maintenance team at https://github.com/VirtusLab/scala-cli
# Running built-in rules...
# Writing project.scala
# Removing directives from Smth.scala
# Built-in rules completed.
# Running scalafix rules...
# 
# Starting compilation server
# Compiling project (Scala 3.6.2, JVM (zulu:23))
# [warn] ./Main.scala:2:7
# [warn] unused local definition
# [warn]   val unused = "unused"
# [warn]       ^^^^^^
# Compiled project (Scala 3.6.2, JVM (zulu:23))
# scalafix rules completed.
```
Former `fix` functionalities are now referred to in the code as the `built-in` rules.
Effectively, `fix` now runes 2 separate sets of rules (both enabled by default): `built-in` and `scalafix`.
They can be controlled via the `--enable-scalafix` and `--enable-built-in` command line options.


modules/cli/src/main/scala/scala/cli/commands/fix/Fix.scala
```diff
@@ -20,347 +23,43 @@ object Fix extends ScalaCommand[FixOptions] {
...
-      val key = directiveLine
-        .stripPrefix("//> using")
-        .stripLeading()
-        .stripPrefix("test.")
-        // separate key from value
-        .takeWhile(!_.isWhitespace)
-
-      directivesOrder(key)
-    }
+    else logger.message("No rules were enabled. Did you disable everything intentionally?")
```
tgodzik, line 63: This feels a bit unclear. Couldn't this also happen if we have an empty scalafix.conf?

tgodzik (approved):
LGTM, just one minor comment

modules/cli/src/main/scala/scala/cli/commands/fix/Fix.scala
Gedochao: Right... I will do a follow-up regarding `.scalafix.conf`, I'll adjust this warning there.
Thanks for noting!

bjaglin:
Not sure if this PR is the right forum to open a discussion (happy to continue wherever is more suited, just let me know!), but I was surprised by this UX change that I found out while reading the release notes, particularly because scalafix [already has a concept built-rules](https://scalacenter.github.io/scalafix/docs/rules/overview.html).

While I share the concern about having several fix commands, I am not sure that users will understand what "built-in rules" refer to and how they can be opted-in/out. As far as I can tell, at the moment, there is no way to opt-out from the _scala-cli_ built-in rule, while all _scalafix_ built-in rules are opt-in (until https://github.com/VirtusLab/scala-cli/issues/3330 at least).

Maybe a way to bring some clarity for users would be to rewrite the _scala-cli_ built-in rule as a _scalafix_ rule (potentially maintained in this repo together with the build), publish it as a regular scalafix external rule but load by default when scalafix is run through scala-cli, allowing users to opt-in/out via `.scalafix.conf` just like any other scalafix rule?

(1) from https://github.com/VirtusLab/scala-cli/pull/2309 looks non-trivial to implement as a Scalafix rule as the Scalafix API is very file-centric, but it should be doable by
- hacking some state persistence in the rule accumulating directive information collected in source files
- creating an empty `project.scala`
- invoke the scalafix CLI with target files explicitly (as positional args) to control the order of evaluation and make sure `project.scala` always gets evaluated last

I am obviously biased as the Scalafix core maintainer, but if this turns out to work without any change in Scalafix (as I hope), we would have another impactful community rule that rule authors can get inspiration from for their own use-case, and if we hit some limitations, we could extend the Scalafix API, which would also benefit rule authors. Maybe that would also be an opportunity to parse/prettyprint scala-cli directives in Scalameta, to reduce the string parsing/building logic here?

tgodzik:
> there is no way to opt-out from the scala-cli built-in rule,

You should be able to do `scala-cli fix --enable-built-in=false .` which should only run scalafix rules. I think we could change the defaults not to run build in ones by default, but we would need to discuss it further. We can for sure rename those flags as it's rather experimental still.

Having a scalafix rules that would do the same would also be quite useful, since it is much more complex I am not sure when we would be able to write it.

bjaglin:
> > there is no way to opt-out from the scala-cli built-in rule,
> 
> You should be able to do `scala-cli fix --enable-built-in=false .` which should only run scalafix rules.

Indeed, since there is only one _scala-cli_ built-in rule at the moment, this flag does the job, but as soon as we have more than one built-in rule, we would likely need a toggle per rule.

> we don't need third-party library that only rely on TASTy (e.g. https://github.com/scalacenter/tasty-query) for type inspection and inference, TASTy as a serialiser is good enough

To start with, we could have a no-op Scalafix rule just for the sake of enabling `.scalafix.conf` and corresponding CLI args to opt-in/out for a consistent UI - scala-cli could still use custom code when detecting the presence of it.

tgodzik:
> Indeed, since there is only one scala-cli built-in rule at the moment, this flag does the job, but as soon as we have more than one built-in rule, we would likely need a toggle per rule.

I don't think we plan on having more and I would prefer to try to use scalafix API next time. If we had more than one rule, then maybe we could migrate to `scala build-fix` command instead :thinking: Or we could run build in rules by name with a more general `--rules` option.

> To start with, we could have a no-op Scalafix rule just for the sake of enabling .scalafix.conf and corresponding CLI args to opt-in/out for a consistent UI - scala-cli could still use custom code when detecting the presence of it.

I think I am a bit lost here. Why would we need it?

bjaglin:
> > To start with, we could have a no-op Scalafix rule just for the sake of enabling .scalafix.conf and corresponding CLI args to opt-in/out for a consistent UI - scala-cli could still use custom code when detecting the presence of it.
> 
> I think I am a bit lost here. Why would we need it?

- Going back to [my original comment](https://github.com/VirtusLab/scala-cli/pull/3400#issuecomment-2628524125), I am concerned about having 2 separate ways of enabling/disabling rules, and the ambiguous name of "built-in", therefore I was suggesting to converge on Scalafix as a distribution/configuration channel, in order using `.scalafix.conf` or the scalafix-specific CLI args for all of them.
- You replied that it's unlikely that the effort of porting the existing _scala-cli_ built-in rule ever gets prioritized.
- As a result, I was suggesting a tradeoff: keep the _scala-cli_ built-in rule logic as-is (i.e. not built against Scalafix APIs), but execute only when a rule of specific name is requested to Scalafix.
  - The actual implementation of that Scalafix rule would be a no-op, just to avoid failing the Scalafix invocation.
  - That way we get a consistent user-facing UI, with the mental model that `scala-cli fix` is a pure wrapper of Scalafix (preloaded with scala-cli-specific rules), potentially with [default rules](https://github.com/VirtusLab/scala-cli/issues/3330) (including the scala-cli one) if none is explicitly requested.

bjaglin:
> If we had more than one rule, then maybe we could migrate to `scala build-fix` command instead

That would definitely reduce the ambiguity. I guess this is a decision that needs to be taken if/when `fix` gets out of experimental mode.

> Or we could run build in rules by name with a more general --rules option.

I think this is where my concern is back: will users understand that the same command can control rules of different type/distribution channels via different CLI options?

tgodzik:
> I think this is where my concern is back: will users understand that the same command can control rules of different type/distribution channels via different CLI options?

Yeah, I guess it's a tough one. Though I wonder if from user perspective it makes any difference and should we make that difference visible or go the other way?

bjaglin:
Thanks for spinning that off to https://github.com/VirtusLab/scala-cli/discussions/3499 @Gedochao, let's follow up there 🙏

Issue #3258, opened 2024-11-04
Implement a mechanism for deprecation of features


Gedochao:
**Describe what needs to be done and why**
We need a mechanism to deprecate individual features of the runner/CLI, similar to the one we have for features which are experimental. 
It should become easy to tag or annotate individual features as deprecated, label them as such in our docs and log warnings.

Features which should be deprecatable:
- [ ] command line options
  - [ ] individual command line option aliases
- [ ] using directives
  - [ ] individual using directive aliases
- [ ] sub-commands
  - [ ] individual sub-command aliases
- [ ] global config keys

Additionally, we need:
- [ ] a warning suppression command line option, which will silence deprecation warnings altogether
- [ ] a corresponding global config key for suppression of deprecation warnings
- [ ] all deprecation warnings should be gathered into a single log (like is the case for experimental features), rather than each being logged separately

**Is your feature request related to a past ticket or discussion?**
- https://github.com/VirtusLab/scala-cli/pull/3257
- https://github.com/VirtusLab/scala-cli/pull/2622

**Describe alternatives you've considered**
What we have in place are rough and/or partial hacks, we need a proper solution.

PR #3406, opened 2025-01-09
Add ways to suppress deprecation warnings

Gedochao:
- covers a subset of the requirements in #3258 
- adds a new `config` key: `suppress-warning.deprecated-features`, which lets globally suppress deprecated feature warnings
```bash
scala-cli config suppress-warning.deprecated-features true
```
- adds a new command line option: `--suppress-deprecated-warnings`, which does the same
```bash
scala-cli project-with-deprecated-stuff --suppress-deprecated-warnings
```
- covers existing ways of deprecating features, including command line options and aliases, deprecation warnings from the using directive parser and deprecated directives
  - others should be covered alongside progress in #3258

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCommand.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCommand.scala
@@ -29,6 +29,9 @@ abstract class PgpCommand[T](implicit myParser: Parser[T], help: Help[T])
   override def shouldSuppressExperimentalFeatureWarnings: Boolean =
     false // TODO add handling for scala-cli-signing
 
+  override def shouldSuppressDeprecatedFeatureWarnings: Boolean =
+    false // TODO add handling for scala-cli-signing
+
   override def logger: Logger = CliLogger.default // TODO add handling for scala-cli-signing
 
   override def hidden = true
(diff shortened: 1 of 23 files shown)


modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCommand.scala
```diff
@@ -29,6 +29,9 @@ abstract class PgpCommand[T](implicit myParser: Parser[T], help: Help[T])
   override def shouldSuppressExperimentalFeatureWarnings: Boolean =
     false // TODO add handling for scala-cli-signing
 
+  override def shouldSuppressDeprecatedFeatureWarnings: Boolean =
+    false // TODO add handling for scala-cli-signing
```
tgodzik, line 33: Do we have an issue for `add handling for scala-cli-signing` ?

tgodzik (approved):

modules/cli/src/main/scala/scala/cli/commands/pgp/PgpCommand.scala
Gedochao: We now do:
- https://github.com/VirtusLab/scala-cli/issues/3417

PR #3424, opened 2025-01-15
Add CLI flag & directive for clang/LLVM target triples when using Scala Native

### LINKED ISSUE #2766: Adding CLI option for cross-compiling on Scala Native

Scala Native has its cross-compiling feature that is came from clang/LLVM.

https://scala-native.org/en/stable/user/sbt.html#cross-compilation

```scala
nativeConfig ~= { _.withTargetTriple("x86_64-apple-macosx10.14.0") }
```

Though Scala CLI has some CLI options that manipulates Scala Native, there is no option to handle with target triple.

(Scala CLI have `--native-target`, but it sets artifact type e.g. application / dynamic lib / static lib)

Add `--native-target-triple` to provide target triple info for Scala Native (and clang/LLVM)

We can convert `foobar.scala.sc` into sbt project, and manually add Scala Native configurations.

**Additional context**

I tried building native binary via Docker Image (virtuslab/scala-cli), and that's quite easy (I wrote build stage, and copied binary to host or another thin stage). If we have cross-compiling feature, that's awesome build infrastructure for small tool (like Go, Rust).

bynux-gh:
This commit is meant to resolve issue #2766 by adding both a directive:
```scala
//> using nativeTargetTriple arch-mnfctr-opsys
```
and a command line option (`--native-target-triple`) that can successfully pass the target triple to Scala Native for cross-compiling projects.

As previously mentioned, this is a *very* experimental option with limited reach. Use with caution, and don't hesitate to provide suggestions for a better approach :)

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
@@ -32,6 +32,11 @@ final case class ScalaNativeOptions(
     nativeGc: Option[String] = None,
 
   @Group(HelpGroup.ScalaNative.toString)
+  @HelpMessage("Set a target triple to which Scala Native can cross-compile")
+  @Tag(tags.should)
+    nativeTargetTriple: Option[String] = None,
+
+  @Group(HelpGroup.ScalaNative.toString)
   @HelpMessage("Path to the Clang command")
   @Tag(tags.implementation)
     nativeClang: Option[String] = None,
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
@@ -33,7 +33,7 @@ final case class ScalaNativeOptions(
 
   @Group(HelpGroup.ScalaNative.toString)
   @HelpMessage("Set a target triple to which Scala Native can cross-compile")
-  @Tag(tags.should)
+  @Tag(tags.experimental)
     nativeTargetTriple: Option[String] = None,
 
   @Group(HelpGroup.ScalaNative.toString)
--- a/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
@@ -282,6 +282,7 @@ final case class SharedOptions(
       modeStr = nativeMode,
       ltoStr = nativeLto,
       gcStr = nativeGc,
+      targetTripleStr = nativeTargetTriple,
       clang = nativeClang,
       clangpp = nativeClangpp,
       linkingOptions = nativeLinking,
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaNative.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalaNative.scala
@@ -16,6 +16,8 @@ import scala.cli.commands.SpecificationLevel
     |
     |`//> using nativeLto` _value_
     |
+    |`//> using nativeTargetTriple` _value_
+    |
     |`//> using nativeVersion` _value_
     |
     |`//> using nativeCompile` _value1_ _value2_ …
@@ -38,6 +40,7 @@ final case class ScalaNative(
   nativeGc: Option[String] = None,
   nativeMode: Option[String] = None,
   nativeLto: Option[String] = None,
+  nativeTargetTriple: Option[String] = None,
   nativeVersion: Option[String] = None,
   nativeCompile: List[String] = Nil,
   nativeLinking: List[String] = Nil,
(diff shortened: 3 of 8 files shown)


modules/cli/src/main/scala/scala/cli/commands/shared/ScalaNativeOptions.scala
```diff
@@ -31,6 +31,11 @@ final case class ScalaNativeOptions(
   @Tag(tags.should)
     nativeGc: Option[String] = None,
 
+  @Group(HelpGroup.ScalaNative.toString)
+  @HelpMessage("Set a target triple to which Scala Native can cross-compile")
+  @Tag(tags.should)
```
Gedochao:
```suggestion
  @Tag(tags.experimental)
```

Gedochao (changes requested):
We need integration tests, if this was to be merged. The right place would be one of:
- https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/RunScalaNativeTestDefinitions.scala
- https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala

Gedochao:
cc @WojciechMazur

WojciechMazur:
The scala-native-lci integration itself LGTM, however as it was previously mentioned we need to have first a working integration test to prove that it can be actually usable. I left my concerns and mentioned problems with this flag in the comment https://github.com/VirtusLab/scala-cli/issues/2766#issuecomment-2592667517 
We should not merge it if we cannot guarantee it is somehow usable and using scala-cli we can actually create a working, cross-compiled binary and execute it.

tgodzik:
@Gedochao do you think it would be good if we merged it under experimental? We don't actually provide any guarantees at that point.

bynux-gh:
You all will be very pleased to hear that **I got cross-compilation working using this setting**! I was able to compile an AArch64 copy of a simple single-file Scala program using an x86_64 machine, then run it in qemu with minimal issues.

The last thing I hadn't accounted for is that Scala-CLI automatically adds system libraries, which is *normally* the desired behavior, but not when they're being added to a redefined sysroot. So, in the latest commit I have it overriding that behavior when a target triple is defined.

After making sure my rootfs was properly configured, I was able to point clang at it with
```sh
--native-linking "--sysroot=/usr/aarch64-linux-gnu"                                      
```

Everything else is handled automatically by `clang`/LLVM when the target triple is handed over, so in total the command I needed was:

```sh
scala-cli package Test.scala --native --native-target-triple aarch64-linux-gnu --native-linking "--sysroot=/usr/aarch64-linux-gnu" --destination $OUTDIR
```

The resulting output can be run with `qemu-aarch64` as long as it is also told where to find its libraries. It also runs without any additional input on AArch64 Arch Linux (using proot on Termux for Android on my Galaxy Note 9).

Someone who already does cross-compilation on the regular and has sysroots set up will have an easier time of making use of this for small shell programs!

Needless to say, integration tests will still be needed, and the feature should remain experimental while we only have confirmation of one particular workflow working as planned, but it's a *massive* step in the right direction. Next challenge will be using `mingw` to see if I can cross-compile Scala code to Windows :)

### TL;DR: *I have it working on two of my machines*, so we just need integration tests to demonstrate reproducible results. I was able to compile a simple one-file Scala program into a native binary that both the qemu emulator *and* a Linux installation on my Android phone can both execute without issues.

Gedochao:
@bynux-gh do you need help with setting up the integration test?

bynux-gh:
> @bynux-gh do you need help with setting up the integration test?

Hello hello, yes! I said something in the Scala Discord server but meant to bring it over to this thread. I haven't had much use for GitHub's CI system until now, so I'm not familiar with how it functions or how it integrates with test code. I could definitely use assistance setting that up and making sure the test code "tells" GitHub's systems what we want to hear.

Gedochao:
> Hello hello, yes! I said something in the Scala Discord server but meant to bring it over to this thread. I haven't had much use for GitHub's CI system until now, so I'm not familiar with how it functions or how it integrates with test code. I could definitely use assistance setting that up and making sure the test code "tells" GitHub's systems what we want to hear.

@bynux-gh Okay, so let's not worry about the CI too much for now. 
Let's start with adding an integration test which tests the feature and passes on your local machine, and then we'll try to figure out how to make that pass on the CI, if there are problems.

As mentioned before, the test would probably be best added to one of:
- https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/RunScalaNativeTestDefinitions.scala
- https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala

More info on our integration tests can be found around here: https://github.com/VirtusLab/scala-cli/blob/main/DEV.md#run-integration-tests-with-the-jvm-launcher

As a side note, the CI is currently failing on reference docs and formatting, the commands for making those pass can be found here: https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md#rules-for-a-well-formed-pr

Let me know if you get stuck.

Gedochao:
@bynux-gh Is this still relevant? Do you mean to keep working on this?

Gedochao:

PR #3460, opened 2025-01-24
Build Linux arm64 launchers on a GitHub runner

Gedochao:
Linux arm64 runners are in preview, let's give them a go!
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


Gedochao:
```
Generation complete.
/data/run-native-image.sh: line 10: /data/cs: cannot execute binary file: Exec format error
/data/run-native-image.sh: line 11: gu: command not found
```
🤔

Gedochao:
Likely caused by this line:

https://github.com/alexarchambault/mill-native-image/blob/79a73bde0cf8d06b9b2560b96b3a35f5c8962a47/plugin/src/io/github/alexarchambault/millnativeimage/NativeImage.scala#L365

Looks similar to https://github.com/alexarchambault/mill-native-image/issues/83, but I suppose we're treading new ground for a Linux arm64 here...

But why doesn't `cs` work? 🤨

PR #3466, opened 2025-01-29
Update Scala 3 LTS to 3.3.5

Gedochao:
https://github.com/scala/scala3/issues/22465


tgodzik (approved):

Gedochao:
Added a band-aid solution for  downgrading LTS to 3.3.4 when used with the default version of Ammonite.
The downgrading logic doesn't fully work as expected, as per https://github.com/VirtusLab/scala-cli/issues/3467.
We may soon stop having to rely on Ammonite for our REPL tests as well, when https://github.com/scala/scala3/issues/21242 gets delivered (likely in Scala 3.7.x), so I'm not sure if this is worth more effort at this point.

Should be ready for re-review.

tgodzik (approved):

PR #2541, opened 2023-11-14
Exclude conflicting dependencies

MaciejG604:
Fixes #2421

Issue #2421, opened 2023-09-23
scala-cli depends on multiple versions of artifacts with conflicting cross-version suffixes


armanbilge:
**Version(s)**
1.0.4

**Describe the bug**
org.virtuslab.scala-cli:cli_3:1.0.4 depends on multiple versions of artifacts with conflicting cross-version suffixes:
```
com.lihaoyi:geny _2.13, _3
org.scala-lang.modules:scala-collection-compat _2.13, _3
com.github.plokhotnyuk.jsoniter-scala:jsoniter-scala-core _3, _2.13
com.lihaoyi:sourcecode _3, _2.13
ai.kien:python-native-libs _3, _2.13
com.lihaoyi:os-lib _2.13, _3
```

**To Reproduce**
Navigate to:
https://central.sonatype.com/artifact/org.virtuslab.scala-cli/cli_3/dependencies
And select "Dependency Type: All". 

**Expected behaviour**

You should not depend on multiple versions of the same artifact with conflicting cross-version suffixes. When multiple versions of the same class are on the classpath the behavior is undefined / non-deterministic.

cc @toniogela

TonioGela:
I'll add that using via scala-cli itself won't trigger the problem:

```scala
//> using scala 3.3.1
//> using dep org.virtuslab.scala-cli::cli::1.0.4

import scala.cli.ScalaCli

object testApp extends App:
  ScalaCli.main(Array("compile", "foo.scala"))
```

while adding the dependency in an sbt project like `"org.virtuslab.scala-cli" %% "cli" % "1.0.4"` won't work. A temporary fix is adding exclusion rules:

```scala
"org.virtuslab.scala-cli" %% "cli" % "1.0.4" excludeAll (
  ExclusionRule("com.lihaoyi:geny_2.13"),
  ExclusionRule("org.scala-lang.modules", "scala-collection-compat_2.13"),
  ExclusionRule(
    "com.github.plokhotnyuk.jsoniter-scala",
    "jsoniter-scala-core_2.13"
  ),
  ExclusionRule("com.lihaoyi", "sourcecode_2.13"),
  ExclusionRule("ai.kien", "python-native-libs_2.13"),
  ExclusionRule("com.lihaoyi", "os-lib_2.13")
)
```

lwronski:
It seems that Mill miss version conflict detection. I can't find a plugin that allows users to set eviction rules, similar to what's available in SBT.

PR #3472, opened 2025-01-31
Remove conflicting Scala 2.13 `io.get-coursier:dependency` dependency & add a CI check

### LINKED ISSUE #3453: scala-cli depends on multiple versions of artifacts with conflicting cross-version suffixes
**Version(s)**
1.6.1

**Describe the bug**
`org.virtuslab.scala-cli:cli_3:1.6.1` depends on multiple versions of artifacts with conflicting cross-version suffixes:
```
[error] Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/workspace/sandbox/"), "sandbox"):
[error]    io.get-coursier:dependency _3, _2.13
```

**To Reproduce**
Navigate to:
https://central.sonatype.com/artifact/org.virtuslab.scala-cli/cli_3/dependencies
And select "Dependency Type: All". 

**Expected behaviour**

You should not depend on multiple versions of the same artifact with conflicting cross-version suffixes. When multiple versions of the same class are on the classpath the behavior is undefined / non-deterministic.

This is a recurring problem:
- https://github.com/VirtusLab/scala-cli/issues/2421

Gedochao:
Fixes #3453

### DIFF
--- a/.github/scripts/check-cross-version-deps.sc
+++ b/.github/scripts/check-cross-version-deps.sc
@@ -1,0 +1,57 @@
+#!/usr/bin/env -S scala-cli shebang
+//> using scala 3
+//> using toolkit default
+
+val modules =
+  os.proc(os.pwd / "mill", "-i", "resolve", "__[]")
+    .call(cwd = os.pwd)
+    .out
+    .lines()
+
+for { module <- modules } {
+  println(s"Checking for $module...")
+  val depRegex            = "[│└─\\S\\s]+\\s([\\w.-]+):([\\w.-]+):([\\w\\s\\S.-]+)".r
+  val scalaDepSuffixRegex = "^(.+?)(_[23](?:\\.\\d{2})?)?$".r
+  val deps = os.proc(os.pwd / "mill", "-i", s"$module.ivyDepsTree")
+    .call(cwd = os.pwd)
+    .out
+    .lines()
+    .map { case depRegex(org, name, depVersion) => (org, name, depVersion) }
+  val scalaVersionsByOrgAndName = deps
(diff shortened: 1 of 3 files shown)


.github/scripts/check-cross-version-deps.sc
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env -S scala-cli shebang
```
tgodzik, line 1: There is no actual mill taks to do it? :O

tgodzik (approved):
Would prefer if mill could provide the functionality :/

.github/scripts/check-cross-version-deps.sc
Gedochao: Didn't find anything... there's `ivyDepsTree --whatDependsOn` to make a specific check, but to do a general one, I didn't find a better way.

And it seems people looked for one before me:
- https://github.com/VirtusLab/scala-cli/issues/2421#issuecomment-1733787851
- https://github.com/VirtusLab/scala-cli/pull/2541#issuecomment-1812690252

Gedochao: I guess another way would be to set up a dedicated Mill plugin 😅

tgodzik: An alternative is to have a small sbt test project that has the dependency on cli and we run it for latest nightly

Gedochao: I'd rather be able to run the check on a PR, though.
The script is simple enough, too, I think.

tgodzik: It has regexes :P that is never simple

Gedochao: 🤷‍♂️

PR #3479, opened 2025-02-05
improvement: Detect objects with main class in scripts

### LINKED ISSUE #3473: Need an option for writing scripts that are compatible across all scala3 versions
This is only concerned with a way to migrate `scala 3` versions to `scala 3.5+` and above.   An alternate title for this might be:
#### Provide a way to execute scripts with ".sc" extension as ".scala" files

Migrating a project with many legacy scripts requires that when bugs are encountered in production, a script can be reverted by editing the `hash-bang` line, and without being renamed.    Also, some scripts cannot be renamed, even after the migration period, so the `scala_legacy` script isn't a universal option.

The need is for a way to _optionally_ treat some `.sc` scripts as if they have the `.scala` filename extension.

Prior to `scala 3.5` all scripts require a main method, so a migration format that can easily switch between old and new script semantics requires a main method.

The primary challenge is that after `3.5`, the main method is not automatically called unless the file is renamed.  In other words, the main method is never called if a script has the `.sc` extension.

Example hypothetical migration format (almost adequate, but has non-portable element on line 5 `Main.main(args)`.
```scala
//> using scala 3.4.3
//> using dep "com.lihaoyi::os-lib::0.11.3"
import os._
Main.main(args) // required by `3.5+`, rejected by earlier scala versions
object Main {
    val root1 = os.root("C:\\")
    printf("cRoot: %s\n", root1)
}
```

Possible approaches to specifying the proposed `.scala` mode option:

- `scala-cli` command line, similar to `--main-class` (maybe `--main-method`)
- system property ( "scala.dot_scala_extension")
- alternate supported extension such as `.sc_` or similar, treated as if `.scala`
- comment `//> using scala-extension-semantics`

Adding support for a new filename extension `.sc_` turned out to be pretty simple, and seems to work nicely, although it doesn't address the case of scripts that cannot easily be renamed.  Here's the implementation:
```diff
# git diff
diff --git a/modules/build/src/main/scala/scala/build/input/Inputs.scala b/modules/build/src/main/scala/scala/build/input/Inputs.scala
index 8e93d1463..4d5b44e55 100644
--- a/modules/build/src/main/scala/scala/build/input/Inputs.scala
+++ b/modules/build/src/main/scala/scala/build/input/Inputs.scala
@@ -275,8 +275,8 @@ object Inputs {
         }
         else if path.last == Constants.projectFileName then
           Right(Seq(ProjectScalaFile(dir, subPath)))
+        else if arg.endsWith(".scala") || arg.endsWith(".sc_") then Right(Seq(SourceScalaFile(dir, subPath)))
         else if arg.endsWith(".sc") then Right(Seq(Script(dir, subPath, Some(arg))))
-        else if arg.endsWith(".scala") then Right(Seq(SourceScalaFile(dir, subPath)))
         else if arg.endsWith(".java") then Right(Seq(JavaFile(dir, subPath)))
         else if arg.endsWith(".jar") then Right(Seq(JarFile(dir, subPath)))
         else if arg.endsWith(".c") || arg.endsWith(".h") then Right(Seq(CFile(dir, subPath)))
```

Experiments were done based on various wrappers around `scala-cli`, the two most successful described here.

1. create an on-the-fly symlink in the script parent directory, but with `.scala` extension, which is deleted-on-exit
2. wrapper to pipe the script  to scala-cli STDIN: (similar to  `echo 'println("Hello, Scala!")' | scala-cli -`)

Approach 1 works well most of the time, but after a crash, symlinks tend to be left behind.

The pipe-to-stdin approach is equivalent to the following conversion:
```
jsrc/sum.sc 1 2 3 
# the equivalent command line when piped to `scala-cli` <stdin>:
tail -n +2 jsrc/sum.sc | scala-cli run -Dscript.path=jsrc/sum.sc @/opt/ue3cps _.scala -- 1 2 3
``` 
It works much of the time, but isn't viable for interactive scripts, or those that read from `stdin`.
The minimum startup time with this approach is about 10.0 seconds,  and if `procPipe.sc` is compiled, about 8.0 seconds.

For comparison, the equivalent `.scala` version of the script starts up in less than 4 seconds, and under 1 second after it's been compiled.

**Additional context**
I work with projects having 1400+ legacy scripts, only about 20% have been migrated to scala 3.6.3.

tgodzik:
Should help with https://github.com/VirtusLab/scala-cli/issues/3473

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
@@ -1,6 +1,6 @@
 package scala.build.internal
 
-case object AppCodeWrapper extends CodeWrapper {
+case class AppCodeWrapper(scalaVersion: String, log: String => Unit) extends CodeWrapper {
   override def mainClassObject(className: Name) = className
 
   def apply(
@@ -12,13 +12,19 @@ case object AppCodeWrapper extends CodeWrapper {
   ) = {
     val wrapperObjectName = indexedWrapperName.backticked
 
+    val mainObject = WrapperUtils.mainObjectInScript(scalaVersion, code)
+    val invokeMain = mainObject match
+      case WrapperUtils.ScriptMainMethod.Exists(name) => s"\n$name.main(args)"
+      case otherwise =>
+        otherwise.warningMessage.foreach(log)
+        ""
     val packageDirective =
       if (pkgName.isEmpty) "" else s"package ${AmmUtil.encodeScalaSourcePath(pkgName)}" + "\n"
     val top = AmmUtil.normalizeNewlines(
       s"""$packageDirective
          |
          |object $wrapperObjectName extends App {
-         |val scriptPath = \"\"\"$scriptPath\"\"\"
+         |val scriptPath = \"\"\"$scriptPath\"\"\"$invokeMain
          |""".stripMargin
     )
     val bottom = AmmUtil.normalizeNewlines(
--- a/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
@@ -5,7 +5,7 @@ package scala.build.internal
   * running interconnected scripts using Scala CLI <br> <br> Incompatible with Scala 2 - it uses
   * Scala 3 feature 'export'<br> Incompatible with native JS members - the wrapper is a class
   */
-case object ClassCodeWrapper extends CodeWrapper {
+case class ClassCodeWrapper(scalaVersion: String, log: String => Unit) extends CodeWrapper {
 
   override def mainClassObject(className: Name): Name =
     Name(className.raw ++ "_sc")
@@ -16,8 +16,16 @@ case object ClassCodeWrapper extends CodeWrapper {
     extraCode: String,
     scriptPath: String
   ) = {
+
+    val mainObject = WrapperUtils.mainObjectInScript(scalaVersion, code)
+    val mainInvocation = mainObject match
+      case WrapperUtils.ScriptMainMethod.Exists(name) => s"script.$name.main(args)"
+      case otherwise =>
+        otherwise.warningMessage.foreach(log)
+        s"val _ = script.hashCode()"
+
     val name             = mainClassObject(indexedWrapperName).backticked
-    val wrapperClassName = Name(indexedWrapperName.raw ++ "$_").backticked
+    val wrapperClassName = scala.build.internal.Name(indexedWrapperName.raw ++ "$_").backticked
     val mainObjectCode =
       AmmUtil.normalizeNewlines(s"""|object $name {
                                     |  private var args$$opt0 = Option.empty[Array[String]]
@@ -33,7 +41,7 @@ case object ClassCodeWrapper extends CodeWrapper {
                                     |
                                     |  def main(args: Array[String]): Unit = {
                                     |    args$$set(args)
-                                    |    val _ = script.hashCode() // hashCode to clear scalac warning about pure expression in statement position
+                                    |    $mainInvocation // hashCode to clear scalac warning about pure expression in statement position
                                     |  }
                                     |}
                                     |
--- a/modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
+++ b/modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
@@ -8,16 +8,22 @@ object WrapperUtils {
     case Exists(name: String)
     case Multiple(names: Seq[String])
     case ToplevelStatsPresent
+    case ToplevelStatsWithMultiple(names: Seq[String])
     case NoMain
 
-    def warningMessage: Option[String] =
+    def warningMessage: List[String] =
       this match
         case ScriptMainMethod.Multiple(names) =>
-          Some(WarningMessages.multipleMainObjectsInScript(names))
-        case ScriptMainMethod.ToplevelStatsPresent => Some(
+          List(WarningMessages.multipleMainObjectsInScript(names))
+        case ScriptMainMethod.ToplevelStatsPresent => List(
             WarningMessages.mixedToplvelAndObjectInScript
           )
-        case _ => None
+        case ToplevelStatsWithMultiple(names) =>
+          List(
+            WarningMessages.multipleMainObjectsInScript(names),
+            WarningMessages.mixedToplvelAndObjectInScript
+          )
+        case _ => Nil
 
   def mainObjectInScript(scalaVersion: String, code: String): ScriptMainMethod =
     import scala.meta.*
@@ -74,7 +80,9 @@ object WrapperUtils {
       case head :: Nil =>
         ScriptMainMethod.ToplevelStatsPresent
       case Nil => ScriptMainMethod.NoMain
-      case seq =>
+      case seq if noToplevelStatements =>
         ScriptMainMethod.Multiple(seq)
+      case seq =>
+        ScriptMainMethod.ToplevelStatsWithMultiple(seq)
 
 }
--- a/modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
+++ b/modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
@@ -1,0 +1,80 @@
+package scala.build.internal
+
+import scala.build.internal.util.WarningMessages
+
+object WrapperUtils {
+
+  enum ScriptMainMethod:
+    case Exists(name: String)
+    case Multiple(names: Seq[String])
+    case ToplevelStatsPresent
+    case NoMain
+
+    def warningMessage: Option[String] =
+      this match
+        case ScriptMainMethod.Multiple(names) =>
+          Some(WarningMessages.multipleMainObjectsInScript(names))
+        case ScriptMainMethod.ToplevelStatsPresent => Some(
+            WarningMessages.mixedToplvelAndObjectInScript
+          )
+        case _ => None
+
+  def mainObjectInScript(scalaVersion: String, code: String): ScriptMainMethod =
+    import scala.meta.*
+
+    val scriptDialect =
+      if scalaVersion.startsWith("3") then dialects.Scala3Future else dialects.Scala213Source3
+
+    given Dialect = scriptDialect.withAllowToplevelStatements(true).withAllowToplevelTerms(true)
+    val parsedCode = code.parse[Source] match
+      case Parsed.Success(Source(stats)) => stats
+      case _                             => Nil
+
+    // Check if there is a main function defined inside an object
+    def checkSignature(defn: Defn.Def) =
+      defn.paramClauseGroups match
+        case List(Member.ParamClauseGroup(
+              Type.ParamClause(Nil),
+              List(Term.ParamClause(
+                List(Term.Param(
+                  Nil,
+                  _: Term.Name,
+                  Some(Type.Apply.After_4_6_0(
+                    Type.Name("Array"),
+                    Type.ArgClause(List(Type.Name("String")))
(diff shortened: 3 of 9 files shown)


modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+            )) => true
+        case _ => false
+    parsedCode.collect {
+      case Defn.Object(_, objName, templ) =>
+        templ.body.stats.find {
+          case defn: Defn.Def =>
+            defn.name.value == "main" && checkSignature(defn)
+          case _ => false
+        }.map(_ => objName.value)
+    }.flatten.headOption
```
Gedochao: IMO it's quite deceptive that only `.headOption` is called here.
If anything, we should return a `Seq[String]` here, and later warn which is used // allow to override which is picked.

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
```diff
@@ -12,13 +12,18 @@ case object AppCodeWrapper extends CodeWrapper {
...
+      case Some(name) => s"\n$name.main(args)"
+    }
     val packageDirective =
       if (pkgName.isEmpty) "" else s"package ${AmmUtil.encodeScalaSourcePath(pkgName)}" + "\n"
     val top = AmmUtil.normalizeNewlines(
       s"""$packageDirective
          |
          |object $wrapperObjectName extends App {
-         |val scriptPath = \"\"\"$scriptPath\"\"\"
+         |val scriptPath = \"\"\"$scriptPath\"\"\"$invokeMain
```
Gedochao, line 27: Ok, so while I actually like this hack very much, it introduces some unexpected side-effects which might need to be addressed before we proceed.

Note how both the main method from an object and top-level definitions are called one after another.
So for:
```scala
// legacy.sc
object Main {
  def main(args: Array[String]): Unit = println("1")
}
object AnotherMain {
  def main(args: Array[String]): Unit = println("2")
}
println("3")
```
the result will be
```bash
scala-cli legacy.sc
Compiling project (Scala 3.6.3, JVM (23))
Compiled project (Scala 3.6.3, JVM (23))
3
1
```

Any top level definitions are called first, then followed by the first (and only the first) method from a main object.

Gedochao: Ofc, in an ideal world, the main method from each object should be treated on par with the main method of the script itself. That being said, I know that this is a hack to support legacy use cases, rather than a cure-all. So let's try to get it down to shape.

Some suggestions to improve this:
- I actually think the main method should be called first, followed by the the output of top-level inputs
- if a main method is called like this, we should print a deprecation warning (suppressable as any other deprecated warning), explaining that both the main method as well as the top-level calls will be executed
- if there are multiple main methods (as in the example, with `Main` and `AnotherMain`), we should have a warning about the first one being picked by default, and suggesting how `--main-class` would allow to pick one of them
- (this one may be a tad hard, it's ok to extract it to an issue) `--main-class`/`//> using mainClass` should allow to switch between the multiple main methods
  - `compile`'s `--list-main-classes` should list these main methods, too
  - `--interactive` should allow to pick them

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
```diff
@@ -1,11 +1,13 @@
 package scala.build.internal
 
+// import scala.meta.parsers.Parsed
```
Gedochao: leftover?

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -69,6 +69,42 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+            |""".stripMargin
+    )
+    inputs.fromRoot { root =>
+      val output = os.proc(TestUtil.cli, extraOptions, "--power", "main.sc").call(cwd =
+        root
+      ).out.trim()
+      expect(output == message)
+    }
+  }
+
```
Gedochao, line 126:
```suggestion

test("main.sc has both an object with a main method and an object wrapper as well as top-level calls") {
    val message1 = "Hello"
    val message2 = "Another hello"
    val inputs = TestInputs(
      os.rel / "main.sc" ->
        s"""|//> using objectWrapper
            |object Main {
            |  def main(args: Array[String]): Unit = println("$message1")
            |}
            |println("$message2")
            |""".stripMargin
    )
    inputs.fromRoot { root =>
      val output = os.proc(TestUtil.cli, extraOptions, "--power", "main.sc")
        .call(cwd = root).out.trim()
      expect(output ==
        s"""$message1
           |$message2""".stripMargin)
    }
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -69,6 +69,42 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+            |""".stripMargin
+    )
+    inputs.fromRoot { root =>
+      val output = os.proc(TestUtil.cli, extraOptions, "main.sc").call(cwd =
+        root
+      ).out.trim()
+      expect(output == message)
+    }
+  }
+
```
Gedochao, line 108:
```suggestion

test("main.sc has both an object with a main method as well as top-level definitions") {
    val message1 = "Hello"
    val message2 = "Another hello"
    val inputs = TestInputs(
      os.rel / "main.sc" ->
        s"""|object Main {
            |  def main(args: Array[String]): Unit = println("$message1")
            |}
            |println("$message2")
            |""".stripMargin
    )
    inputs.fromRoot { root =>
      val output = os.proc(TestUtil.cli, extraOptions, "main.sc").call(cwd =
        root
      ).out.trim()
      expect(output ==
        s"""$message1
           |$message2""".stripMargin)
    }
  }
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -69,6 +69,42 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
     }
   }
 
+  test("main.sc has object") {
```
```suggestion
  test("main.sc has an object with a main method") {
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -69,6 +69,42 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+    )
+    inputs.fromRoot { root =>
+      val output = os.proc(TestUtil.cli, extraOptions, "main.sc").call(cwd =
+        root
+      ).out.trim()
+      expect(output == message)
+    }
+  }
+
+  test("main.sc has object with object wrapper") {
```
```suggestion
  test("main.sc has an object with a main method and an object wrapper") {
```

Gedochao (changes requested):
I like the solution itself, especially as it solves a lot of use cases, even though it doesn't solve the problem fully. I left some comments on how I think we could improve this.

In the context of https://github.com/VirtusLab/scala-cli/issues/3473, it's perhaps good to point out that this only fixes things for objects with a main method, but not for `@main` annotated methods which are top-level.

cc @philwalk, in case you have any thoughts on this.

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
tgodzik: Maybe it makes sense instead to suggest to change the extension to .scala in that case? I wonder if adding main class wouldn't be too much effort compared to the gains here. I will check that out

philwalk:
> this only fixes things for objects with a main method, but not for `@main` annotated methods which are top-level.

The object.main() approach provides a migration format that works with all scala versions, so I don't think supporting  `@main` annotated methods is necessary.

modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
tgodzik: Switched it a bit to return an enum

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
tgodzik: In the end I disabled this case, if there is anything toplevel it will be treated as a script

modules/build/src/main/scala/scala/build/internal/ClassCodeWrapper.scala
tgodzik: removed

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
tgodzik: added

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
tgodzik: added

modules/build/src/main/scala/scala/build/internal/util/WarningMessages.scala
```diff
@@ -105,6 +105,12 @@ object WarningMessages {
   val offlineModeBloopJvmNotFound =
     "Offline mode is ON and a JVM for Bloop could not be fetched from the local cache, using scalac as fallback"
 
+  def multipleMainObjectsInScript(names: Seq[String]) =
+    s"Only single main is allowed within scripts and multiple main classes were found in the script: ${names.mkString(", ")}"
```
Gedochao:
```suggestion
    s"Only a single main is allowed within scripts. Multiple main classes were found in the script: ${names.mkString(", ")}"
```

modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
```diff
@@ -0,0 +1,80 @@
...
+  enum ScriptMainMethod:
+    case Exists(name: String)
+    case Multiple(names: Seq[String])
+    case ToplevelStatsPresent
+    case NoMain
+
+    def warningMessage: Option[String] =
+      this match
+        case ScriptMainMethod.Multiple(names) =>
+          Some(WarningMessages.multipleMainObjectsInScript(names))
```
Gedochao: `WarningMessages.mixedToplvelAndObjectInScript` should also be logged here (as well as `WarningMessages.multipleMainObjectsInScript(names)`, as it is right now).

Currently, for
```scala
// legacy.sc
object Main {
  def main(args: Array[String]): Unit = println("1")
}
object AnotherMain {
  def main(args: Array[String]): Unit = println("2")
}
println("3")
```
I get
```bash
scala-cli-prototype example.sc
# [warn]  Only single main is allowed within scripts and multiple main classes were found in the script: Main, AnotherMain
# Compiling project (Scala 3.6.3, JVM (23))
# Compiled project (Scala 3.6.3, JVM (23))
# 3
```
So I get the warning about multiple main classes being detected (while only one is being supported), but no explanation why top-level declarations are being run instead.

I'd expect:
```bash
scala-cli-prototype example.sc
# [warn]  Only single main is allowed within scripts and multiple main classes were found in the script: Main, AnotherMain
# [warn]  Script contains objects with main methods and top-level statements, only the latter will be run.
# Compiling project (Scala 3.6.3, JVM (23))
# Compiled project (Scala 3.6.3, JVM (23))
# 3
```

Gedochao (approved):
LGTM, just commented on some warning improvements.
otherwise should be good to merge.

modules/build/src/main/scala/scala/build/internal/WrapperUtils.scala
tgodzik: Fixed!

PR #3502, opened 2025-02-17
Add support for running a main method from the test scope

### LINKED ISSUE #2728: There is no command to run a main method in test scope

Currently the only way to run a main method in test scope is through the IDE (e.g. Metals code lens for `run|debug`). It would be good to have a programmatic way to do this from the `scala` command.

The `run` command should accept the `--test` flag, like you can do in the `compile` command. Then this **does not** try to invoke test frameworks, rather scans for main methods in test scope, in the same manner that `run` currently looks for main methods in the main scope.

I guess this should filter out any main methods possibly included from main scope, but it is not a strong requirement.

N/A

**Additional context**
Here is a forum request for this feature (equivalent of `Test/runMain` in sbt): https://users.scala-lang.org/t/getting-started-with-scala-cli/9788

Gedochao:
Closes https://github.com/VirtusLab/scala-cli/issues/2728

Tested to work with:
- JVM
- Scala.js / `--js`
- Scala Native / `--native`
- Spark / `--spark`
- Spark stand / `--spark-standalone`
- Hadoop / `--hadoop`


bishabosha:
love to see it!

modules/integration/src/test/scala/scala/cli/integration/RunWithWatchTestDefinitions.scala
```diff
@@ -89,6 +89,51 @@ trait RunWithWatchTestDefinitions { _: RunTestDefinitions =>
       }
   }
 
+  for {
+    (platformDescription, platformOpts) <- Seq(
+      "JVM"    -> Nil,
+      "JS"     -> Seq("--js"),
+      "Native" -> Seq("--native")
+    )
+    if !Properties.isMac || !TestUtil.isCI // TODO make this pass reliably on Mac CI
```
tgodzik: Maybe link to the issue about improving tests?

tgodzik (approved):
Looks good, I see it's mostly switching to lists etc.

modules/integration/src/test/scala/scala/cli/integration/RunWithWatchTestDefinitions.scala
Gedochao: done

PR #3503, opened 2025-02-17
fix: Don't compile docs if there is no need

ghostbuster91:
This PR adds a little optimization to the publish commands. 

If `doc=false` is passed it doesn't even attempt to compile javadocs which speeds up the whole process. 

This is important in cases where people use bloop for compilation as javadocs compilation does not use bloop which results in effectively compiling everything twice.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/publish/SharedPublishOptions.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/publish/SharedPublishOptions.scala
@@ -2,6 +2,7 @@ package scala.cli.commands.publish
 
 import caseapp.*
 
+import scala.build.compiler.{ScalaCompilerMaker, SimpleScalaCompilerMaker}
 import scala.cli.commands.shared.HelpGroup
 import scala.cli.commands.tags
 
@@ -85,8 +86,15 @@ final case class SharedPublishOptions(
   @HelpMessage("Proceed as if publishing, but do not upload / write artifacts to the remote repository")
   @Tag(tags.implementation)
     dummy: Boolean = false
-)
-// format: on
+){
+  // format: on
+
+  def docCompilerMakerOpt: Option[ScalaCompilerMaker] =
+    if (doc.contains(false)) // true by default
+      None
+    else
+      Some(SimpleScalaCompilerMaker("java", Nil, scaladoc = true))
+}
 
 object SharedPublishOptions {
   implicit lazy val parser: Parser[SharedPublishOptions] = Parser.derive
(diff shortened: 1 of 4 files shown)


majk-p (approved):

modules/cli/src/main/scala/scala/cli/commands/publish/SharedPublishOptions.scala
```diff
@@ -85,8 +86,15 @@ final case class SharedPublishOptions(
   @HelpMessage("Proceed as if publishing, but do not upload / write artifacts to the remote repository")
   @Tag(tags.implementation)
     dummy: Boolean = false
-)
-// format: on
+){
+  // format: on
+
+  def docCompilerMakerOpt: Option[ScalaCompilerMaker] =
+    if (doc.contains(false)) // true by default
```
ghostbuster91, line 93: **thought**: I wonder if the `doc` definition shouldn't be rewritten as:
```scala
    doc: Option[Boolean] = Some(true)
```
whereas now it is:
```scala
    doc: Option[Boolean] = None
```
and the meaning of default value is assigned in https://github.com/ghostbuster91/scala-cli/blob/511084bc97554ef6abe5ef530cfd5a3b5c1d8ddf/modules/cli/src/main/scala/scala/cli/commands/publish/Publish.scala#L585

This would make it easier to understand the meaning of the default value. I haven't done it in the PR as I wanted first to discuss that idea. Maybe there are some reasons why it cannot be made like that.

Gedochao: the main reason would be to potentially differentiate between falling back default value being used when the option wasn't passed and passing it explicitly. 
While it doesn't seem to have any meaning for this particular option, we do it this way in a lot of places, so I'd keep it as is.

Gedochao (commented):
Perhaps we could add a test for the `--doc=false` case?
Perhaps somewhere here: https://github.com/VirtusLab/scala-cli/blob/d2bd538f28994b1bd540f14b5a2e7057ae598411/modules/integration/src/test/scala/scala/cli/integration/PublishLocalTestDefinitions.scala

modules/cli/src/main/scala/scala/cli/commands/publish/SharedPublishOptions.scala
ghostbuster91: I see, that makes sense. Though the resolution of the default value should be somehow deduplicated. Maybe there is already a pattern for that in the codebase. Do you recall something like that?

Gedochao: ...not really, unfortunately.

Gedochao (approved):
LGTM

PR #3517, opened 2025-02-25
Fall back to the `cs` command on `PATH` in the `mill` script

Gedochao:
An attempt at https://github.com/scala-steward-org/scala-steward/issues/3213


tgodzik (approved):
LGTM!

PR #3518, opened 2025-02-25
fix: correctly report error position on unknown directive without values

### LINKED ISSUE #3389: The unrecognized directive error for an unrecognized directive without values reports strings from the next line
**Version(s)**
1.5.4

**To Reproduce**
```scala
//> using optionsdeprecation

//> using options -deprecation
```

By debugging the received BSP diagnostic message, these is the line and column information it contains:
Start line: 0
Start column: 28
End line: 0
End column: 40

Notice that there are no 40 characters in the first line. In fact, (0, 28) is just past the "optionsdeprecation" string.

**Expected behaviour**
Possibly a new error message, just "Unrecognized directive: optionsdeprecation" without the "with values" part and it highlights the unrecognized directive "optionsdeprecation".

The reported diagnostic message as it is today, looks weird in IDEA.
https://youtrack.jetbrains.com/issue/SCL-23343/Scala-CLI-next-code-line-is-highlighted-if-using-directive-is-incorrect

kasiaMarek:
This PR changes positions in unrecognised directives errors.
fixes: https://github.com/VirtusLab/scala-cli/issues/3389

### Previously
Errors used to be reported on values:
```
Unrecognized directive: wrong.directive with values: value
//> using wrong.directive value
                          ^^^^^
```
which resulted also in two errors for two values:
```
Unrecognized directive: wrong.directive with values: value1, value2
//> using wrong.directive value1 value2
                          ^^^^^^
Unrecognized directive: wrong.directive with values: value1, value2
//> using wrong.directive value1 value2
                                 ^^^^^^
```
and [incorrect position for no values](https://github.com/VirtusLab/scala-cli/issues/3389).

### Now
We use directive position for error report:
```
Unrecognized directive: wrong.directive with values: value1, value2
//> using wrong.directive value1 value2
          ^^^^^^^^^^^^^^^
```

### DIFF
--- a/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
+++ b/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
@@ -1,7 +1,7 @@
 package scala.build.internal
 
 import bloop.rifle.VersionUtil.parseJavaVersion
-import coursier.jvm.{JavaHome, JvmChannel, JvmIndex}
+import coursier.jvm.{JavaHome, JvmChannel}
 
 import java.io.IOException
 import java.nio.charset.Charset
@@ -56,7 +56,7 @@ object OsLibc {
   // FIXME These values should be the default ones in coursier-jvm
 
   lazy val jvmIndexOs: String = {
-    val default = JvmIndex.defaultOs
+    val default = JvmChannel.defaultOs
     if (default == "linux" && isMusl.getOrElse(false)) "linux-musl"
     else default
   }
--- a/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
+++ b/modules/core/src/main/scala/scala/build/internals/OsLibc.scala
@@ -1,7 +1,7 @@
 package scala.build.internal
 
 import bloop.rifle.VersionUtil.parseJavaVersion
-import coursier.jvm.{JavaHome, JvmChannel}
+import coursier.jvm.{JavaHome, JvmChannel, JvmIndex}
 
 import java.io.IOException
 import java.nio.charset.Charset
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/scala/build/internals/OsLibc.scala
```diff
@@ -56,7 +56,7 @@ object OsLibc {
   // FIXME These values should be the default ones in coursier-jvm
 
   lazy val jvmIndexOs: String = {
-    val default = JvmChannel.defaultOs
+    val default = JvmIndex.defaultOs
```
Gedochao: ?

modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
```diff
@@ -38,18 +38,10 @@ object DirectiveUtil {
...
     scopedDirective: ScopedDirective
-  ): Seq[Positioned[String]] =
-    scopedDirective.directive.values.map {
-      case v: StringValue =>
-        val pos = position(v, scopedDirective.maybePath)
-        Positioned(pos, v.get)
-      case v: BooleanValue =>
-        val pos = position(v, scopedDirective.maybePath)
-        Positioned(pos, v.get.toString)
-      case v: EmptyValue =>
```
Gedochao, line 49: no need to catch empty values here?

Gedochao (approved):
LGTM!

modules/core/src/main/scala/scala/build/internals/OsLibc.scala
kasiaMarek: Oh, thanks, didn't mean to commit this. I might have an incorrect java version or something and it wasn't compiling for me, so this was the fastest way to fix it locally. I'll revert this.

modules/directives/src/main/scala/scala/build/preprocessing/directives/DirectiveUtil.scala
kasiaMarek: I think we want to skip empty values, I use `collect`, so empty just won't be collected.

PR #3519, opened 2025-02-25
Support the `--test` flag with the `package` sub-command

### LINKED ISSUE #3512: Add support for the `--test` option for `package`
To be consistent with other build methods, it should be possible to package the test scope along with main when using the `package` command.

A `--test` command line option, similar to `repl` and `compile`.

Currently it's possible to compile the test scope separately and then add the output to the classpath for `package`.
It's a hassle though, and requires 2 separate `scala-cli` calls.

**Additional context**
- https://github.com/VirtusLab/scala-cli/issues/2728
- https://github.com/VirtusLab/scala-cli/issues/3483
- https://github.com/VirtusLab/scala-cli/issues/2953

Gedochao:
Fixes #3512


tgodzik (approved):
LGTM

bishabosha:
curious how much closer scala-cli is to full multi module support now? (internally) 🤔

Gedochao:
> curious how much closer scala-cli is to full multi module support now? (internally) 🤔

When `--test` gets migrated to `SharedOptions` (after I get it supported in all build sub-commands), technically speaking, it would become easier to add new scopes, which could then function as virtual modules, being inter-dependent.

Doing so would make it way more complex than we want it, however.
Adding multi-module support isn't on the roadmap right now.
And if we were to add it in the future, it'd likely be in a separate tool dependent on Scala CLI, rather than within the CLI itself, that part hasn't changed.

That being said, if you want to ride my recent efforts towards `--test` being supported in all project-building sub-commands and hack something yourself... I'd wait until `publish` and `bsp` support it. (which hopefully will be ready by v1.8)

bishabosha:
> That being said, if you want to ride my recent efforts towards --test being supported in all project-building sub-commands and hack something yourself... I'd wait until publish and bsp support it. (which hopefully will be ready by v1.8)

cool!

PR #3538, opened 2025-03-04
Support the `--test` flag with the `publish` & `publish local` sub-commands

### LINKED ISSUE #3483: Enable including `test` scope inputs when publishing
**Version(s)** 1.6.1

**Describe the bug**
When packaging sources with `scala-cli publish`, sources within `test` directories are omitted even when they contain regular sources that are not tests. This can lead to missing classes in published jars.

Additionally, this behavior is not documented under https://scala-cli.virtuslab.org/docs/commands/publishing/publish

**To Reproduce**
Clone https://github.com/majk-p/scala-cli-publish-test and run ./package-and-check.sh 

**Expected behaviour**
I appreciate that publishing is an experimental feature and I see how this might just be a very simple implementation for skipping test sources. I ideal world Scala CLI would simply knew if the file contains references to test framework but implementing that sounds impractical for a niche issue like this one. 

A practical solution that would be backwards compatible could be a setting like "always-include-path" that would include a path (one or more) that won't be filtered out.

Anything that can help me include `./test/testapi.scala` from the reproduction repo in published jar will be enough.

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/3483

Depends on:
- #3539 
- https://github.com/VirtusLab/scala-cli/pull/3589


tgodzik (changes requested):
Looks like tests are failing, otherwise looks good

Gedochao:
> Looks like tests are failing, otherwise looks good

@tgodzik should be okay now. a test needed adjusting.

tgodzik (approved):
LGTM then!

PR #3557, opened 2025-03-12
docs: document Scala Native flag options

scarf005:
continuation of #3386 because i forgot to address https://scala-cli.virtuslab.org/docs/reference/directives/#scala-native-options


Gedochao (approved):
LGTM, thanks for the contribution

PR #3558, opened 2025-03-12
Pin & update docker images

Gedochao:
- bumps used Ubuntu docker images to `ubuntu:24.04`
- pins Alpine & `libsodium` versions properly
- updates `alpine` to 3.16 
- we can't update `alpine` any higher without adding `.conda` support to `coursier-cache` (but we don't really need to at the moment)

### DIFF
--- a/.github/scripts/docker/ScalaCliSlimDockerFile
+++ b/.github/scripts/docker/ScalaCliSlimDockerFile
@@ -1,6 +1,6 @@
 FROM debian:stable-slim AS build-env
 
-FROM gcr.io/distroless/base-debian10
+FROM gcr.io/distroless/base-debian12
 ADD scala-cli /usr/local/bin/scala-cli
 COPY --from=build-env /lib/x86_64-linux-gnu/libz.so.1 /lib/x86_64-linux-gnu/libz.so.1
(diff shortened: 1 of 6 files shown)


tgodzik (approved):

alexarchambault:
It seems this builds the Linux launcher on the CI from an Ubuntu 24.04 docker image, rather than 20.04. This generates a launcher that doesn't work on Ubuntu 20.04, with errors like
```text
~/.local/share/coursier/bin/.scala-cli.aux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ~/.local/share/coursier/bin/.scala-cli.aux)
~/.local/share/coursier/bin/.scala-cli.aux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ~/.local/share/coursier/bin/.scala-cli.aux)
```

Gedochao:
@alexarchambault Ubuntu 20.04 is no longer supported by GitHub Actions. It's been EOL in April.
https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

PR #3600, opened 2025-03-31
compileOnly option added to the documentation

yadavan88:
Add `compileOnly` usage example to the reference documentation

PR #3602, opened 2025-03-31
docs: guide for compile only deps

scarf005:
bit late to the party (#3600) but still


Gedochao (approved):
Ah, not at all late, as you're filling the gap in a different spot.

PR #3613, opened 2025-04-03
Fix discovery of project file for exclude

### LINKED ISSUE #3546: `//> using exclude` does not work when excluded dir has `project.scala`
The symptom: `//> using exclude foo` does not stop scanning inside of `foo`, so if `foo` contains `project.scala` with directives, then it will evaluate the directives in `foo/project.scala` and even try to compile code.

(when i remove `foo/project.scala` then code compilation is excluded in `foo` as expected)

Here was an initial idea for a solution:

> well, i would assume `exclude` should tell scala-cli to never even look inside that directory. Currently (if the excluded subdirectory has project.scala) it will try to typecheck my scala code in that subdirectory - and check using directives (such as reporting unresolvable library dependencies).
> 
> I would imagine the solution being an ahead of time scan for `exclude` before evaluating other settings/running compiler. 
> 
> if that is too complex or performance sensitive - then perhaps warn about these semantics in the documentation 

 _Originally posted by @bishabosha in [#3385](https://github.com/VirtusLab/scala-cli/issues/3385#issuecomment-2706077451)_

bishabosha:
pick the outermost project file, so that if an excluded directory is itself a scala-cli project then it is ignored.

result of the Scala Tooling Spree April 3 2025

fixes #3546

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -190,9 +190,14 @@ object CrossSources {
 
     val flattenedInputs = inputs.flattened()
     val allExclude = { // supports only one exclude directive in one source file, which should be the project file.
-      val projectScalaFileOpt = flattenedInputs.collectFirst {
+      val projectScalaFileCandidates = flattenedInputs.collect {
         case f: ProjectScalaFile => f
       }
+
+      // this relies upon the inferred workspace root being the outermost directory.
+      // which is the common case when you pass a single directory.
+      val projectScalaFileOpt =
+        projectScalaFileCandidates.sortBy(_.subPath.segments.size).headOption
       val excludeFromProjectFile =
         value(preprocessSources(projectScalaFileOpt.toSeq))
           .flatMap(_.options).flatMap(_.internal.exclude)
--- a/modules/build/src/test/scala/scala/build/tests/ExcludeTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/ExcludeTests.scala
@@ -213,4 +213,40 @@ class ExcludeTests extends TestUtil.ScalaCliBuildSuite {
     }
   }
 
+  test("exclude nested scala-cli project") {
+    val testInputs = TestInputs(
+      os.rel / "Hello.scala" -> "object Hello",
+      // this project.scala needs to come first so that the inferred workspace is the outermost one.
+      os.rel / "project.scala" ->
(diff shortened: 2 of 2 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -190,9 +190,14 @@ object CrossSources {
...
     val allExclude = { // supports only one exclude directive in one source file, which should be the project file.
-      val projectScalaFileOpt = flattenedInputs.collectFirst {
+      val projectScalaFileCandidates = flattenedInputs.collect {
         case f: ProjectScalaFile => f
       }
+
+      // this relies upon the inferred workspace root being the outermost directory.
+      // which is the common case when you pass a single directory.
+      val projectScalaFileOpt =
+        projectScalaFileCandidates.sortBy(_.subPath.segments.size).headOption
```
Gedochao, line 200: while defaulting to the outermost directory doesn't sound like a bad fallback (and we can keep it as fallback), I'm thinking we should favor the `project.scala` which is closest to the designated workspace for the Scala CLI project - and that may not be the shortest path, actually. Especially since the workspace can be overridden with a command line option.

As in, let's say we have:
```
── partitionA
   └── outerdir
       └── actualWorkspace
           └── project.scala
── partitionB
    └── outerdir
        └── project.scala
```
`partitionA/outerdir/actualWorkspace/project.scala` is longer than `partitionB/outerdir/project.scala`, but it should still be picked.
That being said your change is strictly better than what we have on `main`, so I'm okay with that being fixed as a follow-up.

bishabosha: so closest to the workspace of `Inputs`?

Gedochao: Yep. Just be wary that it can be overridden with `--workspace`.

bishabosha:
@Gedochao is there a reason that the paths aren't sorted (e.g. by outermost) before inferring the workspace? ( i guess changing that would be breaking)

Gedochao:
> @Gedochao is there a reason that the paths aren't sorted (e.g. by outermost) before inferring the workspace? ( i guess changing that would be breaking)

No reason that I know of... It should be covered with tests, in case you want to experiment.

Gedochao:
@bishabosha is this something you mean to get back to? or should we close it for now?

bishabosha:
dont see myself working on this any longer unless i get blocked by it again

Issue #3598, opened 2025-03-31
Additional logging of the scala/java/scala-cli versions


yadavan88:
**Is your feature request related to a problem? Please describe.**
When a scala-cli project is compiled for the first time(or recompiled after change), it shows the scala version and the jdk version. If there is no change in the code, on the next runs, it doesn't log this info. 
Furthermore, this doesn't show the scala-cli version in the log. 
Attaching the screenshot of the output I get when I run a file

<img width="383" alt="Image" src="https://github.com/user-attachments/assets/a8defedd-e668-43de-a47e-cbc9a7f513f2" />

**Describe the solution you'd like**
It would be nice if there is an option or a directive or something like it that could show these info all the time(since the default behaviour is not to have too many logs and lose focus for most people). When working with multiple JDKs and Scala Versions, it would be beneficial to always see which scala and jdk is used to run the program just as a reminder. Also, unless someone remembers to update the scala-cli version, one might use the older version of scala-cli without realizing. For example, I was using 1.5.x in one of my laptops when the current version is 1.7.x. Seeing an info about the versions used might give an hint to check if these are the expected/intereted versions. 

**Describe alternatives you've considered**
Clean compile and manually checking the scal-cli version once in a while

**Additional context**
In discord, the suggestion is to have less logs as the preferred method, but an option to have additional important logs might be nice.

Gedochao:
What I think we could do:
- at increased verbosity (behing a single `-v` flag), log Scala / JVM versions even when no re-compilation happens
  - we currently only do it with Bloop; we probably could also log the information with `--server=false`
- we could have a dedicated verbosity flag (`--log-cli-version`?) and/or global config  (`log-cli-version`) for logging the CLI version as well (or, alternatively, also do it at increased verbosity, perhaps at `-v -v`)

yadavan88:
Is this verbocity flag be added as a directive? I mean in some of my scripts, I want to ensure that this info is always logged. I prefer to keep all as directives in the file (even commented out, so that i can enable it later if needed and don't need to remember the flags or search in internet to find out)

Gedochao:
> Is this verbocity flag be added as a directive? I mean in some of my scripts, I want to ensure that this info is always logged.

As a rule, verbosity can't (and shouldn't) be configured by directives.

@yadavan88 In that case, for your use case, perhaps you'd like to print the information you need yourself and control it in the script.
Then it doesn't have to be controlled by verbosity (which many want to keep at a minimum).
[...]

PR #3614, opened 2025-04-03
Log scala-cli version when --log-cli-version is enabled

yadavan88:
Implementation for https://github.com/VirtusLab/scala-cli/issues/3598


modules/cli/src/main/scala/scala/cli/commands/shared/LoggingOptions.scala
```diff
@@ -20,7 +20,11 @@ final case class LoggingOptions(
   @Group(HelpGroup.Logging.toString)
   @Tag(tags.implementation)
   @HelpMessage("Use progress bars")
-    progress: Option[Boolean] = None
+    progress: Option[Boolean] = None,
+  @Group(HelpGroup.Logging.toString)
+  @Tag(tags.implementation)
+  @HelpMessage("Logs scala-cli version on project build")
```
Gedochao, line 26:
```suggestion
  @HelpMessage(s"Logs ${scala.cli.ScalaCli.fullRunnerName} version alongside JVM and Scala versions when the project gets recompiled by the server.")
```

Gedochao: it still feels hyper-granular to have a command line option just for that... wouldn't it be better to print this when verbosity is set to +1? (as in, just a single `-v`)
If we can live without this command line option, I guess I'd rather not have it.
tagging @tgodzik for second opinion.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -1090,13 +1090,18 @@ object Build {
...
+
       val params = value(options0.scalaParams)
 
       val scopeParams =
         if (scope == Scope.Main) Nil
         else Seq(scope.name)
 
-      buildClient.setProjectParams(scopeParams ++ value(options0.projectParams))
+      val scalaCliVersionLog =
+        if options0.internal.logCliVersion then Seq(s"scala-cli ${Constants.version}") else Nil
```
Gedochao, line 1101: Can't use `scala-cli` here, as contextually it may be just `scala`. What you actually need is `scala.cli.ScalaCli.fullRunnerName`... which is unfortunate, since you log it in the `build` module.
You could maybe pass it via `scala.build.input.ScalaCliInvokeData`... See how it's passed around in the code elsewhere. I wonder if `scala.build.input.ScalaCliInvokeData.progName` would be enough... 

I'm thinking this would look awkward:
```
Compiling project (Scala 3.7.0 JVM (17), scala 1.7.1)
```

We should have the  full runner name, so:
```
Compiling project (Scala 3.7.0 JVM (17), Scala code runner  1.7.1)
```

modules/cli/src/main/scala/scala/cli/commands/shared/LoggingOptions.scala
yadavan88: Yeah, probably too less info for a separate flag. I think it is alos fine with `-v`. Additionally, it might be good if the scala-cli version is also added to buildinfo, so that in cases such as mine, I can explicitly log this info using the buildinfo.

Gedochao: yup, adding it to `BuildInfo` is a good idea, I think.
Feel free to do it in a separate PR (or make a new issue).

yadavan88: Let me see if I can create a separate PR with adding this info in the buildinfo. Reg -v, let me know. It would still be nice to see this info without making any code change in the scripts.

yadavan88: @Gedochao I created a new PR
https://github.com/VirtusLab/scala-cli/pull/3617

Gedochao:

Issue #3596, opened 2025-03-28
Allow for preset scalacOptions


tgodzik:
**Is your feature request related to a problem? Please describe.**
Something similar to https://github.com/typelevel/sbt-tpolecat but instead of using plugins we can create presets (this could acompany toolkit as the best practise)

**Describe the solution you'd like**

`//> using options.preset suggested`

**Describe alternatives you've considered**
Add all options manually (it's quite a lot)

tgodzik:
~We could also detect standard flags such as CI and turn off fatal-warnings if it doesn't exist locally.~

Or alternatively, we could just turn on fatal warnings on CI

yadavan88:
This is really interesting. Should there be option to remove anything from the suggested if not interested? Or that is an overkill?

tgodzik:
> This is really interesting. Should there be option to remove anything from the suggested if not interested? Or that is an overkill?

That would require some kind of DSL, so probably we wouldn't want to add it unfortunately.

yadavan88:
@tgodzik / @Gedochao 
I tried to implement this, here is the [PR](https://github.com/VirtusLab/scala-cli/pull/3615) with basic working solution. 
Let me know if I am in the right direction or did I get it wrong?

PR #3615, opened 2025-04-03
Scalac preset option in directive

yadavan88:
Implementation for https://github.com/VirtusLab/scala-cli/issues/3596

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -56,8 +56,8 @@ the unit tests:
 To can filter unit test suites:
 
 ```bash
-./mill 'build-module.test' 'scala.build.tests.BuildTestsScalac.*'
-./mill 'build-module.test' 'scala.build.tests.BuildTestsScalac.simple'
+./mill 'build-module[].test' 'scala.build.tests.BuildTestsScalac.*'
+./mill 'build-module[].test' 'scala.build.tests.BuildTestsScalac.simple'
 ```
 
 #### Run integration tests with the JVM launcher
--- a/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
+++ b/modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
@@ -72,19 +72,19 @@ final case class ScalacOptions(
 
     presetOptions match {
       case None => explicitScalacOptions
-      case Some(ScalacOpt.PresetOptions.Suggested.preset) =>
+      case Some(ScalacOpt.PresetOption.Suggested.preset) =>
         val presetOptions =
           ScalacOptions.buildOptions(
             ScalacOptions.presetOptionsSuggested(scalaVersion).map(Positioned.none)
           )
         explicitScalacOptions :+ presetOptions.map(_.withEmptyRequirements)
-      case Some(ScalacOpt.PresetOptions.CI.preset) =>
+      case Some(ScalacOpt.PresetOption.CI.preset) =>
         val presetOptions =
           ScalacOptions.buildOptions(
             ScalacOptions.presetOptionsCI(scalaVersion).map(Positioned.none)
           )
         explicitScalacOptions :+ presetOptions.map(_.withEmptyRequirements)
-      case Some(ScalacOpt.PresetOptions.Strict.preset) =>
+      case Some(ScalacOpt.PresetOption.Strict.preset) =>
         val presetOptions =
           ScalacOptions.buildOptions(
             ScalacOptions.presetOptionsStrict(scalaVersion).map(Positioned.none)
@@ -93,7 +93,7 @@ final case class ScalacOptions(
       case Some(other) =>
         List(Left(
           InputsException(
-            s"Unknown preset options: $other. Available options are: ${ScalacOpt.PresetOptions.values.map(_.preset).mkString(", ")}"
+            s"Unknown preset options: $other. Available options are: ${ScalacOpt.PresetOption.values.map(_.preset).mkString(", ")}"
           )
         ))
     }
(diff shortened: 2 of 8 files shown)


modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
```diff
@@ -44,12 +47,39 @@ final case class ScalacOptions(
...
+          ScalacOptions.buildOptions(ScalacOptions.presetOptionsCI.map(Positioned.none))
+        explicitScalacOptions :+ presetOptions.map(_.withEmptyRequirements)
+      case Some("strict") =>
+        val presetOptions =
+          ScalacOptions.buildOptions(ScalacOptions.presetOptionsStrict.map(Positioned.none))
+        explicitScalacOptions :+ presetOptions.map(_.withEmptyRequirements)
+      case Some(other) =>
+        List(Left(
+          InputsException(
+            s"Unknown preset options: $other. Available options are: suggested, ci, strict"
```
Gedochao: We should have an enumeration of the available presets somewhere, with tied option lists.
then we should just call the full list here, rather than hardcoding it like this.

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
```diff
@@ -62,4 +92,9 @@ object ScalacOptions {
         )
       )
     }
+
+  // todo: this should be based on the scala version. This might not be the correct location
```
Gedochao: yup... might be annoying to keep up-to-date as well. 
I actually wonder if the best place for defining the presets wouldn't be the compiler itself... but that could be done as an improvement. 
just theorising, since this is an early draft.

Gedochao: the 3 presets for a start are good I think (having `suggested`, `ci` and `strict`; still need to establish what would each of them entail)

DEV.md
```diff
@@ -56,8 +56,8 @@ the unit tests:
 To can filter unit test suites:
 
 ```bash
-./mill 'build-module.test' 'scala.build.tests.BuildTestsScalac.*'
-./mill 'build-module.test' 'scala.build.tests.BuildTestsScalac.simple'
+./mill 'build-module[].test' 'scala.build.tests.BuildTestsScalac.*'
+./mill 'build-module[].test' 'scala.build.tests.BuildTestsScalac.simple'
```
Gedochao: it's correct to fix this, but maybe should be done in a separate PR 😅

Gedochao (commented):
Note: we should support this both via a directive and as a command line option (similarly to how toolkits are done).
Then, the command line should be able to override what we have in directives (be careful not to concatenate the options, if a different preset is chosen from the command line than in a directive.
Finally, we'll need to test if duplicates are handled correctly (for example, if a user passes an option explicitly, but also uses a preset which contains it, we need to handle the duplicates).
This means quite a bunch test scenarios to write.

DEV.md
yadavan88: Yeah, I noticed that this was not compiling, so I added this change. I can quickly make a separate PR just for this. :)

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
yadavan88: Yeah, I just wanted to add some ideas, I am also not fully clear what should they be, but added there just to make sure that we should handle multiple presets.

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
yadavan88: Yeah, where should this be defined. I was only making a quick draft just to see if I understood the requirement correctly. The issue seems to be really interesting as I tend to add some options manually and I never remember what should be used.

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
Gedochao: Regarding what they should be, I believe the idea was to start with something like what https://github.com/typelevel/sbt-tpolecat has... 

We could also start a discussion for suggestions somewhere...  A thread on https://contributors.scala-lang.org/, maybe... or just a discussion on this repo? 
Either way, if/when suggestions on which options should be included come, we could tweak it gradually.
The feature should first come out as experimental, anyway.

DEV.md
yadavan88: should this be against main or stable, I assume main since this is for dev?

yadavan88: Here is it https://github.com/VirtusLab/scala-cli/pull/3619
I dont understand what is the diff between [] and without it in Mill, but it works!

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
yadavan88: I thought about the duplicates yday, but I saw some test related to duplicate scalaC args are ignored or something like that. SO thought it can be ignored. I will try to handle them. How about getting the scalaVersion? We should have the sugested and other presets based on the scala versions right. I assumes `optsForScala212`, `optsForScala213`,  `optsForScala3` which supports all versions within 2.13, 2.12 and 3.
Can you point me how to get tht scala version within this scalacOptions code?

yadavan88: If the command line args needs to be processed, then is it not possible to add the scalac options in the `ScalacOptions.scala` file right? I mean, tif the preset is also passed as command line option, we need to keep the preset value till the last moment and then decide priority (command line over directive). So, the preset value should be somehow carried over and not the flattened scalac options). Am I right?

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
yadavan88: I added these options as enums

modules/directives/src/main/scala/scala/build/preprocessing/directives/ScalacOptions.scala
yadavan88: @Gedochao Could you have a look at it when you get time?

Gedochao: > Can you point me how to get tht scala version within this scalacOptions code?

You can't. We don't know what's the final Scala version until `using` directives are parsed. 

> If the command line args needs to be processed, then is it not possible to add the scalac options in the ScalacOptions.scala file right? I mean, tif the preset is also passed as command line option, we need to keep the preset value till the last moment and then decide priority (command line over directive). So, the preset value should be somehow carried over and not the flattened scalac options). Am I right?

Yep, the presets can only be applied once the Scala version is finalised, which generally should be after (or within) `scala.build.CrossSources.forInputs`.

> I thought about the duplicates yday, but I saw some test related to duplicate scalaC args are ignored or something like that.

We do have handling for when there's duplicates introduced by the command line vs directives. If duplicates are explicitly added by the user in one context, we don't remove them and let the compiler handle it.
However, for when a preset has a duplicate with whatever explicit options the user passed, you need to make sure it's handled correctly. Just add a test for this case.

Gedochao:

PR #3617, opened 2025-04-04
Add scala-cli version to the BuildInfo

yadavan88:
Add a variable for the scalaCliVersion in the generated BuildInfo


yadavan88:
@Gedochao Here is the PR for adding scala-cli version to buildInfo. Could you please review?

modules/options/src/main/scala/scala/build/info/BuildInfo.scala
```diff
@@ -64,7 +65,11 @@ final case class BuildInfo(
       Seq(
         "/** Project version */",
         "val projectVersion ="
-      ) -> projectVersion
+      ) -> projectVersion,
+      Seq(
+        "/** Scala-cli version used for the compilation */",
```
Gedochao:
```suggestion
        "/** Scala CLI version used for the compilation */",
```

website/docs/reference/build-info.md
```diff
@@ -51,6 +51,8 @@ object BuildInfo {
   val mainClass = Some("Main")
   /** Project version */
   val projectVersion = None
+  /** Scala-cli version used for the compilation */
+  val scalaCliVersion = Some("1.1.1-SNAPSHOT")
```
Gedochao: mental note, that this doc should get updated with some more modern values =)
```suggestion
  val scalaCliVersion = Some("1.7.1")
```

Gedochao (commented):
Commented on some nitpicks.
Also, please include a test.

website/docs/reference/build-info.md
yadavan88: @Gedochao I fixed the tests in SourceGeneratorTests, I hope that is the only place. 
Also, since the constants file was not accessible there, i hardcoded the version 1.1.1-SNAPSHOT in the test. Is it ok?

yadavan88: @Gedochao Could you review this once again. I think there are some test failures, will have a look at it once again. But is the other changes ok?

Gedochao: Yup, LGTM

Gedochao (approved):

PR #3619, opened 2025-04-04
Fixed DEV.md file related to test command

yadavan88:


DEV.md
```diff
@@ -43,21 +43,21 @@ E.g:
...
-./mill 'build-module.test'
+./mill 'build-module[].test'
 ```
 
 If you want to run unit tests for another module, set `module_name` to the name of the module from which you want to run
 the unit tests:
 
 ```bash
-./mill 'module_name.test'
+./mill 'module_name[].test'
```
Gedochao: not all modules are cross-compiled.

yadavan88: ah, this `[]` is for cross compilation? I removed from these

Gedochao: Yes.
https://mill-build.org/mill/fundamentals/cross-builds.html#_cross_modules_defaults

yadavan88: Done

Gedochao (approved):

PR #3621, opened 2025-04-07
Run all found test frameworks, rather than just one

### LINKED ISSUE #3620: `test` should warn when multiple test frameworks are on the classpath
**Version(s)**
Scala code runner version: 1.7.1
Scala version (default): 3.6.4

**Describe the bug**

adding `//> using toolkit default` to scalatest file 

**To Reproduce**

`main.test.scala`
```scala
//> using test.dep org.scalatest::scalatest::3.2.19

import org.scalatest.flatspec.AnyFlatSpec

class ExampleSpec extends AnyFlatSpec:
    "example" should "work" in { assertResult(1)(1) }
```
```sh
scala --cli-version 1.7.1 test .
Compiling project (test, Scala 3.6.4, JVM (23))
Compiled project (test, Scala 3.6.4, JVM (23))
ExampleSpec:
example
- should work
Run completed in 221 milliseconds.
Total number of tests run: 1
Suites: completed 1, aborted 0
Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
```

```scala
//> using test.dep org.scalatest::scalatest::3.2.19
//> using toolkit default

import org.scalatest.flatspec.AnyFlatSpec

class ExampleSpec extends AnyFlatSpec:
    "example" should "work" in { assertResult(1)(1) }
```
```
scala --cli-version 1.7.1 test .
Compiling project (test, Scala 3.6.4, JVM (23))
Compiled project (test, Scala 3.6.4, JVM (23))
```

**Expected behaviour**

importing toolkits shouldn't prevent scalatest from working silently.

Gedochao:
Fixes #3620 

What this does:
- on the JVM, all test frameworks should be detected and ran automatically
- on Scala Native & Scala.js platforms, all recognised test frameworks should be detected and ran automatically
  - unknown (non-recognised) and custom test frameworks are limited to 1 at a time, and not together with others; this is due to deep-searches for unknown test frameworks being very time-consuming;
  - this is still a straight-up upgrade over the past behaviour, when a single test framework would be run at all times.
- the `--test-framework` CLI option and `//> using testFramework` using directive work as before, where **only** the specified test framework is to be run
  - note that this gives the best performance, particularly for Scala Native & Scala.js tests, as this effectively skips test framework detection and runs with what is passed

Potential future follow-ups:
- handle multiple test frameworks on `SBT`/`Mill` export: https://github.com/VirtusLab/scala-cli/issues/3634
- expand the `--test-framework` CLI option and `//> using testFramework` directive to accept multiple values, allowing to preconfigure multiple frameworks and skipping detection: https://github.com/VirtusLab/scala-cli/issues/3632
- add a deep search flag, which would allow someone to actually deep search for test frameworks on Scala Native & Scala.js, if they really care to and don't mind the long loading times (not sure how many use cases that'd have, but it doesn't hurt to add a flag)
- add a `--print-test-frameworks` flag, similar to `--print-classpath`, to make it easier to find what test frameworks are available in the project: https://github.com/VirtusLab/scala-cli/issues/3633
- add a global config for preconfiguring more recognised test frameworks

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/Runner.scala
+++ b/modules/build/src/main/scala/scala/build/internal/Runner.scala
@@ -4,6 +4,7 @@ import coursier.jvm.Execve
 import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
 import org.scalajs.jsenv.nodejs.NodeJSEnv
 import org.scalajs.jsenv.{Input, RunConfig}
+import org.scalajs.testing.adapter.TestAdapter as ScalaJsTestAdapter
 import sbt.testing.{Framework, Status}
 
 import java.io.File
@@ -16,6 +17,7 @@ import scala.build.errors.*
 import scala.build.internals.EnvVar
 import scala.build.testrunner.FrameworkUtils.*
 import scala.build.testrunner.{AsmTestRunner, TestRunner}
+import scala.scalanative.testinterface.adapter.TestAdapter as ScalaNativeTestAdapter
 import scala.util.{Failure, Properties, Success}
 
 object Runner {
@@ -240,22 +242,20 @@ object Runner {
     sourceMap: Boolean = false,
     esModule: Boolean = false
   ): Either[BuildException, Process] = either {
-
-    import logger.{log, debug}
-
-    val nodePath = value(findInPath("node").map(_.toString).toRight(NodeNotFoundError()))
-
-    if (!jsDom && allowExecve && Execve.available()) {
-
+    val nodePath: String =
+      value(findInPath("node")
+        .map(_.toString)
+        .toRight(NodeNotFoundError()))
+    if !jsDom && allowExecve && Execve.available() then {
       val command = Seq(nodePath, entrypoint.getAbsolutePath) ++ args
 
-      log(
+      logger.log(
         s"Running ${command.mkString(" ")}",
         "  Running" + System.lineSeparator() +
           command.iterator.map(_ + System.lineSeparator()).mkString
       )
 
-      debug("execve available")
+      logger.debug("execve available")
       Execve.execve(
         command.head,
         "node" +: command.tail.toArray,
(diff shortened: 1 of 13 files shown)


modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -410,57 +414,72 @@ object Runner {
...
+          adapter
+            .loadFrameworks(foundFrameworkNames.map(List(_)))
+            .flatten
+            .distinctBy(_.name())
+
+        val finalTestFrameworks =
+          loadedFrameworks
+            .filter(
+              !_.name().toLowerCase.contains("junit") ||
+              !loadedFrameworks.exists(_.name().toLowerCase.contains("munit"))
```
tgodzik, line 468: MAybe add a comment why filter junit and munit here?

modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -471,42 +490,61 @@ object Runner {
...
+
+        val finalTestFrameworks =
+          loadedFrameworks
+            // .filter(
+            //  _.name() != "Scala Native JUnit test framework" ||
+            //    !loadedFrameworks.exists(_.name() == "munit")
+            // )
+            // TODO: add support for JUnit and then only hardcode filtering it out when passed with munit
+            // https://github.com/VirtusLab/scala-cli/issues/3627
+            .filter(_.name() != "Scala Native JUnit test framework")
```
tgodzik, line 527: Could we get that title from somewhere? Looks fragile if it ever changes

modules/test-runner/src/main/scala/scala/build/testrunner/FrameworkUtils.scala
```diff
@@ -0,0 +1,223 @@
...
+      finally if (stream != null) stream.close()
+    }
+    else if (keepJars && Files.isRegularFile(classPathEntry)) {
+      import java.util.zip._
+      var zf: ZipFile = null
+      try {
+        zf = new ZipFile(classPathEntry.toFile)
+        zf.entries
+          .asScala
+          // FIXME Check if these are files too
```
tgodzik, line 120: Should we just add a .exists check here? Might be worth updating or resolving this FIXME

tgodzik (approved):
Looks good! I just added some comments, but feel free to ignore me.

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: doesn't seem like it 😕

Gedochao: tbf, I'd like to add support for JUnit on Native at some point, so hopefully we'll be able to get rid of this

modules/test-runner/src/main/scala/scala/build/testrunner/FrameworkUtils.scala
Gedochao: I can try to look into it, but it's out of scope for this PR.
I merely moved this piece of code around.

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: will add it separately.

PR #3639, opened 2025-04-15
Directives API

Gedochao:


tgodzik (commented):
Looks cool, so the next step would be to generate some json structure out of it? Or a separate jar? We could have some code generation that creates a separate jar crosspublished for 2.13 and 3 🤔 

Basically something like:

```
 val all = Directives.getDirectiveHandlers.map {
  handler => 
    s"""|
       |  object ${handler.name} {
       |    def description = "${handler.description}"
       |    def descriptionMd = "${handler.descriptionMd}"
       |    def usage = "${handler.usage}"
       |    def usageMd = "${handler.usageMd}"
       |    def keys = List(${handler.keys.map(key => s""""${key.name}"""").mkString(", ")})
       |  }
       |
       |""".stripMargin
}

  s"""|object Directives {
     |  ${all.mkString("\n")}
     |}
     |""".stripMargin
```

we can put that into a main to invoke from Mill and paste the output into Directives.scala files in a separate module.

Gedochao:
> Looks cool, so the next step would be to generate some json structure out of it? Or a separate jar?

Ideally, I'd like a separate JAR, so that there'd be a way to match definitions with a particular version of Scala CLI.
I'll see what I can hack when I have a moment.

PR #3647, opened 2025-04-24
Add a suffix to the script file to avoid conflict with library name

### LINKED ISSUE #3595: Broken compilation when script name equals to library name
**Version(s)**
1.6.1, 1.7.1

**Describe the bug**
In worksheet mode filename affects compilation.

**To Reproduce**
1) Create file `zio.sc` with code:

```sc
//> using dep "dev.zio::zio:2.1.16"

import zio.ZIO
```

2) Run:  `scala-cli zio.sc`

Got: 
```
[error] ./zio.sc:4:12
[error] value ZIO is not a member of zio$_
[error] import zio.ZIO
[error]            ^^^
```

**Expected behaviour**

No compiler errors.
In non-worksheet mode (if filename is `zio.scala`) the behavior is as expected.

yadavan88:
Fixes https://github.com/VirtusLab/scala-cli/issues/3595

Note. Could not add fallaback for the runner name

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
+++ b/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
@@ -13,7 +13,7 @@ object AmmUtil {
       case i  => fileName.take(i)
     }
 
-    (pkg, Name(wrapper))
+    (pkg, Name(s"${wrapper}_scalacli_wrapper"))
   }
 
   def encodeScalaSourcePath(path: Seq[Name]) = path.map(_.backticked).mkString(".")
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
@@ -105,7 +105,7 @@ case object ScriptPreprocessor extends Preprocessor {
         scriptCode,
         inputArgPath.getOrElse(subPath.toString)
       )
-      
+
       val relPath = os.rel / (subPath / os.up) / s"${subPath.last.stripSuffix(".sc")}.scala"
 
       val file = PreprocessedSource.UnwrappedScript(
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScriptPreprocessor.scala
@@ -105,9 +105,8 @@ case object ScriptPreprocessor extends Preprocessor {
         scriptCode,
         inputArgPath.getOrElse(subPath.toString)
       )
-
-      val className = (pkg :+ wrapper).map(_.raw).mkString(".")
-      val relPath   = os.rel / (subPath / os.up) / s"${subPath.last.stripSuffix(".sc")}.scala"
+      
+      val relPath = os.rel / (subPath / os.up) / s"${subPath.last.stripSuffix(".sc")}.scala"
 
       val file = PreprocessedSource.UnwrappedScript(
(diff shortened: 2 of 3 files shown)


modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
```diff
@@ -13,7 +13,7 @@ object AmmUtil {
       case i  => fileName.take(i)
     }
 
-    (pkg, Name(wrapper))
+    (pkg, Name(s"${wrapper}_scalacli_wrapper"))
```
tgodzik, line 16: We were also thinking about adding a random suffix, but that would make it impossible to run the main class by name. So instead we did a named that should not interfere with any library. This might be breaking so we also wanted to have a fallback to add the suffix if someone is using the old one. @Gedochao any idea where that can be?

Gedochao: Sorry for getting back late.
As for inferring main classes, detecting script main classes and such, the logic is here:
https://github.com/VirtusLab/scala-cli/blob/a4c286ff026a0a77142d2ae226b6a3fd1fb6f692/modules/build/src/main/scala/scala/build/Build.scala#L115-L183

Gedochao: However, I am not convinced we should change the wrapper name - not without explicit configuration, at least.
Maybe via directive/flag.
Changing the default behaviour will be an effective change of syntax, not just for inferring the main class name.
i.e. code like this will stop working:
https://gist.github.com/Gedochao/8902375ad97d24d8ed49c9bb33d5bd08
And this feels quite bad to me, since we've had this syntax since forever.

Gedochao: Maybe something like:
```scala
//> using script.overrideWrapperName <name>
```
?
So that the change of wrapper is committed in the code, to make the changed behaviour predictable.

Gedochao: As for better UX, we could think about detecting the cases such as the reproduction from https://github.com/VirtusLab/scala-cli/issues/3595, and then print a warning, suggesting to override the wrapper... Although detecting such a case would likely mean parsing AST, so that's a tougher cookie.

Gedochao:
@yadavan88 we discussed this solution with @tgodzik offline. Let me sum up.

the TL;DR of it is that the reproduction from https://github.com/VirtusLab/scala-cli/issues/3595 is a `wontfix`.
The solution presented here breaks existing syntax for referring to script code from another input (see [this gist](https://gist.github.com/Gedochao/8902375ad97d24d8ed49c9bb33d5bd08) from [one of my earlier comments](https://github.com/VirtusLab/scala-cli/pull/3647#discussion_r2065780802)), so it's a no-go.

We could technically do a workaround behind a directive/flag (also described [in an earlier comment](https://github.com/VirtusLab/scala-cli/pull/3647#discussion_r2065782939)), but it would require the user to know he/she needs to use the workaround.

We could have a warning suggesting the workaround (or even implicitly adding the suffix to the wrapper), but that would mean looking for conflicts on the classpath (expensive, were we to do it every time a script is invoked), or at the very least parsing the script and looking at the imports (i.e. the `zio.sc` example has the `import zio.ZIO` line, which already implies there could be a conflict with the `zio` classified name, warranting a warning).
Having a warning would be a nice UX improvement, but implementing it is a rather complex undertaking for a relatively small gain.

Overall, as it turns out, this is a rather difficult issue for the Spree (not the first time, I know).

PR #3654, opened 2025-04-30
chore: Update LTS to 3.3.6-RC2

tgodzik:


tgodzik:
Looks like it's only the ammonite issues

Issue #3210, opened 2024-09-26
Test the REPL with the Scala 3 REPL rather than Ammonite


Gedochao:
**Is your feature request related to a problem? Please describe.**
We currently test a number of things in our REPL integration tests using Ammonite (see https://github.com/VirtusLab/scala-cli/pull/3209).
As Scala CLI is closely tied with Scala 3 releases and Ammonite is maintained separately, we should uncouple these 2 things.
The way things are, it's hard to guess what works until there's an Ammonite version for a given Scala version ready (which sometimes happens late or not at all, like with Scala 3.4.0/3.4.1)

**Describe the solution you'd like**
We should test all `repl` sub-command features via the Scala 3 REPL rather than Ammonite REPL.
The only tests relying on Ammonite should be the ones testing the actual integration.

We should run all REPL tests for Scala 3 using the Scala 3 REPL, rather than Ammonite (except for the tests where we do, in fact, test whether the Ammonite integration works as expected).

**Dependencies**
- https://github.com/scala/scala3/issues/21242

**Additional context**
- https://github.com/VirtusLab/scala-cli/pull/3209
- https://github.com/com-lihaoyi/Ammonite/issues/1563
- https://github.com/com-lihaoyi/Ammonite/issues/1395

PR #3664, opened 2025-05-06
Add support for the `--repl-quit-after-init` REPL option

Gedochao:
Follow-up for 3.7 after:
- https://github.com/scala/scala3/pull/22636

Also a pre-requirement for:
- https://github.com/VirtusLab/scala-cli/issues/3210

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
@@ -407,20 +407,15 @@ object Repl extends ScalaCommand[ReplOptions] with BuildCommandHelpers {
       extraProps: Map[String, String] = Map.empty
     ): Unit = {
       val isAmmonite = replArtifacts.replMainClass.startsWith("ammonite")
-      if replArgs.exists(_.noDashPrefixes == ScalacOptions.replInitScript) then
-        scalaParams.scalaVersion match
-          case _ if isAmmonite =>
+      if isAmmonite then
+        replArgs
+          .map(_.noDashPrefixes)
+          .filter(ScalacOptions.replExecuteScriptOptions.contains)
+          .foreach(arg =>
             logger.message(
-              "The '--repl-init-script' option is not supported with Ammonite. Did you mean to use '--ammonite-arg'?"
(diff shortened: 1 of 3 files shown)


modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
```diff
@@ -407,20 +407,15 @@ object Repl extends ScalaCommand[ReplOptions] with BuildCommandHelpers {
...
+          .filter(ScalacOptions.replExecuteScriptOptions.contains)
+          .foreach(arg =>
             logger.message(
-              "The '--repl-init-script' option is not supported with Ammonite. Did you mean to use '--ammonite-arg'?"
+              s"The '--$arg' option is not supported with Ammonite. Did you mean to use '--ammonite-arg -c' to execute a script?"
             )
-          case s
-              if s.coursierVersion < "3.6.4-RC1".coursierVersion &&
-              s.coursierVersion < "3.6.4".coursierVersion &&
-              s.coursierVersion < "3.6.4-RC1-bin-20250109-a50a1e4-NIGHTLY".coursierVersion =>
```
Gedochao, line 419: I'm thinking perhaps it's overkill to hardcode in which exact version was an option introduced here (especially as those REPL options are actually meant chiefly for internal use)

modules/integration/src/test/scala/scala/cli/integration/ReplTestsDefault.scala
```diff
@@ -36,4 +38,23 @@ class ReplTestsDefault extends ReplTestDefinitions
         }
       }
     }
+
+  if (!Properties.isWin)
+    // TODO: run this on Scala 3 LTS when 3.3.7 is out
```
Gedochao: - https://github.com/scala/scala3-lts/pull/272

tgodzik: Could we have a check that runs something on LTS only if above a certain version? We usually do it metals so that we don't forget about it. Or we should create an issue

Gedochao: sure, added a check that should run it on 3.3.7+ when we bump the LTS.
and the Scala Next RC, while I was at it.

tgodzik (approved):

PR #3681, opened 2025-05-15
Support links in using file directive (implements #1328)

### LINKED ISSUE #1328: //> using file directives should allow URLs and git repository paths
It would be useful if the `//> using file` directives could accept URL and git repository paths in addition to local files. This would allow users to refer to their scala-cli module scripts from their executable scripts, no matter what paths the executable scripts were stored in:

```
//> using file "git+ssh://git@gitlab.com/benwbooth/scala-scripts.git?dir=lib/scriptdir/my-script.sc"
```

or

```
//> using file "https://raw.githubusercontent.com/benwbooth/scala-scripts/lib/scriptdir/my-script.sc"
```

[Deno](https://examples.deno.land/import-export) has a similar feature for pulling in dependencies from URLs.

ivan-klass:
Scala Tooling Spree 15.05

see #1328


ivan-klass:
`git+ssh` is not supported by coursier and should be addressed there - currently only HTTP is assumed:

```
coursier.cache.internal.Downloader.Blocking.doDownload
```

ivan-klass:
@majk-p @tgodzik It works! Indeed, dependencies are resolved recursively already.

`out/test.scala`:

```
//> using file https://raw.githubusercontent.com/softwaremill/sttp/refs/heads/master/examples/src/main/scala/sttp/client4/examples/json/GetAndParseJsonCatsEffectCirce.scala
```

`./mill 'cli[3.3.6].run' out/test.scala`:

```
Got response code: 200
Right(HttpBinResponse(87.116.164.191,Map(Accept-Encoding -> gzip, deflate, Host -> httpbin.org, User-Agent -> Java-http-client/17.0.13, X-Amzn-Trace-Id -> Root=1-6826f31e-7030221a5448052d12774af8)))
```

Gedochao:
Note: need to fix formatting & regenerate reference docs for the CI to pass.
https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -403,7 +405,39 @@ object CrossSources {
     fromInputs ++ fromSources ++ fromSourcesWithRequirements
   }
 
-  private def resolveInputsFromSources(sources: Seq[Positioned[os.Path]], enableMarkdown: Boolean) =
+  // TODO: reuse existing one? e.g. scala.cli.commands.shared.SharedOptions.coursierCache
+  lazy val fileCache: FileCache[coursier.util.Task] = FileCache()
```
Gedochao: Yup, I think it's accurate that `SharedOptions.coursierCache` should be used.

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -403,7 +405,39 @@ object CrossSources {
...
+    import scala.build.options.ScalaVersionUtil.fileWithTtl0
+    val artifact = Artifact(pUri.value.toString).withChanging(true)
+    fileCache.fileWithTtl0(artifact)
+      .left
+      .map(err =>
+        new MalformedDirectiveError(err.describe, pUri.positions)
+      ) // TODO: better error type
+      .map(f => os.read.bytes(os.Path(f, Os.pwd))).map(content =>
+        Seq(Virtual(pUri.value.toString, content))
+      )
```
Gedochao: This is very similar to `SharedOptions.downloadInputs`... I wonder if the function itself could be passed all the way here, rather than the cache

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -403,7 +405,39 @@ object CrossSources {
...
+  lazy val fileCache: FileCache[coursier.util.Task] = FileCache()
+
+  private def downloadFile(pUri: Positioned[java.net.URI]) =
+    import scala.build.options.ScalaVersionUtil.fileWithTtl0
+    val artifact = Artifact(pUri.value.toString).withChanging(true)
+    fileCache.fileWithTtl0(artifact)
+      .left
+      .map(err =>
+        new MalformedDirectiveError(err.describe, pUri.positions)
+      ) // TODO: better error type
```
Gedochao: just a new error class would do, I think

Gedochao (commented):
It's fine to skip `git+ssh` IMO, we don't support that anywhere else, either. 
The solution looks good so far, just need to address the TODOs and add tests

As for tests, maybe can reuse stuff from `scala.cli.integration.RunGistTestDefinitions`.

modules/build/src/main/scala/scala/build/CrossSources.scala
ivan-klass: @Gedochao but `scala.cli.commands.shared.SharedOptions.coursierCache` is in `cli` module, while we're in `build` here - so that definition is inaccessible (build doesn't depend on cli)

ivan-klass: That's why we haven't picked it up and left a todo

website/docs/reference/directives.md
```diff
@@ -126,6 +126,8 @@ Manually add sources to the project. Does not support chaining, sources are adde
 #### Examples
 `//> using file utils.scala`
 
+`//> using file https://raw.githubusercontent.com/softwaremill/sttp/refs/heads/master/examples/src/main/scala/sttp/client4/examples/json/GetAndParseJsonCatsEffectCirce.scala`
```
ivan-klass, line 129: @Gedochao I think it would be nice to reference a file within `scala-cli` repo itself

modules/build/src/main/scala/scala/build/CrossSources.scala
Gedochao: ...well, yes, but `cli` depends on `build`.
so what I mean here is that the way to go would be to pass it all the way from `cli` to `CrossSources`.

website/docs/reference/directives.md
Gedochao: I mean... dunno if it matters, as long as the example is useful, to be honest. 🤷

modules/build/src/main/scala/scala/build/CrossSources.scala
Gedochao: if you just leave a TODO, it's very likely to get forgotten.

Gedochao: if you don't have time to fix it, please at least create a ticket for the follow-up and link it back in the code.

ivan-klass: @Gedochao, I think I understand what you mean by passing a download handler from the external. Could you please review the recent change?

Gedochao: Yup, looks good!

Gedochao (commented):
Looks good, we just need a test at this point.
As noted earlier, you should be able to reuse stuff from `scala.cli.integration.RunGistTestDefinitions`.

ivan-klass:
@Gedochao thanks, please take a look on the test added

Gedochao (approved):
LGTM

PR #3697, opened 2025-05-26
Fix Changing Java versions document

tmrkw1497:
The documentation has been fixed because adopt open jdk does not support m1 mac and an error occurred when executing the command.


Gedochao:
Please rebase on top of the `stable` branch.

Gedochao (approved):

PR #3700, opened 2025-05-27
bugfix: Fix mill script on fish

tgodzik:


mill
```diff
@@ -3,6 +3,7 @@
 # Adapted from
 
 coursier_version="2.1.24"
+COMMAND=$@
```
tgodzik, line 6: This is a mystery, on BASH it should always be the program arguments which on fish it's last invoked command's one. When running bash script from fish it seems to be using fish semantics or maybe it's because cs is being invoked in fish and it's changing the variable? 😓

Gedochao (approved):

PR #3701, opened 2025-05-28
Make `test-fish-shell`, `test-hypothetical-sbt-export` and `bloop-memory-footprint` required for publishing

Gedochao:
It was theoretically possible to make a release with those 3 jobs failing.


tgodzik:
> It was theoretically possible to make a release with those 3 jobs failing.

I wouldn't bother about fish, but it also shouldn't fail.

tgodzik (approved):

PR #3702, opened 2025-05-30
[NIT] Refactor Scala CLI CI scripts

Gedochao:


.github/scripts/check-cross-version-deps.sc
```diff
@@ -18,14 +19,22 @@ for { module <- modules } {
...
+      case _                                                   => invalidOrgAndName
     }
+    .filterNot { case (key, _) => key == invalidOrgAndName }
     .map { case (key, entries) =>
-      key -> entries.map { case (_, scalaDepSuffixRegex(_, scalaVersion), _) =>
-        scalaVersion
-      }.distinct
+      val invalidVersion = "invalid"
+      key -> entries
+        .map {
```
tgodzik: Why not do collect?

.github/scripts/check-cross-version-deps.sc
```diff
@@ -18,14 +19,22 @@ for { module <- modules } {
...
+  val invalidOrgAndName = "invalid:invalid"
   val scalaVersionsByOrgAndName = deps
-    .groupBy { case (org, scalaDepSuffixRegex(nameWithoutSuffix, _), _) =>
-      s"$org:$nameWithoutSuffix"
+    .groupBy {
+      case (org, scalaDepSuffixRegex(nameWithoutSuffix, _), _) => s"$org:$nameWithoutSuffix"
+      case _                                                   => invalidOrgAndName
     }
+    .filterNot { case (key, _) => key == invalidOrgAndName }
     .map { case (key, entries) =>
```
tgodzik:
```suggestion
    .collect { case (key, entries) if key != invalidOrgAndName =>
```

tgodzik (changes requested):

.github/scripts/check-cross-version-deps.sc
Gedochao: indeed, why not.
fixed.

tgodzik (approved):

PR #3726, opened 2025-06-07
fix for 3725, 3752, 3766 and 3769

### LINKED ISSUE #3725: Appearance of "!#" in a script causes compile to fail
**Version(s)**
Scala CLI version: 1.8.1

**Describe the bug**
The following script fails if it has a `.sc` filename extension:
```scala

object CliBug {
  // !#  // <<<< this is the culprit
    printf("Hello\n")
  }
}
```
Here's the compile error:
```scala
[error] .\jsrc/.scala-build/jsrc_42edccb669-3b332a4750/src_generated/main/cliBug.scala:16:1
[error] eof expected, but '}' found
[error] }
[error] ^
[error] .\jsrc/.scala-build/jsrc_42edccb669-3b332a4750/src_generated/main/cliBug.scala:4:12
[error] Not found: cliBug_sc
[error] def args = cliBug_sc.args$
[error]            ^^^^^^^^^
```
If either of the characters characters '!#' are overwritten with something else, the problem doesn't appear.

A diff of the expected `.scala-build` cliBug.scala code and the defective code is this:
```diff
# diff /tmp/cliBug.scala jsrc/.scala-build/jsrc_42edccb669-3b332a4750/src_generated/main/cliBug.scala
9,10c9,10
< object CliBug {
<   // "x#"
---
>
>
34c34
<     script.CliBug.main(args) // hashCode to clear scalac warning about pure expression in statement position
---
>     val _ = script.hashCode() // hashCode to clear scalac warning about pure expression in statement position
```
It seems likely that the problem is [here](https://github.com/VirtusLab/scala-cli/blob/bb9c9f8fe1b5980439e253bd48edffd613abdeb9/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala#L13)

**To Reproduce**
The appearance of "!#" is the trigger, even if within a comment.
If the script is renamed to have the `.scala` filename extension, there is no error.

**Expected behaviour**
Should be allowed to have "!#" appear within a script

### LINKED ISSUE #3752: shebang line treated as a regex when partitioning shebang section
**Version(s)**
Scala CLI version: 1.8.1

**Describe the bug**

Crash on legal `/usr/bin/env -S ...` expression in the `shebang` line caused by shebang line being interpreted as a regular expression [here](https://github.com/VirtusLab/scala-cli/blob/646bbe5fcc79dd5209d51791460774927507107e/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala#L16)

**To Reproduce**
create a script with a `/usr/bin/env` shebang line that references a varname, e.g., `@{HOME}/atfile`.
The following command line demonstrates the expected behavior. 
```bash
# echo '[${HOME}/atfile]'
[${HOME}/atfile]

# /usr/bin/env -S 'echo [${HOME}/atfile]'
[/home/philwalk/atfile]
```

**Expected behaviour**
Should expand the varname within the `shebang` line.

### LINKED ISSUE #3766: sclicheck injects extra shebang line in scripts having a shebang
**Version(s)**
current release 1.8.1

**Describe the bug**
`sclicheck` accidentally injects two shebang lines into generated test code, but the bug is hidden by the `SheBang` regex.

An example of the bug is this generated code, from `website/docs/cookbooks/introduction/scala-scripts.md` at line 35 (with some intermediate `//fakeline`s removed for brevity).
```scala
//fakeline

//fakeline
//fakeline
//fakeline
[...]
//fakeline
//fakeline
//fakeline

val sv = scala.util.Properties.versionNumberString

def printMessage(): Unit =
  val message = s"Hello from Scala ${sv}, Java ${System.getProperty("java.version")}"
  println(message)

printMessage()
```

**To Reproduce**
due to #3725. the only way in the current code to see the problem is to examine the generated `HelloScriptSheBang.sc` code.   After #3725 is merged, this bug shows up as failed docs-tests.

### LINKED ISSUE #3769: ActionableDiagnosticTests invalid url String test failures in Windows
**Version(s)**
Scala CLI version: 1.8.3

**Describe the bug**
A temp directory is incorrectly converted to a URL string in Windows [here](https://github.com/VirtusLab/scala-cli/blob/0bda9f068e0322cdaf6f6f4b0f37dc0f7fda3c2b/modules/build/src/test/scala/scala/build/tests/ActionableDiagnosticTests.scala#L163) and [here](https://github.com/VirtusLab/scala-cli/blob/0bda9f068e0322cdaf6f6f4b0f37dc0f7fda3c2b/modules/build/src/test/scala/scala/build/tests/ActionableDiagnosticTests.scala#L225)

**To Reproduce**
```sh
mill -i build-module[].test.test "scala.build.tests.ActionableDiagnosticTests.*"

[1026] ==> X scala.build.tests.ActionableDiagnosticTests.actionable actions suggest update only to stable version  0.121s java.net.MalformedURLException: URL file:C:\Users\philwalk\AppData\Local\Temp\scala-cli-tests-actionable-diagnostic-repo1321821012424301336/org/scala-lang/scala3-compiler_3/3.7.1/scala3-compiler_3-3.7.1.pom doesn't contain an absolute path
[1026] ==> X scala.build.tests.ActionableDiagnosticTests.actionable actions should not suggest update to previous version  0.121s java.net.MalformedURLException: URL file:C:/Users/philwalk/AppData/Local/Temp/scala-cli-tests-actionable-diagnostic-repo7309957884627832308/org/scala-lang/scala3-compiler_3/3.7.1/scala3-compiler_3-3.7.1.pom doesn't contain an absolute path
```

**Expected behaviour**
Tests should succeed when running in Windows

Note: is fixed in #3726

philwalk:
The `Regex` that is used to partition the `shebang` section from script code is too permissive.
The end marker must appear as the first 2 characters of the line it appears on.

The cause of bug #3725 is that the `partitionShebangSection` grabs part of the scala code [here](https://github.com/VirtusLab/scala-cli/blob/bb9c9f8fe1b5980439e253bd48edffd613abdeb9/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala#L13).
The fix is to require that the `Regex` substring `!#` is anchored at the beginning of a line.

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
+++ b/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
@@ -18,7 +18,7 @@ object AmmUtil {
 
   def encodeScalaSourcePath(path: Seq[Name]) = path.map(_.backticked).mkString(".")
 
-  def normalizeNewlines(s: String)(using newLine: String = lineSeparator): String =
+  def normalizeNewlines(s: String): String =
     s.replace("\r", "")
 
   lazy val lineSeparator: String = "\n"
--- a/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
+++ b/modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
@@ -18,5 +18,8 @@ object AmmUtil {
 
   def encodeScalaSourcePath(path: Seq[Name]) = path.map(_.backticked).mkString(".")
 
-  def normalizeNewlines(s: String) = s.replace("\r", "").replace("\n", System.lineSeparator())
+  def normalizeNewlines(s: String)(using newLine: String = lineSeparator): String =
+    s.replace("\r", "")
+
+  lazy val lineSeparator: String = "\n"
 }
--- a/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
@@ -308,24 +308,31 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
           |#! nix-shell -i scala-cli
           |
           |println("Hello World")""".stripMargin,
-      os.rel / "something4.scala" ->
+      os.rel / "something4.sc" ->
         """#!/usr/bin/scala-cli
           |#! nix-shell -i scala-cli
           |
           |!#
           |
           |println("Hello World")""".stripMargin,
-      os.rel / "something5.scala" ->
+      os.rel / "something5.sc" ->
         """#!/usr/bin/scala-cli
           |
           |println("Hello World #!")""".stripMargin,
-      os.rel / "multiline.scala" ->
+      os.rel / "multiline.sc" ->
         """#!/usr/bin/scala-cli
           |# comment
           |VAL=1
           |!#
           |
-          |println("Hello World #!")""".stripMargin
+          |println("Hello World #!")""".stripMargin,
+      os.rel / "hasBangHashInComment.sc" ->
+        """#!/usr/bin/scala-cli
+          |
+          |
+          |
+          |
+          |println("Hello World !#")""".stripMargin
     )
     val expectedParsedCodes = Seq(
       """
@@ -349,7 +356,13 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
         |
         |
         |
-        |println("Hello World #!")""".stripMargin
+        |println("Hello World #!")""".stripMargin,
+      """
+        |
+        |
+        |
+        |
+        |println("Hello World !#")""".stripMargin
     )
 
     testInputs.withInputs { (root, inputs) =>
@@ -375,9 +388,19 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
         sources.inMemory.map(_.content).map(s => new String(s, StandardCharsets.UTF_8))
 
       parsedCodes.zip(expectedParsedCodes).foreach { case (parsedCode, expectedCode) =>
+        showDiff(parsedCode, expectedCode)
         expect(parsedCode.contains(expectedCode))
       }
     }
+  }
+  def showDiff(parsed: String, expected: String): Unit = {
+    if (!parsed.contains(expected))
+      for (((p, e), i) <- (parsed zip expected).zipWithIndex) {
+        val ps = TestUtil.c2s(p)
+        val es = TestUtil.c2s(e)
+        if (ps != es)
+          System.err.printf("%2d: [%s]!=[%s]\n", i, ps, es)
+      }
   }
 
   test("dependencies in .sc - using") {
--- a/modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
+++ b/modules/build/src/main/scala/scala/build/internal/markdown/MarkdownCodeWrapper.scala
@@ -115,15 +115,15 @@ object MarkdownCodeWrapper {
       val fence: MarkdownCodeBlock = snippets(index)
       val classOpener: String      =
         if (index == 0)
-          s"object ${scopeObjectName(scopeIndex)} {${System.lineSeparator()}" // first snippet needs to open a class
+          s"object ${scopeObjectName(scopeIndex)} {${AmmUtil.lineSeparator}" // first snippet needs to open a class
         else if (fence.resetScope)
-          s"}; object ${scopeObjectName(scopeIndex)} {${System.lineSeparator()}" // if scope is being reset, close previous class and open a new one
-        else System.lineSeparator()
+          s"}; object ${scopeObjectName(scopeIndex)} {${AmmUtil.lineSeparator}" // if scope is being reset, close previous class and open a new one
+        else AmmUtil.lineSeparator
       val nextScopeIndex = if index == 0 || fence.resetScope then scopeIndex + 1 else scopeIndex
-      val newAcc         = acc + (System.lineSeparator() * (fence.startLine - line - 1)) // padding
+      val newAcc         = acc + (AmmUtil.lineSeparator * (fence.startLine - line - 1)) // padding
         .:++(classOpener) // new class opening (if applicable)
         .:++(fence.body) // snippet body
-        .:++(System.lineSeparator()) // padding in place of closing backticks
+        .:++(AmmUtil.lineSeparator) // padding in place of closing backticks
       generateMainScalaLines(
         snippets = snippets,
         index = index + 1,
@@ -154,9 +154,9 @@ object MarkdownCodeWrapper {
     if index >= snippets.length then acc
     else {
       val fence: MarkdownCodeBlock = snippets(index)
-      val newAcc = acc + (System.lineSeparator() * (fence.startLine - line)) // padding
+      val newAcc = acc + (AmmUtil.lineSeparator * (fence.startLine - line)) // padding
         .:++(fence.body) // snippet body
-        .:++(System.lineSeparator()) // padding in place of closing backticks
+        .:++(AmmUtil.lineSeparator) // padding in place of closing backticks
       generateRawScalaLines(snippets, index + 1, fence.endLine + 1, newAcc)
     }
 }
--- a/modules/build/src/main/scala/scala/build/internal/markdown/MarkdownOpenFence.scala
+++ b/modules/build/src/main/scala/scala/build/internal/markdown/MarkdownOpenFence.scala
@@ -39,7 +39,8 @@ case class MarkdownOpenFence(
     val start: Int               = tickStartLine + 1
     val bodyLines: Array[String] = lines.slice(start, tickEndLine)
     val body                     = bodyLines.mkString("\n")
-    val (bodyWithNoSheBang, _)   = SheBang.ignoreSheBangLines(body)
+
+    val (bodyWithNoSheBang, _, _) = SheBang.ignoreSheBangLines(body)
     MarkdownCodeBlock(
       info.split("\\s+").toList, // strip info by whitespaces
       bodyWithNoSheBang,
--- a/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/ScalaPreprocessor.scala
@@ -171,7 +171,8 @@ case object ScalaPreprocessor extends Preprocessor {
     allowRestrictedFeatures: Boolean,
     suppressWarningOptions: SuppressWarningOptions
   )(using ScalaCliInvokeData): Either[BuildException, Option[ProcessingOutput]] = either {
-    val (contentWithNoShebang, _)                = SheBang.ignoreSheBangLines(content)
+    val (contentWithNoShebang, _, _) = SheBang.ignoreSheBangLines(content)
+
     val extractedDirectives: ExtractedDirectives = value(ExtractedDirectives.from(
       contentChars = contentWithNoShebang.toCharArray,
       path = path,
@@ -210,7 +211,7 @@ case object ScalaPreprocessor extends Preprocessor {
       logger
     )
 
-    val (content0, isSheBang)                          = SheBang.ignoreSheBangLines(content)
+    val (content0, isSheBang, _)                       = SheBang.ignoreSheBangLines(content)
     val preprocessedDirectives: PreprocessedDirectives =
       value(DirectivesPreprocessor(
         path,
--- a/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
+++ b/modules/build/src/main/scala/scala/build/preprocessing/SheBang.scala
@@ -1,38 +1,47 @@
 package scala.build.preprocessing
 
-import scala.util.matching.Regex
-
 object SheBang {
-  private val sheBangRegex: Regex = s"""(^(#!.*(\\X)?)+(\\X*!#.*)?)""".r
+  def isShebangScript(content: String): Boolean = content.startsWith("#!")
 
-  def isShebangScript(content: String): Boolean = sheBangRegex.unanchored.matches(content)
-
-  /** Returns the shebang section and the content without the shebang section */
-  def partitionOnShebangSection(content: String): (String, String) =
+  def partitionOnShebangSection(content: String): (String, String, String) =
     if (content.startsWith("#!")) {
-      val regexMatch = sheBangRegex.findFirstMatchIn(content)
-      regexMatch match {
-        case Some(firstMatch) =>
-          (firstMatch.toString(), content.replaceFirst(firstMatch.toString(), ""))
-        case None => ("", content)
+      val splitIndex = content.indexOf("\n!#") match {
+        case -1 =>
+          val eolIndex = content.indexOf("\n")
+          content.drop(eolIndex + 1) match {
+            case s if s.startsWith("#!") =>
+              eolIndex + s.indexOf("\n") + 1 // skip over #! nix-shell line
+            case _ =>
+              eolIndex + 1
+          }
+        case index =>
+          var i = index + 1
+          while (i < content.length && content(i) != '\n') i += 1
+          i + 1 // split at start of subsequent line
       }
+      val newLine: String = content.drop(splitIndex - 2).take(2) match {
+        case CRLF => CRLF
+        case _    => "\n"
+      }
+      val (header, body) = content.splitAt(splitIndex)
+      (header, body, newLine)
     }
     else
-      ("", content)
+      ("", content, lineSeparator(content))
 
-  def ignoreSheBangLines(content: String): (String, Boolean) =
-    if (content.startsWith("#!")) {
-      val regexMatch = sheBangRegex.findFirstMatchIn(content)
-      regexMatch match {
-        case Some(firstMatch) =>
-          content.replace(
-            firstMatch.toString(),
-            System.lineSeparator() * firstMatch.toString().split(System.lineSeparator()).length
-          ) -> true
-        case None => (content, false)
-      }
-    }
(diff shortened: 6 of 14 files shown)


philwalk:
Am working on an integration test to verify the fix.

philwalk:
I added two tests, and ~I assume I can run them~ with the following command line:
```bash
 ./mill integration.test.jvm 'scala.build.tests.SourcesTests' 'scala.build.tests.markdown.MarkdownCodeBlockTests'
```
The output shows no error messages, but it doesn't report pass/fail either:
```scala
[1237/1339] test-runner[3.3.6].publishLocalNoFluff
[1237] Publishing Artifact(org.virtuslab.scala-cli,test-runner_3,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1238/1339] runner[3.3.6].publishLocalNoFluff
[1238] Publishing Artifact(org.virtuslab.scala-cli,runner_3,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1239/1339] runner[2.13.16].publishLocalNoFluff
[1239] Publishing Artifact(org.virtuslab.scala-cli,runner_2.13,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1240/1339] test-runner[2.13.16].publishLocalNoFluff
[1240] Publishing Artifact(org.virtuslab.scala-cli,test-runner_2.13,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1241/1339] runner[2.12.20].publishLocalNoFluff
[1241] Publishing Artifact(org.virtuslab.scala-cli,runner_2.12,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1242/1339] test-runner[2.12.20].publishLocalNoFluff
[1242] Publishing Artifact(org.virtuslab.scala-cli,test-runner_2.12,0.0.1-SNAPSHOT) to ivy repo /home/philwalk/workspace/scala-cli-phil/out/repo/0.0.1-SNAPSHOT
[1111/1339] integration.compile
[1111] [info] compiling 1 Scala source to /home/philwalk/workspace/scala-cli-phil/out/integration/compile.dest/classes ...
[1111] [info] done compiling
[1332/1339] integration.test.compile
[1332] [info] compiling 154 Scala sources to /home/philwalk/workspace/scala-cli-phil/out/integration/test/compile.dest/classes ...
[1332] [warn] /home/philwalk/workspace/scala-cli-phil/modules/integration/src/test/scala/scala/cli/integration/ExportScalaOrientedBuildToolsTestDefinitions.scala:28:42: parameter testArgs in method testZioTest is never used
[1332] [warn]   def testZioTest(testClassName: String, testArgs: Seq[String] = Nil): Unit = {
[1332] [warn]                                          ^
[1332] [warn] /home/philwalk/workspace/scala-cli-phil/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala:2358:57: Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method exitCode,
[1332] [warn] or remove the empty argument list from its definition (Java-defined methods are exempt).
[1332] [warn] In Scala 3, an unapplied method like this will be eta-expanded into a function. [quickfixable]
[1332] [warn]             processes.foreach { case (p, _) => expect(p.exitCode == 0) }
[1332] [warn]                                                         ^
[1332] [warn] /home/philwalk/workspace/scala-cli-phil/modules/integration/src/test/scala/scala/cli/integration/BspSuite.scala:170:14: method destroyForcibly in class SubProcess is deprecated: Use destroy(shutdownGracePeriod = 0)
[1332] [warn]         proc.destroyForcibly()
[1332] [warn]              ^
[1332] [warn] /home/philwalk/workspace/scala-cli-phil/modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala:254:14: method destroyForcibly in class SubProcess is deprecated: Use destroy(shutdownGracePeriod = 0)
[1332] [warn]         proc.destroyForcibly()
[1332] [warn]              ^
[1332] [warn] /home/philwalk/workspace/scala-cli-phil/modules/integration/src/test/scala/scala/cli/integration/TestUtil.scala:371:34: method destroyForcibly in class SubProcess is deprecated: Use destroy(shutdownGracePeriod = 0)
[1332] [warn]         if (proc.isAlive()) proc.destroyForcibly()
[1332] [warn]                                  ^
[1332] [warn] 5 warnings found
[1332] [info] done compiling
[1339/1339] =============== integration.test.jvm scala.build.tests.Sourc....build.tests.markdown.MarkdownCodeBlockTests =============== 58s
(base) philwalk@d5:~/workspace/scala-cli-phil$ 
```

philwalk:
~I'm not sure why the `SourcesTests` is failing, AFAICT I left it the way I found it and appended an additional test to the `testInputs` Seq` and a corresponding additional `expectedParseCodes` value.~

It takes 6 minutes to run `./mill 'build-module[3.7.1].test.test'`, if someone can suggest a shortcut that runs minumum number of tests, it would be welcome.   I'll see if I can figure this out tomorrow.

philwalk:
A more subtle fix is needed, working on it.

Gedochao:
> It takes 6 minutes to run `./mill 'build-module[3.7.1].test.test'`, if someone can suggest a shortcut that runs minumum number of tests, it would be welcome. I'll see if I can figure this out tomorrow.

I'm guessing you need something like
```bash
./mill 'build-module[].test.test' 'scala.build.tests.SourcesTests.*'
```
(you can tweak the regex for a specific test)

philwalk:
Here's a working version of the Regex:
```
  private val sheBangRegex: Regex = s"""(?m)^(#![^\\r\\n]*\\R){1,2}(.*?\\R)*?(^!#[^\\r\\n]*\\R)""".r
```
However, there are a handful of failing tests, some of which are "off-by-one" errors, so it will be awhile before I can push the fix.

philwalk:
Also, I have a non-regex equivalent version that is 3 to 5 times as fast, so let me know if there's a preference.

Gedochao:
> Also, I have a non-regex equivalent version that is 3 to 5 times as fast, so let me know if there's a preference.

I mean, you make it sound like the non-regex version is worth a look :shrug:

philwalk:
I need to benchmark it again ...
I'll fix the `SheBang.scala` formatting before Monday.

philwalk:
Of the 3 or 4 test failures from the Friday morning run that I've tried to duplicate,  none are failing on my system.   I won't be able to do a deep dive until tomorrow.

tgodzik:
No worries, @Gedochao is out until next Monday, so if you need help I can also try and help out.

philwalk:
All the `jvm-tests-*` are failing due to a single test, so should be cleared up soon.
I'll work on the docs test failure after that.

philwalk:
Here's a crude benchmark of the speedup factor versus file size, based on 999 scala scripts, none with multi-line shebang sections.

image

philwalk:
This PR also fixes #3752, discovered while testing various scripts.

philwalk:
I'm looking into the docs-tests failure ...

Gedochao:
@philwalk It seems they fail consistently:
```scala
docs-tests[3.3.6].test.test 6 tests failed: 
  sclicheck.DocTests.root release_notes sclicheck.DocTests.root release_notes
  sclicheck.DocTests.cookbook introduction/scala-scripts sclicheck.DocTests.cookbook introduction/scala-scripts
  sclicheck.DocTests.command run sclicheck.DocTests.command run
  sclicheck.DocTests.guide introduction/old-runner-migration sclicheck.DocTests.guide introduction/old-runner-migration
  sclicheck.DocTests.guide scripting/scripts sclicheck.DocTests.guide scripting/scripts
  sclicheck.DocTests.guide scripting/shebang sclicheck.DocTests.guide scripting/shebang
```
Let me know if you need help with those.

philwalk:
@Gedochao 
I'm currently testing in a WSL invironment.    The docs-tests don't seem to be designed for a Windows test environment, where the DockerFile ADDs the Windows `scala-cli` (batch file) to a Linux image before running.

~Is there a way to only run the failing docs-tests individually?   I haven't yet figured out how to do so, if there is.
If not, no problem.~   I figured it out ...

philwalk:
The current version mostly implements #3756, the next commit adds the last piece (in addition to fixing failing docs-tests).

philwalk:
One of the failing `docs-tests` is based on this generated script.
The error is due to the 2nd `shebang` line.
See #3766.

```scala
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline
//fakeline

val sv = scala.util.Properties.versionNumberString

def printMessage(): Unit =
  val message = s"Hello from Scala ${sv}, Java ${System.getProperty("java.version")}"
  println(message)

printMessage()
```

modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
```diff
@@ -18,5 +20,9 @@ object AmmUtil {
 
   def encodeScalaSourcePath(path: Seq[Name]) = path.map(_.backticked).mkString(".")
 
-  def normalizeNewlines(s: String) = s.replace("\r", "").replace("\n", System.lineSeparator())
+  def normalizeNewlines(s: String)(using newLine: String = lineSeparator): String =
+    s.replace("\r", "").replace("\n", newLine)
+
+  lazy val lineSeparator: String =
+    if EnvVar.Misc.shell.valueOpt.nonEmpty then "\n" else System.lineSeparator()
```
Gedochao: Wait... `\n` for all shells, arbitrarily?
why?
shouldn't you handle specific cases here?

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
```diff
@@ -20,6 +20,7 @@ case class AppCodeWrapper(scalaVersion: String, log: String => Unit) extends Cod
         ""
     val packageDirective =
       if (pkgName.isEmpty) "" else s"package ${AmmUtil.encodeScalaSourcePath(pkgName)}" + "\n"
+
```
Gedochao: ?

modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
```diff
@@ -375,10 +388,26 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
...
       }
     }
   }
+  def showDiff(parsed: String, expected: String): Unit = {
+    if (!parsed.contains(expected)) {
+      def c2s(c: Char): String = c match {
+        case '\r' => "\\r"
+        case '\n' => "\\n"
+        case s    => s"$s"
+      }
```
Gedochao: it feels like this part could be extracted `TestUtil`, as it's repeating

Gedochao (commented):
I can see the CI is green already, seems we're close with this being done. 
Just a few things.

modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
Gedochao: I'd imagine smth like
```suggestion
EnvVar.Misc.shell.valueOpt.map(_.split("[\\/]+").last).map {
          case ??? => "\n" // for whichever shell you'd like to handle
          case _  => System.lineSeparator()
        }.getOrElse(System.lineSeparator())
```

philwalk: Carriage return leads to errors:

- `^M: command not found`
- `/bin/bash^M: bad interpreter: No such file or directory`

 where the `\r` is interpreted as part of the command.

If SHELL is defined, 99% likelihood is `bash`, `sh`, `ksh`, `zsh` or `fish`

We could instead make it conditional on [`isShebangCapableShell`](https://github.com/VirtusLab/scala-cli/blob/8298cc06686abcf8bbe190b11cc484271bed99bc/modules/cli/src/main/scala/scala/cli/internal/ProcUtil.scala#L102) but that would be less effective in avoiding these types of errors, since new shells might become popular in the future.

The only case I'm aware of in which a Carriage Return is required is when passing command strings to `CMD.EXE`.  Otherwise, it's not required, even in Windows (my experience).

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
philwalk: the extra newline was unintentional, we can revert this.

modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
Gedochao: ...shouldn't we at least handle `cmd` here, then?

modules/build/src/main/scala/scala/build/internal/AppCodeWrapper.scala
Gedochao: It's a nitpick, but since you don't change anything else here...
```suggestion
```

modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
philwalk: The `AmmUtil.lineSeparator` change only affects line endings of generated wrapper code.  The line endings of wrapped client code aren't modified, so the semantics of client code w.r.t. line endings sent to the console or written to files, etc. aren't affected.   

This is mostly a cosmetic issue, visible in some text editors (see #3756).

I did some experiments with a native-compiled `scala-cli`.   Running a script in a CMD.EXE terminal, with or without SHELL defined had no effect on how the code ran, or on the output, as expected.

However, in Windows, the following tests will fail if `System.lineSeparator` is used, since the tests compare exact literal Strings against expected values:
```scala
[1026] =[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.a simple Scala code block is wrapped correctly  0.032s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.multiple plain Scala code blocks are wrapped correctly  0.003s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.multiple plain Scala code blocks with different scopes are wrapped correctly  0.003s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.a raw Scala code block is wrapped correctly  0.001s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.multiple raw Scala code blocks are glued together correctly  0.01s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.a test Scala snippet is wrapped correctly  0.009s java.lang.AssertionError: assertion failed
[1026] ==> X scala.build.tests.markdown.MarkdownCodeWrapperTests.multiple test Scala snippets are glued together correctly  0.013s java.lang.AssertionError: assertion failed
```

Line endings depend on where a String originates:

- client source code: "\n" in a SHELL environment, unpredictable otherwise
- Triple quote strings:  taken from the enclosing source file

The `scala-cli` source files are normalized to "\n" on commit (by .gitattributes "* text=auto"), so tripled quoted strings in our Test sources will not have `\r` except in some `git` client configurations.

Gedochao: I now understand even less on why do you rely on `EnvVar.Misc.shell` being defined here, however.
Shouldn't it then always be `\n`?

philwalk: > why do you rely on `EnvVar.Misc.shell`

The intent was that if testing from a `CMD.EXE` this causes it to revert to the previous behavior.   Not sure if anyone actually does that, because  some of the tests always fail in a Windows build, AFAIK.  It might be that most `scala-cli` Windows developers work in `WSL`.   

> Shouldn't it then always be `\n`

Your suggestion would certainly simplify things,  by eliminating `\r` from the equation in all cases.    I'll do some experiments.

philwalk: In a clone of this PR code (checked out with CRLF line endings), applied the suggested simplification, and tested in a git-for-windows bash session.   Because `EnvVar.Misc.shell` is no longer referenced, no need to attempt tests in a `CMD` session.   Tests pass in both git-for-windows and in `msys64` bash session with LF-only line endings.

Encountered and fixed #3725.

modules/build/src/main/scala/scala/build/internal/AmmUtil.scala
```diff
@@ -18,5 +18,8 @@ object AmmUtil {
 
   def encodeScalaSourcePath(path: Seq[Name]) = path.map(_.backticked).mkString(".")
 
-  def normalizeNewlines(s: String) = s.replace("\r", "").replace("\n", System.lineSeparator())
+  def normalizeNewlines(s: String)(using newLine: String = lineSeparator): String =
```
Gedochao:
```suggestion
  def normalizeNewlines(s: String): String =
```
`newLine` seems to no longer be used here, no?

Gedochao (approved):
LGTM, commented on one last nitpick

Issue #3742, opened 2025-06-12
New nightlies aren't being published to Sonatype Central


Gedochao:
**Version(s)**
1.8.1
1.8.1-17-gd48c691a8-SNAPSHOT

**Describe the bug**
It seems usage of `--cli-version nightly` no longer works as intended, after we migrated publishing to Sonatype Central Portal.

**To Reproduce**
`--cli-version nightly` points to an old version.
```scala
scala-cli --cli-version nightly -e 'println("Hello")' --native 
Fetching Scala CLI 1.8.1-1-gc0e412a9a-SNAPSHOT
Downloading compiler plugin org.scala-native:::nscplugin:0.5.7
Downloading 2 dependencies
Compiling project (Scala 3.7.1, Scala Native 0.5.7)
Compiled project (Scala 3.7.1, Scala Native 0.5.7)
...
```
1.8.1-17-gd48c691a8-SNAPSHOT got released sufficiently earlier (also notice that it's the 17th nightly after 1.8.1): https://github.com/VirtusLab/scala-cli/actions/runs/15588192852/job/43929147790#step:6:470

Pointing to the new nightly version explicitly also fails:
```scala
scala-cli --cli-version 1.8.1-17-gd48c691a8-SNAPSHOT -e 'println("Hello")' --native
Fetching Scala CLI 1.8.1-17-gd48c691a8-SNAPSHOT
Error downloading org.virtuslab.scala-cli:cli_3:1.8.1-17-gd48c691a8-SNAPSHOT
  not found: https://repo1.maven.org/maven2/org/virtuslab/scala-cli/cli_3/1.8.1-17-gd48c691a8-SNAPSHOT/cli_3-1.8.1-17-gd48c691a8-SNAPSHOT.pom
  not found: https://oss.sonatype.org/content/repositories/snapshots/org/virtuslab/scala-cli/cli_3/1.8.1-17-gd48c691a8-SNAPSHOT/cli_3-1.8.1-17-gd48c691a8-SNAPSHOT.pom
  No fallback URL found
  not found: /Users/pchabelski/.ivy2/local/org.virtuslab.scala-cli/cli_3/1.8.1-17-gd48c691a8-SNAPSHOT/ivys/ivy.xml
```

**Expected behaviour**
I'd expect the new nightlies to be available under `--cli-version` and picked up automatically.

Gedochao:
It seems it's not just nightlies, v1.8.2 stable release is also affected.
[...]

PR #3745, opened 2025-06-13
Add extra logging on Scala CLI artifacts publishing

Gedochao:
Adding some extra logging as a sanity check for:
- #3742

### DIFF
--- a/build.mill.scala
+++ b/build.mill.scala
@@ -1344,6 +1344,11 @@ object `local-repo` extends LocalRepo {
 
 // Helper CI commands
 def publishSonatype(tasks: mill.main.Tasks[PublishModule.PublishData]) = Task.Command {
+  val taskNames = tasks.value.map(_.toString())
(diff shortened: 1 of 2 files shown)


build.mill.scala
```diff
@@ -1344,6 +1344,11 @@ object `local-repo` extends LocalRepo {
 
 // Helper CI commands
 def publishSonatype(tasks: mill.main.Tasks[PublishModule.PublishData]) = Task.Command {
+  val taskNames = tasks.value.map(_.toString())
+  System.err.println(
+    s"""Tasks producing artifacts to be included in the bundle:
```
tgodzik, line 1349: We could get the suze around here, no?

tgodzik (approved):

build.mill.scala
Gedochao: nope... the actual bundle is created much later, inside of Mill API... 😕

PR #3753, opened 2025-06-23
Prepare artifacts in a separate step before they are published to Sonatype

Gedochao:
Relevant to:
- https://github.com/VirtusLab/scala-cli/issues/3742

It will be silly if this solves it.
I basically want to make sure that all the `publishArtifacts` jobs have finished before we start publishing to Sonatype.


Gedochao:
It's been suggested to try `-j1` instead to test for potential race conditions. 
Let me open another PR.

Gedochao:
Trying https://github.com/VirtusLab/scala-cli/pull/3755 first.

PR #3755, opened 2025-06-23
Ensure publishing to Sonatype runs without parallelism

Gedochao:
Relevant to:
- https://github.com/VirtusLab/scala-cli/issues/3742

As suggested in https://discord.com/channels/632150470000902164/940067748103487558/1386702655782387783


tgodzik (approved):
Let's check it out!

PR #3774, opened 2025-07-08
`publish` command with Sonatype Central Portal OSSRH Staging API

Gedochao:
So long story short, I mean to try this out:
https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/


tgodzik (approved):
LGTM Let's test it out!

PR #3776, opened 2025-07-11
Add support for publishing with Scala CLI to Sonatype Central Portal

### LINKED ISSUE #2771: Sonatype's new central portal user's token is not working in scala-cli
**Version(s)**
Scala CLI version: 1.1.3
Scala version (default): 3.3.1

**Describe the bug**
Sonatype's new identity management system requires new user's to register via central portal and use user's token instead of username/password. I couldn't make it to work with scala-cli publishing.

- https://central.sonatype.org/publish-ea/publish-ea-guide/
- https://central.sonatype.org/publish/publish-portal-maven/

**To Reproduce**
There is a Github Action failing here: https://github.com/encalmo/graphs/actions/runs/8125371977/job/22207911819

**Expected behaviour**
Support passing user's token generated from https://central.sonatype.com/account

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/2771

The actual fix was delivered here:
- https://github.com/coursier/publish/pull/127

Also note that this doesn't cover publishing `*-SNAPSHOT` versions, which are tracked separately here:
- https://github.com/VirtusLab/scala-cli/issues/3777

### DIFF
--- a/project/deps/package.mill.scala
+++ b/project/deps/package.mill.scala
@@ -115,7 +115,7 @@ object Deps {
     def coursier                          = coursierDefault
     def coursierCli                       = coursierDefault
     def coursierM1Cli                     = coursierDefault
-    def coursierPublish                   = "0.4.0"
+    def coursierPublish                   = "0.4.1"
     def jmh                               = "1.37"
     def jsoniterScalaJava8                = "2.13.5.2"
     def jsoup                             = "1.21.1"
--- a/project/deps/package.mill.scala
+++ b/project/deps/package.mill.scala
@@ -129,7 +129,7 @@ object Deps {
     def maxScalaNativeForScalaPy          = scalaNative04
     def maxScalaNativeForMillExport       = scalaNative05
     def scalaPackager                     = "0.1.33"
-    def signingCli                        = "0.2.9"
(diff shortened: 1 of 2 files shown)


project/deps/package.mill.scala
```diff
@@ -300,7 +300,7 @@ object Docker {
 def customRepositories =
   Seq(
     coursier.Repositories.sonatype("snapshots"),
-    coursier.MavenRepository("https://s01.oss.sonatype.org/content/repositories/snapshots")
-    // Uncomment for local development
-    // coursier.LocalRepositories.Dangerous.maven2Local
+    coursier.MavenRepository("https://s01.oss.sonatype.org/content/repositories/snapshots"),
+//     Uncomment for local development
+     coursier.LocalRepositories.Dangerous.maven2Local
```
tgodzik: Should this be left uncommented?

tgodzik (approved):

project/deps/package.mill.scala
Gedochao: nope, good catch...

Gedochao: fixed

PR #3790, opened 2025-07-17
fix for 3789

philwalk:
This PR recognizes valid scala scripts without an extension, based on the presence of a hash-bang line, regardless of the terminal type.    It prevents rejecting a run from a Windows CMD.EXE terminal.


philwalk:
closing this for now, need to analyze cause of failed tests.

philwalk:
Manual testing with a generated native `scala-cli` appears to fix the bug, but because numerous tests failed, it may take some time to assess whether the fix is invalid or perhaps the failing tests are brittle.

PR #3794, opened 2025-07-18
fix for 3789 - updated

### LINKED ISSUE #3789: unable to execute script without an extension in Windows
**Version(s)**
# scala-cli version
Scala CLI version: 1.8.4

**Describe the bug**
Given the following script, with path "C:\scriptNoExtension":
```scala
printf("%s\n", scriptPath)
```

UPDATE: this issue is only Windows-specific w.r.t. there being a less informative error message in a CMD.EXE terminal than in a SHELL session.   The bug is that there is inconsistent treatment of scripts identified by `.sc` extension versus those with a valid `shebang` headers, and it applies to all environments.

Each the following should be a valid way to run the script, although only the first two examples succeed:

```scala
scala-cli /tmp/scriptPath.sc
/tmp/scriptPath.sc
```
```scala
c:\opt\ue>scala-cli.bat \tmp\scriptPath.sc
\tmp\scripPath.sc
```
```scala
scala-cli /tmp/scriptNoExtension
[error]  /tmp/scriptNoExtension: scripts with no file extension should be run with the 'shebang' sub-command.
  C:\Users\philwalk\AppData\Local\Coursier\data\bin\.scala-cli.aux.exe shebang /tmp/scripPath
```
```scala
C:\tmp>scala-cli scriptNoExtension
[error]  scriptNoExtension: unrecognized source type (expected .scala or .sc extension, or a directory).
```

When attempting to run the script with no extension in a CMD session, the error message implies that it's not a valid script.    If the same command line is issued from a SHELL session, the error message is more informative, and suggests a workaround:

```
C:\tmp> scala-cli shebang c:\tmp\scriptNoExtension
C:\tmp\scriptNoExtension
```

**Expected behaviour**
```scala
C:\tmp>scala-cli.exe c:\tmp\scriptNoExtension
Compiling project (Scala 3.7.1, JVM (17))
Compiled project (Scala 3.7.1, JVM (17))
c:\tmp\scriptNoExtension
```

philwalk:
updated two other script tests that only tested for ".sc" extension.
now it also checks for shebang.


modules/build/src/main/scala/scala/build/input/ElementsUtils.scala
```diff
@@ -8,6 +8,11 @@ import scala.build.Directories
 import scala.build.internal.Constants
 
 object ElementsUtils {
+  extension (p: os.Path) {
+    def hasShebang: Boolean = os.read.bytes(p, offset = 0, count = 2) == Array('#', '!')
+    def isScript: Boolean   = p.ext == "sc" || p.hasShebang
```
Gedochao:
```suggestion
    def isScript: Boolean   = (p.ext == "sc" || p.hasShebang) && p.ext != "scala")
```

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -573,6 +573,21 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
       }
     }
 
+  test("script file with shebang header and no extension run with scala-cli") {
+    val msg    = "compiled and ran"
+    val inputs = TestInputs(
+      os.rel / "scriptWithShebang" ->
+        s"""|#!/usr/bin/env -S ${TestUtil.cli.mkString(" ")} shebang -S 2.13
```
```suggestion
        s"""|#!/usr/bin/env -S ${TestUtil.cli.mkString(" ")} shebang
```
Why override the Scala version here?

Gedochao (approved):
left a couple nitpick comments.

PR #3802, opened 2025-07-24
script extension can be either '.sc' or empty string

philwalk:
File with no extension only considered a script if file has a shebang header.
This is a follow-up to #3794 which passed CI tests but implemented a less restrictive script definition.

A script with a valid shebang header and no extension is recognized as a valid script even when launched from a shell that isn't shebang capable.  Consequently, it can be run when directly passed to `scala-cli` without the `shebang` command, just like a script with '.sc' extension.

A side-effect of the less restrictive script definition in #3794 is that the wrapper class for a script with `.scala` extension also has a `def scriptPath` defined, which is nice.   However, without knowing all the potential side-effects of that change, it seems safer to restrict the scope here.


philwalk:
@Gedochao - At least some of these test failures are download problems:

```
download error: Caught java.io.IOException (Server returned HTTP response code: 503 for URL: https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.3.1/graalvm-ce-java17-linux-amd64-22.3.1.tar.gz) while downloading https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.3.1/graalvm-ce-java17-linux-amd64-22.3.1.tar.gz
```

```
Run VirtusLab/scala-cli-setup@v1
Install Coursier
Error: Unexpected HTTP response: 503
```

All but 1 of the 78 failures are passing on my Linux box.
The one failing test is `'sclicheck.GifTests.complete-install' which times out.

Reading the first 2 bytes of a script file to compare against `#!` is problematic if the file is a pipe or a stream, so it's been disabled when when processing all files in a directory.  I'm not quite sure I fully understand that use-case, so I don't know if that will be a problem.

Let me know if I should squash everything to a single commit.

modules/integration/src/test/scala/scala/cli/integration/RunScriptTestDefinitions.scala
```diff
@@ -574,17 +574,20 @@ trait RunScriptTestDefinitions { _: RunTestDefinitions =>
...
+    val scriptName = "scriptWithShebang"
+    val inputs     = TestInputs(
+      os.rel / scriptName ->
         s"""|#!/usr/bin/env -S ${TestUtil.cli.mkString(" ")} shebang
             |//> using scala $actualScalaVersion
-            |println(s"$msg")""".stripMargin
+            |println(s"$expected")""".stripMargin
     )
+    val cmd = TestUtil.cli ++ Seq(scriptName)
+    System.err.printf("### cmd[%s]\n", cmd.mkString(" "))
```
```suggestion
```

Gedochao (approved):
LGTM
I'll squash the commits when merging, no worries.

philwalk:
The 2 failing tests pass on my `aarch64` mac f.w.i.w.

-  'scala.cli.integration.BspTests213.setup-ide doesn't pass unrecognised arguments to old --cli-versions: 1.5.0-19-g932866db6-SNAPSHOT'
-  'scala.cli.integration.BspTests213.setup-ide prepares a valid BSP configuration with --cli-version 1.5.0-34-g31a88e428-SNAPSHOT'

Gedochao:
Suspecting intermittent failures on the CI, irrelevant to this PR...

PR #3835, opened 2025-08-15
Fix using directive with URL + query parameters

### LINKED ISSUE #3720: "using file" with URL+query parameters is broken
**Version(s)**
1.8.1

**Describe the bug**

The new `using file` feature with URLs doesn't work when a query parameter is passed.

The request is correctly sent to the server, but Scala CLI fails to use the downloaded code.

**To Reproduce**

Assume the following server:

```scala
//> using dep com.lihaoyi::cask::0.10.2

object Codegen extends cask.MainRoutes:
  @cask.get("/Special.scala")
  def dynamicWithParam(prefix: Option[String] = None): String =
    println("GOT PREFIX: " + prefix)
    val _prefix = prefix.getOrElse("NONE")
    val op = s"def specialPrint(msg: String): Unit = println(\"prefix:\" + msg)"
    s"""|package special
        |object Ops:
        |  $op""".stripMargin

  initialize()
```

The following code works as expected:

```scala
//> using file "http://localhost:8080/Special.scala"

import special.Ops.*

object Crazy:
  @main def run: Unit = specialPrint("bar")
```

But, when a query parameter is added to the `using` clause, it fails:

```scala
//> using file "http://localhost:8080/Special.scala?prefix=foo"

import special.Ops.*

object Crazy:
  @main def run: Unit = specialPrint("bar")
```

Fails with:
```
[error] value Ops is not a member of special
[error] import special.Ops.*
```

Even though the server logs the request correctly.

**Expected behaviour**

Both cases should compile and run.

jgoday:
Correct the InMemory file name downloaded when using a dependency with URL + query parameters.
Simple test added in 'modules/build/src/test/scala/scala/build/tests/SourcesTests.scala'.

fixes #3720.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -382,6 +382,16 @@ object CrossSources {
     crossSources -> finalInputs
   }
 
+  extension (uri: java.net.URI)
+    def asString: String =
+      java.net.URI(
+        uri.getScheme,
+        uri.getAuthority,
+        uri.getPath,
+        null,
+        uri.getFragment
+      ).toString
+
   /** @return
     *   the resource directories that should be added to the classpath
     */
@@ -405,7 +415,7 @@ object CrossSources {
     download(pUri.value.toString).left.map(
       new UsingFileFromUriError(pUri.value, pUri.positions, _)
     ).map(content =>
-      Seq(Virtual(pUri.value.toString, content))
+      Seq(Virtual(pUri.value.asString, content))
     )
 
   type CodeFile = os.Path | java.net.URI
--- a/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/SourcesTests.scala
@@ -125,6 +125,39 @@ class SourcesTests extends TestUtil.ScalaCliBuildSuite {
     }
   }
 
+  test("dependencies in .scala - using URL with query parameters") {
+    val testInputs = TestInputs(
+      os.rel / "something.scala" ->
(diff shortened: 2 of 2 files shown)


modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -401,11 +401,23 @@ object CrossSources {
...
+    *   a java.net.URI to a String without query parameters
+    */
+  private def uriToString(uri: java.net.URI): String =
+    java.net.URI(
+      uri.getScheme(),
+      uri.getAuthority(),
+      uri.getPath(),
+      null,
+      uri.getFragment()
+    ).toString()
```
Gedochao: NIT: would be more readable as an extension method, IMO
```scala
extension (uri: java.net.URI)
  def asString: String =
    java.net.URI(
      scheme = uri.getScheme,
      authority = uri.getAuthority,
      path = uri.getPath,
      query = null,
      fragment = uri.getFragment
    ).toString
```

Gedochao (commented):
Hey, sorry for the late review! 
Commented on a small nitpick, otherwise looks good

modules/build/src/main/scala/scala/build/CrossSources.scala
jgoday: Sure, changed to an extension method as you suggested.
Should this extension method be private or somewhere else?

Gedochao: I'd put it in a util object somewhere, I guess.
But since atm there's one usage, you can leave it where it is for now.

Gedochao (approved):
The failures on the CI are caused by https://github.com/VirtusLab/scala-cli/issues/3841
Will need to rebase on top of `main` once https://github.com/VirtusLab/scala-cli/pull/3842 gets merged.

Gedochao:
(rebased the branch)

PR #3693, opened 2025-05-24
Fix race condition in local repo setup

unlsycn:
Fixes https://github.com/VirtusLab/scala-cli/issues/3110 and https://github.com/VirtusLab/scala-cli/pull/3114#issuecomment-2299341773.

A race condition occurs when process A acquires the lock first, creating the directory after process B's check, but before B acquires the lock while waiting for it. This problem frequently arises when running tests in parallel.

For example, consider two scala-cli process perform the following sequence of operations:
| **A**          | **B**                 |
|----------------|---------------------- |
|                | Check repo dir        |
| Acquire lock   | Wait for lock release |
| Setup repo dir | Wait for lock release |
| Release lock   |                       |
|                | Acquire lock          |
|                | Try setup → Exception |

This patch fixes the issue by adding an additional post-acquisition check. Another possible fix would be to simply move the check into the withLock scope, but this would result in a redundant lock acquisition and a release each time.

PR #3846, opened 2025-08-29
Fix completely broken lock during setuping local repo on Linux

unlsycn:
After recently bumping Scala-CLI, I noticed that the race condition in the local repository setup still persists even after merging #3693. I then identified a critical issue: the `withLock` mechanism does not work at all on Linux.

According to https://bugs.openjdk.org/browse/JDK-7095452, the root cause is that a .lock file cannot be observed by other processes:

> On Solaris/Linux if DELETE_ON_CLOSE is specified, the appropriate file is unlinked from the directory when opened and therefore appears inaccessible via path.

In this case, each Scala-CLI process considers itself holding the lock, leading to a race condition.

This issue is difficult to fully capture with a single test due to its concurrent nature, but adding a `Thread.sleep` before `extractZip`  and run multiple Scala-CLI instances in parallel can reproduce the issue reliably nearly 100% of the time.

### DIFF
--- a/modules/build/src/main/scala/scala/build/LocalRepo.scala
+++ b/modules/build/src/main/scala/scala/build/LocalRepo.scala
@@ -77,11 +77,7 @@ object LocalRepo {
           // potential race conditions between initial check and lock acquisition
           if !os.exists(repoDir) then
             val tmpRepoDir = repoDir / os.up / s".$version.tmp"
-            try os.remove.all(tmpRepoDir)
-            catch {
-              case t: Throwable =>
-                logger.message(s"Error removing $tmpRepoDir: ${t.getMessage}")
-            }
+            os.remove.all(tmpRepoDir)
             using(archiveUrl.openStream()) { is =>
               using(WrappedZipInputStream.create(new BufferedInputStream(is))) { zis =>
                 extractZip(zis, tmpRepoDir)
@@ -101,30 +97,21 @@ object LocalRepo {
       val lockFile = dir.resolve(s".lock-$id");
       Util.createDirectories(lockFile.getParent)
       var channel: FileChannel = null
+      var lock: FileLock       = null
 
       try {
         channel = FileChannel.open(
           lockFile,
           StandardOpenOption.CREATE,
-          StandardOpenOption.WRITE,
-          StandardOpenOption.DELETE_ON_CLOSE
+          StandardOpenOption.WRITE
         )
-
-        var lock: FileLock = null
-        try {
-          lock = channel.lock()
-
-          try f
-          finally {
-            lock.release()
-            lock = null
-            channel.close()
-            channel = null
-          }
-        }
-        finally if (lock != null) lock.release()
+        lock = channel.lock()
+        f
       }
-      finally if (channel != null) channel.close()
+      finally {
+        if (lock != null) lock.release()
+        if (channel != null) channel.close()
+      }
     }
 
 }
--- a/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
@@ -2413,4 +2413,29 @@ abstract class RunTestDefinitions
         expect(!res.err.trim().contains(legacyRunnerWarning))
       }
   }
+
+  for (parallelInstancesCount <- Seq(2, 5, 10))
+    test(
+      s"run $parallelInstancesCount instances in parallel without local repo"
+    ) {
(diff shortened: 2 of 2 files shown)


modules/build/src/main/scala/scala/build/LocalRepo.scala
```diff
@@ -77,11 +77,7 @@ object LocalRepo {
           // potential race conditions between initial check and lock acquisition
           if !os.exists(repoDir) then
             val tmpRepoDir = repoDir / os.up / s".$version.tmp"
-            try os.remove.all(tmpRepoDir)
-            catch {
-              case t: Throwable =>
-                logger.message(s"Error removing $tmpRepoDir: ${t.getMessage}")
-            }
```
Gedochao, line 84: why remove the catch case here?

Gedochao:
> This issue is difficult to fully capture with a single test due to its concurrent nature, but adding a `Thread.sleep` before `extractZip` and run multiple Scala-CLI instances in parallel can reproduce the issue reliably nearly 100% of the time.

We have tests for running multiple Scala CLI instances in parallel right here:
https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala#L2320-L2361

Can you try adding a similar test for your case? Optimally, I'd like to be able to track if this works correctly on all platforms.

modules/build/src/main/scala/scala/build/LocalRepo.scala
unlsycn: The removal operation should not fail if the lock works fine. In such cases (e.g., incorrect file permissions), it should throw an exception immediately rather than just logging a message.

unlsycn:
> > This issue is difficult to fully capture with a single test due to its concurrent nature, but adding a `Thread.sleep` before `extractZip` and run multiple Scala-CLI instances in parallel can reproduce the issue reliably nearly 100% of the time.
> 
> We have tests for running multiple Scala CLI instances in parallel right here: https://github.com/VirtusLab/scala-cli/blob/main/modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala#L2320-L2361
> 
> Can you try adding a similar test for your case? Optimally, I'd like to be able to track if this works correctly on all platforms.

Sure. Could you provide some guidance on accessing localRepoDir in integration tests? Since the setup only occurs in environments without cached local repo, I need to remove the local repo for this test. Alternatively, is there a way to run this test in a sandboxed environment to ensure the setup process executes?

Gedochao:
You can print all relevant locations (including the local repository) with the following command:
```bash
scala-cli directories --power
```
Where the output should look like (example for a Mac):
```
Local repository: ~/Library/Caches/ScalaCli/local-repo
Completions: ~/Library/Application Support/ScalaCli/completions
Virtual projects: ~/Library/Caches/ScalaCli/virtual-projects
BSP sockets: ~/Library/Caches/ScalaCli/bsp-sockets
Bloop daemon directory: ~/Library/Caches/ScalaCli/bloop/daemon
Secrets directory: ~/Library/Application Support/ScalaCli/secret
```

The definition in-code can be found here:
https://github.com/VirtusLab/scala-cli/blob/32b11d945ced95500205f26fd1e722c7253992d4/modules/build/src/main/scala/scala/build/Directories.scala#L40-L41

modules/build/src/main/scala/scala/build/LocalRepo.scala
Gedochao: I suppose if the cause would be a problem with the lock, we might still print some context and rethrow.
But fair enough.

unlsycn:
> You can print all relevant locations (including the local repository) with the following command:
> 
> ```shell
> scala-cli directories --power
> ```
> 
> Where the output should look like (example for a Mac):
> 
> ```
> Local repository: ~/Library/Caches/ScalaCli/local-repo
> Completions: ~/Library/Application Support/ScalaCli/completions
> Virtual projects: ~/Library/Caches/ScalaCli/virtual-projects
> BSP sockets: ~/Library/Caches/ScalaCli/bsp-sockets
> Bloop daemon directory: ~/Library/Caches/ScalaCli/bloop/daemon
> Secrets directory: ~/Library/Application Support/ScalaCli/secret
> ```
> 
> The definition in-code can be found here:
> 
> https://github.com/VirtusLab/scala-cli/blob/32b11d945ced95500205f26fd1e722c7253992d4/modules/build/src/main/scala/scala/build/Directories.scala#L40-L41

Thanks for your help! I've added a test that fails before this patch and passes after it.

Gedochao (approved):

Issue #3859, opened 2025-09-04
`setup-ide` provides erroneous `.bsp/scala.json` when used from `scala` installed via `brew`


Gedochao:
**Version(s)**
1.8.4 installed as `scala` via `brew` with Scala 3.7.2

**Describe the bug**
When using the `brew` installation of `scala` specifically, `setup-ide` puts a reference to a non-self-executable `scala-cli.jar` to `.bsp/scala.json`. As a result, IDEs can't start the BSP server.

```json
{
  "name": "scala",
  "argv": [
    "/opt/homebrew/Cellar/scala/3.7.2/libexec/libexec/scala-cli.jar",
    "--cli-default-scala-version",
    "3.7.2",
    "--repository",
    "file:///opt/homebrew/Cellar/scala/3.7.2/libexec/maven2",
    "--prog-name",
    "scala",
    "--skip-cli-updates",
    "bsp",
    "--json-options",
    "~/project/.scala-build/ide-options-v2.json",
    "--json-launcher-options",
    "~/project/.scala-build/ide-launcher-options.json",
    "--envs-file",
    "~/project/.scala-build/ide-envs.json",
    "~/project"
  ],
  "version": "1.8.4",
  "bspVersion": "2.1.1",
  "languages": [
    "scala",
    "java"
  ]
}
```

**Expected behaviour**
the BSP config json should point to a valid command in `argv`.

cc @tgodzik

Gedochao:
What's interesting, this doesn't happen with the other installations, i.e. via Coursier.

Gedochao:
So... this started happening, because the bootstrapped JVM launcher is no longer self-executable, as a result of rules imposed on us by the new Sonatype Maven Central Portal.
That being said, Brew shouldn't be installing the `scala` runner script referring to that launcher in the first place, but that's yet another issue.

The `setup-ide` sub-command should nonetheless be handling this case, so I'm brew-ing (haha) a fix in the meantime.

PR #3872, opened 2025-09-18
Run the default (Scala 3 Next) suite with the JVM bootstrapped launcher on the CI

Gedochao:
What happens here:
- I got rid of some tests for the bootstrapped JVM launcher which weren't testing what they should be testing
- I ensured the whole execution command is swapped in the `scala` script for tests using it 
  - for context, we have tests which use the Coursier installation of Scala and swap the `scala` script reference to the launcher being tested in our integration tests, to ensure it would work in the context of the script for all the current Scala versions
  - my change actually fixes attaching a debugger to those tests, which is actually useful in its own right
- I added a new job on the CI to run the default suite of integration tests with the bootstrapped JVM launcher (which actually is no longer a launcher, as Sonatype forbids us from making it self executable if we want to keep publishing it on Maven Central
   - Note: the bootstrapped JVM launcher is a fancy name for the self-contained FAT JAR of Scala CLI, as described here: https://scala-cli.virtuslab.org/install#bootstrapped-standalone-fat-jar
   - this is an equivalent of the `jvm-tests-1` for the standard JVM launcher, as I think it's sufficient to run this for the default Scala version only (so the current Scala 3 Next suite)
   - the intention is to preemptively catch problems like https://github.com/VirtusLab/scala-cli/issues/3859
   - I will add dedicated tests for the https://github.com/VirtusLab/scala-cli/issues/3859 along with a fix as a follow-up


modules/integration/src/test/scala/scala/cli/integration/StandaloneLauncherTests.scala
```diff
@@ -1,32 +0,0 @@
...
-      val extraEnv = Map(
-        "JAVA_HOME" -> java8Home.toString,
-        "PATH"      -> ((java8Home / "bin").toString + File.pathSeparator + System.getenv("PATH"))
-      )
-
-      // we need to exit to check if the launcher is actually starting the right version of java
-      os.proc(TestUtil.cli, "--power", "bloop", "exit").call(cwd = root, env = extraEnv).out.trim()
-
-      val output =
-        os.proc(TestUtil.cli, ".").call(cwd = root, env = extraEnv).out.trim()
```
Gedochao, line 27: The funny thing is, this test hasn't been testing anything, since `TestUtil.cli` adds the `java` command to the execution automatically.
So it never mattered what was in `JAVA_HOME`, here.

modules/integration/src/test/scala/scala/cli/integration/StandaloneLauncherTests.scala
```diff
@@ -1,32 +0,0 @@
-package scala.cli.integration
-
-import com.eed3si9n.expecty.Expecty.expect
-
-import java.io.File
-
-class StandaloneLauncherTests extends ScalaCliSuite {
-  test(s"Standalone launcher should run with java 8") {
-    // It should download Java 17 and use it to run itself
```
Gedochao, line 9: If it ran itself, this test would have been failing as soon as we migrated from the old Sonatype portal.
Since the migration forced us to make it non-self-executable.
And rendered this test utter nonsense.

Gedochao:
This is in preparation to https://github.com/VirtusLab/scala-cli/issues/3859, as the error is `jvmBootstrapped`-specific and I have come to the conclusion that perhaps it is worth running the entire default JVM suite with that launcher to catch similar cases.

tgodzik (approved):

PR #3876, opened 2025-09-19
Ensure non-self executable JVM launchers' `setup-ide` produces working BSP connection JSON

### LINKED ISSUE #3859: `setup-ide` provides erroneous `.bsp/scala.json` when used from `scala` installed via `brew`

Gedochao:
Fixes #3859

Depends on:
- https://github.com/VirtusLab/scala-cli/pull/3872

Some extra context:
- this doesn't so much fix the `homebrew` installation (it needs to be separately adjusted to use the native launcher, rather than the bootstrapped fat jar), as it fixes the bootstrapped fat jar used in that installation to produce a working Bloop connection JSON
- the resulting connection JSON detects what Java should be used, rather than using (potentially too old) whatever is in JAVA_HOME
- the change only affects non-self-executable JVM distributions of Scala CLI (like the said bootstrapped fat JAR)


Gedochao:
Extra note: it's a bit tricky to test (potential follow-up), but this also fixes `setup-ide` for cases when Scala CLI is used via [cs launch](https://get-coursier.io/docs/cli-launch), as that had the same problem.

tgodzik (approved):

PR #3879, opened 2025-09-23
add test case for not having test.resourceDir in main scope

btomala:
I found that the resource from the directive `test.resourceDir` is added to the classpath in the main scope. It should be only in the test scope.

I thought that was an easy fix, so I tried to do so. However, I failed. I'm opening a PR with a test case for this. I'm happy to fix it if you could give me some guidance on where to look.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -271,19 +271,8 @@ object CrossSources {
       )
     }).flatten
 
-    val resourceDirectoriesFromDirectives = {
-      val resourceDirsFromCli =
-        allInputs.elements.flatMap { case rd: ResourceDirectory => Some(rd.path); case _ => None }
-      val resourceDirsFromBuildOptions: Seq[os.Path] =
-        buildOptions.flatMap(_.value.classPathOptions.resourcesDir).distinct
-      resourceDirsFromBuildOptions
-        .filter(!resourceDirsFromCli.contains(_))
-        .map(ResourceDirectory(_))
-    }
-    val finalInputs = allInputs.add(resourceDirectoriesFromDirectives)
-
     val defaultMainElemPath = for {
-      defaultMainElem <- finalInputs.defaultMainClassElement
+      defaultMainElem <- allInputs.defaultMainClassElement
     } yield defaultMainElem.path
 
     val pathsWithDirectivePositions
@@ -293,7 +282,7 @@ object CrossSources {
           val baseReqs0 = baseReqs(d.scopePath)
           WithBuildRequirements(
             d.requirements.fold(baseReqs0)(_ orElse baseReqs0),
-            (d.path, d.path.relativeTo(finalInputs.workspace))
+            (d.path, d.path.relativeTo(allInputs.workspace))
           ) -> d.directivesPositions
       }
     val inMemoryWithDirectivePositions
@@ -318,7 +307,7 @@ object CrossSources {
       }
 
     val resourceDirs: Seq[WithBuildRequirements[os.Path]] =
-      resolveResourceDirs(finalInputs, preprocessedSources)
+      resolveResourceDirs(allInputs, preprocessedSources)
 
     lazy val allPathsWithDirectivesByScope: Map[Scope, Seq[(os.Path, Position.File)]] =
       (pathsWithDirectivePositions ++ inMemoryWithDirectivePositions ++ unwrappedScriptsWithDirectivePositions)
@@ -379,7 +368,7 @@ object CrossSources {
       buildOptions,
       unwrappedScripts
     )
-    crossSources -> finalInputs
+    crossSources -> allInputs
   }
 
   extension (uri: java.net.URI)
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -271,8 +271,22 @@ object CrossSources {
       )
     }).flatten
 
+    val resourceDirectoriesFromDirectives = {
+      val resourceDirsFromCli =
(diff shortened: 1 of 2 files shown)


Gedochao:
The directive is declared here:
https://github.com/VirtusLab/scala-cli/blob/418959f00760cee36bc47e2c5ac158e46d56ee15/modules/directives/src/main/scala/scala/build/preprocessing/directives/Resources.scala#L34-L36

and then put into options, here:
https://github.com/VirtusLab/scala-cli/blob/418959f00760cee36bc47e2c5ac158e46d56ee15/modules/directives/src/main/scala/scala/build/preprocessing/directives/Resources.scala#L40
At a glance, test scope is being passed as a requirement...

Gedochao:
We construct the build options with the resource directory here:
https://github.com/VirtusLab/scala-cli/blob/418959f00760cee36bc47e2c5ac158e46d56ee15/modules/directives/src/main/scala/scala/build/preprocessing/directives/Resources.scala#L46-L70

It seems it's then extracted while processing sources here:
https://github.com/VirtusLab/scala-cli/blob/418959f00760cee36bc47e2c5ac158e46d56ee15/modules/build/src/main/scala/scala/build/CrossSources.scala#L398-L412

To debug this, I'd track if build requirements (being the test scope) is respected when this field is being passed around:
https://github.com/VirtusLab/scala-cli/blob/418959f00760cee36bc47e2c5ac158e46d56ee15/modules/build/src/main/scala/scala/build/CrossSources.scala#L47

Let me know if this helps, this is where I'd start.

Gedochao:
(note: I converted the PR to a draft, since it's not ready for review)

modules/build/src/test/scala/scala/build/tests/DirectiveTests.scala
```diff
@@ -230,9 +230,11 @@ class DirectiveTests extends TestUtil.ScalaCliBuildSuite {
           |""".stripMargin
       ) { (build, isTestScope) =>
         val resourcesDirs = build.options.classPathOptions.resourcesDir
+        val sourcesResourcesDirs = build.successfulOpt.get.sources.resourceDirs
         expect(resourcesDirs.exists(_.last == "mainResources"))
         val hasTestResources = resourcesDirs.exists(_.last == "testResources")
         expect(if isTestScope then hasTestResources else !hasTestResources)
+        expect(resourcesDirs.toSet == sourcesResourcesDirs.toSet)
```
btomala: I'm not sure if these tests were testing the resolving of test scope resourceDir; it was testing only if the build options are passed correctly. 
I was expecting that this would be equal, but it's not. It wasn't even before my changes. Now this test is failing.

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -271,19 +271,8 @@ object CrossSources {
...
-    val resourceDirectoriesFromDirectives = {
-      val resourceDirsFromCli =
-        allInputs.elements.flatMap { case rd: ResourceDirectory => Some(rd.path); case _ => None }
-      val resourceDirsFromBuildOptions: Seq[os.Path] =
-        buildOptions.flatMap(_.value.classPathOptions.resourcesDir).distinct
-      resourceDirsFromBuildOptions
-        .filter(!resourceDirsFromCli.contains(_))
-        .map(ResourceDirectory(_))
-    }
-    val finalInputs = allInputs.add(resourceDirectoriesFromDirectives)
```
btomala, line 283: Removing this resolves my issue and fixes the tests. The `resourceDirectoriesFromDirectives` extracts the directives and puts them in the input without scope. Later on in the [resolveResourceDirs](https://github.com/VirtusLab/scala-cli/pull/3879/files#diff-96e7a9117a91536eee2d9b9daaee8a5a92e5ea05bacdcb3930c6e443e960f14eR387), the paths were extracted from input and wrapped into `WithBuildRequirements` without scope (because it wasn't available at this point) using the default `main` scope.

However, I'm not sure what the other consequences of this change are, as the `allInputs` are returned from the `forInputs` method. I don't know where this is used yet.

Gedochao: right... so I'm guessing we need to respect the scope here

btomala: Another option to address the issue will be to pass the `allInputs` to `resoveReourceDirs` instead of `finalInputs`

modules/build/src/main/scala/scala/build/CrossSources.scala
```diff
@@ -318,7 +321,7 @@ object CrossSources {
       }
 
     val resourceDirs: Seq[WithBuildRequirements[os.Path]] =
-      resolveResourceDirs(finalInputs, preprocessedSources)
+      resolveResourceDirs(allInputs, preprocessedSources)
```
btomala, line 324: Using allInpts instead of finalInputs solves the issue

Gedochao: it does for this use case, but I do not think it would for all edge cases... I'm working on an alternative fix, sorry this took so long

Gedochao: that being said, that your fix does work made me realise how much this part of the code is in need of a refactor 🙃

modules/build/src/main/scala/scala/build/CrossSources.scala
Gedochao: So the TL;DR is that this is actually the correct course of action.
As in, this part has to be removed.
However, the `--watch` logic was relying on this piece to add the resource directories to elements, as they also should be watched.
So further changes to `--watch` were necessary.
I will open another PR.

Gedochao:
Solved by https://github.com/VirtusLab/scala-cli/pull/3898, closing this.

PR #3885, opened 2025-09-25
Update MacOS CI

Gedochao:
This does the following:
- updates MacOS x86_64 workflows from `macOS-13` to `macOS-15-intel`
- updates MacOS arm workflows from `macOS-14` to `macOS-15`
- turns off MacOS x86_64 workflows running integration test suites for Scala 2.12, 2.13 and 3.3 LTS
  - we'll still run these suites for MacOS arm
  - the motivation is that MacOS x86_64 & arm runners from GitHub share a limited pool of available runners, which means each and every CI run waits for MacOS tests to finish running, effectively doubling, sometimes tripling CI times
    - this slows down development considerably, at times
  - given how rare it is for us to detect problems for these, I think it's sufficient to limit this platform to Scala 3 Next & Scala 3 Next RC (the cutting edge versions, most likely to detect anything)
  - this means the following jobs will have to be removed from requirements on the CI @tgodzik 
    - `native-macos-tests-2`
    - `native-macos-tests-3`
    - `native-macos-tests-4`
- note that `macOS-13` jobs are being deprecated, and the new `macOS-15-intel` runners will be available until August 2027
  - we will likely stop distributing MacOS x86_64 launchers when GitHub stops supporting them on the CI
  - more reading available at https://github.com/actions/runner-images/issues/13045


Gedochao:
I'm thinking a follow-up renaming `-1`-`-5` test suites into something human-comprehendable might be a good idea 😅 
I'd rather do that in a separate PR, however

tgodzik (approved):

zielinsky (approved):

PR #3892, opened 2025-09-30
Mention the ivy2Local predefined repository

przemek-pokrywka:
The ivy2Local predefined repository is most useful when you publish local artifacts with SBT.

Sometimes you work on a large application built with SBT, yet you still want to write scripts that use a part of the application's logic. Then you can publish your application's artifacts from SBT with `publishLocal` and then consume these artifacts from `scala-cli` using `scala-cli --repository ivy2Local --dep myapp::artifact:1.0-SNAPSHOT`


Gedochao:
The reference part of the docs is mostly generated from code, so you have to change it there, and then regenerate the docs with 
```bash
```
as per https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md#rules-for-a-well-formed-pr

The relevant piece of code, where this doc actually lives: https://github.com/VirtusLab/scala-cli/blob/27cb278eb97ce5480a739294a17cf86d709de6e1/modules/directives/src/main/scala/scala/build/preprocessing/directives/Repository.scala#L41

PR #3920, opened 2025-10-23
NIT Fix more miscellaneous warnings

Gedochao:


modules/scala3-runtime/src/main/scala/scala/cli/runtime/SafeLazyVals.scala
```diff
@@ -5,7 +5,7 @@ import java.lang.reflect.Field
 object SafeLazyVals {
   private val debug = false
 
-  private[this] val unsafe: sun.misc.Unsafe =
+  private val unsafe: sun.misc.Unsafe =
```
tgodzik, line 8: Why do we actually use Unsafe in Scala CLI? Looks like we might need to remove it if possible

tgodzik (approved):
LGTM with one question

modules/scala3-runtime/src/main/scala/scala/cli/runtime/SafeLazyVals.scala
Gedochao: I think it's a legacy hack for Scala 3.0-3.2... which it might be we don't need anymore (and we'd definitely have to remove it when migrating to Scala 3.9 LTS)

Gedochao: Created an issue:
- https://github.com/VirtusLab/scala-cli/issues/3922

PR #3923, opened 2025-10-25
fix for 3307 (windows launcher rejects filenames with utf8 chars)

### LINKED ISSUE #3307: cannot find the path specified with UTF-8 characters on Windows
**Version(s)**
1.4.3, `scala-cli` as `scala` 3.5.2

**Describe the bug**
 > On Windows, if the user directory has an umlaut in it (e.g. 'ä'), the `scala.bat` file will not work properly and simply output:
`The system cannot find the path specified.`

**To Reproduce**

 > ```cmd
>  > scala
> ```
> 
> ```
> The system cannot find the path specified.
> ```

**Workaround**

> There are two problems in the batch file. One is, that the umlaut 'ä' was changed to '�' on the line `SET CMDLINE="C:\Users\<name with ä>`.
> > Note: I don't know where exactly this issue stems from.
>
> Another issue is that the code page has to be set to `65001` to execute properly.
> This can be done by adding the following line at the top of the file:
> ```batch
> @CHCP 65001>nul
> ```
> The command `CHCP 65001` changes the code page to 65001 and the rest disables the lines output (`@echo off` is not enough here).
? > Note: This worked for me, but could break things on other systems.

**Expected behaviour**

> The scala repl starts:
> 
> ```
> Welcome to Scala 3.5.2 (23.0.1, Java OpenJDK 64-Bit Server VM).
> Type in expressions for evaluation. Or try :help.
> 
> scala>
> ```

**Extra context**
- Originally raised by @JulianBohne in https://github.com/scala/scala3/issues/21994

philwalk:
UPDATE:  the final fix was to edit Windows workflows to configure the Windows registry for UTF-8, and to update the mill build script to use `zulu:17` instead of `temurin:17`.

This PR modifies `generate-native-image.sh` to generate `scala-cli.exe` with code page set to 65001 if running in Windows:

- save current code page
- set code page to 65001
- generate scala-cli.exe
- restore the code page atexit

Using a test script named `/opt/ue/jsrc/løp.sc`:
```scala
printf("Hello %s\n", scriptPath)
printf("Hello %s\n", sys.props("scala.sources"))
printf("Hello %s\n", sys.props("scala.source.names"))
```

and with the PATH set to use the generated `scala-cli.exe`:
```bash
PATH="out/cli/3.3.7/base-image/nativeImage.dest:$PATH"
```

#### verify the bug
create a buggy `scala-cli.exe` by running `generated-native-image.sh` temporarily modified to set code page 437
Verify that the bug manifests itself when attempting to execute the script:
```bash

# /opt/ue/jsrc/løp.sc
[error]  File not found: C:\opt\ue\jsrc\løp.sc
```
#### Verify the fix
create fixed `scala-cli.exe` by running `generated-native-image.sh` as proposed in this PR (set code page 65001)
Verify that the bug is fixed:
```bash
# /opt/ue/jsrc/løp.sc
Hello /opt/ue/jsrc/løp.sc
Hello C:/opt/ue/jsrc/løp.sc
Hello løp.sc
```


philwalk:
The error messages for the failing tests are the result of having Windows code page 437 active.
I assume that (at least as of jdk17) we need to set the code page to 65001 in Windows before running this test.

The failure is that a script named `TestÅÄÖåäö.sc` with contents:
```scala
object TestÅÄÖåäö {
    println("Hello from TestÅÄÖåäö")
  }
}
```
failed with this error message:

```bash
[950] Error: Could not find or load main class Test├à├ä├û├Ñ├ñ├╢
[950] Caused by: java.lang.ClassNotFoundException: Test├à├ä├û├Ñ├ñ├╢
```
```bash
# ./mill integration.test.native 'scala.cli.integration.RunTestsDefault.UTF-8'
```
```bash
[949] Produced artifacts:
[949]  C:\Users\philwalk\workspace\scala-cli-3307\out\cli\3.3.7\base-image\nativeImage.dest\scala-cli.build_artifacts.txt (txt)
[949]  C:\Users\philwalk\workspace\scala-cli-3307\out\cli\3.3.7\base-image\nativeImage.dest\scala-cli.exe (executable)
[949] ========================================================================================================================
[949] Finished generating 'scala-cli' in 3m 49s.
[950/950] integration.test.native
[950] >==== Running 'UTF-8' from RunTestDefinitions
[950] Running warmup testà
[950] Compiling project (Scala 3.7.3, JVM (17))
[950] Compiled project (Scala 3.7.3, JVM (17))
[950] Done running warmup test.
[950] Compiling project (Scala 3.7.3, JVM (17))
[950] Compiled project (Scala 3.7.3, JVM (17))
[950] Error: Could not find or load main class Test├à├ä├û├Ñ├ñ├╢
[950] Caused by: java.lang.ClassNotFoundException: Test├à├ä├û├Ñ├ñ├╢
[950] X==== Finishing 'UTF-8' from RunTestDefinitions
[950] scala.cli.integration.RunTestsDefault:
[950] ==> X scala.cli.integration.RunTestsDefault.UTF-8  2.017s os.SubprocessException: Result of C:\Users\philwalk\workspace\scala-cli-3307\out\cli\3.3.7\base-image\nativeImage.dest\scala-cli.exeà: 1
[950]
[950]     at os.proc.call(ProcessOps.scala:232)
[950]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$117(RunTestDefinitions.scala:1080)
[950]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$117$adapted(RunTestDefinitions.scala:1072)
[950]     at scala.cli.integration.TestInputs.$anonfun$fromRoot$1(TestInputs.scala:35)
[950]     at scala.cli.integration.TestInputs$.scala$cli$integration$TestInputs$$withTmpDir(TestInputs.scala:95)
[950]     at scala.cli.integration.TestInputs.fromRoot(TestInputs.scala:33)
[950]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$116(RunTestDefinitions.scala:1072)
[950]     at scala.cli.integration.WithWarmUpScalaCliSuite.$anonfun$test$1(WithWarmUpScalaCliSuite.scala:34)
[950]     at scala.cli.integration.WithWarmUpScalaCliSuite.$anonfun$test$2(WithWarmUpScalaCliSuite.scala:37)
[950/950, 1 failed] ============================== integration.test.native scala.cli.integration.RunTestsDefault.UTF-8 ============================== 265s
1 tasks failed
integration.test.native 1 tests failed:
  scala.cli.integration.RunTestsDefault.UTF-8 scala.cli.integration.RunTestsDefault.UTF-8
philwalk@d5 MSYS ~/workspace/scala-cli-3307
```

Gedochao:
> I assume that (at least as of jdk17) we need to set the code page to 65001 in Windows before running this test.

Do you maybe know how to do it? Can't say I'm familiar with code pages in Windows, myself.
The fix looks good otherwise, we just need to get the tests fixed.

philwalk:
The tests are failing due to very subtle jdk bugs, and it has proven to be quite resistant to being fixed.
TLDR:  Here's the log from a heavily instrumented version of the jdk test (native does the same thing):  
It verifies that the filenames and file contents are utf-8 and although the `os.proc(..).call()` fails,  when I issue the exact same command line in `bash` it is successful.

```scala
# JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF-8" time ./mill integration.test.jvm 'scala.cli.integration.RunTestsDefault.UTF-8'
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
[0934/1026] test-runner.publishLocalNoFluff
[0934] Publishing Artifact(org.virtuslab.scala-cli,test-runner_3,0.0.1-SNAPSHOT) to ivy repo C:\Users\philwalk\workspace\scala-cli-3307-clean\out\repo\0.0.1-SNAPSHOT
[0935/1026] runner.publishLocalNoFluff
[0935] Publishing Artifact(org.virtuslab.scala-cli,runner_3,0.0.1-SNAPSHOT) to ivy repo C:\Users\philwalk\workspace\scala-cli-3307-clean\out\repo\0.0.1-SNAPSHOT
[1026/1026] integration.test.jvm
[1026] Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
[1026] >==== Running 'UTF-8' from RunTestDefinitions
[1026] Running warmup test…
[1026] Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
[1026] Compiling project (Scala 3.7.3, JVM (17))
[1026] Compiled project (Scala 3.7.3, JVM (17))
[1026] Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
[1026] Done running warmup test.
[1026] sun.jnu.encoding = UTF-8
[1026] file.encoding = UTF-8
[1026] Charset.defaultCharset = UTF-8
[1026] Class name = scala.cli.integration.RunTestsDefault
[1026] TestUtil.cli with extraOptions [java -Xmx512m -Xms128m -jar C:\Users\philwalk\workspace\scala-cli-3307-clean\out\cli\3.3.7\standaloneLauncher.dest\launcher.bat --bloop-startup-timeout 2min --bloop-bsp-timeout 1min]
[1026] message[Hello from TestÅÄÖåäö]
[1026] fileName[TestÅÄÖåäö.scala]
[1026] sun.jnu.encoding=UTF-8
[1026] file.encoding=UTF-8
[1026] ### [C:\Users\philwalk\workspace\scala-cli-3307-clean\out\integration\tmpDirBase.dest\working-dir\run-1467434246\test-2\TestÅÄÖåäö.scala]
[1026] ####### str[object TestÅÄÖåäö {
[1026]   def main(args: Array[String]): Unit = {
[1026]     println("Hello from TestÅÄÖåäö")
[1026]   }
[1026] }
[1026] ]
[1026] Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
[1026] [error]  File not found: C:\Users\philwalk\workspace\scala-cli-3307-clean\out\integration\tmpDirBase.dest\working-dir\run-1467434246\test-2\TestÃ…Ã„Ã–Ã¥Ã¤Ã¶.scala
[1026] X==== Finishing 'UTF-8' from RunTestDefinitions
[1026] scala.cli.integration.RunTestsDefault:
[1026] ==> X scala.cli.integration.RunTestsDefault.UTF-8  5.759s os.SubprocessException: Result of java…: 1
[1026]
[1026]     at os.proc.call(ProcessOps.scala:232)
[1026]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$118(RunTestDefinitions.scala:1104)
[1026]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$118$adapted(RunTestDefinitions.scala:1084)
[1026]     at scala.cli.integration.TestInputs.$anonfun$fromRoot$1(TestInputs.scala:35)
[1026]     at scala.cli.integration.TestInputs$.scala$cli$integration$TestInputs$$withTmpDir(TestInputs.scala:95)
[1026]     at scala.cli.integration.TestInputs.fromRoot(TestInputs.scala:33)
[1026]     at scala.cli.integration.RunTestDefinitions.$anonfun$new$116(RunTestDefinitions.scala:1084)
[1026]     at scala.cli.integration.WithWarmUpScalaCliSuite.$anonfun$test$1(WithWarmUpScalaCliSuite.scala:34)
[1026]     at scala.cli.integration.WithWarmUpScalaCliSuite.$anonfun$test$2(WithWarmUpScalaCliSuite.scala:37)
[1026/1026, 1 failed] ============================== integration.test.jvm scala.cli.integration.RunTestsDefault.UTF-8 ============================== 8s
1 tasks failed
integration.test.jvm 1 tests failed:
  scala.cli.integration.RunTestsDefault.UTF-8 scala.cli.integration.RunTestsDefault.UTF-8
Command exited with non-zero status 1
```
Here are the command lines for the 2 failing tests, which pass when issued from `bash`:
```scala
philwalk@d5 MSYS ~/workspace/scala-cli-3307-clean
# scala-cli.exe --bloop-startup-timeout 2min --bloop-bsp-timeout 1min TestÅÄÖåäö.scala
Hello from TestÅÄÖåäö
philwalk@d5 MSYS ~/workspace/scala-cli-3307-clean
# /opt/jdk17/bin/java -Xmx512m -Xms128m -jar 'C:\Users\philwalk\workspace\scala-cli-3307-clean\out\cli\3.3.7\standaloneLauncher.dest\launcher.bat' --bloop-startup-timeout 2min --bloop-bsp-timeout 1min TestÅÄÖåäö.scala
Hello from TestÅÄÖåäö
```
One possible fix might be to run this one test with jdk18 or later (not sure how to do that, if it's possible).

The details on encoding corner cases and other issues that affect jdk17 or earlier is [here:](https://bugs.openjdk.org/browse/JDK-8187041#:~:text=Source%20file%20encoding,That%20is%2C%20Charset.) the first jdk that avoids these problems is jdk 18.

I'll keep investigating ...

Gedochao:
> One possible fix might be to run this one test with jdk18 or later (not sure how to do that, if it's possible).

Depends what "running with JDK18" means here.
If it's the JDK used for user code and/or Bloop, then just check how we do it in other tests, for example here:
https://github.com/VirtusLab/scala-cli/blob/863bec3b28b61bafb13837908452030d8e7cdf9e/modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala#L37-L59

`-Dfile.encoding=UTF-8` should be possible to pass directly to the `scala-cli` native launcher (before the sub-command):
```bash
scala-cli -Dfile.encoding=UTF-8 compile (...)
```

If, however, this would mean the `scala-cli` launcher would require JDK18, that's a tougher cookie... I don't think we'd be bumping the required Java for the time being (not before Scala 3 compiler does it, and the upcoming 3.9 LTS will require JDK 17)

philwalk:
Here's a summary of what I discovered about this error.
Perhaps someone recognizes this behaviour and can suggest where thing might be going wrong.

Here's what I concluded:

- if the filename is utf-8 encodeded (`fileName.getBytes.length > fileName.length)
- the result is "File not found: TestÃ…Ã„Ã–Ã¥Ã¤Ã¶.scala

- if a copy of the source file is created in the test directory with the mojibaked name
- the "File not found" message goes away and is replaced by "source file not found: TestÃƒâ€¦Ãƒâ€žÃƒâ€“ÃƒÂ¥ÃƒÂ¤ÃƒÂ¶.scala

- The exact failing command line passed to `os.proc.call` inside the `mill` test succeeds if passed directly by a script
- so the problem is unlikely to be happening in `os-lib`.

If a filename without non-ascii characters is tested, the test succeeds and the test script (enhanced to report all encoding environment factors) reports utf-8 everywhere, plus code page 65001.  The mill test itself also reports correct encoding settings, although a couple of changes were made to `build.mill.scala` and in the build environment to achive this.

# conclusions
This implies **two distinct mojibake transformations**, each caused by a different misinterpretation of UTF-8 bytes at different layers.

UTF-8 → Win1252 → UTF-8 → Win1252

  UTF-8 → Win1252 : TestÃ…Ã„Ã–Ã¥Ã¤Ã¶.scala
  Win1252 → UTF-8 : TestÃƒâ€¦Ãƒâ€žÃƒâ€“ÃƒÂ¥ÃƒÂ¤ÃƒÂ¶.scala

### 🧠 Working Hypothesis: Layered Encoding Misinterpretation

#### ✅ Step 1: You create a UTF-8 filename — `"TestÅÄÖåäö.scala"`
- UTF-8 bytes: `54 65 73 74 C3 85 C3 84 C3 96 C3 A5 C3 A4 C3 B6`

#### ✅ Step 2: You pre-create a **once-mojibaked** file — `"TestÃ…Ã„Ã–Ã¥Ã¤Ã¶.scala"`
- This simulates UTF-8 bytes being misread as Windows-1252
- It prevents the `"File not found"` error because the subprocess can now find a file matching its misinterpreted name
- a new error message appears "source file not found" (printed by scalac)

#### ✅ Step 3: Mill’s test harness or `scala-cli` passes the filename to a subprocess
- The subprocess receives `"TestÃ…Ã„Ã–Ã¥Ã¤Ã¶.scala"` as a UTF-8 string
- But then **re-encodes it as UTF-8 bytes**, and **misinterprets those bytes again as Windows-1252**
- This produces `"TestÃƒâ€¦Ãƒâ€žÃƒâ€“ÃƒÂ¥ÃƒÂ¤ÃƒÂ¶.scala"`

#### ✅ Step 4: `scalac` receives this doubly-mangled filename
- It cannot find a file with that name
- Hence: `"source file not found: TestÃƒâ€¦Ãƒâ€žÃƒâ€“ÃƒÂ¥ÃƒÂ¤ÃƒÂ¶.scala"`
— the "source file not found" error path

### 🧩 Root Cause Hypothesis

The filename is:
1. Correctly encoded as UTF-8
2. Misinterpreted as Windows-1252 by the shell or `scala-cli`
3. Re-encoded and passed to the `scala-cli.exe` or `java .*launcher.bat` subprocess
4. Misinterpreted again by the subprocess JVM before being passed to `scalac`

This layered misinterpretation possible causes:
- Argument marshaling in Mill’s test harness
- `scala-cli` not validating filenames before passing them to `scalac`
- JVM subprocesses inheriting encoding assumptions from the parent shell

philwalk:
It turns out that when an `os.proc(...).call(...)` is run from `mill` it inherits 2 system properties that cause scripts with non-ascii script names to fail.  They can be set to `UTF-8` in the jvm running the integration test, but the spawned `scala-cli` script reports these values:

`sun.jnu.encoding = Cp1252`
`native.encoding = Cp1252`

If I duplicate the failing test in a `scala-cli` script, the bad encoding values are present if the script is run by `scala-cli` version 1.9.1.   But if I run the script with a new GraalVM `scala-cli.exe` (compiled after setting the code page to 65001) then these values are inherited (apparently from `scala-cli.exe`) and the test succeeds:

`sun.jnu.encoding = UTF-8`
`native.encoding = UTF-8`

The same is true of the launcher `out/cli/3.3.7/standaloneLauncher.dest/launcher.bat` as demonstrated by this bash script.  The implication is that the launcher succeeds or fails depending on whether Windows code page is 65001 or not. 

```bash
# time ./mill integration.test.jvm 'scala.cli.integration.RunTestsDefault.UTF-8' 
# time ./mill integration.test.native 'scala.cli.integration.RunTestsDefault.UTF-8' 
```

Here's the demo shell script.
With  scala-cli 1.9.1 in the PATH and it fails, run it with scala-cli from this PR build and it succeeds.

```bash
#!/bin/bash

cat > "/tmp/testÅÄÖåäö.sc" <<'EOF'
//> using dep com.lihaoyi::os-lib:0.8.1
object TestÅÄÖåäö {
  import java.nio.charset.Charset
  var launcherName = "scala-cli.exe"
  val filename = "testÅÄÖåäö.sc"
  def showEncoding(): Unit = {
    printf("======= jvm encoding configuration:\n")
    import scala.jdk.CollectionConverters.*
    import scala.sys.process.*
    System.err.printf("code-page: [%s]\n", ("chcp.com".!!).trim)
    System.err.printf("JAVA_TOOL_OPTIONS[%s]\n", System.getenv("JAVA_TOOL_OPTIONS"))
    System.err.printf("native.encoding = %s\n",  System.getProperty("native.encoding"))
    System.err.printf("sun.jnu.encoding = %s\n", System.getProperty("sun.jnu.encoding"))
    System.err.printf("file.encoding = %s\n",    System.getProperty("file.encoding"))
    System.err.printf("Charset.defaultCharset = %s\n", java.nio.charset.Charset.defaultCharset())
    System.err.printf("Class name = %s\n", this.getClass.getName)
    printf("fileName[%s]\n", sys.props("scala.source.names"))
    import java.lang.ProcessHandle
    def displayAncestors(ph: ProcessHandle): Unit = {
      import java.util.Optional
      val parentHandle: Optional[ProcessHandle] = ph.parent()
      if (parentHandle.isPresent) {
        val parent = parentHandle.get()
        val command = parent.info().command().orElse("N/A")
        println(s"PID: ${parent.pid()}, Command: $command")
        displayAncestors(parent)
      } else {
        println("--- Reached top of process tree ---")
      }
    }
    println("--- Process Ancestry ---")
    // Display the current process first
    val currentHandle = ProcessHandle.current()
    val currentCommand = currentHandle.info().command().orElse("N/A")
    println(s"PID: ${currentHandle.pid()} (Current Process), Command: $currentCommand")
    displayAncestors(ProcessHandle.current())
  }

    val asciiOnly: Boolean = args.contains("-ascii")

    //val p = if scriptPath.fwd.startsWith("/") then os.Path(scriptPath) else os.Path(scriptPath, pwd)
    showEncoding()
    println(s"Hello from $filename")
    println(s"Hello from $scriptPath")
    printf("launcher: %s\n", launcher)
    val subProcessScriptpath = if asciiOnly then "/tmp/"+filename.replaceAll("[^\\x00-\\x7F]", "") else filename

    printf("subProcessScriptpath [%s]\n", subProcessScriptpath)
  
    val endOfCopy = "=== script cloning ends here ==="
    val p = java.nio.file.Paths.get(scriptPath)
    val osp = osPath(scriptPath)
    val content = os.read(osp, scala.io.Codec.UTF8)
    val topPart = content.split("[\r\n]+").takeWhile(!_.contains(endOfCopy)).filter(!_.contains("launcher")).mkString("\n")+"\n  }\n}"
    os.write.over(osPath(subProcessScriptpath), topPart)

    val extraArgs = Seq("--bloop-startup-timeout", "2min", "--bloop-bsp-timeout", "1min")

    printf("\nos.proc.call cloned copy: subProcessScriptpath [%s]\n", subProcessScriptpath)
    val res = os.proc(
      launcher,
      subProcessScriptpath
    )
      .call()
    val outstr = res.out.text(scala.io.Codec.UTF8).trim
    printf("outstr[%s]\n", outstr)
  }
  import scala.sys.process.*
  lazy val launcher = Seq("where.exe", launcherName).!!.trim.replace('\\', '/').split("[\r\n]+").toSeq.head
  def osPath(path: String): os.Path = {
    import java.nio.file.{Paths, Files}
    val p = Paths.get(path).toAbsolutePath
    os.Path(p.toString)
  }
}
EOF

FNAME=/tmp/testÅÄÖåäö.sc

scala-cli.exe run $FNAME # okay
echo "######################################" 1>&2
java -Xmx512m -Xms128m -jar 'out/cli/3.3.7/standaloneLauncher.dest/launcher.bat' $FNAME
```

The script in the HERE-doc is a self-contained `scala-cli` script that should fail in the same way as the 2 integration tests, but it is unable to demonstrate the problem.
Just like the itegration test, the script  launches a script with non-ascii filename via `os.proc`.
It succeeds if the `scala-cli.exe` on the PATH was built with Windows code page 95001 set. 

Added Note: success/failure of this is also a function of which Java 17+ distribution is in the PATH.

philwalk:
This most recent commit adds extensive logging to the failing integration test for diagnostic purposes.
A much simpler and smaller test function is intended.

@Gedochao - are there any restrictions to using code created in collaboration with Copilot?

Gedochao:
> @Gedochao - are there any restrictions to using code created in collaboration with Copilot?

There are none, go ahead. We use a standard [Apache 2.0 License](https://github.com/VirtusLab/scala-cli/blob/main/LICENSE) and there's no specific policy against code created with LLM-reliant tooling.

Gedochao:
@philwalk I converted the PR to draft format. I will keep track of what's happening here, feel free to mark it back as ready to review whenever. thanks for working on this.

philwalk:
> @philwalk I converted the PR to draft format. I will keep track of what's happening here, feel free to mark it back as ready to review whenever. thanks for working on this.

Thanks.  I'm trying to determine whether the two failing Windows tests for the initial PR commit are actually valid in Windows:
```
integration.test.jvm scala.cli.integration.RunTestsDefault.UTF-8
integration.test.native scala.cli.integration.RunTestsDefault.UTF-8
```
I'm trying to evolve the code into a valid UTF-8 test in the `mill` test environment.   It seems possible if not likely that the tests as originally coded would be able to pass without using UTF-8 encoding, because the non-ascii String  `ÅÄÖåäö` also has valid encodings in the Windows default `Cp1252` as well as various other encodings.   When I force `mill` to use the `utf-8` encoding, it doesn't easily get propagated down to the spawned `scala-cli` environment, although the equivalent test script (above) seems to do so.   I haven't figured out yet why it behaves differently in `mill`.

<h3>✅ Encodings with Valid Byte Representations for <code>ÅÄÖåäö&quot;</code></h3>
<p>These encodings support the string <code>ÅÄÖåäö&quot;</code> with valid byte representations:</p>

Encoding | Byte Representation (Hex) for ÅÄÖåäö | Notes
-- | -- | --
UTF-8 | C3 85 C3 84 C3 96 C3 A5 C3 A4 C3 B6 | Fully supported; portable across platforms
Windows-1252 | C5 C4 D6 E5 E4 F6 | Common on Windows; compatible with Western European locales
ISO-8859-1 | C5 C4 D6 E5 E4 F6 | Similar to Windows-1252 but lacks some extra characters
ISO-8859-15 | C5 C4 D6 E5 E4 F6 | Adds euro symbol and other tweaks to ISO-8859-1
MacRoman | 8A 8E 8F 86 84 94 | Legacy Mac encoding; not recommended for new systems
Latin-9 | C5 C4 D6 E5 E4 F6 | Alias for ISO-8859-15
CP437 | 8F 8E 99 86 84 94 | Original IBM PC encoding; limited compatibility
</h2>

philwalk:
@Gedochao - any idea what led to this error?
```
cs.exe java --jvm temurin:17 -version
  Exception in thread "main" java.lang.Exception: Error while getting https://github.com/coursier/jvm-index/raw/master/index.json: coursier.cache.ArtifactError$DownloadError: download error: Caught java.io.IOException (Server returned HTTP response code: 503 for URL: https://github.com/coursier/jvm-index/raw/master/index.json) while downloading https://github.com/coursier/jvm-index/raw/master/index.json
  	at coursier.jvm.JvmIndex$.$anonfun$load$3(JvmIndex.scala:166)
  	at coursier.jvm.JvmIndex$.$anonfun$load$3$adapted(JvmIndex.scala:164)
  	at coursier.util.Task$.$anonfun$flatMap$extension$1(Task.scala:14)
  	at coursier.util.Task$.$anonfun$flatMap$extension$1$adapted(Task.scala:14)
  	at coursier.util.Task$.wrap(Task.scala:82)
  	at coursier.util.Task$.$anonfun$flatMap$2(Task.scala:14)
  	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
  	at java.base@17.0.9/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
  	at java.base@17.0.9/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
  	at java.base@17.0.9/java.lang.Thread.run(Thread.java:840)
  	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:807)
  	at com.oracle.svm.core.windows.WindowsPlatformThreads.osThreadStartRoutine(WindowsPlatformThreads.java:179)
Error: The process 'C:\hostedtoolcache\windows\cs\2.1.24\x64\cs.exe' failed with exit code 1
```

Using the same version of cs on my system, the command succeeds.

Gedochao:
Oh, hey, I see the CI is green.
@philwalk is it ready for re-review?

philwalk:
ready

.github/workflows/ci.yml
```diff
@@ -778,6 +783,11 @@ jobs:
       with:
         fetch-depth: 0
         submodules: true
+    - name: Import custom registry and verify
+      uses: ./.github/actions/windows-reg-import
+      with:
+        reg-file: .github/ci/windows/custom.reg
+      continue-on-error: true
```
Gedochao: ...why should it continue on error?
won't the tests fail if this errors out?

mill
```diff
@@ -70,7 +70,8 @@ function to_bash_syntax {
 if [[ $IS_WINDOWS ]]; then
   # needed for coursier version < 2.1.8, harmless otherwise
   IFS=$'\n'
-  eval "$(to_bash_syntax `"$cs" java --env --jvm temurin:17` || to_bash_syntax `"$cs" java --env --jvm openjdk:1.17.0`)"
+  # temurin:17 doesn't support utf8 filenames
+  eval "$(to_bash_syntax `"$cs" java --env --jvm zulu:17` || to_bash_syntax `"$cs" java --env --jvm openjdk:1.17.0`)"
```
Gedochao: oh wow... that's an interesting flaw for `temurin` 🤨

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
```diff
@@ -2457,4 +2469,34 @@ abstract class RunTestDefinitions
...
+    }
+  }
+  import java.nio.charset.Charset
+  System.err.println(s"Charset.defaultCharset: ${Charset.defaultCharset}")
+  System.err.println(s"fileName Chars: ${fileName.map(c => f"U+$c%04x").mkString(" ")}")
+  def main(args: Array[String]): Unit = {
+    print("""" + message + """") // no newline needed here
+  }
+}
+""".trim.replaceAll("MAINCLASS", classname).replaceAll("FILENAME", fileName)
```
Gedochao: This is... overkill. 😧 
I mean, don't get me wrong, I can see this works, but we used to just
```scala
object TestÅÄÖåäö {
  println("$message")
  }
}
```
😐

Gedochao: I'd rather you specialcased the old tests behind `if !Properties.isWindows`, and added a dedicated Windows test instead.

.github/workflows/ci.yml
philwalk: didn't know if there would be permission problems.  I will redo it later today.

mill
philwalk: I filed a bug report here:  https://github.com/adoptium/adoptium-support/issues/1382

modules/integration/src/test/scala/scala/cli/integration/RunTestDefinitions.scala
philwalk: We can revert to the unmodified original test now, the other version was only to aid diagnosis in case things didn't work.

.github/workflows/ci.yml
philwalk: removed the continue-on-error lines

philwalk:
@Gedochao - I think it's ready to go now.

mill
philwalk: The issue with Temurin 17 turns out to be subtle.
The version of `temurin:17` in my coursier cache did not support utf8
The version acquired after rebuilding the cache does appear to support utf8.

Perhaps it would be okay on the Windows CI test host.

Here's running java hello world script `testÅÄÖåäö.java` with the old and the newer versions:

With older cache `temurin:17` (build 17+35):

`/f/temurin-jdk17/bin/java.exe testÅÄÖåäö.java`
```diff
-openjdk version "17" 2021-09-14
-OpenJDK Runtime Environment Temurin-17+35 (build 17+35)
-OpenJDK 64-Bit Server VM Temurin-17+35 (build 17+35, mixed mode, sharing)
!error: source file not found: testÃ…Ã„Ã–Ã¥Ã¤Ã¶.java
```

Here's the newer version `temurin:17.0.17`:
`/opt/temurin-jdk17/bin/java.exe testÅÄÖåäö.java`
```diff
+openjdk version "17.0.17" 2025-10-21
+OpenJDK Runtime Environment Temurin-17.0.17+10 (build 17.0.17+10)
+OpenJDK 64-Bit Server VM Temurin-17.0.17+10 (build 17.0.17+10, mixed mode, sharing)
!Hello, world
```

Gedochao (approved):
LGTM!
Thanks for the persistence, this seems to have been quite the quest to get right.

PR #3926, opened 2025-10-28
Migrate integration tests to Scala 3

Gedochao:
## What was done
- the `integration` module is now fully built with Scala 3.3.7
- `scala-async` syntax has been rewritten with vanilla Scala 3 Futures syntax
- a number of minor workarounds were necessary because of issues with the `expecty` dependency
  - https://github.com/eed3si9n/expecty/issues/212 / https://github.com/com-lihaoyi/os-lib/issues/361
  - https://github.com/eed3si9n/expecty/issues/197 / https://github.com/eed3si9n/expecty/issues/59

## Potential follow-ups
- completely remove `expecty` from the project
  - way out of scope for this PR
  - this was tempting to do here, but requires a project-wide refactor
  - tl;dr we now have the workarounds around the issues mentioned above everywhere, and the library is perhaps more trouble than it is worth
- cross-compile the `integration` module against Scala 3 Next versions
  - this would be a good idea both for dogfooding and in preparation to the 3.9 LTS bump
  - it may be a good idea to finally bump to Mill 1.x, first

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
@@ -1,18 +1,18 @@
 package scala.cli.integration
 
+import ch.epfl.scala.bsp4j as b
 import ch.epfl.scala.bsp4j.JvmTestEnvironmentParams
-import ch.epfl.scala.{bsp4j => b}
 import com.eed3si9n.expecty.Expecty.expect
 import com.google.gson.{Gson, JsonElement}
 
 import java.net.URI
 import java.nio.file.Paths
 
-import scala.async.Async.{async, await}
+import scala.cli.integration.TestUtil.await
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
-import scala.concurrent.duration._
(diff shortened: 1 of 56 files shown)


modules/integration/src/test/scala/scala/cli/integration/BspTestDefinitions.scala
```diff
@@ -252,20 +251,25 @@ abstract class BspTestDefinitions extends ScalaCliSuite
...
-        else
-          expect(compileProducts.contains(os.rel / "simple$.class"))
+        if (actualScalaVersion.startsWith("3.")) {
+          val s = os.rel / "simple$_.class"
+          expect(compileProducts.contains(s))
+        }
+        else {
+          val s = os.rel / "simple$.class"
+          expect(compileProducts.contains(s))
+        }
```
zielinsky:
```suggestion
        val classFilePath = os.rel / 
          if (actualScalaVersion.startsWith("3.")) 
          then "simple$_.class"
          else "simple$.class"
        expect(compileProducts.contains(classFilePath))
```

zielinsky (changes requested):
LGTM! Just one minor change from me

zielinsky (approved):

PR #3929, opened 2025-10-29
Drop `scala3-runtime`, `scala3-graal` & `scala3-graal-processor` & deprecate pre-Scala-3.3 native images

### LINKED ISSUE #3922: Handle `sun.misc.Unsafe` in `scala3-runtime`
Why do we actually use Unsafe in Scala CLI? Looks like we might need to remove it if possible

_Originally posted by @tgodzik in https://github.com/VirtusLab/scala-cli/pull/3920#discussion_r2455369850_

It seems we retain an old hack for lazy vals, which used to be necessary pre-Scala 3.3.
We need to
- check if it's still needed for any reason
  - if yes, make sure it'll handle JDK 24+ well (no warnings)
  - if no, perhaps get rid of it altogether

Gedochao:
Closes https://github.com/VirtusLab/scala-cli/issues/3922

## What and why?
These modules have been used historically to handle issues with `sun.misc.Unsafe`/ lazy val implementations in Scala 3.0-3.2.
This is AFAIK no longer needed  in modern Scala versions.
As pre-3.3 Scala is no longer supported, I am dropping those from the repository.
We still keep the dependency on old versions of those modules for the time being, just so that one could build native images with Scala 3.0-3.2, although that code path is deprecated for future removal.

## Changes
- when generating a native image on Graal with Scala 3.0-3.2, the user will now get a warning
  ```
  [warning] building native images with Scala 3 older than 3.3.0 is deprecated.
  [warning] support will be dropped in a future Scala CLI version.
  [warning] it is advised to upgrade to a more recent Scala version
  ```
- removed 3 modules from the project, their usage is frozen at the last published version (1.9.1)
   - `scala3-runtime`
   - `scala3-graal`
   - `scala3-graal-processor`

## Potential follow-ups
Were we to ever bump the internal JDK to 24 or higher, we'd need to remove this completely; for now, seems rather harmless to leave this as status quo.
We can consider removing it when we bump internal Scala to 3.9 - as far as I see it, Scala 3.0-3.2 is EOL, anyways. So perhaps sometime in 2026.

## Further reading
- https://github.com/VirtusLab/scala-cli/pull/830
- https://github.com/alexarchambault/mill-native-image/pull/20
- https://github.com/alexarchambault/mill-native-image/issues/19
- https://github.com/scala/scala3/pull/15296
- https://github.com/scala/scala3/pull/16614


lbialy:
If and when `lazyvalgrade` is done, you might want to consider introducing it instead as with GraalVM native-image situation is way different than with jvm itself - there's literally no reason not to build binaries with a modern GraalVM instead of GraalVM for jdk 1.8 so it would mean we would replace broken lazy vals with the ones from Scala 3.8 before pushing the bytecode into the native-image.

tgodzik (approved):

PR #3931, opened 2025-10-30
Misc unit test fixes

Gedochao:
- applied `scalafix` in unit tests everywhere
- addressed numerous compiler warnings in unit test modules
- other minor things

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/TestLogger.scala
+++ b/modules/build/src/test/scala/scala/build/tests/TestLogger.scala
@@ -3,15 +3,15 @@ package scala.build.tests
 import bloop.rifle.BloopRifleLogger
 import coursier.cache.CacheLogger
 import coursier.cache.loggers.{FallbackRefreshDisplay, RefreshLogger}
-import org.scalajs.logging.{Logger as ScalaJsLogger, NullLogger}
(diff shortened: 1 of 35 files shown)


modules/build/src/test/scala/scala/build/tests/TestLogger.scala
```diff
@@ -76,16 +76,16 @@ case class TestLogger(info: Boolean = true, debug: Boolean = false) extends Logg
     sn.Logger.nullLogger
   def scalaNativeCliInternalLoggerOptions: List[String] =
     List()
-  def bloopBspStderr        = None
-  def bloopBspStdout        = None
-  def bloopCliInheritStdout = false
-  def bloopCliInheritStderr = false
+  def bloopBspStderr: Option[Nothing] = None
+  def bloopBspStdout: Option[Nothing] = None
```
tgodzik: Is this even needed? If the inferred type is Nothing?

tgodzik (changes requested):

modules/build/src/test/scala/scala/build/tests/TestLogger.scala
Gedochao: seems it's not. removed.

tgodzik (approved):
LGTM

PR #3962, opened 2025-11-18
Self-contained docker build with ARM64 publishing

### LINKED ISSUE #2132: Publish Arm64 Docker image
It is possible to publish multi-platform images to Dockerhub: https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images

Would be good to have a linux/arm64 image published, as it would speed up considerably the development process for people using Apple Silicon hardware.

Considerations:
- While it's possible to build arm64 image on x86 "hardware" (github actions), it will be very slow. It's best to utilis existing apple silicon hardware you use for building aarch64 darwing binaries.

keynmol:
Fixes #2132

At the moment, the scala-cli docker image is built using a combination of mill tasks, scripts, and dockerfiles that require externally built binary. What's more, it's only built for x86.

With ARM64 runners on GHA being GA, I propose a simplification - a self-contained multi-stage dockerfile, with a custom Github Workflow that merges the images into a single manifest, meaning that docker will pull the correct image no matter the target platform.

Manifest merging and pushing is a complicated step, and [Docker have been amending their docs with the example](https://github.com/docker/docs/pull/21854), which is what this workflow is based on. 

I've been using this setup in multiple apps, e.g. [Mimalyzer](https://github.com/indoorvivants/mimalyzer/actions/runs/18510179491).

---

There are some aspects of the build I don't really understand – static, mostly static images, customisation for Linux x86, etc.

This proposal attempts to do the simplest possible thing, so that building a docker image is just `docker build . -t VirtusLab/scala-cli`.

Currently, it only publishes to ghcr.io to test out the workflows without disturbing the main image on docker hub. But additional publishing steps are easily added, of course


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM eclipse-temurin:17 as build
+RUN apt update && apt install build-essential libz-dev clang procps git -y
+WORKDIR /workdir
+COPY . .
+RUN ./mill -i copyTo --task 'cli[].base-image.nativeImage' --dest "./docker-out/scala-cli" 1>&2
+
+FROM debian:stable-slim
+RUN apt update && apt install build-essential libz-dev clang procps -y
```
keynmol, line 8: Note that I'm not sure if installing those dependencies in a base image is a good idea, or better left to downstream images. Even though I see the appeal of making native and native-image stuff work out of the box..

Main concern is the size 
CleanShot 2025-11-19 at 09 38 47@2x

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
+name: Create and publish a Docker image
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# Configures this workflow to run every time a change is pushed to the branch called `release`.
```
Gedochao: ...you mean like, `v*` tags and `main`, right?

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+  DOCKERFILE: ./Dockerfile
+  REGISTRY_LOGIN: ${{ github.actor }}
+  REGISTRY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  docker_build:
+    strategy:
+      fail-fast: true
+      matrix:
+        os: ["ubuntu-22.04", "ubuntu-22.04-arm"]
```
Gedochao:
```suggestion
        os: ["ubuntu-24.04", "ubuntu-24.04-arm"]
```
we're using 24.04 everywhere in the build, I think

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ matrix.os }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  docker_release_merge:
+    runs-on: ubuntu-22.04
```
Gedochao:
```suggestion
    runs-on: ubuntu-24.04
```

Dockerfile
Gedochao: I mean, we definitely want Scala Native to build nicely with it, so `clang` is a must.
I'd like to keep them, I think.

Gedochao (commented):
@tgodzik @zielinsky can you take a look as well? I think it'd be good to have another pair of eyes on this (or two)

Gedochao:
> There are some aspects of the build I don't really understand – static, mostly static images, customisation for Linux x86, etc.

Static and mostly static images are in the understanding of GraalVM, as per this doc: https://www.graalvm.org/21.3/reference-manual/native-image/StaticImages/index.html
The way they're built is based on the `mill-native-image` plugin: https://github.com/alexarchambault/mill-native-image

> (...) customisation for Linux x86 (...)

What particular customisation for Linux x86 did you mean?

> (...) etc.

Shoot away, I'll try to answer, or at least direct you in the direction of an answer.

What's there in this area was initially coded by @alexarchambault (who's also the author of `mill-native-image`). While I've tinkered with this here and there, most of it has been lying untouched for a long, long time, and I'm not all that familiar with it either.

tgodzik (commented):
I don't know much about this, I only built basic docker images. Overall look ok as long as it work 😅

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+      #
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
```
zielinsky:
```suggestion
          username: ${{ env.REGISTRY_LOGIN }}
```

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # This step uses [docker/metadata-action](https://github.com/docker/metadata-action#about) to extract tags and labels that will be applied to the specified image. The `id` "meta" allows the output of this step to be referenced in a subsequent step. The `images` value provides the base name for the tags and labels.
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
```
zielinsky: Same situation, a very old version.

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # This step uses the `docker/build-push-action` action to build the image, based on your repository's `Dockerfile`. If the build succeeds, it pushes the image to GitHub Packages.
+      # It uses the `context` parameter to define the build's context as the set of files located in the specified path. For more information, see "[Usage](https://github.com/docker/build-push-action#usage)" in the README of the `docker/build-push-action` repository.
+      # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
```
zielinsky: Same situation, a very old version.

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+          file: ${{ env.DOCKERFILE }}
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache-${{ matrix.os }}
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:buildcache-${{ matrix.os }},mode=max
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      # This step generates an artifact attestation for the image, which is an unforgeable statement about where and how it was built. It increases supply chain security for people who consume the image. For more information, see "[AUTOTITLE](/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds)."
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@v1
```
zielinsky: The `attest-build-provenance` documentation says to use version `@v3`

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
zielinsky:
```suggestion
          username: ${{ env.REGISTRY_LOGIN }}
          password: ${{ env.REGISTRY_PASSWORD }}
```

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
zielinsky:
```suggestion
          username: ${{ env.REGISTRY_LOGIN }}
          password: ${{ env.REGISTRY_PASSWORD }}
```

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+      packages: write
+      attestations: write
+      id-token: write
+      #
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
```
zielinsky: This is a version from two years ago, shouldn't we bump it up to the latest one?

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          pattern: digests-*
+          path: /tmp/digests
+          merge-multiple: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
```
zielinsky: Same here

zielinsky (commented):
A few comments from me 😅

.github/workflows/publish-docker.yml
Gedochao: Suspecting Claude might not be aware of the latest one 😅

keynmol: That's not Claude, I just copied a work flow from my project that I haven't updated in a while.

.github/workflows/publish-docker.yml
keynmol: I purposefully choose the lowest ubuntu version supported by GHA – linking against the newest GLIBC will mean that binaries built by this workflow won't run on Ubuntu 22.04 (at least that's the case with binaries I build with any native lang).

In mostly static/dynamic mode, native image will have the same issue: https://github.com/oracle/graal/issues/6183#issuecomment-1464886838

keynmol:
@Gedochao

> What particular customisation for Linux x86 did you mean?

This line threw me off

https://github.com/keynmol/scala-cli/blob/93079619648e0b475f874ee2676387b488d7a637/project/settings/package.mill.scala#L406

Gedochao:
Just to make sure. 

We currently have the following images, which are updated with each Scala CLI release:
- https://hub.docker.com/r/virtuslab/scala-cli-slim
- https://hub.docker.com/r/virtuslab/scala-cli

@keynmol your intention is to create a third one, yes? 
smth like `virtuslab/scala-cli-platform-native`?

Gedochao:
Also, how can we safely test this?

keynmol:
Here's my plan:

1. Merge this workflow as is, make sure it publishes a working `ghcr.io/virtuslab/scala-cli` image (similar to existing DockerHub `virtuslab/scala-cli`)
2. Modify the workflow in a separate PR to build the slim image in the same way (similar to existing DockerHub `virtuslab/scala-cli-slim`)
3. Add DockerHub publishing steps to both images
4. Remove old docker publishing logic

Up to step 3, the existing docker images on DockerHub are safe and won't be touched, as we only publish to ghcr.io, using it as our testing area.

Steps 1-3 can be done in separate PRs as they won't affect existing dockerhub publishing steps.

In the end, we should have `virtuslab/scala-cli` and `virtuslab/scala-cli-slim` that have a manifest merging two platforms, so users automatically get the right image (see screenshot) 
CleanShot 2025-11-21 at 10 05 06@2x

.github/workflows/publish-docker.yml
zielinsky: @keynmol I reverted to unresolved, you changed it in only one place

Gedochao:
Okay... sounds reasonable.
Let's get the CI green and do it one step at a time.

Gedochao:
The CI is green.
@keynmol the workflow is still on Ubuntu 22.04, while the other workflows already rely on 24.04 (I left comments earlier).
Otherwise, we seem to be good to merge here (might want to rebase on `main` while you're at it)

keynmol:
~~I have responded to the Ubuntu version comment: https://github.com/VirtusLab/scala-cli/pull/3962#discussion_r2549075356~~

Actually scratch that, in this instance it won't have any effect, I will update

Gedochao (approved):
LGTM
@keynmol ready to merge?

Gedochao:
@zielinsky I will wait for your okay before clicking `merge`.

keynmol:
I think so, no doubt it will require more iterations when the work flow actually runs :)

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,132 @@
...
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["v*"]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
```
zielinsky: Doesn't the name of the image have to be in lowercase letters? Currently `github.repository` will return "VirtusLab/scala-cli"  ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts)).

Image

zielinsky (commented):
One question from me before approval.

.github/workflows/publish-docker.yml
Gedochao: @keynmol bump

keynmol: Sorry, missed this - I will amend soon, apparently github actions have no built in (without shelling out) ways to lowercase - so I will hardcode it for now

Gedochao: @keynmol bump

zielinsky (approved):
LGTM for now, let's test it!

keynmol:
Workflow succeeded, and the image is published correctly, containing two platform-specific images: https://github.com/VirtusLab/scala-cli/actions/runs/20428482272/job/58694886877#step:7:1

But I think the visibility of packages needs to be changed in repository and package settings

PR #3975, opened 2025-11-25
Add examples for passing REPL options via using directive to the docs

### LINKED ISSUE #3967: Doc page in REPL init script: add example/explanation of using directive for repl-init-script
**Version(s)**
1.10.1

**Describe the bug**
The [doc page on REPL init script](https://scala-cli.virtuslab.org/docs/commands/repl/#passing-repl-options) should explain how to use this as a using directive as well. People are probably looking her to find a way how to make repl do things every time it starts...

**Expected behaviour**

Include this as an example in the doc-page on REPL init scripts:

```scala
//> using options --repl-init-script "import allmystuff.*" 
```

Gedochao:
Fixes #3967 
cc @bjornregnell

### DIFF
--- a/website/docs/commands/repl.md
+++ b/website/docs/commands/repl.md
@@ -189,8 +189,8 @@ scala> :quit
 Since we started the repl with toolkit enabled, we can use the libraries included in the toolkit directly. In the above example, the `os-lib` library from the toolkit is used to print the current path. 
 
 ### Running snippets directly via the REPL
-It is possible to run code snippets via the REPL (with all the internal quirks of the REPL), rather than using standard runner execution.
-While the difference is subtle (and should generally be invisible to users), it is useful for investigating REPL behavior.
+It is possible to run code snippets via the REPL, with all the internal quirks of the REPL, rather than using standard runner execution.
+While the difference is subtle and should generally be invisible to users, it is useful for investigating REPL behavior.
 The key to doing that is to use the both `--repl-init-script` and `--repl-quit-after-init` options together.
 
 <ChainedSnippets>
--- a/website/docs/commands/repl.md
+++ b/website/docs/commands/repl.md
@@ -119,20 +119,47 @@ scala>
 
 </ChainedSnippets>
 
+- via a using directive, treating them as Scala compiler options:
+
+<ChainedSnippets>
+
+```scala compile title=repl-options.scala
+//> using toolkit default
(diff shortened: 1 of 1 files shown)


website/docs/commands/repl.md
```diff
@@ -147,20 +174,54 @@ scala-cli repl --toolkit default
...
+val res0: os.Path = /current/directory/path
 scala> :quit
 
 ```
 </ChainedSnippets>
 
 Since we started the repl with toolkit enabled, we can use the libraries included in the toolkit directly. In the above example, the `os-lib` library from the toolkit is used to print the current path. 
 
+### Running snippets directly via the REPL
+It is possible to run code snippets via the REPL (with all the internal quirks of the REPL), rather than using standard runner execution.
```
tgodzik:
```suggestion
It is possible to run code snippets via the REPL, with all the internal quirks of the REPL, rather than using standard runner execution.
```
Do we need parenthesis?

website/docs/commands/repl.md
```diff
@@ -147,20 +174,54 @@ scala-cli repl --toolkit default
...
+While the difference is subtle (and should generally be invisible to users), it is useful for investigating REPL behavior.
```
tgodzik:
```suggestion
While the difference is subtle and should generally be invisible to users, it is useful for investigating REPL behavior.
```
same here, it make sense as a sentence just as is

tgodzik (approved):
Two small nitpicks

bjornregnell:
LGTM - (except for optional comment above on "quirks")
many thanks!!

website/docs/commands/repl.md
bjornregnell: A beginner may wonder what "all" these "quirks" are? The REPL is fantastic :)
(Perhaps exemplify some "quirks" such as slower startup, changed semantics e.g. for multiple declarations etc)

bjornregnell:
Added a comment on "quirks" see above.

website/docs/commands/repl.md
bjornregnell: Perhaps use a more neutral word than negative "quirks", e.g. "differences" or similar. Proposal: 
`[...], with all the differences of REPL execution vs a normal run, such as slower startup, toString printing, and changed semantics of double declarations etc.`

or similar

PR #3984, opened 2025-12-01
Fix handling of multiple exclude directives

### LINKED ISSUE #2836: Multiple Exclude directives not working as expected
**Version(s)**
1.2.1

**Describe the bug**
Attempting to exclude more than one directory, appears to report that the directive is wrongly declared.
image

A single directory, works as expected

```sh
simon@Simons-Mac-mini exclude_directive_bug % scala-cli run .
[error]  Found exclude directives in files:
  project.scala:4:20-28, project.scala:4:31-40
exclude directive must be defined in project configuration file: /Users/simon/Code/exclude_directive_bug/project.scala.
```

**To Reproduce**
```sh
git clone https://github.com/Quafadas/exclude_directive_bug.git
scala-cli run .
```

**Expected behaviour**
The docs suggest that this is supported;

https://scala-cli.virtuslab.org/docs/reference/directives#exclude-sources

I would expect the files in the two excluded directories to be ignored, and the project to run through.

tom91136:
This resolves #2836, I've added a test case for it as well.

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -487,18 +487,22 @@ object CrossSources {
     sources: Seq[PreprocessedSource],
     workspaceDir: os.Path
   ): Either[BuildException, Seq[PreprocessedSource]] = {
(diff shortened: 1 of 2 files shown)


Gedochao (approved):

PR #3992, opened 2025-12-04
Rewrite Scala.js linking to enable packaging of projects without a main method

### LINKED ISSUE #3964: `package --js` failed when source contains no main method
**Version(s)**
scala-cli 1.8.1

**Describe the bug**
Command `scala-cli package --js --js-module-kind es index.scala` fails against source code `index.scala` below:

```scala
//> using platform js
//> using scala 3.7

import scala.scalajs.js.annotation.*

object Index {
  @JSExportTopLevel(name = "handler", moduleID = "index")
  def handler(): Unit = {
    println("Hello world!")
  }
}
```

It results:

```
Compiling project (Scala 3.7.1, Scala.js 1.19.0)
Compiled project (Scala 3.7.1, Scala.js 1.19.0)
[error]  Error: main.js not found after Scala.js linking (found index.js)
```

It works when main method is added:

```scala
//> using platform js
//> using scala 3.7

import scala.scalajs.js.annotation.*

object Index {
  @JSExportTopLevel(name = "handler", moduleID = "index")
  def handler(): Unit = {
    println("Hello world!")
  }
}

object Hello {
  def main(args: Array[String]): Unit = 42
}
```

But I need no main method because I just want ESM.

**To Reproduce**
Run command for same source code.

**Expected behaviour**

scala-cli can build scala source code without main method using Scala.js.

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/3964

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -720,14 +720,14 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
     isFullOpt <- builds.head.options.scalaJsOptions.fullOpt
     linkerConfig = builds.head.options.scalaJsOptions.linkerConfig(logger)
     linkResult <- linkJs(
-      builds,
-      destPath,
-      mainClass,
+      builds = builds,
+      dest = destPath,
+      mainClassOpt = mainClass,
       addTestInitializer = false,
-      linkerConfig,
-      isFullOpt,
-      builds.head.options.scalaJsOptions.noOpt.getOrElse(false),
-      logger
+      config = linkerConfig,
+      fullOpt = isFullOpt,
+      noOpt = builds.head.options.scalaJsOptions.noOpt.getOrElse(false),
+      logger = logger
     )
   } yield linkResult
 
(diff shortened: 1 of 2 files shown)


Gedochao:
Random question - @sjrd is what I do here a good idea?
We've been sort of arbitrarily assuming there's a `main.js` during linking (which there is, if there's a main method).
I guess we could try to parse what the defined module is with `scalameta`... and then use whatever we find?
What I do right now is look for a `<module>.js` in the linking directory, blindly.

sjrd:
That's what the `Report` returned by `Linker.link` is supposed to be for. It tells you the set of modules that it produced.

But I'm not sure what you're even trying to achieve here. What is the expected result of that function?

Gedochao:
> That's what the `Report` returned by `Linker.link` is supposed to be for. It tells you the set of modules that it produced.

Ah, that's a piece of information I was missing. I suppose that's a potential follow-up refactor, so that the modules' `.js` files produced by the linker could be verified...

> But I'm not sure what you're even trying to achieve here. What is the expected result of that function?

https://scala-cli.virtuslab.org/docs/commands/package#scalajs
`package --js` produces `.js` files to be run by Node.
It generally seems the old code arbitrarily assumed there'd be a main method somewhere, and thus the `main.js` was obligatory.
The aim is to make Scala CLI to package something the users can use even if there isn't one, as per https://github.com/VirtusLab/scala-cli/issues/3964
Which seems to already work with the current implementation in this PR.

sjrd:
`package --js` itself should indeed not assume there's a main method/`main.js` module. It shouldn't need to. The linker will produce "a bunch of files" in the target directory, and that's it. That's the output.

If you want to *run* the code, that's a bit more difficult. Then you need to know which module to start executing, basically. For that, even the sbt plugin assumes there is a `main` method in `main.js` (marked with the setting `scalaJSUseMainModuleInitializer`). If not, `run` fails.

Gedochao:
> `package --js` itself should indeed not assume there's a main method/`main.js` module. It shouldn't need to. The linker will produce "a bunch of files" in the target directory, and that's it. That's the output.
> 
> If you want to _run_ the code, that's a bit more difficult. Then you need to know which module to start executing, basically. For that, even the sbt plugin assumes there is a `main` method in `main.js` (marked with the setting `scalaJSUseMainModuleInitializer`). If not, `run` fails.

Thanks, that confirms my assumptions. 
The current approach should be good enough, then.

tgodzik (approved):

PR #3981, opened 2025-11-26
fix #3979 by adding save-and-restore to generated run-native-image.bat

philwalk:
The fix was quite simple; tested on my local system.

PR #4006, opened 2025-12-13
fix #4005 -  Windows native-image compile failure caused by SUBST collision

### LINKED ISSUE #4005: native-image compile fails due to a residual SUBST pointing to Coursier‑managed GraalVM ZIP
**Version(s)**
Scala CLI version: 1.11.0
Scala version (default): 3.7.4

**Describe the bug**
On Windows, `scala-cli package --native-image` attempts to reuse existing SUBSTed letters, and also leaks SUBST drive mappings, causing failed --native-image compiles.

**Summary:**  
On Windows, `NativeImage` uses `subst` to shorten long GraalVM home paths before invoking `native-image`. The method for choosing an available drive letter only avoids drive letters that correspond to actual physical storage, but must also check to see if a drive letter has already been mapped via the SUBST cmd.

**To Reproduce**
The following steps should reproduce this bug.
Assuming I: is the first available drive not corresponding to physical storage, map it to an existing path:
```cmd.exe
subst I: c:\tmp
```
The next attempt to compile should crash with a stackdump that reports `Drive already SUBSTed`.
```
scala-cli --power package --native-image hello.sc -- --no-fallback
Drive already SUBSTed
Error: os.SubprocessException: Result of cmd…: 1

For more details, ...
```
The cause is that [NativeImage#availableDriveLetter()](https://github.com/VirtusLab/scala-cli/blob/82cbf6076cc3f60fdc27a135f234d29154165d84/modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala#L118C15-L118C35) only checks *physical* drives and does not consider SUBST mappings. 

Additionally, when a native image compile is interrupted, it sometimes leaves behind a `SUBSTed` drive letter.

**Expected behaviour**
- native-image compiles in Windows should:

1.   Detect when a drive letter is already SUBSTed.
2.   Validate that SUBST succeeded before invoking native-image.
3.   clean up SUBST mappings via a shutdownhook rather than try/finally block

And possibly also reclaim stale mappings that point into Coursier/GraalVM cache directories.

philwalk:
The following changes are made when calling the GraalVM compiler:

- modified `availableDriveLetter()` to exclude existing `SUBSTed` drive letters
- replaced the `try/finally` block for doing cleanup with a java shutdown hook
- cleanup now includes restoring the original code page as well as removing the `SUBSTed` drive letter
- the code for calling Windows command lines inside the shutdown hook doesn't use `os.proc()`

The shutdown hook runs on JVM shutdown, including Ctrl-C and is more reliable than a `finally` block for cleanup.
Because `os.proc` itself creates a shutdown hook, it isn't used inside the cleanup method.

For readability, (2 Windows command lines are called inside the cleanup method)  a new method was created:
```
// execute a fire-and-forget windows command without using os.proc during shutdown.
private def exec(cmd: Seq[String], logger: Logger, desc: String): Unit
```

Although the code page is restored by the `run-native-image.bat` script (see #3981), it is vulnerable to interruption by `Ctrl-C`.


modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala
```diff
@@ -115,13 +115,26 @@ object NativeImage {
...
 
     @tailrec
     def helper(from: Char): Char =
       if (from > 'Z') sys.error("Cannot find free drive letter")
-      else if (mountedDrives.contains(from)) helper((from + 1).toChar)
-      else from
+      else if (mountedDrives.contains(from) || substDrives.contains(from))
+        helper((from + 1).toChar)
+      else
+        from
```
Gedochao: nitpick
```suggestion
      else from
```

Gedochao (commented):
any chance for us to have an integration test for this?
While the solution looks good to me, I think it may be fragile without a test.

philwalk:
> any chance for us to have an integration test for this? 

Working on it.

A unit test to verify that `NativeImage.availableDriveLetter` doesn't return an already aliased drive would be fast and simple,
but it isn't obvious where to refactor the `availableDriveLetter` method so it's visible to both modules `scala.cli.packaging` and `scala.cli.integration`.   This would me a much faster alternative to the first integration test below.

Integration tests to verify `nextAvailableDrive` letter avoids already aliased letters.
- alias the `nextAvailableDriveLetter` (sufficient to cause a crash in earlier versions of code)
- launch a native-image compile
- verify it didn't crash

Integration test to verify the native-image interrupted cleanup:
- set code page to 65001
- launch a native-image compile
- interrupt the compile (after the code page has changed to 437)
- verify it didn't leave behind aliased drive letters
- verify the code page was restored to 65001

philwalk:
I haven't figured out the `mill` command to only tests the new integration tests.   There are 2 tests in `PackageTestDefinitions` called `test("availableDriveLetters")` and `test("native-image-shutdown-hook")`.    What's the recipe for building the command line?

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -998,6 +1066,58 @@ abstract class PackageTestDefinitions extends ScalaCliSuite with TestScalaVersio
...
+        expect(os.isFile(root / actualDest))
+
+        val res    = os.proc(root / actualDest).call(cwd = root)
+        val output = res.out.trim()
+        expect(output == message)
+
+        unaliasDriveLetter(driveLetter) // undo test condition
+        val substedAfter = substedDrives
+        printf("substedBefore: [%s]\n", substedBefore.toSet.mkString("|"))
+        printf("substedAfter: [%s]\n", substedAfter.toSet.mkString("|"))
```
Gedochao: dev leftovers

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
```diff
@@ -842,6 +842,74 @@ abstract class PackageTestDefinitions extends ScalaCliSuite with TestScalaVersio
...
+  private def execWindowsCmd(cmd: Seq[String]): (Int, String) =
+    val pb = new ProcessBuilder(cmd*)
+    pb.redirectInput(ProcessBuilder.Redirect.INHERIT)
+    pb.redirectError(ProcessBuilder.Redirect.INHERIT)
+    pb.redirectOutput(ProcessBuilder.Redirect.PIPE)
+    val p = pb.start()
+    // read stdout fully
+    val output   = scala.io.Source.fromInputStream(p.getInputStream, "UTF-8").mkString
+    val exitCode = p.waitFor()
+    (exitCode, output)
```
Gedochao: NIT: feels like these helper methods should land in `TestUtil`, rather than here

Gedochao (approved):
LGTM

modules/integration/src/test/scala/scala/cli/integration/PackageTestDefinitions.scala
philwalk: will do

philwalk:
There's a problem with the shutdown hook that shows up in manual testing so I'm converting it to a draft while I work on it.

Gedochao (approved):

PR #4019, opened 2025-12-19
Migrate to Mill 1.0.6 (was 0.12.17)

Gedochao:
Yes, this is yet another attempt at getting Scala CLI up to Mill 1.x.y standards.
Bear with me.
Christmas is near, this just might be the one!

PR #4028, opened 2025-12-23
Support `export`-ing to Mill 1.x.y & allow to override Mill version with `--mill-version`

### LINKED ISSUE #3882: Bring `Mill` export up-to-date and support Mill. 1.0.x
I can't tell, if it's just the documentation or the whole feature, but the shown generated Mill projects looks outdated. 

The page: https://scala-cli.virtuslab.org/docs/commands/export/

A screenshot:

> Image

The following issue are obvious:

* The file is `build.sc`, while Mill since version 0.12 defaults to `build.mill`. Currently Mill 1.0.5 is the latest stable version
* The `object test extends Tests` definition was deprecated in Mill 0.12 and will not work for Mill 1.0. Instead it should be `object test extends ScalaTests`.
* The explicit addition of the dependency `os-lib` in the `test` module seems redundant, since it is already defined in the main project and the `test` module depends on the main module, so it will be resolved transitively
* The mentioned bootstrap script `millw` from the linked project `lefou/millw` is archived. The script is now part of Mill itself.
* While the generated `.mill-version` file still works, using a `//| mill-version: 1.0.5` YAML header in the `build.mill` file should be preferred. This also shows the version to the reader of the listing, making it more obvious which Mill version is targetted by the `build.mill`.

Generally, the shown examples should mention the used versions, so the pages can age better.

Gedochao:
Closes https://github.com/VirtusLab/scala-cli/issues/3882

- extracts pieces of https://github.com/VirtusLab/scala-cli/pull/4019 related to Mill export
- defaults to `build.mill.scala` for the main build file when exporting to Mill 1.x.y (stays on `build.sc` for 0.12.x for now, even though later patches support it)
- adds a `--mill-version` command line option for `export`, which does what you'd think it'd do
- refactored `export` integration tests to run them on both Mill 0.12.17 & 1.0.6
- this is just the tip of the iceberg in regard of what Mill `export` would need to be considered fully stable, but at least it'll now work with modern Mill versions

obvious follow-ups (or a list of things I'd like to do, but probably won't have the time - but then, who knows):
- bring Mill 0.12.17/1.0.6 export up to such a high level of up-to-date-ness, that integration tests don't log deprecation warnings
  - then add a check for those deprecation warnings, so that we don't fall back behind with that
- add support for Mill 1.1.x (it's at 1.1.0-RC3 at the time of me writing this, so it's likely not very far off)
  - run the tests for that as well
- add more test cases
- give a similar amount of love to the other build tools
  - SBT
  - Maven


tgodzik (approved):
LGTM

PR #4041, opened 2026-01-08
Bump libsodium to 1.0.21 (partially, except for static launchers locked at 1.0.18)

Gedochao:
- As every once and then, old `libsodium` links are once again unresponsive, effectively killing our CI.
https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip seems dead, as of the time of me writing this, returning a 404.
- However, the freshly released 1.0.21 seems to work, so let's try that: https://github.com/jedisct1/libsodium/releases/tag/1.0.21-RELEASE
- I'm freezing the alpine & conda libsodium versions at 1.0.18 for now, as bumping that would be a considerable effort and should be done separately (there's a different packaging format there, which `coursier` does not support).
- This supersedes https://github.com/VirtusLab/scala-cli/pull/3494

### DIFF
--- a/build.mill.scala
+++ b/build.mill.scala
@@ -559,6 +559,8 @@ trait Core extends ScalaCliCrossSbtModule
          |
          |  def libsodiumVersion = "${deps.libsodiumVersion}"
          |  def libsodiumjniVersion = "${Deps.libsodiumjni.dep.versionConstraint.asString}"
+         |  def alpineLibsodiumVersion = "${deps.alpineLibsodiumVersion}"
+         |  def condaLibsodiumVersion = "${deps.condaLibsodiumVersion}"
          |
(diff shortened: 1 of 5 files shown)


Gedochao:
https://download.libsodium.org/libsodium/releases/
~I'm getting a feeling that 1.0.18 is EOL, as it's plain absent from the repositories after the 1.0.21 release 🤔 
So we basically have to upgrade.~
Well, they're actually under a legacy `old` directory now.
https://download.libsodium.org/libsodium/releases/old
Which is telling. Yeah, we should try to upgrade.
A fallback solution where we just change the download address and ignore it for now is here:
- https://github.com/VirtusLab/scala-cli/pull/4043

tgodzik (approved):

PR #4055, opened 2026-01-16
Add support for exporting to Mill 1.1.x series (tests on Mill 1.1.0-RC4)

Gedochao:
https://github.com/com-lihaoyi/mill/releases/tag/1.1.0-RC4
As the `.mill.scala` format was dropped in 1.1.0-RC1, I changed the default build file output to `.mill` for Mill 1.0 exports as well.
We've yet to change it in the Scala CLI repo itself.

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
@@ -99,11 +99,13 @@ object Export extends ScalaCommand[ExportOptions] {
     cache: FileCache[Task],
     projectName: Option[String],
     millVersion: String,
+    useLatestLaunchers: Boolean = true,
     logger: Logger
   ): MillProjectDescriptor = {
+    val launcherTag       = if useLatestLaunchers then "main" else millVersion
     val launcherArtifacts = Seq(
-      os.rel / "mill"     -> s"https://github.com/com-lihaoyi/mill/raw/$millVersion/mill",
-      os.rel / "mill.bat" -> s"https://github.com/com-lihaoyi/mill/raw/$millVersion/mill.bat"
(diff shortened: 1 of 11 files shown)


modules/cli/src/main/scala/scala/cli/commands/export0/Export.scala
```diff
@@ -99,11 +99,13 @@ object Export extends ScalaCommand[ExportOptions] {
...
     millVersion: String,
+    useLatestLaunchers: Boolean = true,
     logger: Logger
   ): MillProjectDescriptor = {
+    val launcherTag       = if useLatestLaunchers then "main" else millVersion
     val launcherArtifacts = Seq(
-      os.rel / "mill"     -> s"https://github.com/com-lihaoyi/mill/raw/$millVersion/mill",
-      os.rel / "mill.bat" -> s"https://github.com/com-lihaoyi/mill/raw/$millVersion/mill.bat"
+      os.rel / "mill"     -> s"https://github.com/com-lihaoyi/mill/raw/$launcherTag/mill",
+      os.rel / "mill.bat" -> s"https://github.com/com-lihaoyi/mill/raw/$launcherTag/mill.bat"
```
Gedochao, line 108: To work around https://github.com/com-lihaoyi/mill/issues/6632 / https://github.com/com-lihaoyi/mill/issues/6623, we will now default exported mill launchers to the main branch of the Mill repo, with the option to instead use the version tag hidden behind a flag.
Since export is probably forever an experimental endeavour, and some legacy Mill versions would require using a more recent launcher script anyway, this feels like the lesser evil here.

Gedochao:
Superseded by https://github.com/VirtusLab/scala-cli/pull/4086

PR #4086, opened 2026-01-28
Migrate to Mill 1.1.2 (was 1.0.6)

Gedochao:
https://github.com/com-lihaoyi/mill/releases/tag/1.1.2
Here we go again...

- had to change all .mill.scala files to .mill (.mill.scala got dropped in Mill 1.1.x 😕)
- export sub-command now returns build.mill rather than build.mill.scala for all Mill 1.x versions (used to be build.mill.scala for 1.0.x)
- a lot of reflect-config changes for GraalVM's sake, which smh only started breaking with this Mill bump
- some Windows compat stuff, which also started failing here, randomly
- Scala 2 nightly repo times out more often than not, tagged relevant tests as flaky


tgodzik (approved):
LGTM

PR #4123, opened 2026-02-10
Cut down on integration test suites

Gedochao:
Integration test suites grow in size and become increasingly flaky, making our CI sluggish and unresponsive.
While there is a bunch of improvements that could be done around the tests themselves, I do now believe entire suites are no longer worth running.

Specifically:
- Scala 2.12 and 2.13 suites will now only run on JVM and Linux (Scala 2 is essentially legacy support for Scala CLI, and we very rarely encounter problem specific to a particular platform with them)
- Scala 3.3 LTS suites will now only run on JVM, Linux, MacOS Aarch64 and Windows (this is a compromise approach, with MacOS Aarch64 and Windows test suites being the ones catching problems the most often)

All test suites will still run on the JVM and native Linux launchers (which have the highest availability in the action runners' pool, and thus are not a problem, while still providing the coverage).

Additionally, all integration test suites got renamed in an informative manner:
- `*-1` (group 1: default Scala) becomes `*-default`
- `*-2` (group 2: Scala 2.13) becomes `*-scala-2-13`
- `*-3` (group 3: Scala 2.12) becomes `*-scala-2-12`
- `*-4` (group 4: Scala 3 LTS) becomes `*-lts`
- `*-5` (group 5: Scala 3 Next RC) becomes `*-rc`
- `*macos-m1*` becomes `macos-arm64`

@tgodzik this will require adjusting the CI requirements for merging PRs, but should increase readability 😅 
#shouldHaveDoneThisAgesAgo


tgodzik (approved):
Thanks! LGTM Let me know when to update (don't want to block other PRs)

PR #4126, opened 2026-02-12
Add `coursier/cache-action` on the CI

### LINKED ISSUE #4112: Caching dependencies on CI
**Version(s)**
CI related

Currently I don't see any caching of dependencies being done, which might cause the test to be more flaky due to connection issues. If we cached at least the coursier cache, we would at least get rid of those issues.

**Is your feature request related to a past ticket or discussion?**
No

No caching

**Additional context**
Latest release failed a lot on downloading Scala 3.2.0

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4112

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
       with:
         fetch-depth: 0
         submodules: true
+    - uses: coursier/cache-action@v8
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Thanks!

PR #4140, opened 2026-02-19
Restore Scala 2 nightlies

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4107

Scala 2.12/2.13 nightlies are genuinely broken with the past Scala CLI versions.
The number of versions grew so much, the fetched JSON with their list is now truncated and crashes at parsing.
```
Exception in thread "main" com.github.plokhotnyuk.jsoniter_scala.core.JsonReaderException: expected '}' or ',', offset: 0x0001f91f, buf:
996-11] +----------+-------------------------------------------------+------------------+
996-11] |          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
996-11] +----------+-------------------------------------------------+------------------+
996-11] | 0001f8f0 | 20 3a 20 30 2c 0a 20 20 20 20 22 66 6f 6c 64 65 |  : 0,.    "folde |
996-11] | 0001f900 | 72 22 20 3a 20 74 72 75 65 2c 0a 20 20 20 20 22 | r" : true,.    " |
996-11] | 0001f910 | 72 65 6d 6f 74 65 22 20 3a 20 66 7b 0a 20 20 22 | remote" : f{.  " |
996-11] | 0001f920 | 72 65 70 6f 22 20 3a 20 22 73 63 61 6c 61 2d 69 | repo" : "scala-i |
996-11] | 0001f930 | 6e 74 65 67 72 61 74 69 6f 6e 22 2c 0a 20 20 22 | ntegration",.  " |
```
This should fix it + it restores the disabled tests as well.

PR #4142, opened 2026-02-20
Ensure Scala CLI correctly refers to the latest Scala 2.12/2.13 nightly and recovers from errors

Gedochao:
Follow-up to https://github.com/VirtusLab/scala-cli/pull/4140

The previous solution is buggy, as coursier sorts Scala versions by the version string when fetching from the maven repository. The sorting is smart and works for most version formats, but not for Scala 2 nightlies, which have no indication in the version string regarding order (the're just suffixed with commit hashes).

For example:
- Scala 2.12 nightly: 2.12.22-bin-3df0eeb
- Scala 2.13 nightly: 2.13.19-bin-0272d5a
- Scala 3 nightly: 3.8.4-RC1-bin-20260220-3a6dac8-NIGHTLY

(notice how Scala 3 nightlies can easily be sorted by string alone)

Because of that, this PR does the following
- behaviour from before https://github.com/VirtusLab/scala-cli/pull/4140 is restored (the website works correctly again, the JSON is no longer truncated and seems to return correct results)
  - of course, only the logic from https://github.com/VirtusLab/scala-cli/pull/4140 gets reverted, I kept the automatic tests that were brought back in that PR
- if it fails, we now fall back to fetching the version from the maven repository (as in https://github.com/VirtusLab/scala-cli/pull/4140), but rather than trust coursier sorting to get what's the latest, we refer the commit hashes from https://github.com/scala/scala (`2.13.x` and `2.12.x` branches)
- if that fails as well, we produce a meaningful error, rather than crashing

Feels like it might be overkill, but then, the Scala 2 nightly repo has been failing a number of times in the past, so perhaps this approach will prove more resilient...

cc @SethTisue

### DIFF
--- a/modules/options/src/main/scala/scala/build/options/ScalaVersionUtil.scala
+++ b/modules/options/src/main/scala/scala/build/options/ScalaVersionUtil.scala
@@ -108,9 +108,19 @@ object ScalaVersionUtil {
       cache: FileCache[Task]
     ): Either[BuildException, String] = either {
       val webPageScala2Repo = value(downloadScala2RepoPage(cache))
-      val scala2Repo        = readFromArray(webPageScala2Repo)(using Scala2Repo.codec)
-      val versions          = scala2Repo.children
-      val sortedVersion     =
+      val scala2Repo        = value {
+        try Right(readFromArray(webPageScala2Repo)(using Scala2Repo.codec))
+        catch {
+          case e: JsonReaderException =>
+            val msg =
+              s"""|Unable to compute the latest Scala 2 nightly version.
+                  |Throws error during parsing web page repository for Scala 2.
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Looks good!

SethTisue:
Perhaps @lrytz would consider adopting the version number format with the date for Scala 2? I don't think doing it that way was ever even considered, that I can recall.

In addition to being convenient for automation, it's also convenient for human readers for the date to be included like this.

> the website works correctly again, the JSON is no longer truncated and seems to return correct results

We've been having a lot of Artifactory trouble in the last couple months, as seen at https://github.com/scala/community-build/issues/1738 — knock on wood that it will stay fixed.

Regardless, this PR seems mergeable to me.

PR #4176, opened 2026-03-11
Add Wasm support: `--wasm` flag with Node.js, Deno & Bun runtimes

### LINKED ISSUE #3316: Integrate WASM with Scala CLI
We currently support emitting Wasm via Scala.js, but we do not yet integrate it with Scala CLI features.

I'd like to be able to run Scala code with Wasm directly from Scala CLI.
Unsure which runtimes should be supported, + we should likely have a default.
- [ ] support running Wasm from Scala CLI
  - [ ] a CLI flag `--wasm`
  - [ ] a corresponding directive, likely a `wasm` param for the `platform` directive
- [ ] allow to pick a Wasm runtime + have a default
  - [ ] command line option 
  - [ ] directive
- [ ] consider Wasm for other Scala CLI features (likely out of scope) and create follow-up issues

**Additional context**
- initial support for WASM: #3213 

cc @sjrd @tanishiking

lostflydev:
Implements scala-cli issue #3316: integrate WebAssembly with Scala CLI through the experimental Scala.js Wasm backend (https://www.scala-js.org/doc/project/webassembly.html).

What it adds
- --js-emit-wasm flag and //> using wasm directive to enable Wasm output.
- --js-wasm-runtime <runtime> option and //> using wasmRuntime directive. Supported values: node (default), deno, bun — JS-hosted runtimes that embed a Wasm engine (V8 for Node.js/Deno, JavaScriptCore
  for Bun), sharing the same Scala.js Wasm execution path.

Behavior
- Enabling Wasm implies the JS platform and forces ES module output (both required by the backend). These implicit choices are logged, and combining --js-emit-wasm with --platform native raises an AmbiguousPlatformError.
- --experimental-wasm-exnref is the one flag required by the backend; Scala CLI injects it implicitly (on Node.js, and via DENO_V8_FLAGS for Deno) and logs that it did. WasmGC and typed function-references are on by default on supported runtimes; optional/feature flags (--experimental-wasm-jspi, --experimental-wasm-imported-strings, --turboshaft-wasm) are left to the user via NODE_OPTIONS / DENO_V8_FLAGS.                                                                                                               
- No runtime version detection — Scala CLI loosely couples with the runtime and lets it fail if too old.
- Runs on JS-hosted engines only; standalone Wasm VMs (wasmtime, WasmEdge) are not supported yet — see scala-js/scala-js#4991 (https://github.com/scala-js/scala-js/issues/4991).
  
Tests & docs                                                                                                                                
- Wasm integration tests for Node.js run in all jvm-tests CI groups; Deno and Bun are installed in those jobs too (via setup-deno / setup-bun), so their Wasm tests run on CI as well.
- Added a setup guide: guides/advanced/scala-wasm.md (runtimes, flag behavior, examples, limitations).


Gedochao:
cc @tanishiking @Florian3k @sjrd @lbialy

Gedochao:
@lostflydev I will go over the review in the coming days (this is a hefty one, might take a bit), but in the meantime - it seems the reference doc hasn't been generated.
Refer to https://github.com/VirtusLab/scala-cli/blob/main/CONTRIBUTING.md#rules-for-a-well-formed-pr

modules/cli/src/main/scala/scala/cli/internal/WasmRuntimeDownloader.scala
```diff
@@ -0,0 +1,104 @@
...
+import scala.build.errors.BuildException
+import scala.build.internal.FetchExternalBinary
+import scala.util.Properties
+
+/** Resolves Deno binary for WASM execution.
+  *
+  * Deno is first looked up on the system PATH. If not found, it is downloaded from GitHub releases
+  * and cached via Coursier's ArchiveCache.
+  */
+object WasmRuntimeDownloader {
```
bishabosha: what is the current behavior for `scala-cli run --js foo.scala`? i could be wrong but i would think it does not download a runtime

tanishiking: Yeah, I also wondering the current behavior for that command.
I'm not sure about the policy on downloading the runtime through `scala-cli` (I personally don't like it though), but I believe downloading runtime stuff should be in different PR, because it's not very relevant to "supporing wasm".

modules/options/src/main/scala/scala/build/options/WasmRuntime.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+object WasmRuntime {
+  // JS-based runtimes (work now)
+  case object Node extends WasmRuntime("node")
+  case object Deno extends WasmRuntime("deno")
+  // Standalone runtimes (future - requires upstream Scala.js standalone WASM support)
+  case object Wasmtime extends WasmRuntime("wasmtime")
+  case object WasmEdge extends WasmRuntime("wasmedge")
+  case object Wasmer   extends WasmRuntime("wasmer")
```
tanishiking: If they don't work, I feel we don't need to add those options yet

modules/cli/src/main/scala/scala/cli/internal/WasmRuntimeDownloader.scala
Gedochao: I'd treat WASM runtimes similar to how we treat `node` with Scala.js - the user needs to download it themselves, rather than Scala CLI doing it for them.

modules/options/src/main/scala/scala/build/options/WasmRuntime.scala
Gedochao: +1

modules/cli/src/main/scala/scala/cli/internal/WasmRuntimeDownloader.scala
dos65: Scala-cli has `--jvm` flag and downloads jvm by itself. This behaviour with wasm runtime downloading looks similar.

Gedochao:
(converted to a draft, as this clearly needs more work; feel free to change it back when it's ready to review)

modules/cli/src/main/scala/scala/cli/internal/WasmRuntimeDownloader.scala
Gedochao: @dos65 JVM is key to Scala CLI's core functionalities, and thus it's treated as a major dependency.
Node is considered necessary setup for Scala.js, similar to how `clang` is for Scala Native.
I understand this is blurry and vague, but I think we will expect the user to install WASM runtimes on their own, rather than do it out of the box here (certainly out of scope for this PR, but feel free to start a dedicated discussion or issue on this as a follow-up).

dos65: @Gedochao 
Actually it was impelemented because it was written in https://github.com/VirtusLab/scala-cli/issues/3316 - `Additionally, I'd like to be able to choose a runtime and let Scala CLI download it for me.` 

I don't have any preference of having this functionality - I only want to clarify if this thing really isn't required.

Gedochao: You got me there. 😅 
Not sure why did I define the requirements this way when creating the issue.
Will amend it there.
My bad.

Gedochao:
(I can see the post-review changes, but a rebase will be necessary to re-run the CI)

lostflydev:
> (I can see the post-review changes, but a rebase will be necessary to re-run the CI)

Hi @Gedochao ! I`ve rebased onto actual main, could you pls re-run CI/CD

He-Pin:
why not use bun as a runtime

lostflydev:
> why not use bun as a runtime

@He-Pin Thanks for comment, added bun as a runtime

lostflydev:
Hi @Gedochao, could you please re-run the failed native-windows-tests-default (https://github.com/VirtusLab/scala-cli/actions/runs/25658534589/job/75315398257?pr=4176#logs) job? I think it might be flaky
and not caused by the changes in this PR

Gedochao:
Yep, almost certainly flaky. I restarted it.
There's a bunch of flaky tests and GitHub is very unstable this week, ping me if anything else needs restarting.

lostflydev:
Hi @tanishiking @sjrd @Florian3k @lbialy @dos65 👋

Gentle ping — the PR is ready for another look. Addressed previous feedback: dropped runtime auto-download and unsupported standalone runtimes, added Bun support, rebased, fixed style and docs. 
All required CI checks are green.

Whenever you have a spare moment, I'd really appreciate your feedback

modules/cli/src/main/scala/scala/cli/commands/shared/WasmOptions.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.github.plokhotnyuk.jsoniter_scala.core.*
+import com.github.plokhotnyuk.jsoniter_scala.macros.*
+
+import scala.cli.commands.tags
+
+// format: off
+final case class WasmOptions(
+  @Group(HelpGroup.Wasm.toString)
+  @Tag(tags.experimental)
+  @HelpMessage("Enable WebAssembly output (Scala.js WASM backend). Uses Node.js by default. To show more options for WASM pass `--help-wasm`")
```
tanishiking: Let's use "Wasm" instead of "WASM" ;) 

> A contraction of “WebAssembly”, not an acronym, hence not using all-caps. 
https://webassembly.github.io/spec/core/intro/introduction.html#wasm

modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -265,12 +322,25 @@ object Runner {
       )
       sys.error("should not happen")
     }
+    else if (emitWasm) {
+      // For WASM mode with ES modules, run node directly instead of NodeJSEnv.
+      // NodeJSEnv's stdin piping with "-" doesn't work with Input.ESModule.
```
tanishiking: I believe it works with ESModule, and we don't need this `else if (emitWasm)` branch. What didn't work here?

modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -189,6 +189,55 @@ object Runner {
       run(command, logger, cwd = cwd, extraEnv = extraEnv)
   }
 
+  // Detects the major version of Node.js on PATH; cached for the JVM lifetime (lazy val).
+  // Returns None if node is not found or version cannot be parsed.
+  private lazy val nodeMajorVersion: Option[Int] =
```
tanishiking: I'm not sure about scala-cli policy, but I feel like detecting node version if it supports wasm or not is too much.

my 2 cents: scala-cli should loosely couple with the runtime environment, just try to run and let them fail if it's too old.

modules/build/src/main/scala/scala/build/internal/Runner.scala
```diff
@@ -242,14 +293,20 @@ object Runner {
...
   ): Either[BuildException, Process] = either {
     val nodePath: String =
       value(findInPath("node")
         .map(_.toString)
         .toRight(NodeNotFoundError()))
+    if (emitWasm)
+      nodeMajorVersion.foreach { v =>
+        if (v < 22) value(Left(new NodeVersionTooOldForWasmError(v)))
+      }
+    val nodeFlags = if (emitWasm && nodeNeedsWasmFlag) List("--experimental-wasm-exnref") else Nil
```
tanishiking: I don't think tools like `scala-cli` to hardcode Node options, and instead, let users explicitly specify Node options by themselves something like like: `--node-args=--experimental-wasm-exnref,--experimental-wasm-imported-strings` ?

in Node 26, options like `--experimental-wasm-imported-strings` is removed, and `--experimental-wasm-exnref` is now enabled by default (and may eventually be removed as well). If `scala-cli` hardcode options, we'll be in trouble when underlying runtime (node) removes options.

modules/options/src/main/scala/scala/build/options/WasmOptions.scala
```diff
@@ -0,0 +1,18 @@
+package scala.build.options
+
+/** Options for WebAssembly compilation and execution.
+  *
+  * @param enabled
+  *   If true, enable WASM output (Scala.js WASM backend)
+  * @param runtime
+  *   The WASM runtime to use for execution (node, deno)
+  */
+final case class WasmOptions(
```
tanishiking: Making `wasm` a standalone option sounds like it's an another platform like JS/JVM/Native/Wasm, but in reality everything is passed as scala-js options.

It makes more sense to register it as one of the scala-js options, like `--js-wasm` under `ScalaJSOptions`. WDYT?

tanishiking:
Just left some drive by comments. (Also, I think adding `deno` and `bun` should be in a separate issue/PR).
Anyway, good work, thank you :)

modules/cli/src/main/scala/scala/cli/commands/shared/WasmOptions.scala
lostflydev: Thx, fixed it, in all @HelpMessage annotations, scaladocs, error messages, and code comments

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: The branch is necessary, though the original comment was unclear about the reason. Here's what breaks without it: 

The else path (NodeJSEnv) constructs nodeArgs as: nodeFlags ++ (if args.isEmpty then Nil else "-" :: args.toList) The "-" is the stdin-pipe signal for Input.Script mode. When NodeJSEnv runs an Input.ESModule with non-empty args, the final Node command becomes: node --experimental-wasm-exnref - foo bar baz /tmp/scalajs-runner.cjs
Node sees "-" and reads from stdin, ignoring /tmp/scalajs-runner.cjs. Wasm always uses ESModule, so the program never runs when user args are present
 
The direct ProcessBuilder path produces the correct command: node --experimental-wasm-exnref /tmp/main.mjs foo bar baz
Updated the comment to accurately reflect this. This branch is covered by the "Wasm passes arguments to program" integration test

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Removed it, so scala cli lets runtime fail if it is too old

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Fixed. nodeNeedsWasmFlag is now version-aware: private def nodeNeedsWasmFlag: Boolean = nodeMajorVersion.forall(_ < 25)

modules/options/src/main/scala/scala/build/options/WasmOptions.scala
lostflydev: Good point, moved wasm settings into ScalaJsOptions at both build and CLI layers. 

Directive syntax unchanged — //> using wasm and //> using wasmRuntime work as before, imho it is the most user-friendly approach

lostflydev:
> Just left some drive by comments. (Also, I think adding `deno` and `bun` should be in a separate issue/PR). Anyway, good work, thank you :)

About the Deno/Bun scope: they share the exact same V8-based Wasm execution path as Node. Because of this, the implementation required minimal additions. Since supporting multiple runtimes was in the initial requirements, I included them here

@Gedochao wdyt?

Gedochao:
@lostflydev at a glance, I think it's fine to leave them in this PR.
I also mean to do a round of code review when I have some time, I'll try to do that after the weekend (and feel free to ping me again if I don't get to it).

modules/build/src/main/scala/scala/build/internal/Runner.scala
tanishiking: I don't see it's removed, did you forget to push some commits?

modules/build/src/main/scala/scala/build/internal/Runner.scala
tanishiking: Ah sorry, `--experimental-wasm-exnref` being unnecessary in Node 26+ was just an example. The point was that options passed to Node shouldn't be hardcoded on the scala-cli side. (current implementation is fine since it doesn't add invalid options to node though)

Whether to detect the Node version and automatically add options should follow scala-cli's implementation policy. :) FYI @Gedochao 

---

I was thinking about passing Node options from the `scala-cli` side like `--node-args=--experimental-wasm-exnref,--experimental-wasm-imported-strings`, but there's `NODE_OPTIONS` environment variable. Nevermind! 😄

modules/build/src/main/scala/scala/build/internal/Runner.scala
tanishiking: > When NodeJSEnv runs an Input.ESModule with non-empty args, the final Node command becomes: node --experimental-wasm-exnref - foo bar baz /tmp/scalajs-runner.cjs

I don't think this is true, as commented below, `NodeJSEnv` "runs apps by piping JS to node" (write `Input` content to tmp file and piped JS dynamically `import`s that). So there shouldn't be something like `/tmp/scalajs-runner.cjs` on the final command line. (If that's true, scala-cli's ESModule Input should be broken).

> This branch is covered by the "Wasm passes arguments to program" integration test

I see "Wasm passes arguments to program", and it passes without this branch.

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
```diff
@@ -474,229 +474,305 @@ object Run extends ScalaCommand[RunOptions] with BuildCommandHelpers {
...
-                  val jsDom = build.options.scalaJsOptions.dom.getOrElse(false)
-                  if showCommand then Left(Runner.jsCommand(outputPath.toIO, args, jsDom = jsDom))
-                  else {
-                    val process = value {
+          val jsOpts = build.options.scalaJsOptions
+
+          // Check if Wasm mode is requested
+          if jsOpts.jsEmitWasm then {
+            val runtime  = jsOpts.wasmRuntime
+            val esModule = true // Wasm backend uses ES modules
```
Gedochao: maybe worth a `logger.log` about this being implicitly enabled if user hasn't enabled it

modules/directives/src/main/scala/scala/build/preprocessing/directives/Wasm.scala
```diff
@@ -0,0 +1,56 @@
...
+        }
+      }
+    parsedRuntime.map { runtime =>
+      val wasmEnabled = wasm.getOrElse(false) || wasmRuntime.isDefined
+      // When Wasm is enabled, force Platform.JS (Scala.js Wasm backend requires JS compilation)
+      val scalaOptions =
+        if (wasmEnabled)
+          ScalaOptions(platform = Some(Positioned.none(Platform.JS)))
+        else
+          ScalaOptions()
```
Gedochao: might be worth a log about this being implicit.
Also, this might clash if a user explicitly passes `--wasm` with `--platform native`, is this detected anywhere?

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -325,11 +325,284 @@ trait RunScalaJsTestDefinitions { this: RunTestDefinitions =>
...
+          "--js-emit-wasm",
+          "--js-wasm-runtime",
+          "deno",
+          extraOptions
+        ).call(cwd = root).out.trim()
+        expect(output == "Hello from Deno WASM!")
+      }
+    }
+
+  if (TestUtil.fromPath("bun").isDefined)
```
Gedochao, line 487: are the `bun` tests on the CI?

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
```diff
@@ -325,11 +325,284 @@ trait RunScalaJsTestDefinitions { this: RunTestDefinitions =>
...
+        "--",
+        "foo",
+        "bar",
+        "baz"
+      ).call(cwd = root).out.trim()
+      expect(output == "foo bar baz")
+    }
+  }
+
+  if (TestUtil.fromPath("deno").isDefined)
```
Gedochao, line 461: Are the `deno` tests run on the CI?

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: this conversation seems unresolved?

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: likewise, unresolved?

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: We can add some options as implicit if user doesn't specify them and we know they are necessary to run the Wasm build. Just make sure the implicit stuff is logged, so that the user knows what's happening.

Gedochao (commented):
Left some comments. Sorry to be so late to look at this.

Gedochao:
The `docs-tests` `sclicheck.GifTests.beforeAll(sclicheck.GifTests)` failure is due to a change in the GitHub runner, I fixed it in https://github.com/VirtusLab/scala-cli/pull/4288/changes/aa9145533c9f40dab20581062a1657340c2c8e9a
A rebase should help.

Gedochao:
Not necessarily in the scope of this PR, but - some docs for how to set up the Wasm runtimes and example snippets to run (maybe a dedicated guide) would be a great follow-up.

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: @tanishiking This branch was removed in a previous commit.  I recheck it carefully, you were right

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Apologies for the earlier confusion — I said "removed" but hadn't actually pushed the change. In the end I kept the version detection, because @Gedochao's comment below explicitly approved the approach:
"We can add some options as implicit if user doesn't specify them and we know they are necessary to run the Wasm build. Just make sure the implicit stuff is logged."
 
 The current state: nodeMajorVersion remains, --experimental-wasm-exnref is passed only on Node < 25 (where V8 12.x requires it), and when it is injected, Scala CLI now logs:
 "Wasm: adding --experimental-wasm-exnref (required for Wasm exception handling on Node.js < 25)"

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
lostflydev: added logger.log("Wasm mode enabled: using ES module output on JS platform") right after the implicit assignment

modules/directives/src/main/scala/scala/build/preprocessing/directives/Wasm.scala
lostflydev: added a log in Run.scala when the Wasm path is entered ("Wasm mode enabled: using ES module output on JS platform")

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
lostflydev: Deno and Bun tests are conditionally guarded by TestUtil.fromPath — they run only when the runtime is on PATH. GitHub Actions runners don't include Deno or Bun, so these tests are skipped in CI

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
lostflydev: Deno and Bun tests are conditionally guarded by TestUtil.fromPath — they run only when the runtime is on PATH. GitHub Actions runners don't include Deno or Bun, so these tests are skipped in CI

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
tanishiking: We should test those on CI

modules/build/src/main/scala/scala/build/internal/Runner.scala
tanishiking: (While I personally don't like `scala-cli` manage options implicitly), if we do that, we should properly append all necessary flags for GC and EH across every version of `deno` and `Node`. Why don't we automatically append options for Node 22, 23, and 24? and how about every versions of Deno?

Also, do I understand correctly that, we only manage the minimum required flags for execution (exnref, gc, function-references)? I mean, how about other features such as `js-string-builtins`, `JSPI`, (and custom descriptor in future?) Are those flags are expected to be added by users manually through `NODE_OPTIONS` and `DENO_V8_FLAGS` ?

modules/options/src/main/scala/scala/build/options/WasmRuntime.scala
```diff
@@ -0,0 +1,38 @@
...
+  *   - Node: Uses Node.js (V8 engine) with JavaScript loader
+  *   - Deno: Uses Deno (V8 engine) with ES module support
+  *   - Bun: Uses Bun (JavaScriptCore engine) with ES module support
+  */
+sealed abstract class WasmRuntime(val name: String)
+
+object WasmRuntime {
+  case object Node extends WasmRuntime("node")
+  case object Deno extends WasmRuntime("deno")
+  case object Bun  extends WasmRuntime("bun")
```
tanishiking: All of them are JS runtimes (that embed Wasm engine).
maybe we wanna rename this to JSRuntime? and move into JS option group?

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
Gedochao: +1
there ought to be a way to set those up.

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: that sounds like quite the headache to manage implicitly... 
to chip in - doing some of this for the user is fine, as long as it is clear what needs to be passed manually, and what has been picked up implicitly.
the feature is meant to be experimental, so it's okay to require some knowledge from the user (although if we can provide a seamless UX, that's great)

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
Gedochao: note - this could be done as a follow-up, Deno and Bun support could be extracted to separate PRs (as suggested by @tanishiking) with tests done there.
I don't think we want those silently untested and subject to bitrot.

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
Gedochao: Uh, I meant that we should log about it if the assignment is implicit... the added log currently prints regardless if the user passed the es module options or not, right?

modules/options/src/main/scala/scala/build/options/WasmRuntime.scala
lostflydev: Renamed WasmRuntime → JSRuntime and moved the runtime option into the Scala.js (--help-js) option group. The flag stays --js-wasm-runtime

lostflydev:
> Not necessarily in the scope of this PR, but - some docs for how to set up the Wasm runtimes and example snippets to run (maybe a dedicated guide) would be a great follow-up.

Added a Wasm setup guide (guides/advanced/scala-wasm.md)

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
lostflydev: Added denoland/setup-deno@v2 and oven-sh/setup-bun@v2 to all five jvm-tests jobs, so the Deno/Bun Wasm tests now run in every group

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
lostflydev: Added denoland/setup-deno@v2 and oven-sh/setup-bun@v2 to all five jvm-tests jobs

modules/directives/src/main/scala/scala/build/preprocessing/directives/Wasm.scala
lostflydev: The clash is detected: --js-emit-wasm --platform native (either order) raises AmbiguousPlatformError in SharedOptions.buildOptions. The implicit ES-module enabling is now logged when entering Wasm mode

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
lostflydev: Fixed it — the implicit-ES-module log now only fires when the user didn't pass --js-module-kind. If they did, the linker-config warning handles it instead (and that warning no longer fires when they explicitly chose es)

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Done — the flag is injected implicitly and logged. It's unconditional in Wasm mode now, so no version logic remains

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: I removed the Node-version detection entirely. The [Scala.js docs](https://www.scala-js.org/doc/project/webassembly.html) list --experimental-wasm-exnref as always required, and the flag is accepted on every supported Node — I verified 22 through 25: it's required where exnref is still gated, and a harmless no-op where it's already on by default (Node 25+). The old gate coupled to runtime versions (which you preferred to avoid) and rested on an incorrect comment

On scope: --experimental-wasm-exnref is the only flag Scala CLI injects, since it's the one engine feature still gated on V8 runtimes. WasmGC and typed function-references are enabled by default on supported runtimes, and Scala.js polyfills the JS string builtins, so none of those need a flag. The opt-in/optional flags — --experimental-wasm-jspi (for js.async/js.await), --experimental-wasm-imported-strings (perf), --turboshaft-wasm (Node 23.x stability) — are intentionally left to the user via NODE_OPTIONS / DENO_V8_FLAGS. The injected flag is logged, and this split is documented in a new guide

lostflydev:
@Gedochao all review threads addressed and replied; could you approve the workflow run so CI can re-run? Thanks!

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Reworked this after CI surfaced the real problem. I'd first switched to always injecting --experimental-wasm-exnref, but CI showed Deno 2.8 (V8 14.9) removed the flag — passing it makes Deno abort. And it can't be version-gated either: whether a throwing program needs the flag is non-monotonic across versions (runs unflagged on Node 22.22 and 24, crashes on Node 22.0 and 23.11). So Scala CLI now probes each runtime once — runs it with the flag and checks the exit code — and passes the flag only when accepted: injected where exnref is still gated, skipped where it was removed. This is exactly your "you'll be in trouble when the runtime removes options" concern handled directly. Node takes it as a CLI arg, Deno via DENO_V8_FLAGS; both log when applied; Bun (JavaScriptCore) needs nothing. New integration tests exercise real exception handling (a throwing program, since println doesn't trigger exnref) on deno and bun, and CI now installs deno+bun in all five jvm-tests jobs — verified locally against deno 2.8.2.

modules/build/src/main/scala/scala/build/internal/Runner.scala
lostflydev: Reworked this after CI surfaced the real problem. I'd first switched to always injecting --experimental-wasm-exnref, but CI showed Deno 2.8 (V8 14.9) removed the flag — passing it makes Deno abort. And it can't be version-gated either: whether a throwing program needs the flag is non-monotonic across versions (runs unflagged on Node 22.22 and 24, crashes on Node 22.0 and 23.11). So Scala CLI now probes each runtime once — runs it with the flag and checks the exit code — and passes the flag only when accepted: injected where exnref is still gated, skipped where it was removed. This is exactly your "you'll be in trouble when the runtime removes options" concern handled directly. Node takes it as a CLI arg, Deno via DENO_V8_FLAGS; both log when applied; Bun (JavaScriptCore) needs nothing. New integration tests exercise real exception handling (a throwing program, since println doesn't trigger exnref) on deno and bun, and CI now installs deno+bun in all five jvm-tests jobs — verified locally against deno 2.8.2.

modules/options/src/main/scala/scala/build/options/WasmRuntime.scala
tanishiking: Is there any reason why you want to stay `--js-wasm-runtime`, when it's all js runtimes? I think it's better to rename flags as well.

modules/build/src/main/scala/scala/build/internal/Runner.scala
tanishiking: Hmm, checking the exit code isn't great. Anyone using a newer version of Deno (or a future version of Node) would need to run an extra command just to check whether the flag still exists. Also, a non-zero exit could happen when the program just failed (not because of absent of flag), right?


If you really want to inject flags implicitly, then you'd need to check every version of Node and Deno, inject accordingly, and run CI against multiple runtime versions. (and CI should test against multiple versions of those runtimes)
TBH, I don't think that's worth adding because those flags will eventually become defaults (actually with the latest Deno, no flags needed any longer), and implicit injection will be unnecessary anyway in future.

modules/integration/src/test/scala/scala/cli/integration/RunScalaJsTestDefinitions.scala
tanishiking: If we implicitly inject flags based on deno and node versions, we should test against multiple versions of those runtimes. But, as I commented above, I personally don't think adding implicit flags aren't worth it.

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
tanishiking: I'd also recommend against implicitly setting the module kind to `ESModule` when `useWasm`, and just let users set it explicitly.

- I believe configuration behavior should be consistent with Scala.js
- Also, the next version of Scala.js will require ES2022 or higher. Does that mean we should also implicitly inject `esVersion >= ES2022` for scala-js 1.22.0 or later?
- Currently Wasm can only output `ESModule` right now, so it might seem convenient, but what happens when other module kinds like `MinimalWasm` or `WasmComponent` is added?

tanishiking (commented):
FYI, In the coming version of scala-js we will deprecate `useExperimentalWebAssembly`, and put `useWebAssembly` in `ESFeatures` https://github.com/scala-js/scala-js/pull/5361
We might want to wait until next release (probably soon), and follow the new scala-js configuration structure? (It's fine to add `useWebAssembly`, but it should be soon deprecated).

Gedochao:
> @Gedochao all review threads addressed and replied; could you approve the workflow run so CI can re-run? Thanks!

just ran it!

modules/build/src/main/scala/scala/build/internal/Runner.scala
Gedochao: > with the latest Deno, no flags needed any longer

If this is the case, then +1

modules/cli/src/main/scala/scala/cli/commands/run/Run.scala
Gedochao: > Does that mean we should also implicitly inject esVersion >= ES2022 for scala-js 1.22.0 or later?

Optional UX improvement, IMO. Scala CLI does a bunch of similar stuff elsewhere, so we can. Just need to log about what is being done implicitly, if the user hasn't configured it.

Gedochao: > Currently Wasm can only output ESModule right now, so it might seem convenient, but what happens when other module kinds like MinimalWasm or WasmComponent is added?

Eh, this is worse.
But then, Wasm as a platform is very much an experimental and under-development thing, so I guess things changing would be expected...
I'll rely on your opinion here, @tanishiking

Gedochao: > I believe configuration behavior should be consistent with Scala.js

Fair, +1

Gedochao (approved):
As far as I can tell, this already LGTM

The existing CI failures seem like flakes (GitHub CI has become extremely capricious recently, and we rely on downloading a lot of dependencies on the fly). They will likely pass after a restart or two (I am beginning to associate how GitHub runners behave with moon phases, at this stage).

Before we merge this, I believe there are still some unresolved comments from @tanishiking, mostly on which options should be added implicitly.

tanishiking (approved):

tanishiking:
LGTM, thank you for addressing all comments

Gedochao:
Thanks for the patience in delivering this, great work!

PR #4178, opened 2026-03-12
Add `AGENTS.md`

Gedochao:
Some context for speeding up agents, encouraging certain behaviours and such.

Requires/includes:
- https://github.com/VirtusLab/scala-cli/pull/4177

## Checklist

extensively, this is made by agents for agents, purely vibe coded metaprompting ™️ 
used Cursor with Claude

I use this as context for agents when working with this repo and it works reasonably well

## Additional notes
A similar file in the Scala 3 repo (without the detailed one)
https://github.com/scala/scala3/blob/main/AGENTS.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,165 @@
+# AGENTS.md — Guidance for AI agents contributing to Scala CLI
+
+Short reference for AI agents. For task-specific guidance (directives, integration tests), load skills from *
+*[agentskills/](agentskills/)** when relevant.
+
+> **LLM Policy**: All AI-assisted contributions must comply with the
+> [LLM usage policy](https://github.com/scala/scala3/blob/HEAD/LLM_POLICY.md). The contributor (human) is responsible
+> for every line. State LLM usage in the PR description. See [LLM_POLICY.md](LLM_POLICY.md).
+
+## Human-facing docs
+
+- **[DEV.md](DEV.md)**                   — Setup, run from source, tests, launchers, GraalVM.
+- **[CONTRIBUTING.md](CONTRIBUTING.md)** — PR workflow, formatting, reference doc generation.
+- **[INTERNALS.md](INTERNALS.md)**       — Modules, `Inputs → Sources → Build`, preprocessing.
+
+## Build system
+
+The project uses [Mill](https://mill-build.org/). Mill launchers ship with the repo (`./mill`). JVM 17 required.
+Cross-compilation: default `Scala.defaultInternal`; `[]` = default version, `[_]` = all.
+
+### Key build files
+
+| File                            | Purpose                                                                                  |
+|---------------------------------|------------------------------------------------------------------------------------------|
+| `build.mill`                    | Root build definition: all module declarations, CI helper tasks, integration test wiring |
+| `project/deps/package.mill`     | Dependency versions and definitions (`Deps`, `Scala`, `Java` objects)                    |
+| `project/settings/package.mill` | Shared traits, utils (`HasTests`, `CliLaunchers`, `FormatNativeImageConf`, etc.)         |
+| `project/publish/package.mill`  | Publishing settings                                                                      |
+| `project/website/package.mill`  | Website-related build tasks                                                              |
+
+### Essential commands
+
+```bash
+./mill -i clean                                                           # Clean Mill context
+./mill -i scala …args…                                                    # Run Scala CLI from source
+./mill -i __.compile                                                      # Compile everything
+./mill -i unitTests                                                       # All unit tests
+./mill -i 'build-module[].test'                                           # Unit tests for a specific module
+./mill -i 'build-module[].test' 'scala.build.tests.BuildTestsScalac.*'    # Filter by suite
+./mill -i 'build-module[].test' 'scala.build.tests.BuildTests.simple'     # Single test by name
+./mill -i integration.test.jvm                                            # Integration tests (JVM launcher)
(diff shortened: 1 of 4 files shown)


bishabosha:
where is AGENTS_DETAILED.md recommended?

Gedochao:
@bishabosha recommended?
As a best practice, tbh I haven't encountered this division, although both documents seem to be useful to me depending on what I do.
Open to suggestions, of course.

bishabosha:
my thought was that AGENTS.md is hardcoded to be used as context by those agents that support it, so it isnt useful to have a separate file that they arent looking for, perhaps im wrong.

AGENTS.md
```diff
@@ -0,0 +1,113 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents working with code in this repository.
+
+> **LLM Policy**: All AI-assisted contributions must comply with the
+> [LLM usage policy](https://github.com/scala/scala3/blob/HEAD/LLM_POLICY.md).
+> The human contributor is responsible for every line of code. State LLM usage clearly in the PR
+> description. See also [LLM_POLICY.md](LLM_POLICY.md).
+
+For detailed guidance on all topics below, see [AGENTS_DETAILED.md](AGENTS_DETAILED.md).
```
Gedochao: I mean, `AGENTS.md` points to `AGENTS_DETAILED.md`, so the agents will find it accordingly (Cursor finds it).

Gedochao:
After some experiments and research, it seems it will still be handled better by many tools to just have a single file.
One can create summaries when needed on one's own, I suppose.
@bishabosha I suppose I yield.

He-Pin:
Let agents fix all the issues:)

Gedochao:
> Let agents fix all the issues:)

If only it were that simple. 😅

AGENTS.md
```diff
@@ -0,0 +1,424 @@
...
+> **LLM Policy**: All AI-assisted contributions to this project must comply with
+> the [LLM usage policy](https://github.com/scala/scala3/blob/HEAD/LLM_POLICY.md).
+> The contributor (human) is responsible for every line of code submitted. State LLM usage
+> clearly in the PR description. See also [LLM_POLICY.md](LLM_POLICY.md) in this repo.
+
+## Human-facing docs
+
+Read these for broader context; do not duplicate their content in your PRs:
+
+- **[DEV.md](DEV.md)** — Developer setup, how to run the CLI from sources, run tests, generate launchers, IDE import,
```
tgodzik: NIT: Could probably be a skill

AGENTS.md
```diff
@@ -0,0 +1,424 @@
...
+  GraalVM reflection config, and more.
+- **[CONTRIBUTING.md](CONTRIBUTING.md)** — Fork-pull workflow, branch strategy (`main` for code, `stable` for
```
tgodzik: Does it not contain too much data for LLMs? If we create a skill in theory we could split the data to be loaded more lazily

AGENTS.md
```diff
@@ -0,0 +1,424 @@
...
+| `project/settings/package.mill` | Shared traits, utils (`HasTests`, `CliLaunchers`, `FormatNativeImageConf`, etc.)         |
+| `project/publish/package.mill`  | Publishing settings                                                                      |
+| `project/website/package.mill`  | Website-related build tasks                                                              |
+
+### Essential commands
+
+```bash
+./mill -i scala …args…                            # Run Scala CLI from source
+./mill -i __.compile                              # Compile everything
+./mill -i unitTests                               # All unit tests
```
tgodzik: Maybe add how to run a single test?

AGENTS.md
```diff
@@ -0,0 +1,424 @@
...
+> **Always re-run and verify tests locally before submitting.** After any logic change, run the
+> relevant test suites on your machine and confirm they pass. Do not rely on CI to catch failures —
+> CI resources are shared, and broken PRs waste maintainer time.
+
+### Unit tests
+
+Unit tests live inside individual modules as `test` submodules. The test framework is **munit**.
+
+```bash
+./mill -i 'build-module[].test'                                        # All build-module tests
```
tgodzik: I don't think we should repeat it, we already have some commands defined above. We should merge both

AGENTS.md
```diff
@@ -0,0 +1,424 @@
...
+
+Modules with unit tests include: `build-module`, `build-macros`, `cli`, `directives`, `options`, and others. Check for a
+`test` object inside the module trait in `build.mill`.
+
+**When adding a feature**, always start with unit tests. Unit tests for directives and build options live in
+`modules/build/src/test/scala/scala/build/tests/` (e.g. `BuildTests.scala`, `DirectiveTests.scala`). CLI-level unit
+tests live in `modules/cli/src/test/scala/`. Only reach for integration tests when a unit test cannot cover the
+scenario.
+
+### Integration tests
```
tgodzik: Should the agents/md be that long? Are we doing too much context? Could we skip the additional stuff below?

I think there was a discusion that Agends.md being long was not helpful. I would limit it to how to run test and maybe what each module is. We can use more hierachical structure like skills for more context

tgodzik (changes requested):

AGENTS.md
Gedochao: done

AGENTS.md
Gedochao: should be fixed

AGENTS.md
Gedochao: trimmed it down, removed ~150 lines for now

tgodzik (approved):
Let's try it!

PR #4180, opened 2026-03-12
Use Java 17 mapping when generating docs with Scala 3.8+ with `doc`

Gedochao:
Scala 3.8 uses Java 17, adjusted the `doc` command passed external mappings.

## Checklist

extensively, Cursor + Claude

Added a simple unit test.
Verified the mappings work, otherwise (an integration test feels overkill, but let me know if I should add smth anyway).

## Additional notes
Kudos to @hamzaremmal for spotting this

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/doc/Doc.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/doc/Doc.scala
@@ -125,16 +125,26 @@ object Doc extends ScalaCommand[DocOptions] {
     logger.message(s"Wrote Scaladoc to $printableOutput")
   }
 
+  private def javadocBaseUrl(javaVersion: Int): String =
+    if javaVersion >= 11 then
+      s"https://docs.oracle.com/en/java/javase/$javaVersion/docs/api/java.base/"
+    else
+      s"https://docs.oracle.com/javase/$javaVersion/docs/api/"
+
+  private def scaladocBaseUrl(scalaVersion: String): String =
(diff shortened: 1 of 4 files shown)


modules/cli/src/main/scala/scala/cli/commands/doc/Doc.scala
```diff
@@ -125,16 +125,23 @@ object Doc extends ScalaCommand[DocOptions] {
...
-    "-Ygenerate-inkuire",
-    "-external-mappings:" +
-      ".*/scala/.*::scaladoc3::https://scala-lang.org/api/3.x/," +
-      ".*/java/.*::javadoc::https://docs.oracle.com/javase/8/docs/api/",
-    "-author",
-    "-groups"
-  )
+  private[commands] def defaultScaladocArgs(scalaVersion: String): Seq[String] =
+    val javadocBaseUrl =
+      if scalaVersion.coursierVersion.isScala38OrNewer then java17JavadocUrl
```
hamzaremmal: Shouldn't it actually depend on the JVM version being used? and have the URL depend on that version?

hamzaremmal: I would even go further and say the same about the `scaladoc3` link.

Gedochao: done and done

hamzaremmal (approved):
Thanks @Gedochao

zielinsky (approved):

PR #4181, opened 2026-03-13
Warn when `.java` & `.scala` sources are used in a mixed compilation with `--server=false`

### LINKED ISSUE #3897: Missing a meaningful error message when trying to pass non-Scala inputs with `--server=false`
**Version(s)**
1.9.1

**Describe the bug**
Scala CLI doesn't support running non-Scala inputs without Bloop (with `--server=false`), but it doesn't raise a meaningful error, and rather lets the compiler fail.

**To Reproduce**
```java
//Side.java
public  class Side {
    public static String message = "Hello";
}
```

```scala
//Smth.scala
@main def main() = println(Side.message)
```

```bash
scala-cli .
# Compiling project (Scala 3.7.3, JVM (23))
# Compiled project (Scala 3.7.3, JVM (23))
# Hello
scala-cli . --server=false
# Exception in thread "main" java.lang.NoClassDefFoundError: Side
#        at Smth$.<clinit>(Main.scala:2)
#         at Smth.main(Main.scala)
# Caused by: java.lang.ClassNotFoundException: Side
#         at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
#         at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
#         at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:528)
#         ... 2 more
```

**Expected behaviour**
```bash
scala-cli .
# Compiling project (Scala 3.7.3, JVM (23))
# Compiled project (Scala 3.7.3, JVM (23))
# Hello
scala-cli . --server=false
# Error: Scala CLI does not support .java inputs with --server=false
```

Gedochao:
Fixes #3897 

## Checklist

extensively, Claude + Cursor

unit & integration tests provided

## Additional notes
It doesn't make sense to error outright here, as the `.java` files still pass through the Java parser step in the compiler, which may be useful to a power user (well, chiefly to a compiler engineer, but still)

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
@@ -62,9 +62,7 @@ abstract class CompileTestDefinitions
         |""".stripMargin
   )
 
-  test(
-    "java files with no using directives should not produce warnings about using directives in multiple files"
-  ) {
+  {
     val inputs = TestInputs(
       os.rel / "Bar.java" ->
         """public class Bar {}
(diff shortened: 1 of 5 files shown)


modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
```diff
@@ -82,6 +82,84 @@ abstract class CompileTestDefinitions
...
+      val res = os.proc(TestUtil.cli, "compile", "--server=false", extraOptions, ".")
+        .call(cwd = root, check = false, stderr = os.Pipe)
+      val errOutput = res.err.text()
+      expect(errOutput.contains(".java files are not compiled to .class files"))
+      expect(errOutput.contains("Affected .java files"))
+      expect(errOutput.contains("--server=false"))
+    }
+  }
+
+  test("no Java warning with --server=false when only Scala sources") {
```
tgodzik: Could we not add an assertion to existing tests instead?

tgodzik (approved):
Just a nitpick otherwise.

modules/integration/src/test/scala/scala/cli/integration/CompileTestDefinitions.scala
Gedochao: not as simple as adding an assertion, as it requires re-running with `--server=false`...
I refactored a bit to reuse existing inputs across tests, should be better overall

zielinsky (approved):

PR #4197, opened 2026-03-23
Add `java-test-runner` module to support running tests with pure Java

### LINKED ISSUE #3010: Add a pure Java test runner
We should add a pure Java test runner at some point, so that we can keep Java test builds pure as well (we currently have to add Scala to them)
Context:
 - https://github.com/VirtusLab/scala-cli/pull/3009#discussion_r1669880350
 - #2940

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/3010

This essentially adds a new (pure Java) `java-test-runner` module to Scala CLI, which is essentially a Java rewrite of the existing `test-runner`.
This in turn enables to not add Scala to test builds for Java projects.
I also added some extra logging to the old `test-runner` to get it on par with the new `java-test-runner`.

## Checklist

extensively, Cursor + Claude

- unit tests for the new `java-test-runner` module included
- integration tests for running tests with pure Java included

### DIFF
--- a/modules/java-test-runner/src/main/java/scala/build/testrunner/JavaTestRunner.java
+++ b/modules/java-test-runner/src/main/java/scala/build/testrunner/JavaTestRunner.java
@@ -16,16 +16,10 @@ import java.util.List;
 public class JavaTestRunner {
 
     public static List<String> commonTestFrameworks() {
-        return Arrays.asList(
-            "munit.Framework",
-            "utest.runner.Framework",
-            "org.scalacheck.ScalaCheckFramework",
-            "zio.test.sbt.ZTestFramework",
-            "org.scalatest.tools.Framework",
-            "com.novocode.junit.JUnitFramework",
-            "org.scalajs.junit.JUnitFramework",
-            "weaver.framework.CatsEffect"
-        );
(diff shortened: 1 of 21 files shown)


modules/java-test-runner/src/main/java/scala/build/testrunner/JavaTestRunner.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.List;
+
+public class JavaTestRunner {
+
+    public static List<String> commonTestFrameworks() {
+        return Arrays.asList(
+            "munit.Framework",
```
tgodzik: Can we actually run these test from Java runner? Don't we only want junit?

tgodzik (approved):
LGTM, just one questions otherwise

modules/java-test-runner/src/main/java/scala/build/testrunner/JavaTestRunner.java
Gedochao: Good catch, we only want JUnit.
Fixed 👍

Issue #4127, opened 2026-02-12
Debian repository needs a new key, SHA1 no longer accepted


DavidGoodenough:
**Version(s)**
N/a

**Describe the bug**

Current version of apt on sid no longer accepts SHA1 PGP keys so scala-cli is no longer installable.

OpenPGP signature verification failed: https://virtuslab.github.io/scala-cli-packages/debian ./ InRelease: Sub-process /usr/bin/sqv returned an error code (1), error message is: Signing key on E77ADC545062908539D5DBD61B92C29C942DFC9B is not bound:            No binding signature at time 2026-02-06T11:10:41Z   because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance   because: SHA1 is not considered secure since 2026-02-01T00:00:00Z

DavidGoodenough:
If you are visiting the deb building process could you also address the issue I raised a while ago about the missing change log?

DavidGoodenough:
See #3501

Gedochao:
> If you are visiting the deb building process could you also address the issue I raised a while ago about the missing change log?

good call, thanks for the reminder.

DavidGoodenough:
Pending resolution could you advertise the URL so we can get the .deb file

Gedochao:
We have updated the key (thanks @tgodzik for the help), this should be resolved after Scala CLI v1.13.0 is out (probably sometime next week).
Leaving the issue open until then.

DavidGoodenough:
Have you resolved #3501 and added a changelog?

Gedochao:
Not the changelog, but the `signed-by` part is on PR at https://github.com/VirtusLab/scala-cli/pull/4207

PR #4207, opened 2026-03-31
Add `signed-by` support to Debian APT repository

### LINKED ISSUE #3501: Scala-cli .deb file is not signed
**Version(s)**
All versions

**Describe the bug**
On a Debian sid system, APT is getting increasingly insistent that.deb files loaded from repositories should be signed.  Scala-cli is not and should be.

**To Reproduce**
If you use Debian sid with an up to date apt , if you run apt update then it will emark that the repository does not have a  signed by value, especially if you have used the apt modernize-sources command to use the new sources file format.

**Expected behaviour**
A distribution key pair should be generated and the .deb should be signed with the private ket and the public key should be made available for download and use in the signed by clause.

Gedochao:
Fixes #3501 

## Checklist
- [x] ~tested the solution locally and it works~ (no way to do that)

extensively, Cursor + Claude

Can't really test it other than do a new release and re-check APT.

## Additional notes
This is somewhat adjacent (but separate) to #4127, which requires regenerating the key with SHA-256+

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -1849,6 +1849,33 @@ object ci extends Module {
     )
       .call(cwd = debianDir, stdin = pgpPassphrase, stdout = inReleasePath)
 
+    // Export the public key as a binary (non-armored) keyring at the repo root so that
+    // users can reference it via `signed-by` in their APT sources.
+    // Binary format is required per https://wiki.debian.org/DebianRepository/UseThirdParty
+    val keyringPath = packagesDir / "scala-cli-archive-keyring.gpg"
+    os.proc("gpg", "--batch", "--yes", "--export", keyName)
+      .call(stdout = keyringPath)
+
+    // Update the .list file to include the signed-by option pointing at the keyring.
+    // This scopes the key to this repository only, preventing the globally-trusted key
+    // security issue described in the Debian wiki.
+    os.write.over(
+      debianDir / "scala_cli_packages.list",
+      "deb [signed-by=/etc/apt/keyrings/scala-cli-archive-keyring.gpg] https://virtuslab.github.io/scala-cli-packages/debian ./\n"
+    )
+
+    // Also provide a DEB822 .sources file for users on modern Debian (apt modernize-sources).
+    // No Components field: this is a flat repository (Suites: ./).
+    // No Architectures field: avoids breaking when aarch64 packages are added later.
+    os.write.over(
+      debianDir / "scala_cli_packages.sources",
+      """Types: deb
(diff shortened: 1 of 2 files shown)


website/docs/_advanced_install.mdx
```diff
@@ -83,7 +83,8 @@ scala-cli version
 Scala CLI can be installed via [apt](https://wiki.debian.org/Apt) packager tool.
 
 ```bash
-curl -sS "https://virtuslab.github.io/scala-cli-packages/KEY.gpg" | sudo gpg --dearmor  -o /etc/apt/trusted.gpg.d/scala-cli.gpg 2>/dev/null
+sudo mkdir -p /etc/apt/keyrings
+curl -sS "https://virtuslab.github.io/scala-cli-packages/scala-cli-archive-keyring.gpg" | sudo tee /etc/apt/keyrings/scala-cli-archive-keyring.gpg > /dev/null
```
tgodzik, line 87: What is this link? `https://virtuslab.github.io/scala-cli-packages/KEY.gpg` exists, while the new one doesn't. Do you know how that key is even created? It seems to be something different from what we use in the secrets for github actions

tgodzik (changes requested):

build.mill
```diff
@@ -1849,6 +1849,33 @@ object ci extends Module {
     )
       .call(cwd = debianDir, stdin = pgpPassphrase, stdout = inReleasePath)
 
+    // Export the public key as a binary (non-armored) keyring at the repo root so that
+    // users can reference it via `signed-by` in their APT sources.
+    // Binary format is required per https://wiki.debian.org/DebianRepository/UseThirdParty
+    val keyringPath = packagesDir / "scala-cli-archive-keyring.gpg"
```
tgodzik, line 1855: Will this create the key needed in the instructions?

website/docs/_advanced_install.mdx
Gedochao: https://github.com/Gedochao/scala-cli/blob/4c1299a25250b2d09a0eb63e27e703dbf9f04bf9/build.mill#L1852-L1857
it's where the `updateDebianPackages` Mill job should write the keyring file, as per this snippet.
We haven't created this file before, it's part of the purpose of this PR. Perhaps something I should've explained in the description. We need to include the public key as a binary keyring in the distribution for users to refer to.
Admittedly, not something I have actually tested locally (not exactly easy to test end to end).

build.mill
Gedochao: yeah, it should.

tgodzik (approved):
Let's try it

Issue #2937, opened 2024-05-24
Optimise & refactor the testing pipeline


Gedochao:
**Version(s)**
v1.3.2 (although I assure you it'll be relevant for a long while)

**General context**
Our testing jobs have gotten awful long & flaky, while we keep on adding more tests. The time has come for us to deal with this, as false-positive failures on the CI are now a daily occurence, I've begun to expect a restart or 2 on every PR or release, and this isn't going anywhere good.

**Describe what needs to be done and why**
Here's what I'd suggest:
- [ ] add the `--offline` flag in all integration tests which don't need to check online activity, to eliminate unnecessary checking of dependencies and other slowdowns/timeouts due to connection quality on the CI.
- [ ] verify if some tests haven't gone deprecated and could potentially be dropped/updated/optimised.
- [ ] eliminate inevitable duplicate tests.
- [x] go through past test reports, identify the flakiest of tests and try to make them pass consistently.
- [ ] (optional, as this a tall order) move some of the integration tests to unit tests, where and if applicable
- [x] https://github.com/VirtusLab/scala-cli/issues/2930
- [x] https://github.com/VirtusLab/scala-cli/issues/2936
- [x] (after #2936 is done) split the test suites we run on M1 similar to other platforms and run them for all workflows, rather than just on the `main` branch.

PR #4208, opened 2026-03-31
Skip CI steps irrelevant to committed changes on PRs

Gedochao:
Relevant to:
- https://github.com/VirtusLab/scala-cli/issues/2937

This is an attempt to save time on running tests irrelevant to changes on a given pull request.
i.e. when changing a doc or adding release notes, it is essentially pointless to run the whole integration suites, which pessimistically can run for hours, due to all the platforms necessary.

A new `changes` job is added, which classifies changes to categories. Steps in other jobs are then subsequently skipped based on category.
This is a bit sneaky, as the jobs have to still "run" to satisfy our PR checks. The actual status on what was run can be checked based on the time a job elapsed & what was logged.

`main` branch & release tags bypass this, retaining the old behaviour (in fact, all jobs which don't originate from a PR retain the old behaviour, running everything).

Additionally, override keywords have been added to allow to still run certain test suites even if the code in a PR would imply they're irrelevant. They're checked in the PR description and have syntax in the style of the Scala 3 compiler repo. Documented them in `DEV.md`.
Not including them in the PR description, as they would actually change behaviour 🙃 

## Checklist
- [x] ~tested the solution locally and it works~ (can't test locally)
- [x] ~ran the code formatter (`scala-cli fmt .`)~ (irrelevant)
- [x] ~ran `scalafix` (`./mill -i __.fix`)~ (irrelevant)
- [x] ~ran reference docs auto-generation (`./mill -i 'generate-reference-doc[]'.run`)~ (irrelevant)

extensively, Cursor + Claude

Testing it on the CI, impossible to test locally.

### DIFF
--- a/.github/scripts/check-override-keywords.sh
+++ b/.github/scripts/check-override-keywords.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Checks the PR body for [test_*] override keywords.
+# Inputs (env vars): EVENT_NAME, PR_BODY
+# Outputs: writes override=true/false pairs to $GITHUB_OUTPUT and a summary table to $GITHUB_STEP_SUMMARY
+
+if [[ "$EVENT_NAME" != "pull_request" ]]; then
+  echo "Non-PR event, setting all overrides to true"
+  for override in test_all test_native test_integration test_docs test_format; do
+    echo "$override=true" >> "$GITHUB_OUTPUT"
+  done
+  exit 0
+fi
+
+TEST_ALL=false; TEST_NATIVE=false; TEST_INTEGRATION=false; TEST_DOCS=false; TEST_FORMAT=false
+
+check_override() {
+  local keyword="$1"
+  local var_name="$2"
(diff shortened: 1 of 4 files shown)


.github/scripts/check-override-keywords.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
tgodzik, line 2: Should we do the scripts in Scala though?

tgodzik (approved):
LGTM, though it's a lot of changes, so just had a quick look

.github/scripts/check-override-keywords.sh
Gedochao: worth considering.
although we have a whole bunch of existing scripts that could also be rewritten in Scala, and I mean to address that separately.
So let's keep this in bash for now.

Gedochao:
> though it's a lot of changes, so just had a quick look

it's also hard to fully test this without getting it on `main`.
So let's just observe after I merge.

PR #4209, opened 2026-03-31
[doc] publish.md: add info about `publish.scm` directive

spamegg1:
There is no `--scm` command line option as far as I can tell.


website/docs/commands/publishing/publish.md
```diff
@@ -127,6 +127,7 @@ provided as a convenience.
 | License | `publish.license` | `--license` | `Apache-2.0`, `MIT`, `Foo:https://foo.com/license.txt`, … | Run `scala-cli publish --license list` to list pre-defined licenses |
 | URL | `publish.url` | `--url` | | |
 | VCS | `publish.vcs` | `--vcs` | `github:VirtusLab/scala-cli`, `https://github.com/VirtusLab/scala-cli.git` |scm:git:github.com/VirtusLab/scala-cli.git|scm:git:git@github.com:VirtusLab/scala-cli.git` | |
+| SCM | `publish.scm` |         |`github:VirtusLab/scala-cli`                                              | Use to set the SCM information of the release |
```
Gedochao, line 130: my bad, it seems `publish.scm` and `publish.vcs` are synonyms.
so this duplicates the previous line (we don't document aliases here)
https://github.com/VirtusLab/scala-cli/blob/8a04f12c376c742ab21fb4ff81d604f6f6e49af9/modules/directives/src/main/scala/scala/build/preprocessing/directives/Publish.scala#L72-L74
only noticed now 😅

spamegg1: OK I'll close this I guess? :sweat_smile:

Gedochao:
Closing, I guess there's no new information to add here (I know the doc needs to be updated separately)

PR #4214, opened 2026-04-08
Run tests with JDK 26

Gedochao:
We should run the basic JDK compat tests with Java 26, since it's out and about now.

## Checklist

not at all

existing test suites

### DIFF
--- a/project/deps/package.mill
+++ b/project/deps/package.mill
@@ -105,7 +105,7 @@ object Java {
     defaultJava,
     minimumJavaLauncherJava
   ).distinct
-  def mainJavaVersions: Seq[Int] = Seq(8, 11, 17, 21, 23, 24, 25)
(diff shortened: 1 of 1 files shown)


project/deps/package.mill
```diff
@@ -105,7 +105,7 @@ object Java {
     defaultJava,
     minimumJavaLauncherJava
   ).distinct
-  def mainJavaVersions: Seq[Int] = Seq(8, 11, 17, 21, 23, 24, 25)
+  def mainJavaVersions: Seq[Int] = Seq(8, 11, 17, 21, 23, 24, 26)
```
tgodzik: Why drop 25? It's LTS, so we should probably keep it

Gedochao: right 🤦 
fixed.

zielinsky (approved):

tgodzik (approved):

PR #4223, opened 2026-04-11
Add directive `packaging.graalvmJvmId`

zrhmn:
Fixes #4222 

Add `graalvmJvmId` param to the `Packaging` class in `scala.build.preprocessing.directives`.
Add relevant `@DirectiveExamples` and `@DirectiveUsage`.
Add test at `scala.build.tests.PackagingUsingDirectiveTests` validating that `graalvmJvmId` gets passed to `nativeImageOptions`.

## Checklist
- [x] tested the solution locally and it works 
- [x] ran the code formatter (`scala-cli fmt .`)
- [x] ran `scalafix` (`./mill -i __.fix`)
- [x] ran reference docs auto-generation (`./mill -i 'generate-reference-doc[]'.run`)

## How much have your relied on LLM-based tools in this contribution?
I relied moderately on GitHub Copilot. Since I am not a regular developer, it was hard for me to manually browse the codebase, until I ask the Copilot chat for help locating where directives were defined. Making the code change was easy without copilot help because the `case class Packaging` code was pretty readable and self-explanatory. Then I had to ask Copilot where to add tests, and finished the tests with Copilot-autocomplete.

## How was the solution tested?

I created a project with the following directives:
```scala
//> using packaging.packageType graalvm
//> using packaging.graalvmJvmId graalvm-java25:25.0.2
//> using packaging.output "sctest01"
//> using packaging.graalvmArgs --no-fallback
```
... and ran with the built binary
```cmd
../scala-cli/out/cli/3.3.7/standaloneLauncher.dest/launcher --power package .
```
(sans the `graalvm-jvm-id` arg) and it compiled, showing version `GraalVM CE 25.0.2+10.1` during compilation. If I repeat this with the `scala-cli` installed on my system, I get `GraalVM CE 17.0.9+9.1` instead, which confirms that the built `launcher` exhibits desired behavior.

PR #4225, opened 2026-04-13
Add additional packaging.graalvm* directives

### LINKED ISSUE #4224: Add missing `packaging.graalvm*` directives
Continuing the discussion from #4222, which was fixed in #4223; and a [comment on that PR](https://github.com/VirtusLab/scala-cli/pull/4223#issuecomment-4230124180): I think the `graalvm-version` and `graalvm-java-version` options should also be directives (in addition to the already added `graalvm-jvm-id` => `packaging.graalvmJvmId` and `graalvm-args` => `packaging.graalvmArgs` which already existed).

The alternative is to not add those directives, and let them be just command-line args. I considered the alternative but I think it is more consistent to have all of the available GraalVM packaging options as command-line args _and_ directives.

zrhmn:
Fixes #4224.

- Add directive `packaging.graalvmVersion` (string)
- Add directive `packaging.graalvmJavaVersion` (int)
- Validate (valid int and >0) of `graalvmJavaVersion`
- Update docs to reflect the changes (as well as a previously missed doc for `graalvmJvmId`)
- Also updated the docs to reflect the "default" GraalVM versions and IDs -- I did not think it mattered too much, but I thought it was best for the sake of consistency.
- Add unit tests for the newly added directives.

## Checklist

**Moderately.** In a previous PR #4223 I got familiarized with the `preprocessing.directives.Packaging` case class as well as relevant unit tests with the help of GitHub Copilot (using Claude Sonnet 4.6), so I did not need much AI assistance this time around. I did trip on the fact that `graalvmJvmVersion` arg of `NativeImageOptions` was supposed to be an `Option[Int]` while the arg I was getting from the `Packaging` args was an `Option[String]`. I tried using a simple `.flatMap(_.toIntOption)` but then realized if a bad (non-int) `graalvmJvmVersion` is passed, it would be replaced silently. So I used AI help to figure out how to weave-in validation into the `buildOptions` method. A hint was all it took; it was easy enough to follow existing patterns.

Unit tests included.

Additionally, I created a test project with the following lines:
```scala
//> using packaging.packageType graalvm
//> using packaging.graalvmVersion 25.0.2
//> using packaging.graalvmJavaVersion 25
//> using packaging.output "sctest02"
//> using packaging.graalvmArgs --no-fallback
```

And ran with
```cmd
../scala-cli/out/cli/3.3.7/standaloneLauncher.dest/launcher --power package .
```

I saw the version `GraalVM CE 25.0.2+10.1` in the build output, which verified that `graalvmVersion` and `graalvmJavaVersion` were being correctly compiled into a valid JVM ID by the just-compiled launcher.

I also tried with an invalid graalvmJavaVersion and saw the expected error output:
```cmd
[error] ./build.scala:10:40
[error] Malformed graalvm-java-version "notint", expected a positive integer
[error] //> using packaging.graalvmJavaVersion foo
[error]
```
verifying that integer validation of `graalvmJavaVersion` was working as intended.

## Additional notes

I think enhancements could be made to `NativeImageOptions`.

Currently, passing just a whole JVM ID (e.g. `graalvm-community:25.0.2`) works smoothly. If the _JVM ID is not passed,_ and instead a Java version (e.g. 25) _along with a_ JVM version (e.g. 25.0.2) is passed, then `NativeImageOptions` uses that to construct a JVM ID (in case of e.g. it would be `graalvm-java25:25.0.2`). If both a JVM ID _as well as_ the Java version + JVM version pair are passed, then the latter get ignored entirely. This is all expected behavior.

If user mistakenly passes an unmatched Java version and JVM version pair (e.g. `graalvm-version=24.0.2 graalvm-java-version=22`), `NativeImageOptions` naively compiles that to an invalid JVM ID (in case of e.g. `graalvm-java22:24.0.2`). This, too, may be considered expected behavior and an obvious user error.

However, something that IMO could be more common and not an obvious user error, would be that user passes a JVM version but neglects to pass a Java version or vice versa (e.g. user passes `graalvm-version=24.0.2`, resulting in `graalvm-jvm-id=graalvm-java17:24.0.2` or user passes `graalvm-java-version=24`, resulting in `graalvm-jvm-id=graalvm-java24:17.0.9`).

So for the user, passing a JVM ID directly may be the recommended action, rendering the separate Java version and JVM version parameters useless and superficial in practice. Alternatively, we can have `NativeImageOptions` resolve JVM version + Java version pair in a more intelligent, slightly better way. Perhaps, if a JVM version is provided without a Java version, the Java version can be determined from the first part of the JVM version (e.g. for JVM version 24.0.2 assume the Java version will always be 24). If Java version is provided without a JVM version, the latest JVM version for the given Java version could automatically be determined somehow. An error could be shown if the JVM version and Java version are both given but do not match. etc. etc.

This is just something I was considering when making the changes in this PR. For now I think the changes in the PR are fine and in-line with the currently expected behavior. I would welcome comments on the suggested improvements to NativeImageOptions and consider making the changes in the future if I can get folks onboard with my thinking.

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/PackagingUsingDirectiveTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/PackagingUsingDirectiveTests.scala
@@ -34,7 +34,7 @@ class PackagingUsingDirectiveTests extends TestUtil.ScalaCliBuildSuite {
     }
   }
 
-  test("graalvm packaging") {
+  test("graalvm packaging jvmId") {
     val inputs = TestInputs(
       os.rel / "p.sc" ->
         """//> using packaging.packageType graalvm
@@ -48,6 +48,22 @@ class PackagingUsingDirectiveTests extends TestUtil.ScalaCliBuildSuite {
       val nativeImageOpt = maybeBuild.options.notForBloopOptions.packageOptions.nativeImageOptions
       expect(nativeImageOpt.jvmId == "graalvm-community:23.0.2")
       expect(nativeImageOpt.graalvmArgs.exists(_.value == "--no-fallback"))
+    }
+  }
+
+  test("graalvm packaging javaVersion") {
+    val inputs = TestInputs(
+      os.rel / "p.sc" ->
+        """//> using packaging.packageType graalvm
+          |//> using packaging.graalvmJavaVersion 23
+          |//> using packaging.graalvmVersion 23.0.2
+          |
+          |def foo() = println("hello foo")
+          |""".stripMargin
+    )
+    inputs.withLoadedBuild(buildOptions, buildThreads, bloopConfig) { (_, _, maybeBuild) =>
+      val nativeImageOpt = maybeBuild.options.notForBloopOptions.packageOptions.nativeImageOptions
+      expect(nativeImageOpt.jvmId == "graalvm-java23:23.0.2")
     }
   }
 
--- a/modules/build/src/test/scala/scala/build/tests/PackagingUsingDirectiveTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/PackagingUsingDirectiveTests.scala
@@ -51,7 +51,7 @@ class PackagingUsingDirectiveTests extends TestUtil.ScalaCliBuildSuite {
     }
   }
 
-  test("graalvm packaging javaVersion") {
+  test("graalvm packaging valid javaVersion") {
     val inputs = TestInputs(
       os.rel / "p.sc" ->
         """//> using packaging.packageType graalvm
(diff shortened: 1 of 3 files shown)


modules/directives/src/main/scala/scala/build/preprocessing/directives/Packaging.scala
```diff
@@ -120,9 +131,23 @@ final case class Packaging(
       }
       .sequence
       .left.map(CompositeBuildException(_))
+    val maybeGraalVMJavaVersion = graalvmJavaVersion
+      .map { version =>
+        version.value.toIntOption
+          .filter(_ > 0)
```
Gedochao: I don't think GraalVM has ever supported Java older than 8. Correct me if I'm wrong. 
Contextual validation could be done as an (optional) follow-up (as specific GraalVM versions support a particular range), but I think it's safe to assume >7 here, and an appropriate message in the error.

modules/build/src/test/scala/scala/build/tests/PackagingUsingDirectiveTests.scala
```diff
@@ -51,6 +51,22 @@ class PackagingUsingDirectiveTests extends TestUtil.ScalaCliBuildSuite {
     }
   }
 
+  test("graalvm packaging javaVersion") {
```
Gedochao: optional: an extra unit test with invalid `packaging.graalvmJavaVersion` wouldn't hurt, but I'm nitpicking here.

Gedochao (approved):
LGTM, thanks!

PR #4258, opened 2026-05-07
Allow to disable auto-setup-ide in build commands with a command line option & config

### LINKED ISSUE #4256: Provide a way to disable creation of .bsp/scala-cli.json
**Problem with the current implementation**

Scala CLI currently creates `{project-root}/.bsp/scala-cli.json` implicitly during normal use such as run, compile, package commands.

This is problematic in mixed-tool workspaces, especially when the project root is primarily managed by another build tool such as Mill, and the IDE is expected to use that build tool’s BSP connection file, for example:

`.bsp/mill-bsp.json`

If Scala CLI is occasionally run in the same project root, it may also create:

`.bsp/scala-cli.json`

After that, the workspace has multiple BSP providers in the same root, which disturbs IDE tooling such as Metals in VS Code.

**Real-world use case**

I have a Mill project opened in VS Code with Metals.

The intended BSP server is:

`.bsp/mill-bsp.json`

However, some helper tasks occasionally invoke Scala CLI in the project root, for example:

```
scala-cli run ...
scala-cli compile ...
scala-cli package ...

```
Scala CLI then creates:

`.bsp/scala-cli.json`

This introduces an extra BSP provider in the same workspace. In practice, this leads to problems like:
- repeated “import build changes?” prompts in Metals / VS Code
- build server confusion in the IDE
- failures in “go to definition / type definition / implementation”
- unstable IDE behavior until `.bsp/scala-cli.json` is manually removed

**Proposed feature**

Please add a way to disable creation and update of:

`{project-root}/.bsp/scala-cli.json`

This should disable BSP file creation for all normal Scala CLI commands like run, compile, package etc.

**Proposed solutions**

Any of these would solve the problem:

Option 1: command-line flag

```
scala-cli run Main.scala --no-bsp-file
scala-cli compile Foo.scala --no-bsp-file
scala-cli package App.scala --no-bsp-file

```
Option 2: environment variable

`SCALA_CLI_DISABLE_BSP_FILE=1 scala-cli run Main.scala`

Option 3: config setting

Something like:

```
{
  "ide": {
    "writeBspConnectionFile": false
  }
}

```

**Expected behavior**

When this mode is enabled:
- Scala CLI should still run / compile / test / package normally
- .bsp/scala-cli.json should not be created
- existing .bsp/scala-cli.json should not be modified

This new feature is also fully backward compatible if the default behavior remains unchanged.

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4256

## Checklist

Extensively, Cursor + Claude

Added automated tests

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/setupide/SetupIde.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/setupide/SetupIde.scala
@@ -93,17 +93,19 @@ object SetupIde extends ScalaCommand[SetupIdeOptions] {
     previousCommandName: Option[String],
     args: Seq[String]
   ): Unit =
-    writeBspConfiguration(
-      SetupIdeOptions(shared = options),
-      inputs,
-      buildOptions,
-      previousCommandName,
-      args
-    ) match {
-      case Left(ex) =>
-        logger.debug(s"Ignoring error during setup-ide: ${ex.message}")
-      case Right(_) =>
-    }
+    if options.autoSetupIdeEnabled then
+      writeBspConfiguration(
+        SetupIdeOptions(shared = options),
+        inputs,
+        buildOptions,
+        previousCommandName,
+        args
+      ) match {
+        case Left(ex) =>
+          logger.debug(s"Ignoring error during setup-ide: ${ex.message}")
+        case Right(_) =>
+      }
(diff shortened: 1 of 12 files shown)


tgodzik (approved):

zielinsky (approved):
lgtm

PR #4259, opened 2026-05-07
Split the `update-packages` CI job down into individual targets

Gedochao:
This ensures individual targets of `update-packages` can be restarted independently in case of failures.

image

The cost is some setup duplication, but I think it's still worth it.

Extensively, Cursor + Claude

It wasn't. Can't run these jobs outside of a release.


tgodzik (approved):
Looks ok, I guess we'll know when it runs.

PR #4261, opened 2026-05-08
Support for `.test.java`

### LINKED ISSUE #4245: Include .test.java files to the test scope
To make a scala source file a part of the test scope, one can either put it into `test` directory or use `.scala.test` file name extension. However, for java sources the second technique (using `.test.java`) doesn't work, which might be surprising to users.

Files with `.test.java` extension should be considered a part of the test scope sources

Placing `.java` files in `test` directory actually works as a workaround

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4245

This adds full support of `*.test.java` inputs being recognised as test scope sources, automatically.

As public classes in Java inputs have to be in `PublicClassName.java` sources, arbitrarily, Scala CLI generates a `PublicClassName.java` source out of a `PublicClassName.test.java` under the hood to make `javac` happy. It's a bit hacky, but I think it's worth it for the consistent UX.

## Checklist

Extensively, Cursor + Claude

included unit and integration tests for all the ways a Java input can be classified as a test scope source (directive, `.test.java`, `test` dir)

### DIFF
--- a/modules/build/src/main/scala/scala/build/CrossSources.scala
+++ b/modules/build/src/main/scala/scala/build/CrossSources.scala
@@ -135,6 +135,10 @@ final case class CrossSources(
 }
 
 object CrossSources {
+  private val testScopeFileSuffixes: Set[String] = Set(".test.scala", ".test.java")
+
+  private def hasTestScopeSuffix(fileName: String): Boolean =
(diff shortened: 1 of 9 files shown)


tgodzik (approved):
LGtM

PR #4262, opened 2026-05-08
Add support for JShell in `repl`

Gedochao:
https://github.com/user-attachments/assets/6caae413-4718-445b-a26e-5b8da95144b7

Note that JShell requires Java 9 or higher.

Some ricochet fixes for Java projects:
- pure Java projects will now use the default Scala versions for declared Scala dependencies (this enables the usage of Scala Toolkit in JShell, for example)
- `--repl-init-script-file` enables to pass `--repl-init-script` via a file containing the script
  - this is a workaround for making initialization scripts on Windows
  - currently, it's a low effort implementation which pastes the contents of the script into the REPL (enforcing it has to be a JShell script for Jshell and a Scala script for Scala)
  - a possible follow-up could add extra handling for those scripts (like validation of contents for a given REPL), but I consider it out of scope for this PR

## Checklist

Extensively, Cursor + Claude

new automated tests


modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
```diff
@@ -482,22 +556,60 @@ object Repl extends ScalaCommand[ReplOptions] with BuildCommandHelpers {
...
+        value(
+          JShellRunner.run(
+            jshellCommand,
+            logger,
+            allowExecve = allowExit,
+            dryRun = dryRun
+          )
+        )
+
+      case ReplBackend.Ammonite =>
```
tgodzik, line 584: Could we remove it now 😅

modules/cli/src/main/scala/scala/cli/commands/repl/SharedReplOptions.scala
```diff
@@ -41,6 +48,13 @@ final case class SharedReplOptions(
   @Hidden
     ammoniteArg: List[String] = Nil,
 
+  @Group(HelpGroup.Repl.toString)
+  @Tag(tags.implementation)
+  @Tag(tags.inShortHelp)
+  @ValueDescription("path")
+  @HelpMessage("Read the REPL init script (--repl-init-script) from a file. Mutually exclusive with --repl-init-script.")
+    replInitScriptFile: Option[String] = None,
```
tgodzik, line 56: Why do we need this now?

tgodzik (approved):
LGTM just two questions from me

modules/cli/src/main/scala/scala/cli/commands/repl/Repl.scala
Gedochao: Yes. If I was wiser, I'd have done it before I started work on this PR. I will do that as a follow-up.

modules/cli/src/main/scala/scala/cli/commands/repl/SharedReplOptions.scala
Gedochao: Necessary to run tests for JShell on Windows. And I wanted to know if it does, indeed, work on Windows.
More on this in the PR description - the option I add here is a bare-bones bridge.

SethTisue:
very cool :+1:

PR #4272, opened 2026-05-14
Improve dependency update actionable diagnostic handling for slow/unresponsive custom repositories

### LINKED ISSUE #3793: `scala-cli compile` attempts to fetch `toolkit_3` from private GitHub Packages repositories, causing compilation to hang
**Describe the bug**
Users of `scala-cli` version `1.8.4` and `cs` version `2.1.25-M16` are experiencing a recurring issue when compiling `.sc` scripts that use private GitHub Packages repositories. Compilation hangs while trying to fetch `org/scala-lang/toolkit_3` from these private repositories – even though `toolkit_3` should be available in public Maven Central repository.

It appears that `scala-cli` is trying to resolve `toolkit_3` dependencies using the private repositories specified via `//> using repository` directives, instead of first checking public repositories.

---

**To Reproduce**
Example script: `example.sc` (excerpt):

```scala
#!/usr/bin/env -S scala-cli shebang --quiet

// note that we don't really know where in the dep tree toolkit is included transitively
// my best guess is script-utils library 

//> using scala 3.5.2
//> using jvm 21
//> using toolkit 0.5.0
//> using repository https://maven.pkg.github.com/the-company/repo-1
//> using repository https://maven.pkg.github.com/the-company/repo-2
//> using dep org.scala-lang.modules::scala-parallel-collections:1.0.4
//> using dep io.github.the-company::models:1.4.127
//> using dep "io.github.the-company::aws-client:1.4.77,exclude=software.amazon.awssdk%apache-client,exclude=software.amazon.awssdk%netty-nio-client"
//> using dep org.user::script-utils:0.9.5
```

Run with:

```bash
scala-cli compile ./scripts/aws/dynamodb/example.sc
```

Output hangs at:

```
https://maven.pkg.github.com/the-company/repo-1/org/scala-lang/toolkit_3/
    0.0% [          ] 0B (0B / s)
https://maven.pkg.github.com/the-company/repo-2/org/scala-lang/toolkit_3/
    0.0% [          ] 0B (0B / s)
```

---

**Expected Behavior**
`scala-cli` should attempt to fetch `toolkit_3` from public repositories (e.g., Maven Central, Sonatype) *before* trying to fetch it from private repositories, which are unlikely to contain this artifact.

---

**Environment**

* `scala-cli` version: 1.8.4
* `cs` version: 2.1.25-M16
* The script used to work previously; the issue is now recurring
* No custom mirrors or resolve order configuration in use

---

**Additional Context**
Adding the following directives does **not** resolve the issue:

```scala
//> using repository central
//> using repository https://repo1.maven.org/maven2
```

The hanging compilation also impacts Metals, which fails to load the script.

Issue passed to me by the affected user in a direct conversation.

Gedochao:
Fixes #3793 

- checking for the latest version of deps (for the actionable diagnostic mentioning a possible bump) now omits `scala-lang.org` `groupId` for user custom repositories
- checking for the latest version of deps now has a per-repo timeout in case the connection is slow (in which case this feature is generally too expensive to do)

### DIFF
--- a/modules/build/src/test/scala/scala/build/tests/ActionableDiagnosticTests.scala
+++ b/modules/build/src/test/scala/scala/build/tests/ActionableDiagnosticTests.scala
@@ -1,7 +1,15 @@
 package scala.build.tests
 
 import com.eed3si9n.expecty.Expecty.expect
+import com.sun.net.httpserver.HttpServer
+import coursier.cache.FileCache
+import coursier.util.Task
 import coursier.version.Version
+
+import java.net.InetSocketAddress
+import java.util.UUID
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{ConcurrentLinkedQueue, Executors}
 
 import scala.build.Ops.*
 import scala.build.Position.File
@@ -9,6 +17,7 @@ import scala.build.actionable.ActionableDiagnostic.*
 import scala.build.actionable.ActionablePreprocessor
 import scala.build.options.{BuildOptions, InternalOptions, SuppressWarningOptions}
 import scala.build.{BuildThreads, Directories, LocalRepo}
+import scala.jdk.CollectionConverters.*
 
 class ActionableDiagnosticTests extends TestUtil.ScalaCliBuildSuite {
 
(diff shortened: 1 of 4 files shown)


tgodzik (approved):

tgodzik:
Is it possible that this caused the actionable diagnostics stop showing up in the IDE? The update is failing on this in https://github.com/scalameta/metals/pull/8618 @Gedochao

tgodzik:
Looks like it stop showing when Scala CLI is started from jar launcher 🤔

PR #4276, opened 2026-05-15
Address updateCentosPackages `GPG_TTY` warnings

Gedochao:
This should solve the `GPG_TTY` warnings from the `update-centos` CI job.

```
warning: Could not set GPG_TTY to stdin: Inappropriate ioctl for device
```

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -2006,6 +2006,7 @@ object ci extends Module {
     val cmd = Seq[os.Shellable](
       "docker",
       "run",
+      "-t",
       "-v",
       s"$packagesDir:/packages",
       "-w",


build.mill
```diff
@@ -2006,6 +2006,7 @@ object ci extends Module {
     val cmd = Seq[os.Shellable](
       "docker",
       "run",
+      "-t",
```
Gedochao, line 2009: `-t` allocates a pseudo-TTY so `rpmsign` can resolve `ttyname(stdin)` and set `GPG_TTY` (untested, really - the only way to test is to run the next release)

Gedochao:
After offline discussion, decided this is not worth the hassle.

PR #4281, opened 2026-05-19
Improve error for when JVM-only dependencies are not found on other platforms

### LINKED ISSUE #4017: scala-cli test --native fails with "No framework found by Scala.js test bridge"
**Version(s)**
1.10.1

**Describe the bug**
Hello - and many thanks for your work! 

Running scala-cli test --native with any test lib compiles and links successfully ... but fails with some misleading error about Scala.js test bridge, even though no JS platform is involved.

**To Reproduce**
```scala
//> using scala 3.5
//> using platform jvm native
//> using dep "org.scalameta::munit::1.2.1"

class SimpleTest extends munit.FunSuite {
    test("foo") {
        assertEquals(1 + 1, 2)
    }
}
```
```bash
scala-cli test . --platform native
```
**Expected behaviour**
Tests run on Native

**Actual**
```scala
 [error]  No framework found by Scala.js test bridge
```

**Environment**
```
  - Scala Native: 0.5.9
  - OS: Linux
```

Gedochao:
Fixes #4017 

New error looks like this:
```
[error]  No framework found by Scala.js test bridge.
The following JVM-only Scala dependencies are on the Scala.js classpath and likely caused the failure:
  - utest_3-0.8.3.jar
Use the platform-aware dependency syntax (extra ':' before the version, e.g. 'org::name::version' / '//> using dep org::name::version') so the Scala.js artifact is fetched instead of the JVM one.
```
The aim is to easier catch the culprit dependency with the improved error message.

## Checklist

Extensively, Cursor + Claude

Added new tests

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/Runner.scala
+++ b/modules/build/src/main/scala/scala/build/internal/Runner.scala
@@ -413,6 +413,25 @@ object Runner {
     if frameworks.nonEmpty then Right(frameworks) else Left(new NoTestFrameworkFoundError)
   }
 
+  private def jvmOnlyScalaJarsOnPlatformClassPath(
+    classPath: Seq[Path],
+    scalaBinaryVersion: String,
+    platformSuffix: String,
+    userDeclaredDepNames: Set[String]
+  ): Seq[String] =
+    if userDeclaredDepNames.isEmpty then Nil
+    else {
+      val jvmSuffix      = s"_$scalaBinaryVersion"
+      val platformMarker = s"_${platformSuffix}_"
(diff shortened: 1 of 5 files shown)


lbialy:
A good question is how many other situations like this should also have this warning. We've seen similar issues with SN linker when one of application deps is added in single colon syntax too.

Gedochao:
> A good question is how many other situations like this should also have this warning. We've seen similar issues with SN linker when one of application deps is added in single colon syntax too.

out of scope here, but I agree it's worth looking into.
I bet a lot of people get lost with this colon syntax and the error messages don't always help.

tgodzik (approved):

PR #4288, opened 2026-05-26
Bump Scala 3 Next RC to 3.8.4-RC3

Gedochao:
https://github.com/scala/scala3/issues/26129

## Checklist

not at all

existing tests

### DIFF
--- a/gifs/Dockerfile
+++ b/gifs/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:24.04
 RUN curl -fsSL https://deb.nodesource.com/setup_12.x | bash -
 RUN apt-get update
 RUN apt-get install ca-certificates-java openjdk-17-jdk openjdk-17-jre -y
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y pv curl clang rubygems nodejs python3-pip &&\
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y pv curl clang rubygems ruby-dev nodejs python3-pip &&\
     rm -rf /var/lib/apt/lists/* &&\
     gem install rouge &&\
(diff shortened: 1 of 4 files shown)


tgodzik (approved):

zielinsky (approved):

modules/integration/src/test/scala/scala/cli/integration/FixScalafixRulesTestDefinitions.scala
```diff
@@ -295,6 +295,16 @@ trait FixScalafixRulesTestDefinitions {
     }
   }
 
+  if actualScalaVersion.contains("RC")
+  then // seems to be a no-op for 3.8.4-RC3, specifically
+    test("explicit-result-types".flaky) {
```
Gedochao, line 300: seems to be a `scalafix` upstream problem, as this rule seems to be a no-op for 3.8.4-RC3, but is working fine with 3.8.4-RC2.
The test will run with stable 3.8.4, so we can re-evaluate it then.

gifs/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:24.04
 RUN curl -fsSL https://deb.nodesource.com/setup_12.x | bash -
 RUN apt-get update
 RUN apt-get install ca-certificates-java openjdk-17-jdk openjdk-17-jre -y
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y pv curl clang rubygems nodejs python3-pip &&\
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y pv curl clang rubygems ruby-dev nodejs python3-pip &&\
```
Gedochao, line 6: `sclicheck.GifTests.beforeAll` seems to have been crashing due to this dependency missing.

PR #4289, opened 2026-05-26
Add a diagnostic for when a script's wrapper name shadows dependency packages

### LINKED ISSUE #4285: Script filename (doodle.sc) shadows dependency package (doodle) — possible scala-cli naming collision
Environment

    scala-cli: 1.14.0
    Scala: 3.8.3
    JVM: Temurin 21
    OS: Windows 10
    Dependency: org.creativescala:doodle-core_3:0.34.0

Description
A script file named doodle.sc causes imports from the doodle dependency to be shadowed. Importing doodle.core._ fails with "value core is not a member of doodle". Renaming the script (e.g., doodle1.sc) fixes the issue. This only occurs for scripts.

Minimal reproduction (copy-paste)

    File: doodle.sc

//> using dep "org.creativescala::doodle-core:0.34.0"
import doodle.core._
println("repro")

    Command: scala-cli run -S 3 doodle.sc --jvm temurin:21

Observed

    Compilation fails with: "value core is not a member of doodle$_"

Expected

    The import should resolve to the package inside the downloaded JAR and script should compile and run (print "repro"). Script filename must not create a generated symbol that shadows dependency packages.

Notes / diagnosis

    Renaming the file to doodle1.sc immediately resolves the problem.
    Likely cause: scala-cli generates a wrapper package/object from the script filename that collides with dependency package names, and that generated symbol takes precedence in name resolution.

Requested additional info for triage

    Full compiler error output (including any stack trace)
    Output of: scala-cli -v run -S 3 doodle.sc --jvm temurin:21
    Any generated wrapper file contents or temp build artifacts (if accessible)

Suggested investigation

    Verify how script filenames are converted to generated package/object names.
    Confirm name-resolution precedence between generated wrapper code and classpath packages.
    Add a regression test for script filename ↔ package name collisions.

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4285

As per https://github.com/VirtusLab/scala-cli/issues/4285#issuecomment-4541583650, added detection for clashes between a script name and dependency packages.
In case of a clash, the build isn't outright failed, and instead a warning is printed.

For example, for a given `cats.sc` script:
```scala
//> using dep org.typelevel::cats-core:2.12.0
```
we'll now get:
```
[warn] Script 'cats.sc' generates a top-level symbol 'cats' that shadows the 'cats' package from dependencies: 'cats-core_3-2.12.0.jar', 'cats-kernel_3-2.12.0.jar'.
[warn] Imports of 'cats.*' from the dependency will not resolve.
[warn] Consider renaming the script (e.g. to 'cats1.sc') or moving it into a subdirectory.
```

## Checklist

Extensively, Cursor + Claude

Integration tests included

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/ScriptUtils.scala
+++ b/modules/build/src/main/scala/scala/build/internal/ScriptUtils.scala
@@ -1,0 +1,41 @@
+package scala.build.internal
+
+import scala.build.{Logger, Sources}
+
+object ScriptUtils {
+  private val ignoredPackageRoots = Set("scala", "java", "javax", "META-INF")
+
+  final case class ScriptDescriptor(
+    name: String,
+    subPath: os.SubPath,
+    filePath: os.Path,
+    shadowedDependencyJars: Seq[String] = Nil
+  )
+
+  def findShadowedDependencyPackages(
+    sources: Sources,
+    classPath: Seq[os.Path],
+    logger: Logger
+  ): Seq[ScriptDescriptor] = {
+    val scripts = sources.scriptTopLevelNames.filterNot(s => ignoredPackageRoots(s.name))
+    if scripts.isEmpty then Nil
+    else
+      val packageRoots =
+        classPath
+          .filter(_.last.endsWith(".jar"))
+          .flatMap(path =>
+            JarUtils.walkClassEntries(path, logger) { (name, _) =>
+              val slashIdx = name.indexOf('/')
+              if slashIdx > 0 then Iterator.single(name.take(slashIdx))
+              else Iterator.empty
+            }.toSet.map(_ -> path)
+          )
+          .groupMap((root, _) => root)((_, path) => path)
+          .view.mapValues(_.toSet).toMap
+      scripts.flatMap { script =>
+        packageRoots.get(script.name).map { jars =>
+          script.copy(shadowedDependencyJars = jars.toSeq.map(_.last).sorted)
+        }
+      }
+  }
+}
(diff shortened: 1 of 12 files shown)


modules/build/src/main/scala/scala/build/internal/ScriptUtils.scala
```diff
@@ -0,0 +1,41 @@
...
+    sources: Sources,
+    classPath: Seq[os.Path],
+    logger: Logger
+  ): Seq[ScriptDescriptor] = {
+    val scripts = sources.scriptTopLevelNames.filterNot(s => ignoredPackageRoots(s.name))
+    if scripts.isEmpty then Nil
+    else
+      val packageRoots =
+        classPath
+          .filter(_.last.endsWith(".jar"))
```
tgodzik: Wouldn't the same happen if we had locally package with the same name, is it limited to jars?

modules/build/src/main/scala/scala/build/internal/ScriptUtils.scala
```diff
@@ -0,0 +1,41 @@
...
+  private val ignoredPackageRoots = Set("scala", "java", "javax", "META-INF")
+
+  final case class ScriptDescriptor(
+    name: String,
+    subPath: os.SubPath,
+    filePath: os.Path,
+    shadowedDependencyJars: Seq[String] = Nil
+  )
+
+  def findShadowedDependencyPackages(
```
tgodzik: Are we walking the classpath every time? Does it make running scripts much slower?

Or maybe we could only run it if any errors are present?

PR #4301, opened 2026-06-04
Ensure scala-cli.sh logs to stderr

alexarchambault:
This makes the `scala-cli.sh` script send the messages it prints to stderr rather than stdout. These messages could be a problem if users pipe the output of a scala-cli command to a file or to another command.

This also fixes an undefined command (`echoerr`, which is undefined) in the script in some cases, and makes it respect `COURSIER_CACHE` in the env if it's set.

## Checklist
- [X] tested the solution locally and it works 
- [X] ran the code formatter (`scala-cli fmt .`)
- [X] ran `scalafix` (`./mill -i __.fix`)
- [X] ran reference docs auto-generation (`./mill -i 'generate-reference-doc[]'.run`)

LLM tools were used to write a first version of the added integration tests. These have been tweaked manually after that.

Manually + added integration tests

### DIFF
--- a/modules/integration/src/test/scala/scala/cli/integration/ShellLauncherTests.scala
+++ b/modules/integration/src/test/scala/scala/cli/integration/ShellLauncherTests.scala
@@ -1,0 +1,63 @@
+package scala.cli.integration
+
+import com.eed3si9n.expecty.Expecty.expect
+
+import scala.concurrent.duration.{Duration, DurationInt}
+import scala.util.Properties
+
+class ShellLauncherTests extends ScalaCliSuite {
+
+  // downloading the launcher + a JVM + the compiler on a cold cache can take a while
+  override def munitTimeout: Duration =
+    10.minutes
+
+  private lazy val launcherScript: os.Path = {
+    val path = Option(System.getenv("SCALA_CLI_SHELL_LAUNCHER")).getOrElse {
+      sys.error("SCALA_CLI_SHELL_LAUNCHER not set")
+    }
+    os.Path(path)
+  }
+
+  private def hasCachedLaunchers(cache: os.Path): Boolean =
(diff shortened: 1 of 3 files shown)


alexarchambault:
(Still a draft only to check what the CI says, no further pushes planned if it's green)

alexarchambault:
(Seems the CI error aren't related to those changes, right?)

Gedochao:
> (Seems the CI error aren't related to those changes, right?)

likely a flake, I restarted it just now

modules/integration/src/test/scala/scala/cli/integration/ShellLauncherTests.scala
```diff
@@ -0,0 +1,63 @@
...
+      appRelPath -> s"""println("$appMessage")"""
+    ).fromRoot { root =>
+      val cache = root / "cs-cache"
+      os.makeDir.all(cache)
+
+      // sanity check: the launcher really isn't in the cache to begin with
+      expect(!hasCachedLaunchers(cache))
+
+      def runLauncher(): os.CommandResult =
+        os.proc(launcherScript, "run", "--server=false", appRelPath).call(
```
Gedochao, line 46: why `--server=false` here?

alexarchambault: I think that makes the test download less things (Bloop, maybe a JVM specifically for it too), that we're not interested in testing here

Gedochao: fair enough.

Gedochao:
(rebased in an attempt to reign in the CI)

Gedochao (approved):
LGTM

PR #4315, opened 2026-06-12
Isolate unit-test Bloop servers per JVM fork

Gedochao:
This should deal with unit test flake failures related to Bloop hanging

## Checklist

extensively, Cursor + Claude

existing tests


Gedochao:
While it behaves better, this seems to still not be enough, lemme experiment some more...

PR #4330, opened 2026-06-16
[WIP] Run tests with Bloop when the build server wasn't disabled

Gedochao:
When the build server wasn't disabled with `--server=false`, we could be running tests with Bloop.
Of course, `test-runner` and `java-test-runner` would still be used with `--server=false`.

## Checklist

Extensively

split the `test` command integration tests into dedicated suites, with and without Bloop tested separately


Gedochao:
This needs a Bloop bump when it's out to bring https://github.com/scalacenter/bloop/pull/2938 in.

PR #4333, opened 2026-06-17
Cut MacOS integration tests down to just sanity tests

Gedochao:
As MacOS runners have become very unstable, we are cutting on running integration tests on MacOS to the absolute bare minimum.

## Checklist

Extensively

ran the entire suite, only the tagged tests ran.


Gedochao:
@tgodzik note: `native-macos-arm64-tests-lts` needs to be removed from required tests (we won't be running it on PRs anymore)

tgodzik (approved):
LGTM

PR #4348, opened 2026-06-24
Add `sloth` support

Gedochao:
Adds Sloth support behind `--sloth` / `//> using sloth`
https://github.com/VirtusLab/sloth

Scala 3, up till 3.7.x series, has been using the (now terminally deprecated) `sun.misc.Unsafe` API. This in turn results in some ugly warnings on JDK 24+ (and likely errors in the future).

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$ 
(file:(...)/maven2/org/scala-lang/scala3-library_3/3.7.4/scala3-library_3-3.7.4.jar)
WARNING: Please consider reporting this to the maintainers of class scala.runtime.LazyVals$
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```

Now, while Scala 3.8+ uses a new implementation, and Scala 3.3 allows for the new implementation on JDK 9+ behind the `-Yfuture-lazy-vals` compiler flag, the problem is not entirely solved. Even if your app is on the latest Scala, it likely still depends on libraries built with Scala 3.3 (or some other <3.8 version). That in turn brings the problem back. And since it is not widespread to suffix Scala 3 libraries with the minor, as we did in Scala 2 days, it is not entirely easy to spot which libraries will pose a problem.

Sloth is a bytecode patcher, which is capable of patching all past lazy val implementations (including the pre-3.3 one) up to the current one. Kudos to @lbialy for developing it.

Behind the `--sloth` flag, this PR allows to post-process and patch the bytecode of your app/artifacts to silently get rid of the problem.
The `--sloth-agent` flag runs Sloth as a Java agent instead, allowing it to i.e. run in the JVM your tests live.
More details on how it works and what it does at https://github.com/VirtusLab/sloth

This feature is brought to the following sub-commands:
- run
- compile
- doc
- fix
- package
- publish
- repl (including the REPL artifacts themselves for Scala <3.8)
- run
- test

## Checklist

extensively, Cursor + Claude

```bash
./mill -i integration.test.jvm '*sun.misc.Unsafe*'
```


Gedochao:
@lbialy how close are we to publishing a stable version of `lazyvalgrade` to Maven Central?

SethTisue:
Why is it called "lazyvalgrade", what does "grade" mean here?

Gedochao:
@SethTisue I believe that's a pun on "lazy val upgrade".
We've already been discussing releasing it under a different name, coming very soon 😅

Gedochao:
https://central.sonatype.com/search?q=org.virtuslab+sloth 
Sloth artifacts are up!

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
```diff
@@ -210,6 +210,11 @@ final case class SharedOptions(
   @HelpMessage("Force object wrapper for scripts")
   @Tag(tags.experimental)
     objectWrapper: Option[Boolean] = None,
+  @HelpMessage("Patch Scala 3.0-3.7.x lazy val bytecode on the classpath for JDK 26+ compatibility")
+  @Name("lazyvalgrade")
```
lbialy, line 214: prolly rename to sloth?

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sloth.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.cli.commands.SpecificationLevel
+
+@DirectiveExamples("//> using sloth")
+@DirectiveUsage("//> using sloth", "`//> using sloth`")
+@DirectiveDescription(
+  "Patch Scala 3.0-3.7.x lazy val bytecode on the classpath for JDK 26+ compatibility"
+)
+@DirectiveLevel(SpecificationLevel.EXPERIMENTAL)
+final case class Sloth(
+  @DirectiveName("lazyvalgrade")
```
lbialy, line 15: prolly rename to sloth?

website/docs/reference/cli-options.md
```diff
@@ -1708,6 +1708,12 @@ Exclude sources
 
 Force object wrapper for scripts
 
+### `--sloth`
+
+Aliases: `--lazyvalgrade`, `--patch-lazy-vals`
```
lbialy, line 1713: --sloth?

tgodzik:
Do we now have adopt a sloth in a zoo? And have sloth plushies?

Gedochao:
We should. 🦥

website/docs/reference/cli-options.md
Gedochao: it is --sloth, --lazyvalgrade is an alias.

modules/cli/src/main/scala/scala/cli/commands/shared/SharedOptions.scala
Gedochao: it is --sloth, --lazyvalgrade is an alias.

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sloth.scala
Gedochao: it is sloth, lazyvalgrade is an alias.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
```diff
@@ -1025,6 +1078,8 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
       .withPreambleOpt(preambleOpt)
     value(alreadyExistsCheck())
     AssemblyGenerator.generate(params, destPath.toNIO)
+    val patchedDest = value(SlothPatcher.patchJarFile(destPath, options, logger))
+    if patchedDest != destPath then os.copy.over(patchedDest, destPath, createFolders = true)
```
warcholjakub, line 1106: Not sure whether this behaviour is intentional, but even if it is, we should probably warn the user.

With preamble:
```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> cat Main.scala
//> using scala 3.3.8

object Greetings:
  lazy val uno: String = "Heloł"
  lazy val dos: String = "Greetings"

@main def run(): Unit =
  import Greetings.*
  println(s"$uno & $dos")
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala --power package --assembly --sloth Main.scala -o app.jar
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> java -jar app.jar
Error: Invalid or corrupt jarfile app.jar
```
Without preamble:
```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade) [1]> ./mill -i scala --power package --assembly --preamble=false --sloth Main.scala -o app2.jar
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> java -jar app2.jar
Heloł & Greetings
```

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
```diff
@@ -0,0 +1,233 @@
...
+    else
+      val tmpJar = os.temp(prefix = "sloth-entries-", suffix = ".jar", deleteOnExit = false)
+      try
+        writeZipEntries(tmpJar, entries)
+        patchJarFile(tmpJar, options, logger).map(readZipEntries)
+      finally os.remove(tmpJar)
+
+  private def patchIfJar(path: os.Path, logger: Logger): os.Path =
+    if path.ext == "jar" then patchJar(path, logger)
+    else path
```
warcholjakub: Since we allow package outputs without a .jar extension, we shouldn't rely solely on the extension to determine whether it's a jar. At the very least, we should warn the user when Sloth does not patch an artifact.

```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala --power package --assembly --force --sloth Main.scala -o app_no_jar_ext

jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> java -jar app_no_jar_ext
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$ (file:/Users/jwarchol/Documents/GitHub/scala-cli/app_no_jar)
WARNING: Please consider reporting this to the maintainers of class scala.runtime.LazyVals$
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
Heloł & Greetings
```

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sloth.scala
warcholjakub, line 1: NIT: I've noticed that `//> using sloth false` results in Sloth being enabled. I think it'd be a better UX, if we handled `true` & `false` or informed user, that it doesn't take arguments.

warcholjakub (commented):
Those are the issues I managed to spot, while experimenting with it in the terminal. I'll look at the code next.

modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala
```diff
@@ -97,8 +100,11 @@ object NativeImage {
     )
 
     if cacheData.changed then {
```
warcholjakub, line 102: NIT: cacheData doesn't include the fact whether someone used sloth. So if you build an image without sloth, and then try with it enabled, it won't rebuild the image.

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
```diff
@@ -0,0 +1,233 @@
...
+        if os.isFile(path)
+      do
+        val relativePath = path.relativeTo(dir).toString.replace('\\', '/')
+        val entry        = ZipEntry(relativePath)
+        entry.setLastModifiedTime(FileTime.fromMillis(os.mtime(path)))
+        val content = os.read.bytes(path)
+        entry.setSize(content.length)
+        jos.putNextEntry(entry)
+        jos.write(content)
+        jos.closeEntry()
```
warcholjakub: We create a manifest in line 117, but if the user has supplied their own, we later try to add it, which causes a duplication error.

warcholjakub:
I'm wondering about one thing. Let's assume this hypothetical scenario:
1. Person 1 makes a library, and signs it.
2. Person 2 wants to use this library in his app. They also use sloth.
3. Sloth modifies the library, hence the hashes don't match up.
4. Person 2 runs his app, and verification fails.

Tried to reproduce it, and it indeed failed, but not sure how realistic and common this problem would be. Also I don't really see any straightforward solution. But yeah - just something to think about.

EDIT: `--sloth-agent` seems to avoid this issue

warcholjakub (commented):
Final comments

modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala
Gedochao: oooh, this is a good catch

lbialy:
@warcholjakub Can you elaborate about what do you mean by sloth modifying the dependency library? In what circumstances the verification you mentioned happens? What exactly fails verification?

warcholjakub:
> @warcholjakub Can you elaborate about what do you mean by sloth modifying the dependency library? In what circumstances the verification you mentioned happens? What exactly fails verification?

I mean something like this:

```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> cat SignedLib.scala
//> using scala 3.3.8
package signedlib
object SignedLib:
  lazy val message = "hello from signed library"
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> cat Main.scala 
import signedlib.SignedLib
@main def main(): Unit =
  println(SignedLib.message)
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala --power package SignedLib.scala --library  -o "signed.jar"
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> jarsigner -keystore "key.p12" -storepass changeit "signed.jar" test
jar signed.
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala --power run Main.scala --sloth --classpath signed.jar
[...]
Exception in thread "main" java.lang.SecurityException: SHA-384 digest error for Greetings$.class
        at java.base/sun.security.util.ManifestEntryVerifier.verify(ManifestEntryVerifier.java:254)
        at java.base/java.util.jar.JarVerifier.processEntry(JarVerifier.java:248)
        at java.base/java.util.jar.JarVerifier.update(JarVerifier.java:235)
        at java.base/java.util.jar.JarVerifier$VerifierStream.read(JarVerifier.java:453)
        at java.base/jdk.internal.loader.Resource.getBytes(Resource.java:106)
        at java.base/jdk.internal.loader.URLClassPath$JarLoader$1.getBytes(URLClassPath.java:745)
        at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:773)
        at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:691)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:620)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:578)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:490)
        at Main$package$.run(Main.scala:9)
        at run.main(Main.scala:7)
```

As I said, I'm not sure whether it's a realistic scenario - just smth I noticed.

EDIT: Modified the scenario a bit.

lbialy:
I did some research and you're right that the problem exists but fortunately:

  a) signing is now quite rare

  b) the whole signing mechanic is contingent on the presence of 

```
META-INF/*.SF
META-INF/*.DSA
META-INF/*.RSA
META-INF/*.EC
META-INF/SIG-*
```
files.

@Gedochao would it be possible to include a step where deps are copied to `.scala-build`, have these files stripped and then used as the runtime/package classpath when `--sloth` is specified?

Gedochao:
> @Gedochao would it be possible to include a step where deps are copied to `.scala-build`, have these files stripped and then used as the runtime/package classpath when `--sloth` is specified?

Yes, I think it should be possible.
Just a bit of a headache.
Lemme cook for a bit and see where I arrive.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
Gedochao: should be fixed

modules/cli/src/main/scala/scala/cli/packaging/NativeImage.scala
Gedochao: Should be fixed

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
Gedochao: Should be fixed. We now detect by content, rather than by extension.

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
Gedochao: this is technically out of scope, but I ended up solving it here. we now merge the manifests in this case.

Gedochao:
> @Gedochao would it be possible to include a step where deps are copied to `.scala-build`, have these files stripped and then used as the runtime/package classpath when `--sloth` is specified?

@lbialy we strip them in the global Sloth cache, rather than `.scala-build` - but yep, this tactic seems to work.

modules/directives/src/main/scala/scala/build/preprocessing/directives/Sloth.scala
Gedochao: fixed

modules/cli/src/main/scala/scala/cli/commands/test/Test.scala
warcholjakub, line 1: TLDR: tests seem to not execute properly with `--sloth`

```scala
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> cat /tmp/sloth_test.scala 
//> using scala 3.3.8
//> using dep org.scalameta::munit::1.3.4

class MainTest extends munit.FunSuite:
  test("must execute"):
    println("TEST_BODY_EXECUTED")
    assertEquals(1, 1)
```
### No sloth:
```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala test /tmp/sloth_test.scala
[...]
TEST_BODY_EXECUTED
Test run MainTest started
MainTest: finished 0.011s
Test run MainTest finished: 0 failed, 0 ignored, 1 total 0.015s
```

### Sloth:
```bash
jwarchol@vl-d-1635 ~/D/G/scala-cli (feature/lazyvalgrade)> ./mill -i scala test --sloth /tmp/sloth_test.scala
[...]
924] scala
924/924, SUCCESS] /Users/jwarchol/Documents/GitHub/scala-cli/millw scala test --sloth /tmp/sloth_test.scala 4s
```

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
```diff
@@ -0,0 +1,589 @@
...
+  private val cacheDir = Directories.directories.cacheDir / "sloth"
+
+  private[build] def shouldPatchProjectClasses(
+    hasJava: Boolean,
+    hasScala: Boolean,
+    scalaVersions: Seq[String]
+  ): Boolean =
+    val isPureJavaProject = hasJava && !hasScala
+    !isPureJavaProject &&
+    scalaVersions.exists(v => !v.coursierVersion.isScala38OrNewer)
```
warcholjakub, line 35: Shouldn't we exclude also Scala 2 from patching?

warcholjakub (commented):
Last two comments 🦥

modules/build/src/main/scala/scala/build/postprocessing/SlothPatcher.scala
warcholjakub: We came to the conclusion offline that it isn't worth it (for now at least).
EDIT: More detailed explanation below ↓↓↓

Gedochao: discussed offline.
TL;DR: this needs more experimentation. Theoretically a Scala 2.13 project could have 3.3-built artifacts on the classpath. 
Will investigate this as a follow-up, but treating as out-of scope here, leaving as-is for now.

modules/cli/src/main/scala/scala/cli/commands/test/Test.scala
Gedochao: fixed

warcholjakub (approved):
LGTM 🦥

PR #4363, opened 2026-07-09
Temporarily label all Scala 3 nightly tests as flaky

Gedochao:
It seems Scala 3 nightlies are broken (both Next and 3.3 LTS), so I'm labelling them as flaky to unblock the CI.

## Checklist

moderately

existing tests


tgodzik (approved):

Gedochao:
No longer necessary, nightlies are fixed.

PR #4393, opened 2026-07-23
Only prepend snapshot repos for `*-SNAPSHOT` Bloop versions

### LINKED ISSUE #4392: `scala-cli --power config repositories.default` doesn't override global default repo since `1.15.0`
**Version(s)**
1.15.0 (for 1.14.0 and earlier - OK)

**Describe the bug**
Hi. We get a problem in our infrastructure: we use private artifactory (as proxy for maven), and on gitlab runner we restrict access to the public Internet. So starting from version 1.15.0, for some reason, the repositories defined via `config repositories.default` and env `COURSIER_REPOSITORIES` stopped working - by default, when downloading `"Downloading compilation server 2.1.0"` (log after commad `run`), first attempt is "central.sonatype.com ", then "repo.scala-lang.org " and only after all those attempts fail does it try our artifactory.

**To Reproduce**
We use `virtuslab/scala-cli:1.15.0` as base image for gitlab runner, so I'll show how to reproduce at that docker image. `<repo>` - url for our artifactory
```bash
docker run -it  --entrypoint /bin/bash --env COURSIER_REPOSITORIES=<repo> --rm virtuslab/scala-cli:1.15.0
echo -e "//> using scala 3.8.4\n\n@main def foo(): Unit = {}" > test.scala
scala-cli --power config repositories.default <repo>
scala-cli run test.scala
```
so after start `run` - cli uses resolvers fst: "central.sonatype.com ", snd: "repo.scala-lang.org ", low priority: `<repo>`

**Expected behaviour**
I expects `scala-cli` would use `<repo>` (provided by `env` `COURSIER_REPOSITORIES` and command `scala-cli --power config repositories.default` ) as first priority resolver

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4392

The TL;DR of it is that we've been prepending snapshot repositories for Bloop downloads regardless if a snapshot Bloop was used or not.
We should only prepend them for actual Bloop snapshot builds.

## Checklist

extensively

included an automated test

### DIFF
--- a/modules/build/src/main/scala/scala/build/Bloop.scala
+++ b/modules/build/src/main/scala/scala/build/Bloop.scala
@@ -78,11 +78,13 @@ object Bloop {
       val res = value {
         Artifacts.artifacts(
           Seq(Positioned.none(dep)),
-          Seq(
-            coursier.Repositories.centralMavenSnapshots,
-            RepositoryUtils.snapshotsRepository,
-            RepositoryUtils.scala3NightlyRepository
-          ),
+          if dep.version.endsWith("SNAPSHOT") then
+            Seq(
+              coursier.Repositories.centralMavenSnapshots,
+              RepositoryUtils.snapshotsRepository,
+              RepositoryUtils.scala3NightlyRepository
+            )
+          else Nil,
           Some(params),
           logger,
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM

PR #4404, opened 2026-07-31
sloth: bump to 0.1.0-M2 & patch loose class files at `-cp`

Gedochao:
This makes Sloth patch loose classfiles passed via `-cp`.
It also bumps Sloth to 0.1.0-M2: https://github.com/VirtusLab/sloth/releases/tag/v0.1.0-M2

## Behaviour without this change

Given a lazy.scala file:
```scala
lazy val hah = true
def slothful: Boolean = hah
```
when
```bash
scala-cli compile --server=false lazy.scala -S lts --jvm 8 -d cp
scala-cli run --server=false -e 'println(slothful)' -cp cp --sloth --power --jvm 25  
```
we still get the warning
```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$ 
(file:(...)/scala-library-3.8.4.jar)
WARNING: Please consider reporting this to the maintainers of class scala.runtime.LazyVals$
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```

## Checklist

extensively

included automated tests

### DIFF
--- a/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
+++ b/modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
@@ -459,7 +459,13 @@ object Package extends ScalaCommand[PackageOptions] with BuildCommandHelpers {
         case PackageType.LibraryJar =>
           val libraryJar0 = Library.libraryJar(builds, mainClassOpt)
           val libraryJar  = value(
-            SlothPatcher.patchJarFile(libraryJar0, builds.head.options, logger)
+            SlothPatcher.patchJarFile(
+              libraryJar0,
+              builds.head.options,
+              logger,
+              hierarchyClassPath = builds.flatMap(_.fullClassPath).distinct,
+              source = SlothPatcher.SlothSource.Project
+            )
           )
           value(alreadyExistsCheck())
           if force then os.copy.over(libraryJar, destPath, createFolders = true)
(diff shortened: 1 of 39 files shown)


modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
```diff
@@ -67,7 +71,7 @@ object BuildCommandHelpers {
                 output,
                 successfulBuild.options,
                 logger,
-                shouldPatch = SlothPatcher.shouldPatchProjectClasses(Seq(successfulBuild))
+                shouldPatch = successfulBuild.options.notForBloopOptions.sloth
```
warcholjakub, line 74: Hypothetically - we have Scala 3.8.4, `destExisted` is false and `wasPatchedInThisProcess` is also false. Previously it wouldn't be patched (and that's expected), because of `shouldPatchProjectClasses`. And now `!destExisted && SlothPatcher.wasPatchedInThisProcess[...]` is `true && false`, so it isn't skipped. And there is no guard in form of `shouldPatchProjectClasses`. So it would try to patch it, and it wouldn't find anything to patch. Unless I'm misunderstanding something.

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
warcholjakub, line 923: Since `patchedPath` could be a temporary jar with a random name, wouldn't it cause packaging to be nondeterministic. I mean different runs would result in different hashes.

Gedochao:
The native image failures seem to be caused by a bug in Sloth's JarProcessor.
They're not directly related to the fix here, but seem to have been activated by the changes.
While we could theoretically work around it here, I'm thinking it's actually best to fix it upstream first:
- https://github.com/VirtusLab/sloth/issues/1

Moving this PR to draft for now.

warcholjakub (commented):
Things I noticed before the PR got drafted

Gedochao:
Upstream fix: 
- https://github.com/VirtusLab/sloth/pull/2

modules/cli/src/main/scala/scala/cli/commands/util/BuildCommandHelpers.scala
Gedochao: should be fixed

modules/cli/src/main/scala/scala/cli/commands/package0/Package.scala
Gedochao: should be fixed

PR #4414, opened 2026-08-04
Maintenance/agents tests

Gedochao:
Integration tests get flaky when run in parallel, which in turn gives false positives when agents are running.

moderately

ran some agent tasks with this

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -128,7 +128,6 @@ object `scala-cli-bsp` extends JavaModule
 }
 object integration extends CliIntegration {
   object test extends IntegrationScalaTests {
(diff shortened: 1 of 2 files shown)


build.mill
```diff
@@ -128,7 +128,6 @@ object `scala-cli-bsp` extends JavaModule
 }
 object integration extends CliIntegration {
   object test extends IntegrationScalaTests {
-    override def testParallelism: T[Boolean]           = !isCI
```
Gedochao, line 131: this is unnecessary, as it's inherited from `ScalaCliTests`

tgodzik (approved):

zielinsky (approved):

PR #4418, opened 2026-08-10
Support JEP 512 flexible main methods on JDK 25+ / JDK21+ with `--enable-preview`

### LINKED ISSUE #3886: Main class detection: recognize simpler main methods in Java 25 (JEP-512)
```
$> cat T.scala
class A {
  def main(): Unit = println(1)
}

$> scala-cli T.scala
Compiling project (Scala 3.7.3, JVM (25))
Compiled project (Scala 3.7.3, JVM (25))
[error]  No main class found

$> sc T.scala
> scala compile --server=false -S 2.13 -d . -release 8 T.scala

$> jar cf run.jar A.class

$> scala-cli -S 2.13.16 run.jar
[error]  No main class found
```

cf https://github.com/sbt/sbt/issues/7384

Gedochao:
Closes #3886

Recognize and launch JEP 512 main methods (`static`/`instance`, with/without `String[]`) when targeting the JVM on JDK 25+ (or JDK 21+ with `--enable-preview`).

```scala
//> using jvm 25
class Hello {
  def main(): Unit =
    println("Hello from an instance main")
}
```

```scala
//> using jvm 21
//> using javaOpt --enable-preview
class Hello {
  def main(): Unit = println("Hello on JDK 21 preview")
}
```

Ofc, works with Java, too.

```java
//> using jvm 25
// Hello.java
void main() {
    System.out.println("Hello from a compact source");
}
```

```java
//> using jvm 25
// HelloWithExplicitClss.java
public class HelloWithExplicitClss {
    void main() {
        System.out.println("Hello from an instance main");
    }
}
```

Worth noting that on lower JDKs we still note those main classes down in the error, since detection works just fine, they're just not runnable.

```text
[error]  No main class found
Class Hello declares a main method that requires JDK 25 or newer (JEP 512); the current JVM is 17.
```

## LLM usage
extensively, Cursor / Claude

## Known limitations
- Default `package` / bootstrap cannot launch JEP-512 mains until coursier’s
  bootstrap launcher is updated; use `--assembly` or `--library`
- doesn't work with Scala.js / Scala Native for now

### DIFF
--- a/modules/build/src/main/scala/scala/build/Build.scala
+++ b/modules/build/src/main/scala/scala/build/Build.scala
@@ -65,14 +65,76 @@ object Build {
       sources.resourceDirs ++ artifacts.compileClassPath
     def fullClassPath: Seq[os.Path]        = Seq(output) ++ dependencyClassPath
     def fullCompileClassPath: Seq[os.Path] = fullClassPath ++ dependencyCompileClassPath
-    private lazy val mainClassesFoundInProject: Seq[String] = MainClass.find(output, logger).sorted
-    private lazy val mainClassesFoundOnExtraClasspath: Seq[String] =
-      options.classPathOptions.extraClassPath.flatMap(MainClass.find(_, logger)).sorted
+    private lazy val mainClassCandidatesInProject: Seq[MainClass.MainClassCandidate] =
+      MainClass.find(output, logger).filterNot(isScriptWrapperUserCodeClass)
+    private lazy val mainClassCandidatesOnExtraClasspath: Seq[MainClass.MainClassCandidate] =
+      options.classPathOptions.extraClassPath.flatMap(MainClass.find(_, logger))
     private lazy val mainClassesFoundInUserExtraDependencies: Seq[String] =
       artifacts.jarsForUserExtraDependencies.flatMap(MainClass.findInDependency).sorted
+
+    private def isScriptWrapperUserCodeClass(candidate: MainClass.MainClassCandidate): Boolean =
+      // ClassCodeWrapper puts user code in a class named `{script}$_`; its instance main must not
+      // surface as a selectable entry point alongside the real `{script}_sc` wrapper main.
+      candidate.className.endsWith("$_")
+
+    private lazy val hasEnablePreview: Boolean =
+      options.javaOptions.javaOpts.toSeq.exists(_.value.value == "--enable-preview")
+
+    private lazy val isJvmPlatform: Boolean = options.platform.value == Platform.JVM
+
+    private lazy val jvmVersion: Int = options.javaHome().value.version
+
+    /** JEP 512 main method shapes are only launchable by a new enough JVM; a classic
+      * `static void main(String[])` is launchable on every platform.
+      */
+    private def isMainMethodSupportedByTargetRuntime(kind: MainClass.MainMethodKind): Boolean =
+      !kind.requiresJep512 ||
+      (isJvmPlatform && kind.isSupportedByJvm(jvmVersion, hasEnablePreview))
+
+    private lazy val allMainClassCandidates: Seq[MainClass.MainClassCandidate] =
+      mainClassCandidatesInProject ++ mainClassCandidatesOnExtraClasspath
+
+    private def supportedMainClassNames(candidates: Seq[MainClass.MainClassCandidate])
+      : Seq[String] =
+      candidates.collect {
+        case c if isMainMethodSupportedByTargetRuntime(c.kind) => c.className
+      }.sorted
+
+    private lazy val mainClassesFoundInProject: Seq[String] =
+      supportedMainClassNames(mainClassCandidatesInProject)
+    private lazy val mainClassesFoundOnExtraClasspath: Seq[String] =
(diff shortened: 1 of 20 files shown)


modules/build/src/test/scala/scala/build/tests/MainClassTests.scala
```diff
@@ -0,0 +1,413 @@
...
+  )
+
+  val defaultOptions: BuildOptions = baseOptions.copy(
+    scalaOptions = baseOptions.scalaOptions.copy(
+      scalaVersion = Some(MaybeScalaVersion(Constants.defaultScalaVersion)),
+      scalaBinaryVersion = None
+    )
+  )
+
+  private def writeClass(
```
tgodzik: Do we need until tests like this? Writing class files manually seems like a super fragile thing. Maybe have just run tests or compile using normal ways.

tgodzik (dismissed):

modules/build/src/test/scala/scala/build/tests/MainClassTests.scala
Gedochao: Agreed, it's overkill.
Rewrote those to just use `.java` inputs.

modules/build/src/main/scala/scala/build/internal/MainClass.scala
```diff
@@ -12,12 +12,39 @@ import scala.build.{Logger, retry}
 
 object MainClass {
 
-  private def stringArrayDescriptor = "([Ljava/lang/String;)V"
+  /** Shape of a `main` entry point recognised in bytecode.
+    *
+    * Case declaration order is the JEP 512 resolution order: `main(String[])` before `main()`, and
+    * within each signature the static form before the instance form.
+    *
+    * Inherited mains are still not detected — the ASM scan is per-class and has no classpath-wide
```
warcholjakub: Aren't they? They seem to be detected just fine. My theory is that it's because info about the main method is also present in the class that inherits it.
```
jwarchol@vl-d-1635 /t/javap-test> javap -p -s classes/InheritsMain.class
Compiled from "inh-test.scala"
public class InheritsMain implements InheritedMain {
  public InheritsMain();
    descriptor: ()V

  public void main();
    descriptor: ()V
}
```

warcholjakub: Tho there's probably a case, where this doesn't happen and it doesn't detect the main method. But still - imo the comment needs updating.

modules/build/src/main/scala/scala/build/Build.scala
```diff
@@ -65,15 +65,77 @@ object Build {
...
+      MainClass.find(output, logger).filterNot(isScriptWrapperUserCodeClass)
+    private lazy val mainClassCandidatesOnExtraClasspath: Seq[MainClass.MainClassCandidate] =
+      options.classPathOptions.extraClassPath.flatMap(MainClass.find(_, logger))
     private lazy val mainClassesFoundInUserExtraDependencies: Seq[String] =
       artifacts.jarsForUserExtraDependencies.flatMap(MainClass.findInDependency).sorted
+
+    private def isScriptWrapperUserCodeClass(candidate: MainClass.MainClassCandidate): Boolean =
+      // ClassCodeWrapper puts user code in a class named `{script}$_`; its instance main must not
+      // surface as a selectable entry point alongside the real `{script}_sc` wrapper main.
+      candidate.className.endsWith("$_")
```
warcholjakub, line 78: NIT: in theory we get a warning from a compiler that we shouldn't use `$`, but it still lets you name a class `Test$_` (and as a result it'd skip a legit main class). I bet it won't affect anyone, since it's such a niche case, but yeah... some food for thought

modules/build/src/main/scala/scala/build/internal/MainClass.scala
Gedochao: Good catch - and yeah, it does work for a main method inherited from a Scala trait.
It doesn't however, if it's inherited from a Java class or interface - I meant to try that as a follow-up.
Updated the comment + added a test for the case which does work.

Gedochao: Example case for when it does not work:
```java
// C.java
public abstract class C {
    void main() {
        System.out.println("Hello from an instance main");
    }
}
```

```java
// A.java
public class A extends C
```

warcholjakub (approved):
LGTM

modules/build/src/main/scala/scala/build/Build.scala
Gedochao: Tbf, that main method remains runnable, it's just excluded from detection.
So if someone decides to indeed name a class `*$_` (what an odd choice that'd be), then they could still run it, Scala CLI's own runner simply won't detect it automatically.
The workaround filters those classes here is to work around the `*.sc` script wrapper here. 
Maybe I could narrow it down as a follow-up, but I think it's out of scope here.

warcholjakub: Yeah, true

tgodzik (approved):
LGTM

PR #4420, opened 2026-08-11
Split CI down into smaller workflows

Gedochao:
This splits the old `ci.yml` workflow down into smaller, reusable ones - with `ci.yml` being kept as an orchestrator.

triggers, permissions, secrets and concurrency are kept. Platform/check/release logic lives in `workflow_call` workflows so everything still runs in **one** workflow run.

## Why do we need this?
Chiefly b/c our test suites rely on downloading dependencies from the internet a lot, and that causes a whole lot of flakes. Those can then be restarted easier in a given scope, rather than for the entire mega-workflow.
IMO, readability also improves and it should be easier to manually find the job one wants to edit.

## New structure
| Workflow | Contents |
|----------|----------|
| `changes.yml` | Change classification + override outputs |
| `jvm.yml` | `unit-tests` + all `jvm-*` |
| `linux.yml` | Linux x86_64/arm64 generate + `native-linux-*` |
| `macos.yml` | macOS intel/arm64 generate + `native-macos-*` |
| `windows.yml` | Windows generate + `native-windows-*` + `vc-redist` |
| `static.yml` / `mostly-static.yml` | Static / mostly-static generate + tests (+ docker push) |
| `misc.yml` | format, checks, docs, fish, sbt export, bloop, reference-doc |
| `release.yml` | publish, launchers, `update-*`, SDKMAN |

Gating flags from `changes` are passed as raw `workflow_call` inputs; job bodies are otherwise unchanged.

## Checklist
N/A

Extensively — Cursor was used to plan the split, extract jobs into reusable workflows, rewrite `ci.yml` as an orchestrator, and validate completeness against the original job set.

Locally, via `actionlint`
Otherwise, not much choice but run it on the actual CI.


Gedochao:
@tgodzik note that most branch protection checks would need a slight update if/when we go through with this, like `CI / native-macos-tests-rc` is now `CI / macos / native-macos-tests-rc`

tgodzik (approved):
The idea is sound, though I think I will need to trust you with this.

PR #4424, opened 2026-08-12
Auto detect user-created main methods in classes named with the `*$_` suffix

Gedochao:
Follow-up to https://github.com/VirtusLab/scala-cli/pull/4418#discussion_r3758403448

Now this was easier than expected. We'll only filter out the classes in the generated script wrappers, and still detect user-created ones with main methods and the `*$_` suffix.
Why we'd need this is beyond me - why not is the likely answer.

## Checklist

extensively, Cursor

automated test included


warcholjakub (approved):
LGTM

PR #4425, opened 2026-08-12
Auto detect main methods inherited from Java sources (including JEP 512 ones)

Gedochao:
Follow-up to https://github.com/VirtusLab/scala-cli/pull/4418#discussion_r3758513282

This enables auto-detection of main methods inherited from Java classes/interfaces. Including ones added by JEP 512.

```java
// C.java
public abstract class C {
    void main() {
        System.out.println("Hello from an instance main");
    }
}
```

```java
// A.java
public class A extends C
```

## Checklist

extensively

added automated tests

### DIFF
--- a/modules/build/src/main/scala/scala/build/internal/MainClass.scala
+++ b/modules/build/src/main/scala/scala/build/internal/MainClass.scala
@@ -7,6 +7,7 @@ import java.io.{ByteArrayInputStream, InputStream}
 import java.nio.file.NoSuchFileException
 import java.util.jar.{Attributes, JarFile}
 
+import scala.annotation.tailrec
 import scala.build.internal.zip.WrappedZipInputStream
 import scala.build.{Logger, retry}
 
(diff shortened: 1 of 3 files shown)


warcholjakub (commented):
Two main things I found:
- First, not caused by this PR:
We don't check whether the main method is protected (and it's not allowed for non JEP512 Java versions). And in result:
```java
public class Test {
  protected static void main(String[] args) {
    System.out.println("hello");
  }
}
```
yields:
```
./mill -i scala run /tmp/test-repro/Test.java --jvm 17
[...]
Error: Main method not found in class Test, please define the main method as:
   public static void main(String[] args)
or a JavaFX application class must extend javafx.application.Application
924/925, 1 FAILED] /Users/jwarchol/Documents/GitHub/scala-cli/millw scala run /tmp/test-repro/Test.java --jvm 17 3s
```
I'll try to prepare a fix.
EDIT1: https://github.com/VirtusLab/scala-cli/pull/4430
EDIT2: We'll probably have to make sure it works properly in inheritance too. So, imho, it'd probably be best to merge it before this PR, and address it here.
- Second, related to the changes:
```java
package far;

public abstract class FarBase {
  static void main(String[] args) {
    System.out.println("FarBase.main");
  }
}
```
```java
package near;

public abstract class NearBase extends far.FarBase {
  private static void main(String[] args) {
    System.out.println("NearBase.main");
  }
}
```
```java
package near;

public class Child extends NearBase {}
```
results in Child being incorrectly classified as a main class.

Gedochao:
Rebased on top of https://github.com/VirtusLab/scala-cli/pull/4430 plus added handling for the missing case.
Should be ready to re-review.

warcholjakub (commented):
One more thing that I've overlooked previously is that when we take the example from my previous review and replace `private static void main(String[] args)` with `public static int main(String[] args)`, then the same thing happens. But other than that - looks good.
Repro:
```java
package far;

public abstract class FarBase {
  static void main(String[] args) {
    System.out.println("FarBase.main");
  }
}
```
```java
package near;

public abstract class NearBase extends far.FarBase {
  public static int main(String[] args) {
    return 0;
  }
}
```
```java
package near;

public class Child extends NearBase {}
```

warcholjakub:
Is it ready for re-review?

Gedochao:
> Is it ready for re-review?

actually, yep!

warcholjakub (approved):
LGTM!

PR #4427, opened 2026-08-13
fix: reject malformed toolkit coordinates

warcholjakub:
Malformed toolkit coordinates such as `:`, `::`, `a:b:c` used to pass and resulted in a weird behaviour. This PR fixes this, and makes it that toolkit only accepts `version` or `flavour:version`.

Related: https://github.com/scala/scala3/pull/26788

## Checklist

Moderately

Automated tests


Gedochao (approved):
lgtm

PR #4428, opened 2026-08-13
Add support for passing individual files as resources

### LINKED ISSUE #792: Support for `--resource` flag
For now, Scala CLI supports only `--resource-dir` flags, which allows specifying resources directory. It would be nice, if it will be available new flag `--resource` to pass resources file instead of whole directory.

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/792

This effectively adds `--resource` as an alias for `--resource-dir`, as I handle both individual files and directories in one place (there's seemingly no reason not to, really).

## Checklist

Extensively

Added automated tests


warcholjakub (commented):
Overall - looks good, but I've got two comments:
- [NIT] If we support also files besides directories, then `resourcesDir: Seq[os.Path]` is a little bit misleading. Perhaps a rename to `resourcePath` would be better? (not sure whether the same applies for the `resourcesVirtualDir: Seq[os.SubPath]`, but maybe?)
- Exporting a project with `//> using resource path/file.txt` to a SBT project yields `Compile / unmanagedResourceDirectories ++= Seq(file("path/file.txt"))` which results in it not working. Not sure, but I suppose it also could apply to other forms of exporting.

Gedochao:
> * [NIT] If we support also files besides directories, then `resourcesDir: Seq[os.Path]` is a little bit misleading. Perhaps a rename to `resourcePath` would be better? (not sure whether the same applies for the `resourcesVirtualDir: Seq[os.SubPath]`, but maybe?)

The main thing is, I left it there for backwards compat. We could deprecate that alias, I suppose - but I'd have to do it in a new minor (and I mean for the next Scala CLI release to be 1.16.1).

> * Exporting a project with `//> using resource path/file.txt` to a SBT project yields `Compile / unmanagedResourceDirectories ++= Seq(file("path/file.txt"))` which results in it not working. Not sure, but I suppose it also could apply to other forms of exporting.

Oh wow, you tested exporting. I'll fix that after the weekend :+1:  Although I bet there's all sort of shenenigans like that, as `export` doesn't have a very good test coverage and needs some love. A lot of love, really. :sweat_smile:

warcholjakub:
> The main thing is, I left it there for backwards compat. We could deprecate that alias, I suppose - but I'd have to do it in a new minor (and I mean for the next Scala CLI release to be 1.16.1).

I completely forgot to include the one sentence, that I intended to write 🤦‍♂️. I meant to specify that I referred to `resourcesDir: Seq[os.Path]` that's an argument of `final case class ClassPathOptions`. My bad.

EDIT: (Unless I misunderstood that you misunderstood me lol)

Gedochao:
Rebased.
Did the renames, as suggested.
Ofc, the `export` sub-command was producing broken builds for the individual files as resources on all build tools we support, not just SBT.
Should be fixed.

modules/cli/src/main/scala/scala/cli/internal/CachedBinary.scala
```diff
@@ -37,7 +37,7 @@ object CachedBinary {
     val classpathResourceDirsIt =
       build.options
         .classPathOptions
-        .resourcesDir
+        .resourcePaths
         .flatMap(dir => hashResourceDir(dir))
```
warcholjakub: If the resource is a file, we still delegate it to the `hashResourceDir` and it tries to call `os.walk` on it. We should handle the file case too.

warcholjakub (commented):
One last thing I've noticed

modules/cli/src/main/scala/scala/cli/internal/CachedBinary.scala
Gedochao: should be fixed

warcholjakub (approved):
LGTM

PR #4430, opened 2026-08-14
fix: non-public main detection (JEP 512)

warcholjakub:
Fixes a case where:
```java
public class Test {
  protected static void main(String[] args) {
    System.out.println("hello");
  }
}
```
would be classified as a non-JEP512 main method.

Related: https://github.com/VirtusLab/scala-cli/pull/4425#pullrequestreview-4937695519

## Checklist

Minimally

Automated tests


modules/build/src/test/scala/scala/build/tests/MainClassTests.scala
```diff
@@ -17,6 +17,7 @@ class MainClassTests extends TestUtil.ScalaCliBuildSuite {
 
   private val jep512MinJava        = scala.build.internal.Constants.jep512MinJavaVersion
   private val jep512PreviewMinJava = scala.build.internal.Constants.jep512PreviewMinJavaVersion
+  private val preJep512Java        = jep512MinJava - 1
```
Gedochao, line 20: NIT it's used in one test, so perhaps the simple subtraction can be done in-place there, where it makes more sense.

Gedochao (approved):
lgtm

PR #4436, opened 2026-08-20
Fix REPL on JDK 24+ for Scala versions with split JLine modules

lrytz:
The Scala 2 branch passed `--add-modules org.jline`, which is the automatic
module name of the `org.jline:jline` uber-jar. Scala 2.13.19 switches to the
individual JLine modules (as Scala 3 already did), where no `org.jline` module
exists, so the REPL dies with:

  java.lang.module.FindException: Module org.jline not found

https://github.com/scala/scala/pull/11275

Use `--add-modules ALL-MODULE-PATH` instead of naming a module. The module path
assembled here only ever holds JLine artifacts, so this works for both shapes
and needs no per-Scala-version logic; the Scala 2 and Scala 3 branches collapse
into one. `--enable-native-access` still needs a real module name, picked from
whether the split-out `jline-native` artifact is present.

Extracted as `jlineJavaOpts` so it can be unit tested.

### DIFF
--- a/modules/build/src/main/scala/scala/build/ReplArtifacts.scala
+++ b/modules/build/src/main/scala/scala/build/ReplArtifacts.scala
@@ -6,6 +6,7 @@ import coursier.util.Task
 import dependency.*
 
 import java.io.File
+import java.nio.file.Paths
 
 import scala.build.EitherCps.{either, value}
 import scala.build.errors.BuildException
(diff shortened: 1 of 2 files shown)


SethTisue:
Does it work with 2.12 too?

Gedochao:
> Does it work with 2.12 too?

~~Let's run integration tests on this.~~ They're currently gated to run on Scala 2.13+
https://github.com/VirtusLab/scala-cli/blob/1b9b93695504046f2517144185227c9bee144dae/modules/integration/src/test/scala/scala/cli/integration/RunJdkTestDefinitions.scala#L136

EDIT: nvm, Scala 2.12 isn't relevant here.

modules/build/src/main/scala/scala/build/ReplArtifacts.scala
```diff
@@ -117,4 +99,29 @@ object ReplArtifacts {
...
+    * earlier) or as the individual modules (Scala 3, 2.13.19+), so no single module name works for
+    * `--add-modules`; `ALL-MODULE-PATH` adds whatever is on the module path instead, see
+    * https://openjdk.org/jeps/261. `--enable-native-access` has no equivalent, hence the check.
+    */
+  private[build] def jlineJavaOpts(
+    jlineArtifacts: Seq[String],
+    javaVersion: Int
+  ): Seq[String] =
+    if javaVersion >= 24 && jlineArtifacts.nonEmpty then {
+      val hasSplitModules =
```
Gedochao:
```suggestion
      val hasJLineNative =
```
NIT more readable, IMO

Gedochao:
> Does it work with 2.12 too?

Yep, it does.

modules/build/src/main/scala/scala/build/ReplArtifacts.scala
Gedochao: fixed

Gedochao (approved):
lgtm

PR #4466, opened 2026-09-11
Build Scala CLI with Scala 3 Next

Gedochao:
- builds Scala CLI with Scala 3 Next, so currently 3.9.0 (well, most of it - see below)
- pins the `config` module at 3.9.0, as it's used by other tools as a dep (i.e. coursier)
- pins the `specification-level` module at 3.3 LTS until its EOL
- similarly, `runner` and `test-runner` stay at 3.3 LTS until its EOL
- all modules at 3.3 LTS will now be built with `-Yfuture-lazy-vals` and `-java-output-version:17` settings
- sources were migrated to newer syntax, where applicable (mostly compiler rewrites, and not a lot of it)
- Scala CLI's `directives-parser` gets replaced by Scala 3's `scala3-directives-parser`
- fixes the GraalVM native build to work around some issues with standard library internals getting initialised 

## Checklist

Yes, although far less than I initially assumed I would 😅 

existing tests should suffice

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -105,8 +105,8 @@ object `build-macros` extends Cross[BuildMacros](Scala.scala3MainVersions)
     with CrossScalaDefaultToInternal
 object `directives-parser` extends Cross[DirectivesParserModule](Scala.scala3MainVersions)
     with CrossScalaDefaultToInternal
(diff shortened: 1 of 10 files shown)


warcholjakub (approved):
LGTM

PR #4482, opened 2026-09-18
Support newer `esVersion` strings; hard error on unknown `esVersion`

### LINKED ISSUE #4478: latest es version strings are not supported
**Version(s)**
1.17.0

**Describe the bug**
when defining a jsEsVersionStr that is newer that es2021 i get the error `Warning: unrecognized argument: es2022 for --js-es-version parameter, use default value: ES2015`. I have also scala 3.9.0 and jsVersion 1.22.0 configured.

**To Reproduce**
define any of es2022 - es2026 as values for jsEsVersionStr directive or --js-es-version parameter.

**Expected behaviour**
scala js 1.22 and scala 3.9.0 added support for es2022 - es2026, but scala-cli seems not to respect them. I expect any of these values to work and be forwarded to the scala js compiler

Gedochao:
Fixes https://github.com/VirtusLab/scala-cli/issues/4478
- adds missing support for `ES2022`-`ES2026`
- now hard errors when an unknown `esVersion` is passed, rather than falling back to an ancient version

## Checklist

extensively

added automated tests


warcholjakub (approved):
LGTM

